PR #6, opened 2015-03-19
BF: remove executable permission from binary, text, and configuration files

yarikoptic:
done via
```
find global/ -type f -perm /u+x | grep -v '\.\(sh\|py\)$' | xargs chmod a-x
```
having any of those files executable when not intended to be such could is not kosher to say the least ;-)


tbbrown:
@yarikoptic Thank you.

There were two files in the global/scripts directory that were not named ending with .py or .sh, but still need to have executable permission: epi_reg_dof (which is a bash script) and mtrendout.awk (which we might get away with not being executable, but I'm going to leave it as is for now.)  So I didn't use your find command.  But I did change all the image files, text files, and configuration files so that they no longer have executable permission.

This change is now in the head of the master branch.  So I'm closing this pull request.

PR #7, opened 2015-03-26
Fixed bug for copying rh.curv.pial not lh.curv.pial

healthonrails:
In the FreeSurfer Pipeline, the last copy command should copy rh.curv.pial instead of lh.curv.pial.


glasserm:
Thanks you aren't the first to mention this one and it should be merged.  Tim, can you do that?

Matt.

tbbrown:
@healthonrails Thank you for pointing this out.

As @glasserm points out you are not the first to mention this, and when it was previously mentioned, I made the change in the branch I was working on at the time (the 7THCP branch).  

I have merged the 7THCP branch into the master branch. So I believe this pull request is no longer necessary.  Therefore, I'm closing this pull request.  @healthonrails please feel free to check the head of the master branch and verify that the change that you noticed was necessary is now included there.

Thanks again.

Tim

PR #8, opened 2015-04-02
Merge 7T HCP changes along with some Prisma 3T changes into master

tbbrown:


glasserm:
Thanks @tbbrown .  Can you give me some clarification on changes number 3 (remove --fsl-dof) and number 10 (fix ${FinalfMRIResolution}).  Were these just fixes of yours to a previous version or bugs of mine?

tbbrown:
Number 3 (remove --fsl-dof) 

Full check-in comment:
* Removed --fsl-dof option from calls to bbregister in 
  DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
* Moved the version of epi_reg_dof which is a modification of the FSL v5.0.7
  version of epi_reg to the global/scripts directory
* Removed all versions of epi_reg_dof from the fMRIVolume/scripts directory
* Updated DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
  to invoke epi_reg_dof in the global/scripts directory

These changes were part of the changes for using a higher degree of freedom in EPI registration for the 7T data and (as I recall) were guided by Keith Jamison.

In particular, bbregister calls in the DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh used both the --${dof} and --fsl-dof ${dof} options.  As I recall it, I was informed (I don't recall by whom, but could find the email chain if necessary) that an option like "--fsl-dof 6" was unnecessary or redundant when an option like --6 was specified.  

Number 10 (fix ${FinalfMRIResolution}) 

Full check-in comment:
Fixed FinalfMRIResolution in GenericfMRISurfaceProcessingPipeline7TBatch.sh to match value in GenericfMRIVolumeProcessingPipeline7TBatch.sh

This is a change to one of the example batch scripts I created for running the pipelines on the 7T data. It changed the FinalfMRIResolution from 2 mm to 1.60 mm.

glasserm:
Thanks @tbbrown

PR #13, opened 2015-07-22
Modified to work with scans that don't start with b0 volume

kjamison:
DiffusionPreprocessing/scripts/basic_preproc.sh does not properly match diffusion volumes to b0 scans when the diffusion volumes precede the first b0 volume: index.txt ends up with 0 and similar problem for the first Neg scans.  This patch makes matches pre-b0 diffusion volumes to the first b0 in the scan.


mharms:
Very old PR.  Closing.

PR #15, opened 2015-10-06
Fix globbing and expansion bug in wb_command -metric-math (PostFreeSurfer)

lennartverhagen:
wb_command -metric-math requires a mathematical expression stated in
quotes as input. Unfortunately passing such an argument is prone to
glitches when it includes spaces. In fact, it will glob and expand
using CentOS 6.7 and bash 4.1.2(1). The fix is easy, just remove the
spaces in the expression. I did so in: PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh. I haven't checked elsewhere.


coalsont:
Is there an easy test for this bug in bash, or a centos bug report?  Globbing inside of quotes sounds like an extremely serious shell bug, so I'd want some verification that this is the cause of the behavior (and that your fix isn't still attempting globbing that just doesn't match any filenames due to the removal of spaces).

Which version and build of wb_command are you using, and is the wb_command being called in your configuration a wrapper script?  Very old versions of these wrapper scripts provided with releases did not correctly quote the input arguments, which might result in undesired globbing.  Specifically, the versions that should work correctly use the string "$@" (including quotes), while beta 0.7 had $*.  Beta 0.8 and later seem to have a script that should work correctly, so I don't know how you could be using such an affected script, but it is worth checking.

lennartverhagen:
Hi Tim, let me see if I can get a quick script for this potential bug and also test it locally on my OS X. But a few things I can answer directly.

First, I'm using wb_command version 1.1.1 (2015-07-20).
Second, it doesn't so much expand the path while inside double quotes, as that it doesn't pass along the double quotes to wb_command. So when stating
wb_command -metric-math "A / B" foobar.shape.gii -var A $foo -var B $bar
it doesn't keep "A / B" but splits at the spaces and parses it as three separate arguments A, /, and B. Hence if you state:
wb_command -metric-math "A * B" foobar.shape.gii -var A $foo -var B $bar
it will break on spaces and expand the path of the wildcard. That's why simply removing spaces solved the issue for me.
Does that explain the problem a bit better? Let me know.

coalsont:
I double checked the redhat release zip of 1.1.1, and its wrapper script has the correct "$@" syntax, so if you unzipped the entire zip file when installing, it should have the correct script somewhere.  The symptoms you describe look like what I would expect with a faulty wrapper script in between the command and the binary.

So, with the variables set as they would be in the script, I would like to know what the result of this command is:

file ${CARET7DIR}/wb_command

If the result says something like "shell script" or "ASCII text", then please also give the output of:

grep wb_command ${CARET7DIR}/wb_command

lennartverhagen:
Of course, that must be it! Stupid of me. I'm working on the Jalapeno server of FMRIB in Oxford. Here CARET7DIR is set to /opt/fmrib/bin to allow wb_command to find other libraries there. And /opt/fmrib/bin/wb_command is a symlink to wb_wrapper. This is a POSIX shell script. The grep command to has no output (neither for the wb_command symlink or for the wb_wrapper script).

lennartverhagen:
By the way, the contents of the wb_wrapper scripts are:
    #!/bin/sh
    # Workbench wrapper
    . /usr/local/fmrib/etc/setup_local
    d=`basename $0`
    cd $FMRIBPLAT/workbench/bin_rh_linux64
    exec ./$d $@

coalsont:
That wrapper script is definitely missing the quotes I mentioned, it needs the last line to be this:

exec ./$d "$@"

I would also say that it is very unfriendly that it does a cd before executing, as this will break on commands that use relative paths for files.

I think this pull request can be closed.

glasserm:
Closed at Tim's Request.

PR #16, opened 2016-03-11
new bias field options for fMRI, improve distortion jacobian support …

coalsont:
…and use by default in most cases


coalsont:
There is some cleanup to do on this, including at least a small change of commands in one script, but it should be fine for testing.

PR #18, opened 2016-03-28
use -cifti-resample to do subcortical parcel cubic resampling

coalsont:
should check that it resolves the variance patterns on resolution change, as the applywarp version did

### DIFF
--- a/fMRISurface/scripts/SubcorticalProcessing.sh
+++ b/fMRISurface/scripts/SubcorticalProcessing.sh
@@ -30,50 +30,50 @@ echo "${script_name}: VolumefMRI: ${VolumefMRI}"
 Sigma=`echo "$SmoothingFWHM / ( 2 * ( sqrt ( 2 * l ( 2 ) ) ) )" | bc -l`
 echo "${script_name}: Sigma: ${Sigma}"
 
-unset POSIXLY_CORRECT
+#NOTE: wmparc has dashes in structure names, which -cifti-create-* won't accept
+#ROIs files have acceptable structure names
 
-if [ 1 -eq `echo "$BrainOrdinatesResolution == $FinalfMRIResolution" | bc -l` ] ; then
-	echo "${script_name}: Doing volume parcel resampling without first applying warp"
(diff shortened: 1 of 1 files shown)


fMRISurface/scripts/SubcorticalProcessing.sh
```diff
@@ -30,50 +30,50 @@ echo "${script_name}: VolumefMRI: ${VolumefMRI}"
...
+then
+    echo "${script_name}: Smoothing and resampling"
+    #this is the whole timeseries, so don't overwrite, in order to allow on-disk writing, then delete temporary
+    ${CARET7DIR}/wb_command -cifti-smoothing ${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii 0 ${Sigma} COLUMN ${ResultsFolder}/${NameOffMRI}_temp_subject_smooth.dtseries.nii -fix-zeros-volume
+    #resample, delete temporary
+    ${CARET7DIR}/wb_command -cifti-resample ${ResultsFolder}/${NameOffMRI}_temp_subject_smooth.dtseries.nii COLUMN ${ResultsFolder}/${NameOffMRI}_temp_template.dlabel.nii COLUMN ADAP_BARY_AREA CUBIC ${ResultsFolder}/${NameOffMRI}_temp_atlas.dtseries.nii -volume-predilate 10
+    rm -f ${ResultsFolder}/${NameOffMRI}_temp_subject_smooth.dtseries.nii
 else
-	echo "${script_name}: Creating subcortical ROI volume in original fMRI resolution"
-	cp "$GrayordinatesSpaceDIR"/Atlas_ROIs."$FinalfMRIResolution".nii.gz "$ROIFolder"/Atlas_ROIs."$FinalfMRIResolution".nii.gz
```
coalsont: Minor difference: my version doesn't require atlas ROIs in the fMRI resolution at all, but Keith's expects one to exist and copies such a file to the subject folder.  This may have been overlooked previously, as we don't often use a fMRI resolution that doesn't have a matching greyordinates space.

coalsont: Major difference: keith's version uses the unset variable $GrayordinatesSpaceDIR (it is used in postfreesurfer, but not in generic surface), which causes keith's version to error.  My version doesn't use this variable, but instead expects the atlas ROIs to have already been copied to the subject folder, like the original script (this copy happens in postfreesurfer, scripts/FreeSurfer2Caret...).

PR #23, opened 2016-06-06
If NHP, use brain-extracted scout for registration to T1w space

cjdonahue:
Addition to fMRIVolume that checks for NHP subject and, in that case, will use brain-extracted scout as opposed to the entire image for registration into T1w space.


cjdonahue:
Fixed broken fieldmap distortion correction.

PR #25, opened 2016-06-07
override locale settings and unset posixly_correct in example setup script

coalsont:


tbbrown:
Made these changes in the current version of Examples/Scripts/SetUpHCPPipeline.sh (which had diverged a bit from it's state when this pull request was originally made)

PR #26, opened 2016-07-05
Fixing small typo "distorction"

hlaubisch:
was "distorction", should probably be "distortion"?


coalsont:
Typo appears to have been fixed during an update on that file (9d3c7700fbb792e13667b81c4b3ae8e81de6304a).

PR #33, opened 2017-04-12
Minor updates to fMRIVolume/OneStepResampling

hfxcarl:
I'm quite new to GitHub, hopefully I'm submitting my suggested improvements correctly. I apologize if not.  I've simply updated the OneStepResampling.sh to have the /postvols/ filenames numbering zeropadded to four places, and changed the bash while-loop into a slightly cleaner for-loop. Very minor details.


glasserm:
This looks fine, though were you having any issues with the original code?

hfxcarl:
The pipelines were working fine for me. I found the minor issue when I was troubleshooting some empty resting-state volumes back through the fMRIVolume PP scripts and found it tricky trying to sort the volumes in the postvols folder. The prevols are number-named better so it seemed an easy fix. After fixing it on my local codebase, I thought it might be useful for everyone. Replacing the while-loop was simply me being trivial and liking the look of the clean, compact for-loop.

glasserm:
I'll let Tim Brown make the all here.  I think the naming improvement is a good idea but he might have to make some changes to his scripts for that.

tbbrown:
The change of the while loop to a three-expression for loop is a very good idea. As long as Matt has no objections to the changing of the postvols file names, then I'm happy with that also.

glasserm:
I am okay with it.

Issue #48, opened 2017-08-14
Pos=AP? or PA?


kfbs:
In DiffusionPreprocessingBatch.7T.sh, 
		PosData="xxxx_AP.nii.gz"
		NegData="xxx_PA.nii.gz"
whereas in DiffPreprocPipeline_PreEddy.sh
		basePos="PA"
		baseNeg="AP"
I assume this does not make any difference for the result, but I want to make sure.
Also, the content of the files created in the Diffusion/rawdata directory does not match those in the unprocessed/Diffusion directory.

glasserm:
Hi All,

There is a ticket below about the 7T diffusion script.  Is it possible that AP/PA mean something different at 3T and 7T because of the other settings sign being flipped?

Matt.

glasserm:
Hi

I cannot remember which convention we used for Pos/Neg as AP/PA.

What is important however is to use as Pos the phase encoding direction that is acquired first in the scan. The final data are concatenated as Pos_Neg prior to eddy, so it makes more sense to use as first volume in this concatenated dataset (and therefore as reference space for distortion correction) the first b0 that is acquired during the scan.

Cheers
Stam


On 14 Aug 2017, at 18:49, Glasser, Matthew <glasserm@wustl.edu<mailto:glasserm@wustl.edu>> wrote:

Hi All,

There is a ticket below about the 7T diffusion script.  Is it possible that AP/PA mean something different at 3T and 7T because of the other settings sign being flipped?

Matt.

glasserm:
We had tried to always make the field map have the correct polarity which does depend on the order of these things.

Matt.

From: Stamatios Sotiropoulos <stamatios.sotiropoulos@ndcn.ox.ac.uk<mailto:stamatios.sotiropoulos@ndcn.ox.ac.uk>>
Date: Monday, August 14, 2017 at 3:01 PM
To: Matt Glasser <glasserm@wustl.edu<mailto:glasserm@wustl.edu>>
[...]

PR #59, opened 2018-02-21
Various changes

mharms:


mharms:
1. Example Batch scripts now use a consistent polarity scheme.
2. Numerous other improvements/updates in the comments in the batch scripts to make them more useful to users.
3. Resolved #48 and clarified some comments in the underlying topup scripts while at it.

PR #70, opened 2018-05-12
Feature/taskfmri analysis combine comment

gcburgess:
1. add and clarify comments for better understanding by user
2. move and consolidate commands into 'functional blocks' of code that perform the same function
3. improve syntax of string comparisons in _if_ statements
     - quote all variables to avoid syntax error if variable is null
     - consolidate multiple tests [ ! "$a" = "$b" ] into single test [ "$a" != "$b" ]
4. modify output scalar names to include "zstat" and "cope"

### DIFF
--- a/TaskfMRIAnalysis/TaskfMRIAnalysis.sh
+++ b/TaskfMRIAnalysis/TaskfMRIAnalysis.sh
@@ -47,28 +47,10 @@ set -e
 # Load function libraries
 source ${HCPPIPEDIR}/global/scripts/log.shlib  # Logging related functions
 source ${HCPPIPEDIR}/global/scripts/opts.shlib # Command line option functions
+source ${HCPPIPEDIR}/global/scripts/fsl_version.shlib	# Function for getting FSL version
 
-# Create utility functions to determine and test FSL versions
-get_fsl_version()
-{
-	local fsl_version_file
-	local fsl_version
-	local __functionResultVar=${1}
 
-	fsl_version_file="${FSLDIR}/etc/fslversion"
-
-	if [ -f ${fsl_version_file} ]
-	then
-		fsl_version=`cat ${fsl_version_file}`
-		log_Msg "INFO: Determined that the FSL version in use is ${fsl_version}"
-	else
-		log_Msg "ERROR: Cannot tell which version of FSL you are using."
-		exit 1
-	fi
-
-	eval $__functionResultVar="'${fsl_version}'"
-}
-
+# function to test FSL versions
 determine_old_or_new_fsl()
 {
 	# NOTE: 
@@ -128,7 +110,7 @@ determine_old_or_new_fsl()
 }
 
 
-########################################## READ COMMAND-LINE ARGUMENTS ##################################
+########################################## READ_ARGS ##################################
 
 # Explcitly set tool name for logging
 log_SetToolName "TaskfMRIAnalysis.sh"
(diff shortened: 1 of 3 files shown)


glasserm:
Hi Greg,

Before I approve this pull request I want to know if it was tested for dense only, dense + volume, and parcellated only and that the code is working, as this is quite a lot of changes.  It looks reasonable but there is enough here in terms of moving things around that I cannot be sure everything works from just looking at it.  

Thanks,

Matt.

gcburgess:
I did make a few changes from the last working version I had. I’ve submitted a few jobs on the NRG cluster, but they haven’t finished yet. I can let you know once they have finished and I’ve double-checked everything.

--Greg

gcburgess:
Yes, I have tested those conditions and they all work. The following will not work 1) CIFTI+Parcellated, 2) CIFTI+Parcellated, and 3) Volume only. Making these work will require (at least) changing arguments passed to the scripts and changing logic statements that determine which analyses will run.

glasserm:
I think that Mike's stuff should probably be split into a separate topic as Greg doesn't have time for it now.  The only time I found it useful to run volume by itself was when I was making some unsmoothed volume analyses to look at denoising effects outside of greymatter.  I did hack my local version to enable this.  In single subjects, unsmoothed volume analyses are always reasonable.  At the group level, that sort of thing should only be used in specialized circumstances (e.g. interested in effects outside of greymatter).  Analyses with volume-based smoothing are never a good idea and that needs to be clearly documented in the pipeline (or ideally would be disabled).  I'm not sure there is high value in parcellated and dense at the same time vs just running the pipeline twice with different settings.

glasserm:
I'm not in favor of renaming the files AllMerged, but am in favor of deleting the single contrasts.

mharms:
Any particular reason (or using just "All")?  We have already established a certain precedent in using "All" to annotate a merged file of some sort (e.g., from the Group data analyses).  Doing the same here would thus be a consistent approach to take.

glasserm:
The path+filenames get too long already and All is reserved for all subjects (e.g. if specific contrasts are concatenated across all subjects).

mharms:
Well, if we eliminate the individual contrasts at least, then there will be less ambiguity about the `*{cope,zstat}_hp*` files to begin with and they won't be as easy to "miss".  So that seems a good intermediate solution.

gcburgess:
The MAIN block in TaskfMRILevel2.sh continues to the end of the script. The larger functional blocks have long strings of #’s, while the smaller functions only have “###".

The term “main” is a convention used in other programming languages where the main function is called first and automatically. Right now, the smaller functions are part of MAIN, which is why the log messages say “MAIN: FUNCTION_NAME”.

Based on other conversations with Tim, I would like to eventually move the smaller functional blocks into separate functions (i.e., separate text scripts). Eventually, those functions will be called by the main script. At that point, the log messages will change to only “FUNCTION_NAME”, because the function’s code won’t be located in MAIN anymore. But until then, they are part of MAIN, making the the multi-level log messages logical.

--Greg

glasserm:
I would prefer we didn't do that, as it makes it much harder for me to follow what is going on.

mharms:
Ok, now that I've had a chance to review your "parcellated,volume" test case, there are some issues with the organization of that output.  Namely, the "volume" analysis gets dumped into a .feat directory with a "ParcellationString" as part of the directory name, and the ensuing *.volume.dscalar.nii files get named with a "ParcellationString" as well.  Neither is appropriate (albeit perhaps the same as how the code functioned previously). I think the optimal solution here is just that any requested volume analysis should go into its own separate "level1vol.feat" and "level2vol.feat" directories, where the volume analysis can live happily by itself (and the "ParcellationString" needs to be removed from its file/directory names).  Alternatively, we could simply prohibit running a VBA concurrently with a parcellated analysis to avoid the current non-sensical naming structure that ensues for the volume data.  i.e., the allowed combinations would be restricted to: (1) dense only, (2) dense+volume, (3) parcellated only.  parcellated+volume would not be a supported combination, and could easily be rejected via some 'if' statements at the front of both the Level1 and Level2 scripts.

Re "MAIN": If "MAIN" really encompasses all of Level1.sh and Level2.sh at this point in time, then I don't see the point.  (If this is confusing me, it is going to confuse others).  Also, if that is the case, then why don't *all* of the log_Msg statements in Level2.sh starting from L64 start with "MAIN"?

glasserm:
Parcellated + Volume was not supposed to be a supported use case.

PR #73, opened 2018-07-05
Optionally run DiffPreprocPipeline for each blip-up/down data pair

MichielCottaar:
When the field is reshimmed between the acquisition of different blip-up/ down diffusion MRI pairs, it might be advantageous to run topup and eddy on each blip-up/down pair separately to correct for any changes in the susceptibility field. Here we add a new script DiffPreprocPipeline_Split.sh, which has a very similar interface to DiffPreprocPipeline.sh. It calls DiffPreprocPipeline.sh separately for each diffusion MRI data pair and then merges the results.

The scripts have been setup in such a way that each call of DiffPreprocPipeline.sh can be run in parallel (i.e., they don't read/write to the same directory). To allow this a minor change was made to DiffusionToStructural.sh .

### DIFF
--- a/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
@@ -115,7 +115,7 @@ get_options()
 				InputNames=${argument#*=}
 				index=$(( index + 1 ))
 				;;
-			--output_dwiname=*)
+			--dwiname=*)
 				OutputName=${argument#*=}
 				index=$(( index + 1 ))
 				;;
@@ -211,7 +211,7 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
   --path=<study-path>     path to subject's data folder
   --subject=<subject-id>  subject ID
   --input_dwinames=<input_names>    @-seperated list of the DWI output directories.
-  [--output_dwiname=<output_name>]
+  [--dwiname=<output_name>]
                           Name of the directory containing the merged diffusion data
                           Defaults to Diffusion
   [--printcom=<print-command>]
--- a/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
@@ -62,7 +62,7 @@
 #
 #~ND~END~
 #
-# Merges the individually processed data pairs into a single diffusion image
+# Merges the individually processed diffusion MRI data into a single diffusion image
 #
 
 # Set up this script such that if any command exits with a non-zero value, the
@@ -170,8 +170,8 @@ validate_environment_vars()
 		error_msgs+="\nERROR: HCPPIPEDIR_dMRI environment variable not set"
 	fi
 
-	if [ ! -e ${HCPPIPEDIR_dMRI}/merge_pairs.sh ] ; then
-		error_msgs+="\nERROR: HCPPIPEDIR_dMRI/merge_pairs.sh not found"
+	if [ ! -e ${HCPPIPEDIR_dMRI}/merge_split.sh ] ; then
+		error_msgs+="\nERROR: HCPPIPEDIR_dMRI/merge_split.sh not found"
 	fi
 
 	if [ -z ${FSLDIR} ] ; then
@@ -247,10 +247,9 @@ main()
     validate_environment_vars
 
     set -e
-    echo -e "\n START: merge_split"
 
-	# Establish tool name for logging
-	log_SetToolName "${SCRIPT_NAME}"
+    # Establish tool name for logging
+    log_SetToolName "${SCRIPT_NAME}"
 
     outdir=${StudyFolder}/${Subject}/T1w/${OutputName}
 
@@ -259,7 +258,7 @@ main()
 
     log_Msg "merging ${indirs} into ${outdir}"
 
-	${runcmd} ${HCPPIPEDIR_dMRI}/merge_pairs.sh ${outdir} ${indirs}
+    ${runcmd} ${HCPPIPEDIR_dMRI}/merge_split.sh ${outdir} ${indirs}
 }
 
 
--- a/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
@@ -29,14 +29,14 @@
 #
 # ## Description
 #
-# This script, <code>DiffPreprocPipeline_Eddy.sh</code>, implements the second
-# part of the Preprocessing Pipeline for diffusion MRI describe in
-# [Glasser et al. 2013][GlasserEtAl]. The entire Preprocessing Pipeline for
-# diffusion MRI is split into pre-eddy, eddy, and post-eddy scripts so that
-# the running of eddy processing can be submitted to a cluster scheduler to
-# take advantage of running on a set of GPUs without forcing the entire diffusion
-# preprocessing to occur on a GPU enabled system.  This particular script
-# implements the eddy part of the diffusion preprocessing.
+# This script, <code>DiffPreprocPipeline_Merge.sh</code>, merges the output
+# of the Preprocessing Pipeline for diffusion MRI as described in
+# [Glasser et al. 2013][GlasserEtAl]. This script will usually
+# not be called as part of the diffusion preprocessing pipeline. However,
+# if the scanner reshimmed between the acquisition of different parts of
+# the diffusion MRI data, it can be advantageous to run the pipeline on
+# each shim group separately.
+# This script merges the output for each shim group in a single dataset.
 #
 # ## Prerequisite Installed Software for the entire Diffusion Preprocessing Pipeline
 #
@@ -50,7 +50,7 @@
 #
 # ## Prerequisite Environment Variables
 #
-# See output of usage function: e.g. <code>$ ./DiffPreprocPipeline_Eddy.sh --help</code>
+# See output of usage function: e.g. <code>$ ./DiffPreprocPipeline_Merge.sh --help</code>
 #
 # <!-- References -->
 #
--- a/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
@@ -1,0 +1,266 @@
+#!/usr/bin/env bash
+#!/bin/bash
+#~ND~FORMAT~MARKDOWN~
+#~ND~START~
+#
+# # DiffPreprocPipeline_Merge.sh
+#
+# ## Copyright Notice
+#
+# Copyright (C) 2012-2016 The Human Connectome Project
+#
+# * Washington University in St. Louis
+# * University of Minnesota
+# * Oxford University
+#
+# ## Author(s)
+#
+# * Michiel Cottaar, FMRIB Analysis Group, Oxford University
+# * Matthew F. Glasser, Department of Anatomy and Neurobiology, Washington University in St. Louis
+# * Michael Harms, Department of Anatomy and Neurobiology, Washington University in St. Louis
+#
+# ## Product
+#
+# [Human Connectome Project][HCP] (HCP) Pipelines
+#
+# ## License
+#
+# See the [LICENSE](https://github.com/Washington-University/Pipelines/blob/master/LICENSE.md) file
+#
+# ## Description
+#
+# This script, <code>DiffPreprocPipeline_Eddy.sh</code>, implements the second
+# part of the Preprocessing Pipeline for diffusion MRI describe in
+# [Glasser et al. 2013][GlasserEtAl]. The entire Preprocessing Pipeline for
+# diffusion MRI is split into pre-eddy, eddy, and post-eddy scripts so that
+# the running of eddy processing can be submitted to a cluster scheduler to
+# take advantage of running on a set of GPUs without forcing the entire diffusion
+# preprocessing to occur on a GPU enabled system.  This particular script
+# implements the eddy part of the diffusion preprocessing.
+#
+# ## Prerequisite Installed Software for the entire Diffusion Preprocessing Pipeline
+#
+# * [FSL][FSL] - FMRIB's Software Library (version 5.0.11)
+#
+#   FSL's environment setup script must also be sourced
+#
+# * [FreeSurfer][FreeSurfer] (version 5.3.0-HCP)
+#
+# * [HCP-gradunwarp][HCP-gradunwarp] - (HCP version 1.0.2)
+#
+# ## Prerequisite Environment Variables
+#
+# See output of usage function: e.g. <code>$ ./DiffPreprocPipeline_Eddy.sh --help</code>
+#
+# <!-- References -->
+#
+# [HCP]: http://www.humanconnectome.org
+# [GlasserEtAl]: http://www.ncbi.nlm.nih.gov/pubmed/23668970
+# [FSL]: http://fsl.fmrib.ox.ac.uk
+# [FreeSurfer]: http://freesurfer.net
+# [HCP-gradunwarp]: https://github.com/Washington-University/gradunwarp/releases
+#
+#~ND~END~
+#
+# Merges the individually processed data pairs into a single diffusion image
+#
+
+# Set up this script such that if any command exits with a non-zero value, the
+# script itself exits and does not attempt any further processing.
+set -e
+
+# Load Function Libraries
+source ${HCPPIPEDIR}/global/scripts/log.shlib     # log_ functions
+source ${HCPPIPEDIR}/global/scripts/version.shlib # version_ functions
+
+get_options()
+{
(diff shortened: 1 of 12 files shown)


MichielCottaar:
Well spotted on the use of -Tmean in grad_dev.nii.gz (I did not include the gradient non-linearities when testing the pipeline, so I completely missed that). I have adjusted how the grad_dev and nodif_brain_mask are averaged, so that they no longer rely on -Tmean.

On the version number: I don't think I use any new FSL features, so I just kept the same version number as in the DiffPreprocPipeline.sh . Do we want to update that version number as well?

MichielCottaar:
Do you require anything more from my side?

glasserm:
Might as well update the version number.  For some reason I didn't get your first comment.  Also would like Mike to look this over.

DiffusionPreprocessing/DiffPreprocPipeline_Split.sh
```diff
@@ -0,0 +1,385 @@
...
+	# report
+	echo "-- ${SCRIPT_NAME}: Environment Variables Used - Start --"
+	echo "   HCPPIPEDIR_dMRI: ${HCPPIPEDIR_dMRI}"
+	echo "   FSLDIR: ${FSLDIR}"
+	echo "-- ${SCRIPT_NAME}: Environment Variables Used - End --"
+}
+
+#
+# Function description
+#  Main processing of script if --split is set
```
mharms: A `--split` flag is mentioned in several places, but I don't see anything that actually checks for it or uses it as part of any conditionals.

DiffusionPreprocessing/DiffPreprocPipeline_Split.sh
```diff
@@ -0,0 +1,385 @@
...
+
+	# verify positive and negative datasets are provided in pairs
+	if [ ${Pos_count} -ne ${Neg_count} ] ; then
+		log_Msg "Wrong number of input datasets! Make sure that you provide pairs of input filenames."
+		exit 1
+	fi
+
+	# verify that none of the files are empty
+	for Image in ${PosInputImages} ; do
+		if [ ${Image} = EMPTY ] ; then
```
mharms: Does `EMPTY` have a predefined meaning in bash?  I couldn't find anything on this construction on the web...

DiffusionPreprocessing/DiffPreprocPipeline_Split.sh
```diff
@@ -365,7 +365,7 @@ main()
 		preproc_pipeline_cmd+=" --subject=${Subject} "
 		preproc_pipeline_cmd+="${PassOnArguments}"
 		log_Msg "Invoking Preprocessing pipeline for positive/negative data pair ${pair}"
-		${preproc_pipeline_cmd}
+		echo ${preproc_pipeline_cmd}
```
mharms: By using 'echo' here, nothing will be actually run.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -79,8 +79,9 @@ ${FSLDIR}/bin/imcp "$WorkingDirectory"/"$regimg"2T1w_restore "$RegOutput"
 ${FSLDIR}/bin/fslmaths "$T1wRestoreImage".nii.gz -mul "$WorkingDirectory"/"$regimg"2T1w_restore.nii.gz -sqrt "$QAImage"_"$regimg".nii.gz
 
 #Generate 1.25mm structural space for resampling the diffusion data into
-${FSLDIR}/bin/flirt -interp spline -in "$T1wRestoreImage" -ref "$T1wRestoreImage" -applyisoxfm ${DiffRes} -out "$T1wRestoreImage"_${DiffRes}
-${FSLDIR}/bin/applywarp --rel --interp=spline -i "$T1wRestoreImage" -r "$T1wRestoreImage"_${DiffRes} -o "$T1wRestoreImage"_${DiffRes}
+${FSLDIR}/bin/flirt -interp spline -in "$T1wRestoreImage" -ref "$T1wRestoreImage" -applyisoxfm ${DiffRes} -out "$WorkingDirectory"/T1w_restore_acpc_dc_${DiffRes}
```
mharms: Why hard code name of the `*_${DiffRes}` file names, rather than use `$T1wRestoreImage`?

MichielCottaar:
Thanks for these comments. I like your idea of using new DataShimGroup flags to handle this issue. I'll adjust the code to handle this.

On the registration issue: in the postEddy script all ShimGroups get aligned to the T1w (using BBR registration), so in the final output they will be aligned with each other. However, aligning the diffusion data with the T1w is likely to be less precise than any direct registration between the diffusion data itself. So a slightly better registration pipeline might be:
1. Run flirt to find the affine matrix that maps each ShimGroup to the first ShimGroup (this is within one modality, so it's probably better not to use the BBR cost function)
2. Use BBR registration to find the affine matrix from the first ShimGroup to T1w space
3. Apply the correct combination of affine matrices to transform each ShimGroup to T1w space and merge them
This has the nice properties that we register the diffusion data internally rather than relying on separate registration to T1w and that the diffusion data is only interpolated once to get to T1w space. Making these changes would take a while, as it would involve quite involved changes in the postEddy script. Does this sound like a reasonable approach @mharms @glasserm?

glasserm:
I think that is reasonable.  Not sure if BBR is accurate enough for different shim groups or not.

mharms:
I reviewed our email train from April 2018, and I see at that time that I had suggested that 

> we basically plan on calling DiffusionPreprocessing multiple times and then just write a separate module for combining at the end (as opposed to building directly into DiffusionPreprocessing itself the ability to parse its inputs into separate topup/eddy runs).

which of course is not what I suggested in this code review above. :)  Reflecting further, I'm wondering if my proposal back in April isn't the best way to go in terms of keeping it simple. Namely, rather than trying to do everything via ShimGroup info added into DiffPreprocPipeline.sh, simply run DiffPreprocPipeline.sh multiple times with the posData and negData inputs limited to a single ShimGroup per run of DiffPreprocPipeline.sh (and a unique DWIName would obviously need to be provided for each run, but that capability already exists).

Under this model, all that would be needed would be a script that could be called separately (e.g., DiffPreprocPipeline_Merge.sh) to handle the final registration and merging of the individual runs of DiffPreprocPipeline.sh.

I like this approach because it seems very modular, and would accomplish what we need in what seems to me the easiest possible manner.

MichielCottaar:
This is closer to what I have been doing. The main point from DiffPreprocPipeline_Split.sh was to call DiffPreprocPipeline.sh multiple times for each diffusion data pair and then to call scripts/merge_pairs.sh to merge all the diffusion data after it has been registered to T1w space (and hence roughly aligned).

If we are happy with registering the different ShimGroups to T1w independently, then everything I would need to change to get to your solution is:
1. Get rid of DiffPreprocPipeline_Split.sh (or adjust it so it takes arbitrary shim groups using the DataShimGroup you suggested earlier)
2. Adjust scripts/merge_pairs.sh so that it accepts the DWIName chosen by the user (and rename it to DiffPreprocPipeline_Merge.sh).
This would definitely be the easier solution. I suspect that independent registrations to T1 should be good enough, so I would be happy to go with this solution.

The more difficult route would be to register the different shim groups to the average space and only then to T1 space. In that case the whole post-eddy part will have to change, because we don't want to register each ShimGroup to the T1w independently.
1. Add a flag to DiffPreprocPipeline.sh which causes it to skip the post-eddy part to be skipped.
2. Incorporate the parts of the PostEddy pipeline not involved with registration to T1 space into a new merge script
3. Register the different ShimGroups to an average diffusion space
4. Find the affine transformation from this average diffusion space to T1 space
5. Transform the different ShimGroups in one step into T1 space
This would involve a lot more changes to the code and also be harder to maintain in the future (because part of the PostEddy analysis is now replicated in both the original PostEddy script for a single ShimGroup and in the Merge script for multiple ShimGroups).

MichielCottaar:
I'm testing it now. I'll let you know afterwards. I'll also still have to remove DiffPreprocPipeline_Split.sh or adjust it according to Mike's earlier suggestion (i.e., accept `--posDataShimGroup=A@A@B` and `--negDataShimGroup=A@B`).

MichielCottaar:
It seems to be working now. As I discuss in the accomponying email, I tested it on a dataset with as far as I know no reshimming (103818). Treating dir97 separately from dir95 and dir96 did slightly decrease the correlation between their b0's and mean diffusion-weighted images, but this was a weak trend, that is barely visible if flipping through the images.

I think we still need decisions on the following points:
- I kept `DiffPreprocPipeline_Split.sh`, which accepts flags like `--posDataShimGroup=A@A@B` and `--negDataShimGroup=A@B` and then runs the regular diffusion preprocessing piplines followed by the merging script. Would you prefer to keep this or would you rather just run the diffusion preprocessing pipeline manually? An example of the former option is available at https://git.fmrib.ox.ac.uk/ndcn0236/pipeline_split/blob/master/run_single.sh, an example of the latter option at https://git.fmrib.ox.ac.uk/ndcn0236/pipeline_split/blob/master/run_manual.sh .
- Would you prefer the code that updates the registration to their mean space? It is available at https://github.com/MichielCottaar/Pipelines/tree/register_shim , but as there it greatly complicates the code and seems to have no benefit, I would not recommend it.
- Should we test this on a subject with actual reshimming before merging?

mharms:
I believe that above you were intending to to refer to https://git.fmrib.ox.ac.uk/ndcn0236/pipeline_split/blob/master/run_split.sh (rather than "run_single.sh"), right?
Did you compare the "run_split.sh" and "run_manual.sh" outputs, and confirm that they were identical?

In terms of whether we should favor the more complicated code that updates the registration to the mean space, since that code is in fact already completed, we should let the decision be driven by the data.  For that, we probably need to compare a couple datasets that actually have different shims between the different dMRI pairs.  Do those examples need to come from HCP-YA (for which structural preprocessing has already been run), or could they come from the Lifespan HCD/HCA projects (for which structural preprocessing would probably need to be run)? The latter would be easier for me to track down some examples, and also a more relevant test, since it is the HCD/HCA projects for which we are actually considering applying this feature.

DiffusionPreprocessing/DiffPreprocPipeline_Split.sh
mharms: This mention of a --split flag/option should be removed.

DiffusionPreprocessing/DiffPreprocPipeline_Merge.sh
```diff
@@ -0,0 +1,265 @@
...
+#
+# [Human Connectome Project][HCP] (HCP) Pipelines
+#
+# ## License
+#
+# See the [LICENSE](https://github.com/Washington-University/Pipelines/blob/master/LICENSE.md) file
+#
+# ## Description
+#
+# This script, <code>DiffPreprocPipeline_Eddy.sh</code>, implements the second
```
mharms: This section needs to be revised to be reflective of the purpose/role of this particular script.

DiffusionPreprocessing/DiffPreprocPipeline_Split.sh
```diff
@@ -0,0 +1,390 @@
+#!/bin/bash
+#~ND~FORMAT~MARKDOWN~
+#~ND~START~
+#
+# # DiffPreprocPipeline.sh
```
mharms: Script name here, and elsewhere in this section needs to be updated.

mharms: In fact, this entire markdown section probably needs to be updated to be relevant for this specific script.

DiffusionPreprocessing/DiffPreprocPipeline_Split.sh
```diff
@@ -0,0 +1,390 @@
...
+
+	# Validate environment variables
+	validate_environment_vars $@
+
+	# Establish tool name for logging
+	log_SetToolName "${SCRIPT_NAME}"
+
+
+    input_dwinames=""
+    while read -r line ; do
```
mharms: Comment preceding this section would be helpful.
i.e., that "while read -r line" is being feed the output "split_shimgroups.py" as input lines.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -79,8 +79,9 @@ ${FSLDIR}/bin/imcp "$WorkingDirectory"/"$regimg"2T1w_restore "$RegOutput"
 ${FSLDIR}/bin/fslmaths "$T1wRestoreImage".nii.gz -mul "$WorkingDirectory"/"$regimg"2T1w_restore.nii.gz -sqrt "$QAImage"_"$regimg".nii.gz
 
 #Generate 1.25mm structural space for resampling the diffusion data into
-${FSLDIR}/bin/flirt -interp spline -in "$T1wRestoreImage" -ref "$T1wRestoreImage" -applyisoxfm ${DiffRes} -out "$T1wRestoreImage"_${DiffRes}
-${FSLDIR}/bin/applywarp --rel --interp=spline -i "$T1wRestoreImage" -r "$T1wRestoreImage"_${DiffRes} -o "$T1wRestoreImage"_${DiffRes}
+${FSLDIR}/bin/flirt -interp spline -in "$T1wRestoreImage" -ref "$T1wRestoreImage" -applyisoxfm ${DiffRes} -out "$WorkingDirectory"/T1w_acpc_dc_restore_${DiffRes}
```
mharms: What is the need for the changes here?  Also, even if the output is going into the "$WorkingDirectory" we shouldn't be hard-coding the file name as "T1w_acpc_dc_restore".  It should respect/use the variables defined internally in the script.

glasserm: We don't want this change.  The T1w volume at DWI resolution is an expected packaged output.  It needs to be resampled with splines which did not originally work when using flirt and -applyisoxfm (hence the need for two lines), but perhaps that bug has since been solved.  

I wasn't expecting changes to the registration code (which was most of what I wrote for this pipeline).  Are there more changes here?

mharms: @glasserm  I think the way that I inserted my comment gave you an incomplete view of the overall change proposed by @MichielCottaar.  You should look at the full set of proposed changes, which I still don't understand the purpose of.  There is still an 'applywarp', but the output is going into the "WorkingDirectory", only to get immediately moved to the location/name of the file created in the original version.

glasserm: Okay, so then I also don't understand the purpose of that.

glasserm:
Perhaps we should require FSL 6.0.1?

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: This change was made, so that you can safely run the diffusion preprocessing pipeline in parallel for the different shims before merging the output. For all the rest of the preprocessing code it is irrelevant whether the preprocessing pipeline for the different shims are run in sequence or in parallel, because the have different $WorkingDirectory. However, in this part of the code the script writes to the general T1w directory and hence the separate runs of the DiffusionToStructural.sh will actually write to the same file. 

In the original version $T1wRestoreImage temporarily contains the result from the flirt command and then the results from the applywarp command. If the preprocessing pipelines for the different shims are run in parallel, this means that at no point in the script can we be certain which of these files is actually stored on disk (as the other parallel script might at any point run the first flirt command). The new version ensures that the file stores on disk is always the result from the applywarp command, whether the scripts are run in parallel or sequence.

In practice this is very unlikely to lead to a bug, but if it does it will be very hard to find the origin of the bug (as it will depend on the timings of parallel scripts and hence not be reproducible). So it seemed like a good idea to make these changes to ensure that this problem never arises...

MichielCottaar:
Thanks for the comments. I'll carefully check the comments sections again and make the suggested changes. 

It would make sense to me to require 6.0.1, especially if we are going to add the susceptibility-by-motion soon.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
glasserm: Ok, my concern was indeed because I did not see that the file was still being made.

glasserm:
Sounds good.

MichielCottaar:
I've made the requested changes to the script's documentation. Let me know once you have a dataset with reshimming on which I can test the code.

MichielCottaar:
I closed this without merging because this pull requested added a lot of complexity to the pipeline without any significant gains. The reason that the gains of implementing this approach are neglible are:
1. Any difference in the susceptibility field is partially already corrected for by susceptibility-by-motion.
2. It does not address the major issue of combining two diffusion datasets of the same subject, which is not the differences in distortions due to a different susceptibility field, but the difference in the bias field.

Potentially, this issue could be revisited once the bias field has been corrected for (either using FAST on the b0 or ideally by inclusion of bias-field estimation in eddy).

mharms:
@MichielCottaar Thanks for all the work you put into this, even though it didn't get incorporated at this time.

PR #74, opened 2018-07-12
Add commented out line indicating what TopUpConfig should be set to

mharms:


glasserm:
Was thinking of another pipeline.

PR #78, opened 2018-10-24
Necessary updates to ICAFIX

mharms:
The contents of ICAFIX don't work currently for general users outside of WUSTL. This PR restores ICAFIX to a state that should work for outside users, and includes some other incidental cleanup related to ICAFIX).


glasserm:
This will need to be reviewed by Tim Coalson who may have made some similar changes and is working on these issues.

coalsont:
The edits to hcp_fix will conflict, except for the T2 image test.  The melodic location edit should copy from hcp_fix_multi_run, which first checks the PATH.

We can leave this PR open until I push, then see what needs to be removed from this PR.  Or, you can separate the hcp_fix edits elsewhere (I haven't touched the ones marked for older versions, so they can stay) and go ahead with the rest.

mharms:
Can you put your version in a branch, so I can see how it conflicts?  What I know is that this version actually _works_.

glasserm:
Tim's version will be fully tested as well.

coalsont:
See here for a snapshot of my current changes to (only) hcp_fix:

https://github.com/coalsont/Pipelines/commit/f36ee6d78a7232f475041f82e228ff0b9c270c34

glasserm:
Maybe you can ask Steve about that?

coalsont:
Here the main places that will conflict when merging my edits with yours:

https://github.com/coalsont/Pipelines/commit/f36ee6d78a7232f475041f82e228ff0b9c270c34#diff-7d77125a77398e8608290ef9fb58d745R114
https://github.com/coalsont/Pipelines/commit/f36ee6d78a7232f475041f82e228ff0b9c270c34#diff-7d77125a77398e8608290ef9fb58d745R218

mharms:
Won’t those just get handled automatically?  Will there actually be a conflict that prevents a fast-forward merge?

--
Michael Harms, Ph.D.
-----------------------------------------------------------
Associate Professor of Psychiatry
Washington University School of Medicine
Department of Psychiatry, Box 8134
660 South Euclid Ave.                        Tel: 314-747-6173
St. Louis, MO  63110                          Email: mharms@wustl.edu

coalsont:
No, we both edited the same lines from the same version, to result in something different, there is no possible way for git to know what to do with whatever set of edits gets applied second to those same lines, because the starting point won't match what it expects for the second set.  Fast-forward is impossible any time that the commits don't already form a perfect chain, and a merge conflict is more problematic than just "not fast forward" - it requires someone to manually edit the thing to whatever "should" be there in the end.

mharms:
So, there isn't an argument to tell git to treat your version as "primary" for any conflicting lines?

coalsont:
If you know in advance that there will be conflicts, and think that "ignore/revert everybody else's edits to that area" is a good idea, yes you can tell git to do that (however, it looks like in order to get git to actually try something intelligent on other conflicting files, you have to specify multiple strategies).  That isn't the usual procedure (for good reason), and doesn't mean there aren't conflicts, it just tells git to forcibly "resolve" all conflicts a certain way (which may revert some desirable edits).

There is really no substitute for actually looking at the conflicts.

coalsont:
If you think it is urgent to change hcp_fix in master, then this can be merged for now and I can sort out the conflicts before I push.

glasserm:
Tim can you see what of these changes should get merged?  I think some of them are out of date, but others, e.g. to the settings and example batch script might be useful.

coalsont:
Rebasing this branch on the new master is the easiest way to look for all conflicts.

coalsont:
I have done a git rebase and resolved the conflicts as I think they should be resolved, see here:

https://github.com/coalsont/Pipelines/tree/pull-78

If Mike thinks it looks okay, we can push it, or Mike can overwrite his master with it to update this PR for further review.

glasserm:
There is a version number for FIX that should be updated.  Also am not sure what these files are:

https://github.com/Washington-University/HCPpipelines/blob/master/ICAFIX/hcp_fix.for_fix1.062
https://github.com/Washington-University/HCPpipelines/blob/master/ICAFIX/hcp_fix.for_fix1.06a
https://github.com/Washington-University/HCPpipelines/blob/master/ICAFIX/hcp_fix_multi_run.original
https://github.com/Washington-University/HCPpipelines/blob/master/ICAFIX/settings.sh.WUSTL_CHPC2

coalsont:
I don't know what file/line "a version number for FIX that should be updated" refers to.

glasserm:
Line 4 of the settings file.

mharms:
This seems good, although I have a couple suggestions:
1) Add comments at the top, indicating how this version of hcp_fix differs from ${FSL_FIXDIR}/hcp_fix, to help users understand the differences.  In particular:

- Includes more verbose output.
- Adds a flag to control whether motion regression is applied (and appropriate conditional code based on that choice).
- Defaults to using whatever version of 'melodic' is primary in $PATH, rather than the $FSLDIR/bin version.
- Removes the restriction of 'melodic' to 250 dimensions.
- Adds flexibility for user to specify their own training data file.
- Replaces the brainmask used as part of deriving the 'veins' image from one derived from $FSL_FIXDIR/mask_files/hcp_0.7mm_brain_mask to the subject's specific T1w/brainmask_fs.nii.gz.

2) I'm not overly familiar with the ((...)) syntax, but for the two "if (( hp < 0 ))" lines, can we write that as "if (( $hp < 0 ))" instead?  (I think both work, but the latter is more visually consistent with the notion that hp is an existing variable).

3) Do we need the ICAFIX/hcp_fix.for_fix1.062 and ICAFIX/hcp_fix_multi_run.original files, or can those be deleted?

And one questions:
Are there any potential implications in switching the origin of the brainmask used for the 'veins' image calculation?  In particular, perhaps I'm wrong, but I seem to recall that the subject specific brainmask_fs.nii.gz sometimes include holes inside the mask?

coalsont:
The ((...)) syntax is integer math expressions, a feature built into bash (use the manpage), everything in them that isn't a number or a math operator is a variable, by elimination.  I am not inclined to add a superfluous $ to them, because math expressions are easier to read without extra symbols that just mean "this is a variable".  I don't particularly like the established convention of ${var} either, the form "$var" is actually more useful, because it is whitespace-safe.

I don't have much of an opinion on the other points.

mharms:
In that case, how aboue we just add the bullets/comments that I suggested (or your favored rewording of them) to your version.  I'd be happy to do it myself, but that will just lead to another PR.

Regarding the use of the ((...)) syntax, what you have is perfectly valid.  That said, it is confusing to novices when code uses the bracket form of conditionals in some places and the ((...)) form in others.  Unless we are trying to force novice bash scripters to learn different but analogous syntactic constructs, why introduce a new variant?  That's my $0.02 on that issue.

coalsont:
If you pull the branch replacing your fork's master, this PR will change to reflect the changes I made while rebasing.  Depending on what you did when making the PR, I might be able to do this for you.

On the flip side, why should we adopt conventions that teach novice bash coders bad habits?  Following your logic further, we shouldn't even attempt to be whitespace safe anywhere, because otherwise people editing the pipelines would have to learn about quotes, and that bash has arrays.  The ((...)) syntax is simpler and more readable for integer math expressions, and is a good habit as long as the script is intended only for bash.  The [[...]] syntax is less specialized, and therefore using it for math expressions is more awkward.  The [...] form is legacy sh convention, and is a bad habit for bash-specific scripts because it is treated like an executable, not as shell syntax ("[[ $a || $b ]]" works, "[ $a || $b ]" is a cryptic syntax error).  I want to put good practice and good habits first.

mharms:
For the record regarding my earlier question, I see that `FreeSurfer2CaretConvertAndRegisterNonlinear.sh` is in fact set up to ensure that brainmask_fs.nii.gz doesn't contain any "holes", so using that for the 'veins' image calculation should have no practical implications on the derivation of the veins image (even if it isn't strictly identical to the previous approach of resampling $FSL_FIXDIR/mask_files/hcp_0.7mm_brain_mask)

PR #79, opened 2018-11-15
BUG: run bet on a single B0 image rather than the full diffusion data

MichielCottaar:
In FSL 6 running bet on a 4D diffusion MRI dataset no longer only loads the first volume,
but actually produces a 4D file with the first volume containing a mangled mask


glasserm:
Thanks Michiel.

PR #81, opened 2018-11-30
Some extra cleanup after pull #78

coalsont:
The hp 0 logic from _multi_run shouldn't be used in selecting training data in hcp_fix, because hcp_fix doesn't have a detrend-only option.  Also, I think Matt wanted the fix version number to actually be set to the latest.

I don't remember if there were other things in the fix stuff.


mharms:
Can we get a function description and descriptions of the input parameters on `functionhighpassandvariancenormalize` and its associated helper functions (esp. `icaDim`)?

PR #82, opened 2018-11-30
Revert "Some extra cleanup after pull #78"

mharms:
Reverts Washington-University/HCPpipelines#81
I believe what we had previously was correct and preferable.

### DIFF
--- a/ICAFIX/hcp_fix
+++ b/ICAFIX/hcp_fix
@@ -230,19 +230,27 @@ else
 	# user has not specified a training data file
 	#TSC: so, let's look for it and give a useful error, rather than pretending they didn't want to do what they said
 	#TSC: if you WANT to use, say, hp150 with hp2000 training, it should be explicitly requested by the user
+	#TSC: exception: we will recommend hp 0 (linear) for performance, and hp 2000 is already trained, and largely indistinguishable
+	training_hp="${hp}"
 	if [[ ! -f "${FSL_FIXDIR}/training_files/HCP_hp${hp}.RData" ]]
 	then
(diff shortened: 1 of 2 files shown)


coalsont:
Using hp 0 in hcp_fix does not add a _hp0 onto the filename, so it won't do what is desired for a linear detrend (wrong filenames at least).  fix itself is only passed a value for highpass in the -h subargument to -m (so probably only applied to the motion regressors), so when we don't tell fix to do motion correction, it can't even know what the highpass value actually is.  I don't know if fix always does a detrend anyway.

I don't know what the version thing is about (though it does make fix itself claim to be version 1.06 in log files), and our reapply launches the fix_3_clean.m matlab code directly, which doesn't look like it contains the check that would produce the warning.  I think it only matters between subsequent calls to fix itself that use the -f, -c, and -a options (hcp_fix uses the no-switch form that automatically does all three stages).

mharms:
The check for FIXVERSION is part of the `fix` script itself, and the warning message that I was referencing _will_ get envoked.  I'm pretty sure that not including the minor version as part of FIXVERSION was intentional on Steve's part.

I see the issue now with hp=0 in the context of hcp_fix.  Perhaps we can add that functionality to hcp_fix later.

glasserm:
Regarding this: it seems that hp=0 should not be used with hcp_fix because the data do not get detrended before running ICA.  As a result, the ICA component timeseries never get detrended and will not match the motion regressors and data in cleanup which will get detrended if hp=0.   This is in addition to the file naming issues that Tim brought up.

PR #83, opened 2018-12-07
Minor tweaks to FIX-related stuff

mharms:


mharms:
I think I’ve resolved the naming clash issue.  I tried to be careful, but if Tim could test it, that would be great.  I also went through and implemented a series of readability improvements (e.g., reducing the use of embedded sub-shell commands considerably; switching to a cleaner way to add to an existing string).  There is one place where I’ve suggested “ADD COMMENT”.

Assuming this works, here are some other things we should consider before we position MR-FIX as "ready to go":
1) In functionhighpassandvariancenormalize.m, shouldn't the NIFTI and CIFTI VN *maps* include the "hp" string as part of their name (if hp>=0)?  [Changing that will require corresponding changes to hcp_fix_multi_run, but now that we have _vnts and _vn as distinct filenames, it should no longer be necessarily to artificially exclude "hp" from the VN map name just to avoid file name collisions (which is, I assume, the reason it wasn't included previously)].  [If you are on-board with that change, I should do that before Tim runs his test].
2) functionhighpassandvariancenormalize.m should have a description and usage added.
3) As of yet, there is no removal of the demeaned and VN'ed time series files (both CIFTI and NIFTI), which is presumably a waste of space.

ICAFIX/functionhighpassandvariancenormalize.m
```diff
@@ -93,16 +95,19 @@
...
-        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_hp' num2str(hp) '_vn.nii.gz'],'f',[1 1 1 1]); 
-        call_fsl(['fslcpgeom ' fmri '.nii.gz ' fmri '_hp' num2str(hp) '_vn.nii.gz -d']); 
+        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_hp' num2str(hp) '_vnts.nii.gz'],'f',[1 1 1 1]); 
+        call_fsl(['fslcpgeom ' fmri '.nii.gz ' fmri '_hp' num2str(hp) '_vnts.nii.gz -d']); 
     else
-        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_vnf.nii.gz'],'f',[1 1 1 1]); 
-        call_fsl(['fslmaths ' fmri '.nii.gz -sub ' fmri '.nii.gz -add ' fmri '_vnf.nii.gz ' fmri '_vnf.nii.gz']); 
-        %call_fsl(['fslcpgeom ' fmri '.nii.gz ' fmri '_vnf.nii.gz -d']); 
+        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_vnts.nii.gz'],'f',[1 1 1 1]);
+        % Following use of 'fslmaths' rather than 'fslcpgeom' is a hack due to FSL 6.0.0 version of the latter not being able to handle very large files.
```
coalsont: This comment is wrong, it is fsl 5 that has a problem with large files.  The problem in fsl 6.0.0 is in not setting all the header fields it is supposed to.

mharms: In that case, why aren't all the other instances of 'fslcpgeom' in that script problematic as well?

coalsont: Some of them probably were, but didn't get tested on as large of datasets.

coalsont: As for the fsl 6.0.0 bug, we have a patched version of fslcpgeom.

mharms: Do you know/recall which specific header fields aren't getting set correctly?

coalsont: Specifically, it only looks at the reference file's qform, and sets only the target file's sform.  There is a patched version they sent us privately.

ICAFIX/functionhighpassandvariancenormalize.m
```diff
@@ -76,6 +77,7 @@
 end
 
 %Save VN
+%% MPH: Shouldn't these VN maps in the hp>=0 condition have 'hp' as part of their file name???
```
glasserm: Yes that would be fine if you want to change it

ICAFIX/functionhighpassandvariancenormalize.m
```diff
@@ -93,16 +95,19 @@
...
-        call_fsl(['fslcpgeom ' fmri '.nii.gz ' fmri '_hp' num2str(hp) '_vn.nii.gz -d']); 
+        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_hp' num2str(hp) '_vnts.nii.gz'],'f',[1 1 1 1]); 
+        call_fsl(['fslcpgeom ' fmri '.nii.gz ' fmri '_hp' num2str(hp) '_vnts.nii.gz -d']); 
     else
-        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_vnf.nii.gz'],'f',[1 1 1 1]); 
-        call_fsl(['fslmaths ' fmri '.nii.gz -sub ' fmri '.nii.gz -add ' fmri '_vnf.nii.gz ' fmri '_vnf.nii.gz']); 
-        %call_fsl(['fslcpgeom ' fmri '.nii.gz ' fmri '_vnf.nii.gz -d']); 
+        save_avw(reshape(cts,ctsX,ctsY,ctsZ,ctsT),[fmri '_vnts.nii.gz'],'f',[1 1 1 1]);
+        % Following use of 'fslmaths' rather than 'fslcpgeom' is a hack due to FSL 6.0.0 version of the latter not being able to handle very large files.
+        call_fsl(['fslmaths ' fmri '.nii.gz -sub ' fmri '.nii.gz -add ' fmri '_vnts.nii.gz ' fmri '_vnts.nii.gz']);
```
glasserm: We don't need this hack with FSL 6.0 any more so can switch back to the commented out line

ICAFIX/hcp_fix_multi_run
```diff
@@ -212,47 +221,30 @@ for fmri in $fmris ; do
...
+	#Demean movement regressors, volumes, and CIFTI data
+	demeanMovementRegressors Movement_Regressors.txt Movement_Regressors_demean.txt
+	MovementTXTMergeSTRING+="$(pwd)/Movement_Regressors_demean.txt "
+
 	${FSLDIR}/bin/fslmaths $fmri -Tmean ${fmri}_mean
 	${FSLDIR}/bin/fslmaths $fmri -sub ${fmri}_mean ${fmri}_demean
 
-	demeanMovementRegressors Movement_Regressors.txt Movement_Regressors_demean.txt
-	MovementTXTMergeSTRING=`echo "${MovementTXTMergeSTRING}$(pwd)/Movement_Regressors_demean.txt "`
+	${FSL_FIX_WBC} -cifti-reduce ${fmri}_Atlas.dtseries.nii MEAN ${fmri}_Atlas_mean.dscalar.nii
```
glasserm: needs to be ${fmriNoExt} like above instead of ${fmri}

ICAFIX/hcp_fix_multi_run
```diff
@@ -212,47 +221,30 @@ for fmri in $fmris ; do
...
-  
-  log_Inform "Dims: $(cat ${fmri}_dims.txt)"
-  
-	fslmaths $(pwd)/${fmri}_hp$hp.ica/mc/prefiltered_func_data_mcf_conf.nii.gz -Tmean $(pwd)/${fmri}_hp$hp.ica/mc/prefiltered_func_data_mcf_conf_mean.nii.gz
-	fslmaths $(pwd)/${fmri}_hp$hp.ica/mc/prefiltered_func_data_mcf_conf.nii.gz -sub $(pwd)/${fmri}_hp$hp.ica/mc/prefiltered_func_data_mcf_conf_mean.nii.gz $(pwd)/${fmri}_hp$hp.ica/mc/prefiltered_func_data_mcf_conf.nii.gz
-	$FSLDIR/bin/imrm $(pwd)/${fmri}_hp$hp.ica/mc/prefiltered_func_data_mcf_conf_mean.nii.gz
+	
+	log_Inform "Dims: $(cat ${fmri}_dims.txt)"
+
+	# ADD COMMENT: Why are the following 3 lines necessary?
```
glasserm: Demean the movement regressors

ICAFIX/hcp_fix_multi_run
```diff
@@ -335,31 +337,30 @@ log_Debug "End of melodic version log, help, and checksum"
...
-#melodic_cmd="${FSLDIR}/bin/melodic -i $concatfmri -o ${concatfmri}.ica/filtered_func_data.ica --nobet --report --Oall --tr=$tr"
-
-# Added "-m `remove_ext ${ConcatName}`_brain_mask" to avoid memory issue - Takuya Hayahsi July 2018
-melodic_cmd="${MELODIC} -i ${concatfmri}_vnf -o ${concatfmri}.ica/filtered_func_data.ica --nobet --report --Oall --tr=$tr --vn --dim=${Dim} --verbose --debug -m `remove_ext ${ConcatName}`_brain_mask"
+## Run melodic on concatenated file
+# Use the Wishart-based dim estimate instead of the melodic-based estimate (--dim=${Dim})
+# Also turn off melodic VN and feed in the input that has already been VN'ed instead (--vn flag, which turns OFF melodic's vn)
+# Added "-m ${ConcatNameNoExt}_brain_mask" to avoid memory issue - Takuya Hayahsi July 2018
+melodic_cmd="${MELODIC} -i ${concatfmri}_vnts -o ${concatfmri}.ica/filtered_func_data.ica --nobet --report --Oall --tr=$tr --vn --dim=${Dim} -m ${ConcatNameNoExt}_brain_mask"
+if [ "${DEBUG}" = "TRUE" ] ; then
```
glasserm: delete vnts after use to save space?

ICAFIX/hcp_fix_multi_run
```diff
@@ -273,29 +263,41 @@ else
...
+${FSL_FIX_WBC} -cifti-average ${ConcatNameNoExt}_Atlas_mean.dscalar.nii ${MeanCIFTISTRING}
+${FSL_FIX_WBC} -cifti-average ${ConcatNameNoExt}_Atlas_vn.dscalar.nii ${VNCIFTISTRING}
+${FSL_FIX_WBC} -cifti-math "TCS + MEAN" ${ConcatNameNoExt}_Atlas.dtseries.nii -var TCS ${ConcatNameNoExt}_Atlas_demean.dtseries.nii -var MEAN ${ConcatNameNoExt}_Atlas_mean.dscalar.nii -select 1 1 -repeat
+${FSL_FIX_WBC} -cifti-merge ${ConcatNameNoExt}_Atlas_hp${hp}_vn.dtseries.nii ${CIFTIhpVNMergeSTRING}
+${FSL_FIX_WBC} -cifti-math "TCS * VN" ${ConcatNameNoExt}_Atlas_hp${hp}.dtseries.nii -var TCS ${ConcatNameNoExt}_Atlas_hp${hp}_vn.dtseries.nii -var VN ${ConcatNameNoExt}_Atlas_vn.dscalar.nii -select 1 1 -repeat
+fslmaths ${ConcatNameNoExt}_SBRef -bin ${ConcatNameNoExt}_brain_mask # Inserted to create mask to be used in melodic for suppressing memory error - Takuya Hayashi
+
+# Disambiguate the concatenated time series created following the "1st pass" VN from the _hp_vnts file
+# that will get created by running functionhighpassandvariancenormalize on the concatfmri file ("2nd pass")
+$FSLDIR/bin/immv ${ConcatNameNoExt}_hp${hp}_vnts ${ConcatNameNoExt}_hp${hp}_vnts0
```
glasserm: just delete since it has been used?

ICAFIX/hcp_fix_multi_run
mharms: Actually, it doesn't at that point in the code, because starting at L205, $fmri is redefined to exclude any extension.  There is a comment at L206 to this effect.  I left the following code "as is" where possible, and only added $fmriNoExt where it was explicitly needed earlier in the code (to consolidate all the sub-shell calls to 'remove_ext').

ICAFIX/hcp_fix_multi_run
mharms: Added comment to that effect.

ICAFIX/hcp_fix_multi_run
mharms: Deleted.

ICAFIX/hcp_fix_multi_run
mharms: Deleted.

coalsont:
So, the edits after changing the names of things appear to finish successfully (specifically, at commit 451bfa5e7dab72c7f33a4286bcfae14791d582fc).  Matt, you can check the output at myelin/tim/docker/testdata**4**/100307 .  I'll start a new clean run from the current PR.

ICAFIX/functionhighpassandvariancenormalize.m
```diff
@@ -96,21 +102,15 @@
 if dovol > 0
     cts=cts./repmat(Outcts.noise_unst_std,1,ctsT);
 	% Use '_vnts' (volume normalized time series) as the suffix for the volumetric VN'ed TCS
-    if hp>=0
```
glasserm: I believe this change will result in _hp0_hp0, which is not what is wanted.  The distinction here needs to be restored I think.

mharms: I don't think so. 'hpstring' is defined as an empty string in the hp<0 condition, which eliminates the need to have two separate 'save_avw' commands that only differ in that one included 'hp' as part of the file name (in the hp>=0 condition) and the other one didn't.

mharms:
With my latest commit a3a553e, I think I have resolved items (1) and (3) that I raised above. [i.e., (1) functionhighpassandvariancenormalize now includes the "hp" string as part of the VN map names, and hcp_fix_multi_run was accordingly modified; and (3) the demeaned and VN'ed time series are deleted when no longer necessary].

functionhighpassandvariancenormalize still needs a description/usage statement (item (2)).

Related to functionhighpassandvariancenormalize, why in the hp<0 condition is that function NOT
(i) computing a CIFTI VN map,
(ii) writing out the NIFTI and CIFTI VN maps
(iii) writing out the CIFTI VN'ed TCS,

While those items may not be necessary for hcp_fix_multi_run itself, the specific needs of hcp_fix_multi_run should not be driving the behavior of functionhighpassandvariancenormalize. Rather, functionhighpassandvariancenormalize should be written as a "stand-alone" function that acts in a consistent manner with its inputs (i.e., writes out a consistent set of files regardless of the value of the hp parameter), and any ensuing naming conflicts resolved in hcp_fix_multi_run.

[Perhaps there are good (other) reasons for functionhighpassandvariancenormalize not computing/writing the 3 items mentioned above?? -- if so, that should be commented in the function].

glasserm:
Thanks Mike.  I checked again and found the following:
*Can do the same vnts file deletes for CIFTI as for volume to save space
*Need to do VN map _hp renames in the last loop that does the splitting
*I checked the development version of RSS and we need ${fMRIName}.nii.gz and .dtseries.nii and ${fMRIName}_hp{HighPass}.nii.gz and .dtseries.nii together obviously with ${fMRIName}_hp{HighPass}_clean.nii.gz and .dtseries.nii.  All other volume and CIFTI timeseries can be deleted.
*Couple of remove_ext are left that you may wish to change to your new form

glasserm:
As for your other comments, functionhighpassandvariancenormalize is a wrapper for icaDim that is specific to MR+FIX, which is the general function that is used in other contexts.  Thus its only job is to detrend / highpass filter various data and call icaDim to compute VN files and dimensionalities.  In the second call, it is behaving like the internals of melodic only on the volume file (VN and dim estimate) and that is why it doesn't output the other files.

coalsont:
FYI, a run using commit 71682942edfcf97cc08f8edf609802e98b35f4f3 completed without error.

coalsont:
Commit a3a553e214ac7da37495ac0de6aa3474a1bde268 MRFIX fails (from a clean data copy without a previous run) with this:


Mon Dec 17 19:20:17 CST 2018 - hcp_fix_multi_run - DEBUG - Moving tfMRI_WM_GAMBLING_MOTOR_RL_LR_hp0.ica to tfMRI_WM_GAMBLING_MOTOR_RL_LR_Atlas_hp0_clean.dtseries.nii
Mon Dec 17 19:20:17 CST 2018 - hcp_fix_multi_run - INFORM - Start=1 Stop=405
Mon Dec 17 19:20:17 CST 2018 - hcp_fix_multi_run - INFORM - Start=1 Stop=405
Mon Dec 17 19:20:17 CST 2018 - hcp_fix_multi_run - DEBUG - cifti merging
Mon Dec 17 19:20:17 CST 2018 - hcp_fix_multi_run - DEBUG - cifti merging

While running:
/usr/local/workbench/bin_linux64/../exe_linux64/wb_command -cifti-math '((TCS / VNA) * VN) + Mean' /basedir/data/100307/MNINonLinear/Results/tfMRI_WM_RL/tfMRI_WM_RL_Atlas_hp0_clean.dtseries.nii -var TCS /basedir/data/100307/MNINonLinear/Results/tfMRI_WM_RL/tfMRI_WM_RL_Atlas_hp0_clean.dtseries.nii -var VNA tfMRI_WM_GAMBLING_MOTOR_RL_LR_Atlas_vn.dscalar.nii -select 1 1 -repeat -var VN /basedir/data/100307/MNINonLinear/Results/tfMRI_WM_RL/tfMRI_WM_RL_Atlas_vn.dscalar.nii -select 1 1 -repeat -var Mean /basedir/data/100307/MNINonLinear/Results/tfMRI_WM_RL/tfMRI_WM_RL_Atlas_mean.dscalar.nii -select 1 1 -repeat

ERROR: failed to open file '/basedir/data/100307/MNINonLinear/Results/tfMRI_WM_GAMBLING_MOTOR_RL_LR/tfMRI_WM_GAMBLING_MOTOR_RL_LR_Atlas_vn.dscalar.nii', file does not exist, or folder permissions prevent seeing it

mharms:
> Thanks Mike. I checked again and found the following:
> *Can do the same vnts file deletes for CIFTI as for volume to save space

Done.  Are the _individual run_ VN'ed time series needed?  I went ahead and deleted those as well.
> *Need to do VN map _hp renames in the last loop that does the splitting

Done.  It took me a while to understand this section. I added a number of comments there and elsewhere to help users figure it out more quickly.
> *I checked the development version of RSS and we need ${fMRIName}.nii.gz and .dtseries.nii and ${fMRIName}_hp{HighPass}.nii.gz and .dtseries.nii together obviously with ${fMRIName}_hp{HighPass}_clean.nii.gz and .dtseries.nii. All other volume and CIFTI timeseries can be deleted.

Per above, I deleted the individual run VN'ed time series.  Hard for me to tell if there are others that can be deleted from the code.  I suggest we check the actual output, and see if we've missed anything else.  (Presumably, we want to keep the vn maps around, right?)
> *Couple of remove_ext are left that you may wish to change to your new form.

Done.  (They weren't necessary in the first place, since concat_fmri_orig already had no extension).

mharms:
> Commit [a3a553e](https://github.com/Washington-University/HCPpipelines/commit/a3a553e214ac7da37495ac0de6aa3474a1bde268) MRFIX fails (from a clean data copy without a previous run) with this:

Yes, that is the issue that Matt pointed out in the splitting section, which I had missed when adding the hp string elsewhere in the script.  It is hopefully now resolved.

mharms:
> As for your other comments, functionhighpassandvariancenormalize is a wrapper for icaDim that is specific to MR+FIX, which is the general function that is used in other contexts. Thus its only job is to detrend / highpass filter various data and call icaDim to compute VN files and dimensionalities. In the second call, it is behaving like the internals of melodic only on the volume file (VN and dim estimate) and that is why it doesn't output the other files.

I think this is unfortunate, and makes functionhighpassandvariancenormalize harder to understand, since what it chooses to output seems a bit arbitrary.  What if someone wanted to apply VN but not use high-pass?  Then the function as currently written becomes worthless.

mharms:
It's rather hard to say that the hp=2000 via fslmaths was nothing less than a pure linear detrend.  And regardless, the code should have the _capability_ to do a fast detrend that is more than just a linear detrend.  If someone wants to detrend via a 4th order polynomial, they should have the option to do so.  I don't anticipate that this will be a difficult addition.

glasserm:
How do you propose to make that happen?

mharms:
It's just a simple regression.  In terms of getting the order of the polynomial into the function, we could repurpose hp=0 to mean no high pass (which is the logical meaning of zero, although I understand that you were just modeling after what fix_3_clean chose to do), and negative integers represent the order for a polynomial detrend.  Or, if you don't like that approach, I'll just add a vargin argument that is specific to the polynomial detrend feature.  We'll need a way to label this via the "hp" string as well -- perhaps something like "hppd#" (for "high pass polynomial detrend order #").

glasserm:
I think hp=0 needs to stay what it is in FIX.  -1 also has a special meaning in FIX

mharms:
Sure, we can avoid "breaking" that convention, although I'm not a fan that FIX chose to do it that way.

glasserm:
Is there a "correct" amount of polynomial detrend?

mharms:
I consider that an empirical question.  It'll depend on how stable your scanner is, the length of the run, etc.

coalsont:
Commit 35e6a36349963155aabe6164ce139d0df3ea9a86 MRFIX appears to have finished without error.

mharms:
Yeah.  Thanks for checking.  Are there any further files that should be deleted?

glasserm:
${StudyFolder}/${Subject}/MNINonLinear/Results/${ConcatName}/${ConcatName}_hp${HighPass}.ica/filtered_func_data.ica/alldat.nii.gz
${StudyFolder}/${Subject}/MNINonLinear/Results/${ConcatName}/${ConcatName}_hp${HighPass}.ica/filtered_func_data.ica/concat_data.nii.gz

Are large and are unnecessary outputs from melodic.  

${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas_hp${HighPass}.dtseries.nii
${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_hp${HighPass}.nii.gz

Are not required for MR enabled RSS and could be easily regenerated from the concatenated files.  

Otherwise this seems to have worked.

mharms:
I'm not seeing those alldat.nii.gz and concat_data.nii.gz files in a recent run of single-run FIX that used the melodic in FSL 6.0.0.  Nor do they seem to be generated by any of the MR-FIX specific code.  Where are those two files getting generated?  (Are you still using a custom version of 'melodic' in your testing?)

In terms of other possible files to delete, I'd rather delete the concatenated versions (see below)
and leave the individual run hp's generated by the first pass through functionhighpassandvariancenormalize.m
That's keeping with the principle that the individual runs are "primary".  (If we ever need to build back up the concatenated versions, that is easy to do).

Come to think of it, are there _any_ time series (NIFTI or CIFTI) that we need to leave behind in ${ConcatFolder} (including the cleaned files themselves)?  Those are all going to be very large in aggregate.

Looking at the output of Tim's most recent run, I'd suggest that we delete the following from ${ConcatFolder}:
```
${ConcatNameNoExt}.nii.gz
${ConcatNameNoExt}_hp${hp}.nii.gz
${ConcatNameNoExt}_hp${hp}_clean.nii.gz
${ConcatNameNoExt}_Atlas.dtseries.nii
${ConcatNameNoExt}_Atlas_hp${hp}.dtseries.nii
${ConcatNameNoExt}_Atlas_hp${hp}_clean.dtseries.nii
```

I can imagine that perhaps the current version of MR enabled RSS requires some of those files.  But in that case, I'd argue that the solution is to modify RSS to re-create the necessary files by concatenating the individual run files; not by letting those large concatenated files linger solely for the purpose of RSS.

mharms:
In creating my candidate list above for deletion, I realized that I can't figure out the purpose for two of them in the 'hcp_fix_multi_run' script.
Namely

- ${ConcatNameNoExt}.nii.gz 
created at L280:
`fslmaths ${ConcatNameNoExt}_demean -add ${ConcatNameNoExt}_mean ${ConcatNameNoExt}`

and its CIFTI analog

- ${ConcatNameNoExt}_Atlas.dtseries.nii
created at L285:
`${FSL_FIX_WBC} -cifti-math "TCS + MEAN" ${ConcatNameNoExt}_Atlas.dtseries.nii -var TCS ${ConcatNameNoExt}_Atlas_demean.dtseries.nii -var MEAN ${ConcatNameNoExt}_Atlas_mean.dscalar.nii -select 1 1 -repeat`

Where do either of those ever actually get used?

glasserm:
I want the concatenated files to be available without having to redo the demean and VN for a variety of things including RSS.  Those files other files are used by RSS.  I don't want to have to run the MR+FIX loop every time we need concatenated data.  Let's stick with the original plan for this.  Perhaps Tim can make those changes and then we can merge this?  I'm getting a lot of questions about MR+FIX privately and I think we need to push this out.

coalsont:
@glasserm I'm not really following what files mean what, or which changes you mean.  Whoever understands what should be done, please do it.

mharms:
This is very space inefficient.  Why would you need to re-run the MR+FIX loop?  Those large files in ${ConcatFolder} are readily assembled from files preserved in the individual runs.

I would like us to preserve the individual run hp files.  Sri uses those in some of his scripts.

Where are alldat.nii.gz and concat_data.nii.gz coming from?  I want to make sure that we've tested this using a stock version of 'melodic' available in an FSL release.

mharms:
BTW: Which of the concatenated time series files are needed for RSS?  We should add a comment to that effect in the script.

glasserm:
Okay that is fair if you are using single run HP we can keep them.  I am guessing that you haven't fully worked through the differences between the concatenated and split (or unmerged) files or you wouldn't be asking me to delete them.  You can do that or trust me that they aren't the same and will serve different purposes.  

Those other files come from the output all option of melodic, which is useful for some of its outputs, but not those.  

Copied from above: *I checked the development version of RSS and we need ${fMRIName}.nii.gz and .dtseries.nii and ${fMRIName}_hp{HighPass}.nii.gz and .dtseries.nii together obviously with ${fMRIName}_hp{HighPass}_clean.nii.gz and .dtseries.nii. All other volume and CIFTI timeseries can be deleted.

mharms:
When you say we need various ${fMRIName}.* files for RSS, I assume that you're referring to the _concatenated_ versions of each of those?

glasserm:
Right

mharms:
Ok. Working on a version that will also support polynomial detrending.  Should have done this afternoon.

mharms:
Ok.  Getting close.  Now, just need to modify 'hcp_fix_multi_run' to support the polynomial detrend approach.  I'll try to get that done tomorrow, and then hopefully Tim can test it for the last time.

glasserm:
I think those changes look ok.

glasserm:
Did you remove the ability to use 0 as a linear detrend or just make that option harder to see?

glasserm:
I think I'd rather not change the naming for hp=0 at least, as some folks have already been using it.  I don't see the value in messing them up and changing how regular FIX and MR+FIX behave.

coalsont:
@mharms oxford sent us a patched version of fslcpgeom.  It was quite broken, such that virtually anything else using it would not have gotten the desired result.  The hack I made with a script and wb_command was not put in the FSL folder in my container when it was needed.

glasserm:
Was the patched one broken also?  I don't remember you mentioning that.

coalsont:
The patched version works perfectly as far as I know.  There was a separate patch to most of FSL, where there was a library dependency that was more complicated than they had expected, which was before the fslcpgeom problem was found.

mharms:
@coalsont  I'm not following.  Do we have a working version of fslcpgeom (from Oxford) or not?  My main point is, whether it is a patched version from Oxford, or the "replacement" version that you crafted using wb_command, how is the necessary version getting "installed"?

coalsont:
Yes, the patched version from oxford works fine.  Since it is really an FSL binary (not a hacky workaround), and the one it replaces would have caused problems for any script, my container setup currently does put it into the fsl folder, same as unzipping the other patch for the missing library situation in FSL 6.0.0, but this is largely irrelevant to the pipeline scripts.  The HCP scripts really shouldn't have to know about bugs in specific versions of FSL.  FSL 6.0.1 should fix both of these issues when it is released.

coalsont:
There are existing patches to get FSL 6.0.0 to a working state, the only question is how public they are (which can be changed).  I don't think we need our scripts to test what version of FSL is being used (maybe simply print it to the log without doing anything about it, like the version of melodic and wb_command).

mharms:
I don't think that the OpenBLAS patch, or the fslcpgeom patch, are public at all.  Given that multiple aspects of hcp_fix_multi_run won't work without FSL 6.0.x, I think a simple version check within the script is well warranted.

coalsont:
Have you made changes that won't work in FSL 5.x?  The only issue I am aware of is that fslcpgeom in 5.x used to fail on large files, and the scripts had workarounds for this.

mharms:
Well, given that we are assuming that most instances of MR-FIX are going to involve a large number of time points, then multiple aspects of FSL 5.x don't work (including melodic).

coalsont:
We were using a special build of melodic when we were using FSL 5.x.  I only remember that the default one was slow, not that it would give wrong answers or error.  I wouldn't want the scripts to suddenly decide not to even try with our previously working setup (I have not yet installed FSL 6 on our processing machines).

glasserm:
There were some bugs with large files and it was a lot slower for PCA.

coalsont:
@mharms The polynomial detrend stuff doesn't seem like it belongs in this PR, which seemed like it was close to being ready to merge.  Can you make a new branch and rewind your master (which this PR uses) to before the polynomial detrend code, and then open a new PR on the new branch for the polynomial detrend implementation?

coalsont:
FYI, commit 137761c135edaf19a718e78bd690d9f66de0a3d4 appears to have run through MR FIX and postfix without error.

mharms:
@glasserm  Per your request, using "hp=0" as the input specification will now continue to use "_hp0" in the output file names.

@coalsont  At this point, I have other stuff beneficial changes intermingled with the polynomial detrend code, so I'd rather we just wrap this all up together.  If you test once with "hp='0'", and once with "hp=pd2", then I think we have our bases covered.  Thx.

The biggest issue remaining, in my opinion, is that we need Oxford to release a FSL 6.0.1 that incorporates the various patches that we are relying on users to have.  (And I still think we should add an explicit check for the FSL version, since MR-FIX is not going to work for probably 90+% of use cases otherwise).

glasserm:
I've asked Tim to run those tests.

coalsont:
hp=0 fails quickly (and silently), on line 282 in octave.  The log file says:

error: functionhighpassandvariancenormalize: Invalid specification for the order of the polynomial detrending

I still disagree with the script trying to predict whether it will fail, and your particular suggestion would have it expect to fail on a setup I know to have worked before your changes.  It is complication and headaches without obvious benefit, and will quickly be useless once FSL makes another release.

glasserm:
That is when hp=-1 is specified.

coalsont:
For clarity, here is what hcp_fix_multi_run runs when the octave code fails:

```
+ /usr/local/fix1.067/call_matlab.sh -l .fix.functionhighpassandvariancenormalize.log -r 'addpath('\''/basedir/Pipelines/ICAFIX'\''); functionhighpassandvariancenormalize(0.720000 , '\''0'\'', '\''tfMRI_WM_RL'\'', '\''/usr/local/workbench/bin_linux64/wb_command'\'');'
```

mharms:
Sorry, introduced an error relative to the successful 137761c test.  Please try again.

mharms:
Regarding a version check, I don't see how it is useless.  Are you saying that MR-FIX works with FSL 5.X _with large concatenated files_?  Because that's the use case that we need to support.  It is irrelevant in my opinion that MR-FIX may complete successfully with FSL 5.X in the case of using files that are less than 2 GB, since that represents a tiny fraction of the MR-FIX use cases.  Basically, the only FSL version for which MR-FIX will work unencumbered by various restrictions is starting with FSL 6.0.1, so we should check for that version, and abort with an appropriate message if the correct version isn't found.

coalsont:
Another error in the same octave step:

error: load: unable to find file tfMRI_WM_RL_hppd1.ica/mc/prefiltered_func_data_mcf.par

With a special version of melodic, and not using fslcpgeom, yes, MR FIX worked with large files when FSL 5 was installed.  "This script needs FSL version 6.0.1 (or patched versions of X and Y)" is something to put in the documentation, not as a check in the code.

coalsont:
To put it another way: how exactly is a new user, who has vanilla FSL 5.x or 6.0.0 installed, going to manage to launch MR FIX without reading any of our documentation?  That is the only situation where having the script say "I'm sorry Dave, I'm afraid I can't do that" would be obviously better than documentation (comments, usage, example script, wiki, etc).

coalsont:
"Not harming" is good enough justification for "don't do it", which is why it is the default position and doesn't need to be defended.  Having the check would make things more painful for us - your proposal would refuse to run on even my currently working setup.  If the user *is* going to read the documentation, if nothing else to figure out what arguments to pass, why should we add a redundant stumbling block that has known false positives?

glasserm:
Would a warning message be a satisfactory compromise?

coalsont:
Also, what happens when we make a public container with patched versions of software (because new releases aren't made yet), and then we add a check like this to a pipeline?  Then, despite having been previously tested as working, a proposal like Mike's would add a check that would break a newer version of the pipelines on the tested-working container image (because it doesn't have the release that has the patch already applied), without anything actually being wrong.

If it was only a message, without aborting, that is better, though it would probably need to pause for a few seconds to give the user time to read the message before normal pipeline output scrolls it off the screen.  This would still at least be somewhat irritating for people who already know their setup works.

mharms:
The only people we are helping by not including an explicit version check, and aborting if that isn't satisfied, is ourselves.  We need to move beyond ourselves, and think whether it would be beneficial to _other_ users, which it would.

mharms:
I would argue that we simply shouldn't be releasing a public container created from our internally patched version of the software.

coalsont:
And what if the new release of that software has a bug in a different part of the same software?  What if the new release of that software takes an extra two months?  That doesn't sound like a good rule.

I have heard only you argue that it would be helpful in any realistic scenario, and you dismissed my scenario where it demonstrably harmed other users.  You need consensus on what is helpful to others before claiming the high ground.

mharms:
What other user is my proposal harming (not counting you, who would need to create a new container starting from FSL 6.0.1)?

coalsont:
A concrete, current example, is Alan A., who has been working on getting MR FIX working also, and also has the patches that get FSL 6.0.0 into working shape.

mharms:
I think that Alan would actually support my position of looking at this from the perspective of general, naive users.

coalsont:
Why don't you ask him in email, and then we can stop going off-topic in this PR.

coalsont:
Different error:

```
/basedir/Pipelines/ICAFIX/hcp_fix_multi_run: line 550: syntax error near unexpected token `then'
```

mharms:
Fixed.  Got tripped up by the 'then' being in a different line from the 'if' statement.

coalsont:
hp=0 seems to have worked, pd2 is in progress.

mharms:
FWIW, I still think we should have a version check embedded in the script. (@aanticev ?)
And it remains unclear how we are going to position this if FSL 6.0.1 isn't forthcoming soon...

glasserm:
I don't see those issues as holding up the merge however, but rather are questions to be dealt with before a release announcement.

mharms:
Do we want to explore Sri creating a compiled matlab version, and including that as part of what we distribute?

glasserm:
Sure if he is willing to.

coalsont:
Since ReApply is not in sync with these changes, I have reverted master to before the merge, after saving the PR state from before the merge as "pull83" (ignore the revert-83-master branch, that was created when I clicked the github revert button, which actually starts making a new PR to revert).

Mike, please make a new PR, you can either use your master again (if you haven't pulled after the merge), make your own branch starting from bc90cf1652223df76a73ed3e8541b94f1d63ede4, or you can use the pull83 branch.  I could make the PR, but then you'd need to set up to push to a different upstream to add edits to the PR.

mharms:
Following the reversion mentioned above, this PR has been re-generated as PR #87.

PR #88, opened 2019-01-17
Feature/fsl v6

junilc:
Fixed (standard_in) 1: syntax error for Functional Preprocessing pipeline with FSL-6.0.0 - converting scientific notation for bc
Tested with both FSL-6.0.0 and FSL-5.0.11


glasserm:
Thanks Junil

PR #89, opened 2019-01-25
Feature/fsl v6

junilc:
sed command for scientific notation ‘e’, ‘E’, ‘e+’, and ‘E+’


glasserm:
Tim says: "Using * rather than \? is a little weird, and would make it also accept things like "E+++".  It should work for the cases stated, though (+ appears to actually need to be escaped to have its special regexp meaning)."

junilc:
He is right that will also take +++, Tim B suggested to use \?

PR #90, opened 2019-02-06
Add capability to use BIDS convention (i/j/k) for specifying axes

mharms:
This will allow for pulling values directly from BIDS-style sidecar json files, without needing to remap the axes into the x/y/z nomenclature.

### DIFF
--- a/PreFreeSurfer/scripts/T2wToT1wDistortionCorrectAndReg.sh
+++ b/PreFreeSurfer/scripts/T2wToT1wDistortionCorrectAndReg.sh
@@ -32,10 +32,10 @@ Usage() {
   echo "            [--SEPhaseNeg=<input spin echo negative phase encoding image>]"
   echo "            [--SEPhasePos=<input spin echo positive phase encoding image>]"
   echo "            [--seechospacing=<effective echo spacing of SEPhaseNeg and SEPhasePos, in seconds>]"
-  echo "            [--seunwarpdir=<direction of distortion of the SEPhase images according to voxel axes>]"
+  echo "            [--seunwarpdir=<direction of distortion of the SEPhase images according to *voxel* axes: {x,y,x-,y-} or {i,j,i-,j-}>]"
   echo "            --t1sampspacing=<sample spacing (readout direction) of T1w image - in seconds>"
   echo "            --t2sampspacing=<sample spacing (readout direction) of T2w image - in seconds>"
-  echo "            --unwarpdir=<direction of distortion of T1 and T2 according to voxel axes (post reorient2std)>"
+  echo "            --unwarpdir=<direction of distortion of T1 and T2 according to *voxel* axes (post fslreorient2std): {x,y,z,x-,y-,z-}, or {i,j,k,i-,j-,k-}>"
   echo "            --ot1=<output corrected T1w image>"
   echo "            --ot1brain=<output corrected, brain-extracted T1w image>"
   echo "            --ot1warp=<output warpfield for distortion correction of T1w image>"
@@ -216,10 +216,13 @@ case $DistortionCorrection in
         # -- Spin Echo Field Maps --  
         # --------------------------
 
-        if [[ ${SEUnwarpDir} = "x" || ${SEUnwarpDir} = "y" ]] ; then
-          ScoutInputName="${SpinEchoPhaseEncodePositive}"
-        elif [[ ${SEUnwarpDir} = "-x" || ${SEUnwarpDir} = "-y" || ${SEUnwarpDir} = "x-" || ${SEUnwarpDir} = "y-" ]] ; then
-          ScoutInputName="${SpinEchoPhaseEncodeNegative}"
+        if [[ ${SEUnwarpDir} = [xyij] ]] ; then
+			ScoutInputName="${SpinEchoPhaseEncodePositive}"
+        elif [[ ${SEUnwarpDir} = -[xyij] || ${SEUnwarpDir} = [xyij]- ]] ; then
+			ScoutInputName="${SpinEchoPhaseEncodeNegative}"
+		else
+			echo "${SCRIPT_NAME} - ERROR - Invalid entry for --seunwarpdir ($SEUnwarpDir)"
+			exit 1
         fi
 
         # Use topup to distortion correct the scout scans
@@ -247,14 +250,18 @@ case $DistortionCorrection in
         exit 1
 esac
 
+# FSL's naming convention for 'convertwarp --shiftdir' is {x,y,z,x-,y-,z-}
+# So, swap out any {i,j,k} for {x,y,z} (using bash pattern replacement)
+# and then make sure any '-' sign is trailing
+UnwarpDir=${UnwarpDir//i/x}
+UnwarpDir=${UnwarpDir//j/y}
+UnwarpDir=${UnwarpDir//k/z}
 if [ "${UnwarpDir}" = "-x" ] ; then
   UnwarpDir="x-"
 fi
-
 if [ "${UnwarpDir}" = "-y" ] ; then
   UnwarpDir="y-"
 fi
-
 if [ "${UnwarpDir}" = "-z" ] ; then
   UnwarpDir="z-"
 fi
--- a/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
+++ b/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
@@ -37,7 +37,7 @@ Usage() {
   echo "             --SEPhaseNeg=<input spin echo negative phase encoding image>"
   echo "             --SEPhasePos=<input spin echo positive phase encoding image>"
   echo "             --echospacing=<effective echo spacing of fMRI image, in seconds>"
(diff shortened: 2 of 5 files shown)


glasserm:
Is this tested?

mharms:
I tested the bash syntax employed in a simple test script. I was careful not to make any typos, but if @tbbrown and @coalsont can just review the changes, as a second pair of eyes to make sure that there are no logic errors, that seems sufficient to me here.

glasserm:
It looked fine to me, but it would be nice if Tim C. checks too.

coalsont:
I glanced over it and didn't see anything obvious in the syntax or logic.

The way I use xyz and ijk are that xyz are spatial, while ijk relate to the way the voxels are ordered inside the file.  Depending on when reorientation to standard occurs, the original values from the BIDS sidecar may no longer be appropriate if the acquisition uses a different axis ordering.  I don't know if xyz in these scripts actually was referring to the voxel order to begin with.

coalsont:
Oh, also, since the FSL standard nifti orientation is -x, +y, +z, it is possible that +i could actually mean -x, if these scripts do in fact treat x as a spatial direction.  Isn't volume orientation fun...

mharms:
All of the FSL tools, to my knowledge, use {x,y,z,x-,y-,z-} as indices into the *voxel* (NIFTI) axes, not the actual spatial directions in "RAS" space.  The HCPpipelines adopted that same nomenclature.  That's why there is a one-to-one mapping between "{x,y,z,x-,y-,z-}" within the context of the HCPpipelines and {i,j,k,i-,j-,k-} that does NOT depend on the orientation of the NIFTI (although the correct value DOES depend on the NIFTI axis ordering).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -37,7 +37,7 @@ Usage() {
   echo "             --SEPhaseNeg=<input spin echo negative phase encoding image>"
   echo "             --SEPhasePos=<input spin echo positive phase encoding image>"
   echo "             --echospacing=<effective echo spacing of fMRI image, in seconds>"
-  echo "             --unwarpdir=<unwarping direction: x/y/z/-x/-y/-z>"
+  echo "             --unwarpdir=<PE direction for unwarping according to the *voxel* axes: {x,y,x-,y-} or {i,j,i-,j-}>"
```
tbbrown: Is there a reason why this no longer mentions z/z- and k/k- ?

global/scripts/TopupPreprocessingAll.sh
```diff
@@ -216,7 +217,7 @@ if [[ $UnwarpDir = "x" || $UnwarpDir = "x-" || $UnwarpDir = "-x" ]] ; then
     i=`echo "$i + 1" | bc`
   done
 # Y direction phase encode
-elif [[ $UnwarpDir = "y" || $UnwarpDir = "y-" || $UnwarpDir = "-y" ]] ; then
+elif [[ $UnwarpDir = [yj] || $UnwarpDir = [yj]- || $UnwarpDir = -[yj] ]] ; then
```
tbbrown: What happens in this script if the unwarp direction is specified as z/-z or k/-k or 'q' for that matter? 
I think it will neither enter the unwarpdir = [xi] [xi]- -[xi] block nor the unwarpdir = [yj] [yj]- -[yj] block. Is that the right thing to do, do nothing and then move on? Is this be an error that is captured and reported?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: Because, in principle for 2D EPI data, at least when converted with 'dcm2niix', the unwarpdir will never be the z/k axis.  That's why I removed z/k as listed options from the --unwarpdir argument in TopPreprocessingAll.sh as well.  In the case of TopPreprocessingAll.sh, z/k is explicitly not a supported option, and I just added an error in that script if --unwarpdir is not a supported value.

In the case of DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh, there is no error check on the allowed values for --unwarpdir currently.  I'll do whatever others think best.  (i.e., either change the usage back to include z/k, add an explicit check for allowed values, or just leave as is).

global/scripts/TopupPreprocessingAll.sh
mharms: Just changed the script to error out in that situation.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
tbbrown: I would always vote for an explicit check for allowed values.

mharms: Easy enough to add an explicit check.  I just need @glasserm and @coalsont thoughts on whether z/k should be an allowable option or not in DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh.  I guess we could nominally allow z/k within that particular script (consistent with how it was), although note that if using SEFMs as the distortion correction approach, the use of z/k will get kicked out as an error once TopPreprocessingAll.sh is called.

coalsont:
If the tools/scripts can support it easily, personally I would want to allow z/k as input, just in case someone converts with some other tool or some weird sequence puts the dicoms together strangely, or...

mharms:
Ok, there seems no reason in principle that 'topup' can't support z/k as the unwarpdir, so I've added support for that possibility in TopPreprocessingAll.sh.  Please review the logic for any errors.

PreFreeSurfer/scripts/T2wToT1wDistortionCorrectAndReg.sh
```diff
@@ -216,10 +216,13 @@ case $DistortionCorrection in
         # -- Spin Echo Field Maps --  
         # --------------------------
 
-        if [[ ${SEUnwarpDir} = "x" || ${SEUnwarpDir} = "y" ]] ; then
-          ScoutInputName="${SpinEchoPhaseEncodePositive}"
-        elif [[ ${SEUnwarpDir} = "-x" || ${SEUnwarpDir} = "-y" || ${SEUnwarpDir} = "x-" || ${SEUnwarpDir} = "y-" ]] ; then
-          ScoutInputName="${SpinEchoPhaseEncodeNegative}"
+        if [[ ${SEUnwarpDir} = [xyij] ]] ; then
```
coalsont: This looks like it will still error on z/k.

glasserm:
I believe topup does not support z/k.  The reason probably is that z/k will not ever be the phase encoding direction in an EPI acquisition (if someone acquired coronal or sagittal slices they would have to make the readout direction z/k to avoid aliasing).  Thus I think we can safely disallow z/k and might as well error as soon as possible.  Readout distortion is a separate matter that is most likely to be z/k.

mharms:
There is nothing in the topup documentation indicating the z/k is not a supported axis -- the third values in the --datain text-file allows for the phase-encoding direction to be the z axis.  The point here is that since z/k is the *voxel* axis, then any PE axis is possible, depending entirely on the ordering of the NIFTI voxel axes.  Given that  180fbe5 implements this already, I'd rather leave it as feature enhancement at this point.

PreFreeSurfer/scripts/T2wToT1wDistortionCorrectAndReg.sh
mharms: I'll fix this once @glasserm agrees that 180fbe5 is ok.

glasserm:
I believe I attempted to use topup with that direction for readout distortion correction of reversed structurals long ago and it didn't work.  I had to rotate the images to get topup to run.  

Also as I said, one couldn't use z/k in EPI data anyway, which is what this pipeline is for.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: I've left z/k as allowed values in this specific script, because presumably, under the path that uses gradient echo field maps (epi_reg), that is still an allowed PE dir.

PreFreeSurfer/scripts/T2wToT1wDistortionCorrectAndReg.sh
mharms: This stays "as is", given that Jesper has confirmed that 'topup' indeed doesn't allow z as the PE direction.

mharms:
Ok, after Jesper's clarification that `topup` indeed isn't designed to work with 'z' as the PE direction, the argument checking in the scripts should now be appropriate.

tbbrown:
Seems like the questions raised have been addressed, right? Are we good with merging this into the master branch?

glasserm:
Thanks Mike

PR #97, opened 2019-03-08
add support for multiple MR reapply groups in a single call, example script for MR FIX

coalsont:
Matt wants this in the release to be able to do the MR FIX groups without calling dedrift multiple times.  I ran a quick test of the dedrift edits, and of an edited copy of the new example script for MR FIX (not dedrift).


coalsont:
It looks like the existing IcaFixProcessingBatch.sh can't do multiple groups of runs per subject (day 1, day 2).  @glasserm?

glasserm:
I had forgotten that we had changed the Batch.  Tim's point is important too.  Could IcaFixProcessingBatch.sh be made to support this usecase?

glasserm:
We should update the Examples, Scripts, DeDriftAndResampleBatch file with the new instructions also

mharms:
If @coalsont could add that use case into the existing IcaFixProcessingBatch.sh, that would be optimal IMO (to reduce the number of example scripts that we need to maintain).

coalsont:
A multi-group setup for MR FIX needs a more involved argument format than normal FIX, but currently the trigger for whether to do MR FIX is whether the concat name is empty.  So, what do we expect the normal FIX to do if someone has a multi-group MR FIX setup and then sets the concat name to empty?

glasserm:
I guess each could be run individually or it could be an error.

coalsont:
I have tested the updated example script, there was a typo in a variable name in the previous MR FIX section.

coalsont:
Found some bugs, retesting.

Examples/Scripts/IcaFixProcessingBatch.sh
```diff
@@ -132,14 +132,14 @@ main() {
 		export FSL_FIXDIR=${FixDir}
 	fi
 
-	# set list of fMRI on which to run ICA+FIX
+	# set list of fMRI on which to run ICA+FIX, separate groups with @ (for MR FIX concatenation)
```
mharms: I'm thinking that users might be a little confused what is meant by "groups" without more context.  Could we add more explanation, and perhaps a (commented out) example of what `fMRINames` would look like in the multiple "groups" condition.

Part of the reason this could be confusing is that usually when the '@' construction is used, it has to be used between _all_ the inputs (so that there are no whitespaces at all).  In fact, because the situation is different here, perhaps we should use '%' rather than '@' to separate the groups -- which would also match nicely to the separator between groups used in the `--multirun-fix-names` argument of `DeDriftAndResamplePipeline.sh`

glasserm: I think making the usages similar is a good idea if they really do parallel eachother.

Examples/Scripts/IcaFixProcessingBatch.sh
```diff
@@ -157,53 +157,68 @@ main() {
...
+			    done
 			
-			echo "  InputFile: ${InputFile}"
+			    echo "  InputFile: ${InputFile}"
 
-			cmd="${queuing_command} ${FixScript} ${InputFile} ${bandpass} ${ConcatFileName} ${domot} ${TrainingData} ${FixThreshold}"
-			echo "About to run: ${cmd}"
-			${cmd}
+			    cmd=("${queuing_command[@]}" "${FixScript}" "${InputFile}" ${bandpass} ${ConcatFileName} ${domot} "${TrainingData}" ${FixThreshold})
+			    echo "About to run: ${cmd[*]}"
```
mharms: This is for my curiosity -- why do you use `${cmd[*]}` in the echo statement, but `${cmd[@]}` to actually invoke the command -- what's the difference?

coalsont: Partly habit - the logging functions at least used to only use $1, so if you passed them multiple arguments only the first one would be used - using @ within "" is special, it expands to multiple arguments, while * expands to a single argument with spaces between elements (if the default IFS is in effect):

```
tim@timsdev:~$ test=(a b c)
tim@timsdev:~$ echo "testing ${test[@]}"
testing a b c
tim@timsdev:~$ function countargs() { echo $#; }
tim@timsdev:~$ countargs "testing ${test[@]}"
3
tim@timsdev:~$ countargs "testing ${test[*]}"
1
```

Examples/Scripts/IcaFixProcessingBatch.sh
coalsont: dedrift takes the string as an input argument, if it used spaces it would need to be quoted, and some pipeline authors don't like being space-safe.  Here, it is in a variable that used to be space delimited in previous versions, and assignment to variables is generally already quoted to work correctly with spaces.  As it currently stands, people can copy their settings from a previous version of this example script and it will work, but if we go with @ and % to match dedrift, then previous settings won't work.

coalsont:
Commit 8907b69 is tested working for MR FIX, I am testing the change to % (and spaces) now, I expect it to work.

coalsont:
I need suggestions of a default example of two MR FIX groups, which some/many users will probably take as an endorsement of setting the MR FIX groups that way.  Are rest1_* and rest2_* collected on different days?

glasserm:
Use the task data example.



Sent from my Verizon, Samsung Galaxy smartphone


-------- Original message --------

coalsont:
I was hoping for something shorter and therefore easier for users to look at and comprehend, but okay.

mharms:
Yes, in the context of HCP-YA at least, REST1 and REST2 were collected within different sessions (generally on different days).  Can we please use `bandpass=pd2` as the example default, with a comment that `pdX` represents a polynomial detrend of order X?  It is very common for there to at least be a quadratric component to the drift.

coalsont:
Be aware that ordinary hcp_fix doesn't support hp=0 (crashes after a little bit with errors mentioning a folder named just ".ica"), and I'm guessing not "pd2" either.

mharms:
Arg... One caveat that we need to mention though, since we have a single Examples script that in inteded to guide users for bot single- and multi-run FIX, is that `bandpass=pdX` is only supported currently for MR-FIX.  Using bandpass=0 is not a solution to this either, since hp=0 isn't supported in single-run FIX (hcp_fix) either.

mharms:
You must not have the most up-to-date version of hcp_fix (now part of the freesurfer_v6_latest branch), because I changed it to specifically error out early on if bandpass=0.

coalsont:
Yes, this branch is based off master, not off freesurfer_v6_latest.

mharms:
If you are testing any ICAFIX related stuff, you should be using the freesurfer_v6_latest branch at this point.

Examples/Scripts/IcaFixProcessingBatch.sh
```diff
@@ -143,8 +143,10 @@ main() {
 	ConcatNames=""
 	ConcatNames="tfMRI_WM_GAMBLING_MOTOR_RL_LR tfMRI_LANGUAGE_SOCIAL_RELATIONAL_EMOTION_RL_LR"  ## Use space to separate concatenation groups
 
-	# set temporal highpass full-width (2*sigma) to use, in seconds
-	bandpass=0
+	# set temporal highpass full-width (2*sigma) to use, in seconds, cannot be 0 for single-run FIX
+	bandpass=2000
+	# MR FIX also supports 0 for linear, "pd2" for quadratic
```
mharms: Just so users know what is fully available, can we say:
```
# MR FIX also supports 0 for a linear detrend, or "pdX" for a polynomial detrend of order X 
# e.g., bandpass=pd1 is linear detrend (functionally equivalent to bandpass=0)
# bandpass=pd2 is a quadratic detrend
```

glasserm:
Not if all the code doesn't work with that.



Sent from my Verizon, Samsung Galaxy smartphone


-------- Original message --------

coalsont:
Octave doesn't like the freesurfer_v6_latest branch:

```
Thu Mar 14 14:33:45 CDT 2019:hcp_fix_multi_run: /usr/local/fix1.067/call_matlab.sh -r addpath('/basedir/Pipelines/ICAFIX/scripts'); functionhighpassandvariancenormalize(0.720000 , 'pd2', 'rfMRI_REST1_LR', '/usr/local/workbench/bin_linux64/wb_command');
<some warnings of shadowing functions>
Adding /matlabextras to MATLAB path
error: 'normalise' undefined near line 108 column 15
```

Examples/Scripts/IcaFixProcessingBatch.sh
coalsont: What is the verdict here?  Do we make it hard to copy paste between dedrift and this example, or do we make it hard for previous single-run fix users to get started with MR FIX?  Or do we require preserved spaces in the arguments to dedrift instead?

coalsont: It should also be possible to support both `a@b%c@d` and `a b%c d` conventions in this script without much trouble.

coalsont:
The example script that supports both @ and space for the last-level delimiter appears to launch the right commands, running it against master again to be sure (since the freesurfer 6 branch appears to be missing a matlab path for "normalise").

mharms:
Testing a fix now https://github.com/Washington-University/HCPpipelines/pull/97#issuecomment-473027545.
We had to change to using normalise.m in functionhighandpassvariancenormalize.m so that we could compile against $HCPPIPEDIR/global/matlab (and not need to compile against the FIX distribution).  But that broke interpreted mode, because $HCPPIPEDIR/global/matlab was not added to the path.  (I didn't re-test interpreted mode again after we got compiled working, so thanks for catching that).

Examples/Scripts/IcaFixProcessingBatch.sh
```diff
@@ -132,17 +132,23 @@ main() {
...
+	ConcatNames=""
+	ConcatNames="tfMRI_WM_GAMBLING_MOTOR_RL_LR tfMRI_LANGUAGE_SOCIAL_RELATIONAL_EMOTION_RL_LR"  ## Use space (or @) to separate concatenation groups
 
-	# set temporal highpass full-width (2*sigma) to use, in seconds
+	# set temporal highpass full-width (2*sigma) to use, in seconds, cannot be 0 for single-run FIX
 	bandpass=2000
+	# MR FIX also supports 0 for a linear detrend, or "pdX" for a polynomial detrend of order X
+	# e.g., bandpass=pd1 is linear detrend (functionally equivalent to bandpass=0)
+	# bandpass=pd2 is a quadratic detrend
+	bandpass=pd2
```
glasserm: I don't want to default to pd2 at this time.  I don't think that is compatible with all of the different pieces of code and I have not personally seen any data showing its advantages or that it is working.

Examples/Scripts/IcaFixProcessingBatch.sh
mharms: That's a helpful example.  But just to make sure I understand, in this particular case, since `"${cmd[@]}"` includes both the command itself, and its arguments (i.e., we aren't actually passing that into a function/script), you would get the same result if you used `"${cmd[*]}"` to invoke the command, right?  Conversely, for the echo statement, `echo "About to run: ${cmd[@]}"` would work, right?

coalsont: For specifically echo, yes, as shown in the example, it does work to use @.  The result of echo is just a string (like any other command's output), and echo uses every argument you give it.  However, `"${cmd[*]}" would not work, as it would try to run a massively long executable name that contains spaces in it, as it would put all the arguments into a single lump with the actual intended executable name, consider:

```
tim@timsdev:~$ echo hi
hi
tim@timsdev:~$ "echo hi"
echo hi: command not found
```

Executing commands in bash is NOT just based on a simple string (side note, windows command line actually does just pass a single string, and each executable has to parse that string in order to pretend that it was being passed multiple arguments - posix exec functions use an array of strings to start an executable, so its arguments truly are separate at the lowest level, and that separation is under the control of the shell itself).  Bash needs to know that each element of the array should be used separately (first element is the executable, second is the first argument, etc).

Using @ or * as an index inside double quotes is the same as the difference between `"$@"` and `"$*"`: `"$@"` expands to `"$1" "$2" "$3"`, etc, each as separate parts of the command, while `"$*"` expands to `"$1 $2 $3"`, all passed as a single argument containing spaces.

Using a bare `${cmd[*]}` without quotes would work to run the command as long as there aren't any spaces in any arguments, since without quotes, bash splits on whitespace after variable expansion.  `"${cmd[@]}"` is safe even when the arguments contain internal whitespace, because expansion inside quotes occurs after whitespace splitting.

mharms: Thanks. Very helpful.  So, all the instances in the scripts in which we use `main $@` should be `main "$@"` instead, to be "whitespace safe"?

coalsont: Yes, that would be better.

coalsont: Note, though, that if quotes aren't used around variable names containing files inside main itself when executing commands, spaces inside arguments will still break scripts, it will just cause commands to fail, rather than the arguments to main being split incorrectly.

PR #98, opened 2019-03-09
Feature/compiled matlab

mharms:
A set of changes to support compiled matlab with MR-FIX.

This needs to be tested yet, and the compiled version of `functionhighpassandvariancenormalize` has yet to be created, but I'm issuing a PR since that is the easiest way (to my knowledge) to see the differences, and comment on any obvious problems.


mharms:
That and a series of other commits allow for the situation in which someone specifies hp=-1 (no hp filter) in which case there should not be an "_hp" string in the file and directory names.  I tested that functionality this afternoon, and it works.

glasserm:
Do you have anything else to add to this pull request or should I merge it?

glasserm:
In general is there anything else outstanding from your end?

mharms:
Don't merge it yet.  I don't have the compiled MR-FIX working yet.  Have to run, but will resume testing later this evening.

PR #103, opened 2019-04-17
Replaced \-\-verbose option with \-v option in cp, rm, mkdir, and mv …

tbbrown:
Related to an issue reported by a user on the HCP-Users mailing list

The MacOSX versions of the `cp`, `rm`, `mkdir`, and `mv` commands do not support the `--verbose` option. The outputs generated by the use of the `--verbose` option are still useful in debugging. So rather than completely remove the use of that option, I've replaced it with uses of the `-v` option which is supported by both the Linux versions of those commands (according to the man pages) and the MacOSX versions of those commands according to:

* https://ss64.com/osx/cp.html
* https://ss64.com/osx/rm.html
* https://ss64.com/osx/mkdir.html
* https://ss64.com/osx/mv.html

I do not have access to a Mac on which to run tests to verify that this works. So I'm relying on the documentation at the above links.

### DIFF
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -768,7 +768,7 @@ main()
 
 					if [[ ! -e ${DownSampleFolder}/${Subject}.atlas_MyelinMap_BC.${LowResMesh}k_fs_LR.dscalar.nii ]] ; then
 						if [[ ! -z ${MyelinTargetFile} ]] ; then
-							cp --verbose ${MyelinTargetFile} ${DownSampleFolder}/${Subject}.atlas_MyelinMap_BC.${LowResMesh}k_fs_LR.dscalar.nii
+							cp -v ${MyelinTargetFile} ${DownSampleFolder}/${Subject}.atlas_MyelinMap_BC.${LowResMesh}k_fs_LR.dscalar.nii
 						else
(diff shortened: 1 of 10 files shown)


mharms:
We don't use the verbose option for those commands throughout most of the code (so we aren't consistent), but this seems fine in the instances where we were so that it'll work across platforms.

tbbrown:
I belive that the places where `--verbose` was previous added were places where I had some need to see the verbose output for the command in the debugging process in the past. I haven't thought it was necessary to add a `--verbose` or `-v` flat to every `cp`, `mv`, `mkdir`, and `rm` in the entire set of scripts. But I also haven't gone back and removed every place that I added the flag for debugging. We can certainly just remove the `-v` or `--verbose` from all of these commands if that is preferable to replacing the `--verbose` with `-v`.

mharms:
Up to you.  Unless those particular instances esp. need the extra verbose info still at this point, maybe we should just go with removing it.

mharms:
In this same vein, any instances of `mkdir --parents` should probably be changed to `mkdir -p`.  e.g., L473 of MSMall/scripts/SingleSubjectConcat.sh (as raised in a recent post to the HCP-User list).  Not sure if there are others as well...

tbbrown:
Removed the `-v` options. If they have to be added back for some debugging purpose later, then we'll do it at that time.

All `mkdir --parents` have been changed to `mkdir -p`

PR #104, opened 2019-04-17
Make invocations of remove_ext consistent is using full path

tbbrown:
A local (WashU) user has reported that calls to `remove_ext` in the scripts do not work as `remove_ext` is not found on the `PATH`.

There is an outstanding call to change the way FSL (and FreeSurfer too I suppose) tools are invoked in the scripts to rely on the `PATH` mechanism consistently. As of now, this is not the way it is done. Instead we rely on specifying full paths to commands like `${FSLDIR}/bin/fslmaths`. Until we decide to make a complete switch over to using the `PATH` for finding FSL, FreeSurfer, etc. commands, we should probably be consistent and use full paths. 

So I've converted the few places I see `remove_ext` being used relying on the `PATH` to using `${FSLDIR}/bin/remove_ext` instead.

### DIFF
--- a/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
+++ b/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
@@ -120,8 +120,8 @@ log_Msg "mri_watershed --version: ${mri_watershed_version}"
 
 # Start work
 
-T1wImageFile=`remove_ext $T1wImage`;
-T1wImageBrainFile=`remove_ext $T1wImageBrain`;
+T1wImageFile=$(${FSLDIR}/bin/remove_ext ${T1wImage})
+T1wImageBrainFile=$(${FSLDIR}/bin/remove_ext ${T1wImageBrain})
 
 PipelineScripts=${HCPPIPEDIR_FS}
 
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -605,7 +605,7 @@ main()
 		# ----------------------------------------------------------------------
 		log_Msg "Thresholding T1w image to eliminate negative voxel values"
 		# ----------------------------------------------------------------------
-		zero_threshold_T1wImage=$(remove_ext ${T1wImage})_zero_threshold.nii.gz
+		zero_threshold_T1wImage=$(${FSLDIR}/bin/remove_ext ${T1wImage})_zero_threshold.nii.gz
 		log_Msg "...This produces a new file named: ${zero_threshold_T1wImage}"
 
 		fslmaths ${T1wImage} -thr 0 ${zero_threshold_T1wImage}


coalsont:
Personally, I would go the opposite direction, and maybe have the example setup script modify PATH to add these directories (and tell the user their PATH was wrong - if fsl isn't in your PATH, then it isn't fully installed).

The main case I am aware of for specifying a full path is that FSL and freesurfer both contain a version of some (originally FSL?) utilities, in particular bbregister, and one of them works better than the other.  I would prefer that no other utilities are specified with a full path, as it makes the scripts less readable.

However, calling other pipeline scripts probably should remain by full path, as adding subdirectories of the pipelines to PATH should probably be avoided.

tbbrown:
Another case I've encountered for having the full path used in the scripts is when the user account under which the scripts are running has a different version of a tool enabled/installed and set up on their `PATH` for regular interactive use than they want to use for running the pipelines. For example, the user wants their account configured to use FSL v5.0.9 and their interactive commands to use `PATH` to find commands like `fslview`. (Maybe they want to keep using `fslview` instead of `fslview_deprecated`.) But for whatever reason (testing maybe), they want a particular run of a pipeline script to be using FSL v6.0.1 and maybe even a separate run of a pipeline script to be using FSL v5.0.9. 

This is a case where FSL is in your `PATH`, but it's not the FSL you want to use for a particular run of a pipeline script. With careful setting of `PATH` values in the setup scripts, I believe this can be handled.

But when mistakes are made in the setup scripts (e.g. adding `${FSLDIR}/bin` to the end of the `PATH` when maybe it should have been added to the beginning of the `PATH`), then what is likely to happen is that commands in the scripts that use the full path based on `FSLDIR` value end up using binaries and executables from version `x` of FSL and commands in the scripts that rely on normal `PATH` searching end up using binaries and executables from version `y` of FSL. The problem with this is that it often goes undetected because everything still "works". 

It also seems like since it is necessary that people `source` the setup script and people may want/need to simply run a pipeline script interactively, having the setup script alter their `PATH` might not be such a good idea. Assuming that I interactively source the setup script, then if that setup script is appending or prepending things to my `PATH` it's a good bit annoying that it alters my interactive environment and may append or prepend the same thing over and over if I do this multiple times. If the setup script doesn't append or prepend things to the `PATH` and instead completely resets the `PATH` as it needs to be for the pipeline script run, then I'm likely to get really annoyed that I've lost all the other things on my `PATH`.

coalsont:
Interactively inspecting outputs is easily done by having a second terminal window for running things in the user's preferred non-pipeline setup.  I expect very few users are running the pipelines by pasting in one line at a time (and those that are, are likely advanced, and may already be using more than one terminal).  Since the purpose of altering PATH in the setup script would be to force specific versions to be found first, it is obvious that modifications should always be made to the front of the path, so I don't see "mistakes" as a compelling argument (users shouldn't be modifying that part).

Sourcing the same file repeatedly is usually easy to avoid, and even if the user doesn't avoid it, having something on your path redundantly doesn't actually cause a problem (maybe if you got up to 100s of times and the shell wasn't smart enough to notice duplicates, then it could slow things down a little).  It would also be possible for the setup script to compare the front of PATH to the additions it would add, solving this problem entirely.

tbbrown:
@coalsont Your points are all valid.

If we want to go the route of having the setup scripts modify the `PATH` and relying on standard `PATH` search behavior, then I'm inclined to wait to do that "all at once". That is, remove all the uses of `${FSLDIR}/bin/<command>`, `${FREESURFER_HOME}/bin/<command>`, `${CARET7DIR}/<command>`, etc. throughout the scripts. This doesn't seem to be something to do "half-way".

I was thinking that the changeover to relying on standard `PATH` search behavior is a relatively large set of changes, will require a good bit of testing, could potentially break a lot of things, and will require changes to a lot of external scripts that are used to run HCPpipelines scripts. Therefore, I hadn't really intended to have this little PR with only 2 files changed turn into that full set of changes.

mharms:
In both of the scripts that @tbbrown modified, other FSL commands are used (e.g., flirt, applywarp, fslmaths) under the assumption that they in the PATH, so why does it make sense to only change `remove_ext` into `${FSLDIR}/bin/remove_ext`?  How is the rest of the script going to work if `remove_ext` isn't found in the PATH?

More generally, I agree that the inconsistent treatment of the FSL commands is bad practice, and I'm partial to @coalsont suggestions that, longer term, we go in the direction of relying on the PATH setting to find the commands.  (And same should apply to finding `wb_command` in the PATH, rather than using ${CARET7DIR}).  I suspect that we inherited this practice from the FSL scripts themselves, which, IIRC, frequently use the full path approach.  Might be worth asking the Oxford folks why they went with that approach long ago...

tbbrown:
Yep! I was too narrowly focused and just concentrating on the uses of `remov_ext` because those where what the user had reported as a problem. These scripts are already "inconsistent" with the general approach of using the full path to commands and already rely on the `PATH` search mechanism. 

I believe @mharms is correct that the full path practice was inherited from the FSL scripts themselves.

I suggest we just ignore this PR and start a separate effort to make all the HCPpipelines scripts consistently use the `PATH` and not specify full paths. 

Anybody else vote for asking Oxford why they did things the way they did, or do we just not bother because the opinion on this end is that we shouldn't do it that way?

coalsont:
It can't hurt to ask, and we might learn some other conflicts (besides bbregister) to keep an eye out for.

coalsont:
Another side benefit of using PATH to resolve nearly all commands is that if there is a bug in some utility, a patched executable can be substituted in just by modifying the PATH, rather than editing the pipeline scripts or the FSL or freesurfer installation directory.

mharms:
I'm not seeing a `${FSLDIR}/bin/bbregister` for FSL 6.0.1, or any of the FSL 5.0.X versions that I just checked.  Not saying that there aren't potential matching binary/script names between FSL and FS (or other tools), but I'm not sure if `bbregister` is one of them.

coalsont:
Another alternative for this PR is to actually make the edits to the setup script to ensure the PATH finds what we want it to, and then we wouldn't have to worry about inconsistent use of full paths to utilities any more.  Since a different PATH won't break the existing absolute paths, we don't need to go about it in an all-or-nothing fashion.

As for bbregister, I was going from memory of something Matt said a long time ago about testing freesurfer and FSL registrations, maybe for the SBRef to the T1w image.  I haven't personally run into a utility conflict between packages.

coalsont:
Another point: if we make a huge change switching to PATH and deleting all paths to utilities, and then find out something broke, we'd have to revert all of it, while if we start by changing PATH without removing absolute paths everywhere in the scripts, we can see if that by itself causes problems before we put the effort into decluttering scripts.

coalsont:
Talking with Matt, it may also be good to add some code to the setup script to sanity check FSLDIR, HCPPIPEDIR, and other vital things (look for the existence of fslmaths and similar), which could make it more obvious to users when they got a path wrong.

glasserm:
I've asked Alan's opinion and want to see what Doug's experience has been with this approach.

PR #105, opened 2019-04-17
Corrected use of path for interpreted MATLAB usages

tbbrown:
Fixed problem with not properly setting up MATLAB path to include the `${HCPPIPEDIR}/global/matlab` directory when invoking interactive MATLAB.

### DIFF
--- a/MSMAll/scripts/MSMAll.sh
+++ b/MSMAll/scripts/MSMAll.sh
@@ -847,11 +847,12 @@ M_PROG
 					1)
 						# Use interpreted MATLAB
 						mPath="${HCPPIPEDIR}/MSMAll/scripts"
-
+						mGlobalPath="${HCPPIPEDIR}/global/matlab"
+						
 						matlab -nojvm -nodisplay -nosplash <<M_PROG
-addpath '$mPath'; MSMregression('${inputspatialmaps}','${inputdtseries}','${inputweights}','${outputspatialmaps}','${outputweights}','${Caret7_Command}','${Method}','${Params}','${VN}',${nTPsForSpectra},'${BC}','${VolParams}');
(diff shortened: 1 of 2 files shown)


coalsont:
Side note: why do the log_Msg statements happen AFTER the calls to matlab?  If we enable set -e, then when the matlab fails, the matlab call won't get written to the log...

tbbrown:
Yep. It certainly would be better to have the log_Msg statements before the calls to matlab. I've moved them.

mharms:
This looks fine, although I'll mention that there are some inconsistencies with the ICAFIX-related pipeline code, where we (1) use interpreted matlab as the default mode, and (2) in the ReApplyFix* scripts we stopped writing to a separate log file for the compiled matlab output (using the existing stdout/stderr instead).  On top of that, writing log files to ${StudyFolder} is a very non-intuitive place to put them.  Could I suggest that we change the logging to just use the existing stdout/stderr instead?

tbbrown:
What should the default MATLAB run mode be? Interpreted, compiled, ?

Of course, changing the default value for any one of these scripts (`MSMAll.sh`, `ReApplyFixPipeline.sh`, or `ReApplyFixMultiRunPipeline.sh`) will mean we need to check and modify any other scripts that call these scripts. 

Changing the log file output to stdout/stderr would not give us a separate log file for each iteration in the `MSMAll.sh` script. (Notice that in `MSMAll.sh`, the log file name contains an iteration number.) I don't know if it is valuable to have separate MATLAB log files for each iteration. But if it is, we'll still need to redirect the output, just maybe into a different location than `${StudyFolder}`.

glasserm:
As to the questions, I think that the default should be interpreted and ideally Octave and compiled are supported, though Octave support might be more important than compiled.  For logging, I agree separate logfiles are harder to debug and prefer a single logfile.   Perhaps a new pull request can have these changes in it?  Tim C. also has various changes to make to the MSMAllPipeline as a part of this release to make it interact better with MR+FIX.  He and I have discussed these.

PR #109, opened 2019-04-30
support for octave in MSMAll, change to here-string convention and simplify

coalsont:


MSMAll/scripts/MSMAll.sh
```diff
@@ -75,9 +75,10 @@ PARAMETERs are [ ] = optional; < > = user supplied value
    --rerun=TBW
    --reg-conf=TBW
    --reg-conf-vars=TBW
-  [--matlab-run-mode={0, 1}] defaults to 0 (Compiled MATLAB)
+  [--matlab-run-mode={0, 1, 2}] defaults to 0 (Compiled MATLAB)
```
mharms: I thought we were going to try to default to using Interpreted MATLAB across all scripts, for consistency.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -63,9 +63,10 @@ PARAMETERs are [ ] = optional; < > = user supplied value
    --revert-bias-field=<YES | NO> revert the bias field or not
         Requires having run the RestingStateStats pipeline and is not necessary if 
         computing variance normalization above
-  [--matlab-run-mode={0, 1}] defaults to 0 (Compiled MATLAB)
+  [--matlab-run-mode={0, 1, 2}] defaults to 0 (Compiled MATLAB)
```
mharms: Ditto.  Should this default to Interpreted MATLAB?

glasserm:
Agree we should change the comments and make sure that 1 is indeed the default.

mharms:
Also, need to update the `MSMAll/README` file.  (Can copy the relevant portions of `ICAFIX/README`).

Unrelated, but as a minor simplification of the pipelines directory structure, could we move `MSMRemoveGroupDrift/MSMRemoveGroupDrift.sh` into the `MSMAll` folder (or perhaps the `DeDriftAndResample` folder would be more appropriate)?  A search doesn't reveal anything that is actually calling `MSMRemoveGroupDrift.sh`, so that minor reorg wouldn't break anything.

glasserm:
It could go in dedrift.

mharms:
The residual issues mentioned above are resolved in PR #111.

mharms:
@glasserm One lingering issue is that MSMAll/README needs a brief introduction of some sort.  Can you or @coalsont add that?

glasserm:
Okay I added some text.

PR #110, opened 2019-05-03
Fixed variable name typo (eddy_cuda variable)

dicemt:
Removed the not needed underscore in variable name: "g_gpu_EnabledEddy" should be "g_gpuEnabledEddy"


glasserm:
Thanks for the fix.

PR #113, opened 2019-05-18
modify msmall scripts to support MR FIX data

coalsont:


MSMAll/MSMAllPipeline.sh
```diff
@@ -109,61 +121,58 @@ get_options()
...
 			--matlab-run-mode=*)
 				p_MatlabRunMode=${argument#*=}
-				index=$(( index + 1 ))
 				;;
 			*)
 				usage
 				log_Err_Abort "unrecognized option: ${argument}"
 				;;
 		esac
+#		index=$(( index + 1 ))
```
mharms: I'm probably missing something, but isn't `index` incremented as part of the `for` loop construct?  Do we really want to increment it again here?

MSMAll/scripts/MSMAll.sh
```diff
@@ -1239,37 +1213,17 @@ main()
...
-				#       to die with a "too many parameters" error message.
-				log_Msg "Run Matlab command: ${matlab_cmd}"
-				echo "${matlab_cmd}" | bash
+				matlab_function_arguments=("${inputspatialmaps}" "${inputdtseries}" "${inputweights}" "${outputspatialmaps}" "${outputweights}" "${Caret7_Command}" "${Method}" "${Params}" "${VN}" "${nTPsForSpectra}" "${BC}" "${VolParams}")
+
+				matlab_cmd=("${matlab_exe}" "${matlab_compiler_runtime}" "${matlab_function_arguments[@]}")
+
+				#don't log to a separate file, separate log files have never been desirable
+				log_Msg "Run Matlab command: ${matlab_cmd[*]}"
+				echo "${matlab_cmd[@]}" | bash
```
mharms: Why are we invoking compiled matlab different here (by piping to bash) compared to the other instances in the script?

MSMAll/MSMAllPipeline.sh
coalsont: It is commented out.  I can delete it.

MSMAll/scripts/MSMAll.sh
coalsont: Good catch, apparently I lost track of the edits to all the separate matlab sections.

PR #117, opened 2019-05-23
Make dedrift and concat regname optional in DeDriftAndResample

coalsont:
This has not been tested.

DeDriftAndResample doesn't appear to actually detect NONE for the dedrift files option, instead it detects whether the regname and concatregname are the same.  I have therefore made both concateregname and dedrift files optional (if unset, concat regname gets set to regname).

This also makes additional edits to the msmall readme which @mharms and I discussed here:

https://github.com/Washington-University/HCPpipelines/commit/9f003a56b321608cd53a2474dd898b7a172ffe37#commitcomment-33644720

### DIFF
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -60,10 +60,9 @@ PARAMETERs are [ ] = optional; < > = user supplied value
        (e.g. 32@59)
    --registration-name=<regname> String corresponding to the MSMAll or other registration sphere name 
        (e.g. \${Subject}.\${Hemisphere}.sphere.\${RegName}.native.surf.gii)
-   --dedrift-reg-files=</Path/to/File/Left.sphere.surf.gii@/Path/to/File/Right.sphere.surf.gii> 
-       Path to the spheres output by the MSMRemoveGroupDrift pipeline or NONE
-   --concat-reg-name=<regname> String corresponding to the output name of the concatenated registration 
-       (i.e. the dedrifted registration)
+   [--dedrift-reg-files=</Path/to/File/Left.sphere.surf.gii@/Path/to/File/Right.sphere.surf.gii>]
+       Path to the spheres output by the MSMRemoveGroupDrift pipeline
+   [--concat-reg-name=<regname>] String corresponding to the output name of the dedrifted registration
    --maps=<non@myelin@maps> @-delimited map name strings corresponding to maps that are not myelin maps 
        (e.g. sulc@curvature@corrThickness@thickness)
    [--myelin-maps=<myelin@maps>] @-delimited map name strings corresponding to myelin maps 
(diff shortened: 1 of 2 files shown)


glasserm:
This will need to be tested, but also I expect making ConcatRegName optional will break the script if not provided or set to equal the input RegName.

coalsont:
> (if unset, concat regname gets set to regname)

Did this not cover your concern already?

glasserm:
What line is that?  I might have missed it...

coalsont:
https://github.com/Washington-University/HCPpipelines/pull/117/files#diff-29e440b6c98c0b8335ae2638db8ece93R267

I edited what used to be an error check to now set the concatregname if it is empty.

glasserm:
A test would still be a good idea.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -60,10 +60,9 @@ PARAMETERs are [ ] = optional; < > = user supplied value
        (e.g. 32@59)
    --registration-name=<regname> String corresponding to the MSMAll or other registration sphere name 
        (e.g. \${Subject}.\${Hemisphere}.sphere.\${RegName}.native.surf.gii)
-   --dedrift-reg-files=</Path/to/File/Left.sphere.surf.gii@/Path/to/File/Right.sphere.surf.gii> 
-       Path to the spheres output by the MSMRemoveGroupDrift pipeline or NONE
-   --concat-reg-name=<regname> String corresponding to the output name of the concatenated registration 
-       (i.e. the dedrifted registration)
+   [--dedrift-reg-files=</Path/to/File/Left.sphere.surf.gii@/Path/to/File/Right.sphere.surf.gii>]
```
mharms: Shouldn't the Usage statement indicate which arguments have default settings (and what those defaults are)?  Otherwise, how are users supposed to know which are optional?

glasserm:
[] means optional, but noting defaults is probably a good idea.

coalsont:
None of the other optional things that default to "NONE" have their default documented.  Therefore, --dedrift-reg-files should not mention a default.  --concat-reg-name is only different because of how the script chose to use the variables, but conceptually, not specifying it means it doesn't use the dedrifting files and just uses the normal regname for outputs, so it acts like a NONE from the user's perspective.

coalsont:
To put it another way, lets say a user has read the MSMAll readme and knows that they shouldn't dedrift for their case.  They read the usage for --concat-reg-name, see that it says "dedrifted registration" and know they therefore shouldn't use it, but let's say the usage did say "defaults to --registration-name"; now they are confused, why does an option that is only used for dedrifting default to a non-null value?

The answer is, because the script was written sort of strangely, which is unrelated to whether they need to specify the option.  Options and their explanations should be about what the effect of using the option is, not about the script internal variables or other details.

The only practical difference between the script being rewritten to take NONE for --concat-reg-name to mean "don't dedrift", is that the current script can't do a dedrift and then reuse the normal regname for its dedrifted outputs.  Since this is an uncommon use case and would be bad practice even if it worked, I don't think such details are worth mentioning here.

mharms:
That's all fine.  By looking only at the 'diff', I missed the `[]` aspect in the Usage statement, indicating the default arguments.
One other little thing I just noticed though -- It would be nice if `rfMRINames` was replaced with `FIXNames`, and `tfMRINames` with `dontFixNames` throughout the script, to bring the script's internal variable names into alignment with the current (preferred) name of the argument flags.

coalsont:
This has run to completion with a previous script I had tested it with (with one call modified to not provide the now-optional parameters).  I don't plan to edit variable names and retest it.

PR #122, opened 2019-06-28
Qunex debugging

sivcek:
Added a debugging library and loading of the library in Pipelines scripts.


mharms:
So, what happens in the sections of code where we currently explicitly check the exit status of a command, and then print an informative message before exiting?  I'm assuming that this "error trap" takes precedence, and would prevent our existing message from occurring?

mharms:
Have you confirmed this works as expected on Macs?

sivcek:
One option is to turn off the error trap for that section of code.

sivcek:
The trap code was tested on Grace cluster (linux) we've used it on Mac servers in the past.

sivcek:
I just tested it on macOS and it works ok. The trap would be triggered in a case:

``` sh
exitcode="./exit_with_error"
```
But the script would exit as well if `set -e` is turned on, which is the case on most if not all Pipeline scripts.

sivcek:
@mharms 

The following change to the debug.shlib:

```
trap 'on_error $LINENO "$BASH_COMMAND" $?' ERR

on_error()
{   
    if [ ! $usetrap == 'off' ]; then
        me=`basename $0`
        line=($2)
        com=${line[0]}
        arg=${line[@]:1}
        echo
        echo -e "\033[33m##### ERROR: Execution stopped prematurely #####"
        echo "---> script: $me"
        echo "---> stopped at line: $1"
        echo "---> command: $2"
        echo "---> expanded: $com" `eval echo $arg`
        echo "---> exit status: $3"
        echo -e "##### -------------------------------- ##### \033[0m "
        echo
        exit $3
    fi
}
```

would enable setting

```
usetrap=off
```
when you don't want the trap to be used, and then setting it back on with

```
usetrap=on
```

It is possible to get even more fancy and integrate the specific message to be displayed in case of an error within the trap code.

sivcek:
@mharms 
with the change in the function:
``` bash
on_error()
{   
    if [ ! $usetrap == 'off' ]; then
        me=`basename $0`
        line=($2)
        com=${line[0]}
        arg=${line[@]:1}
        echo
        echo -e "\033[33m##### ERROR: Execution stopped prematurely #####"
        echo "---> script: $me"
        echo "---> stopped at line: $1"
        echo "---> command: $2"
        echo "---> expanded: $com" `eval echo $arg`
        echo "---> exit status: $3"
        echo -e "##### -------------------------------- ##### \033[0m "
        echo
        exit $3
    elif [ ! -z $command ]; then
        log_Err_Abort "$command command failed with return code: $3"
        if [ ! $exit_on_error = 'no' ]; then
            exit $3
        fi
    fi
}
```

the current code

``` bash
log_Msg "Creating ${output_file} with fslmaths_cmd: ${fslmaths_cmd}"
${fslmaths_cmd}
return_code=$?
if [ "${return_code}" != "0" ]; then
    log_Err_Abort "fslmaths command failed with return code: ${return_code}"
fi
```

could be changed to

``` bash
command=fslmaths
exit_on_error=no
${fslmaths_cmd}
```

mharms:
This sort of thing seems mostly in @coalsont's wheelhouse to me.  That said, I do think we need to decide if we want to preserve the existing status check functionality (and customized messages) at certain isolated places in the code by finding all those instances and replacing with something along the lines of what @sivcek suggested above.

Also, can you clarify how this `on_error` function interacts with `set -e`.  Superfically, having both seems redundant, but that apparently isn't the case.

sivcek:
@mharms I am happy to add the proposed extension to the `on_error()` function. Finding the places with customized messages is rather easy and if we agree on the form, I can easily implement that myself and include in the pull request.

In terms of the form, we have several options. Two most obvious ones: 
a) whenever the check already exists it only does what the check would do (so no change in output), 
b) on error, the full debugg message as provided with `on_error()` is printed out along with the message provided with the existing code (e.g. `log_Err_Abort "fslmaths command failed with return code: ${return_code}"`).

`on_error()` provides detailed error information, whereas with `set -e` the script just exits. If `on_error()` trapping is used, `set -e` is not really needed, but it does not hurt either.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,45 @@
...
+#   — line number
+#   - command ran
+#   - exit status
+
+trap 'on_error $LINENO "$BASH_COMMAND" $?; exit' ERR
+
+on_error()
+{
+    me=`basename $0`
+    line=($2)
```
coalsont: I would at least put double quotes around `$0` since it is passed to basename.  It looks like `line=($2)` could be fooled by spaces in a folder name, but it isn't clear that bash provides a way around that.  I would probably not separate the command from its arguments if I can't guarantee it is done correctly.  Since the executable part of the command may itself contain a variable, perhaps it should just expand it as `echo "---> expanded: $(eval echo "$2")"` (yes, quotes do nest properly like that).

coalsont:
I would not necessarily expect any scripts that don't currently have set -e to simply work with a `trap ... ERR` that exits.  For instance, grep and which will exit with nonzero status under conditions which may not be fatal to the script.

I would prefer if the debug messages (and anything else) didn't use terminal color codes, as these sorts of things will generally be in log files, where the color codes will probably just look like garbage characters.  An extra blank line to make something stand out from the rest is the approach we have generally used.

I am wary of a shlib defining a function with as short a name as 'ceho' (which could also be typo'ed by accident), in addition to my contention that color codes should be avoided in log files.  In our other shlib files, we have used a prefix on every global defined within the file, so perhaps using the function name `debug_on_error` and similar for any additional things would be more appropriate.

I do use `trap ... EXIT` for cleaning up temporary files, but it looks like having both ERR and EXIT traps set results in both executing (even when the ERR trap does an exit), so that should be okay.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,45 @@
...
+    com=${line[0]}
+    arg=${line[@]:1}
+    echo
+    echo -e "\033[33m##### ERROR: Execution stopped prematurely #####"
```
coalsont: Error messages should probably be printed to standard error, not standard out, and I'd prefer not terminal putting color codes into log files.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,45 @@
...
+#
+#   Function for printing colored output to be used same as echo.
+#
+# Input:
+#
+#   — text to be printed
+
+ceho ()
+{
+    echo -e "\033[31m $1 \033[0m"
```
coalsont: While I'm not fond of color codes anyway, you should at least use $* here instead of $1, so that lazy things like `ceho a b c` work as expected.

coalsont:
> ```shell
>     elif [ ! -z $command ]; then
>         log_Err_Abort "$command command failed with return code: $3"
>         if [ ! $exit_on_error = 'no' ]; then
> ```

log_Err_Abort does in fact do an exit, so the lines after it will not get executed.  Being able to disable the trap (while making sure $? is still set as expected!) should be all that is really needed - some of the informative messages include actions to take to resolve the error.  I don't think the formatting of a generic "this command exited nonzero" message is all that important, so messages without any explanatory information can be replaced with your more verbose messages.  I would consider reworking your messages to use `log_Err`, though, for log uniformity and timestamping.

sivcek:
Tim, thanks for great suggestions! I'll add them to the pull request.

In regards to color coding and output to stdout, this was a conscious decision after some deliberation and testing. Color codes do look weird, when you open the log as a text document, but our experience was that it helped significantly when developing, testing and debugging code, and tracking the progress in the terminal using `tail -f` or just printing the log out. The advantages in our experience outweighted the cons.

coalsont:
You could at least make the coloring optional, such that something like `export HCP_PIPE_COLOR=0` in the setup script would avoid using any color codes.

It does appear that our log.shlib currently only outputs to stdout.  I probably would have implemented it differently (for instance, logging anything with "Err" to both stderr and stdout).  However, the log.shlib does already handle things like the toolname (which now defaults to `basename "$0"`, now that I have added the missing quotes), and the time the error happened, so we should think about whether the error messages from nonzero exit status should really have a significantly different look/behavior from every other kind of error the pipelines give.

sivcek:
Optional coloring makes lots of sense. We can make non-colored output the default. Similarly for outputing any 'Err' log entry to both stderr and stdout.

I do think that it make sense to differentiate between the two error types.

sivcek:
Providing the debugging details w/o the regular logging "prefix" details makes it stand out, easier to notice and process, especially in the terminal environment.

coalsont:
To expand on a point I made earlier, but obliquely: This pull would be less likely to break anything if it didn't source the new debug.shlib in scripts that weren't already `set -e` (I haven't checked whether the scripts you added the `source` line to all already have `set -e`).  Also note that at least one FIX-related script actually does `set +e` to disable it just before sourcing a settings file from the FIX distribution (they used `which` without care), so it would need to be set to temporarily ignore this new trap also.

coalsont:
Defaulting to no coloring sounds good to me.

I talked to Matt about the error formatting, and we think a possible compromise would be to first have one line printed with our `log_Err` function with only a simple "command exited with non-zero status $?" message (which I will update to duplicate the message to both stderr and stdout), followed by the extra info in your current function (without using log_Err for those lines, and with optional coloring on them).  This would allow it to stand out visually, while still making it clear that the pipeline script is what is outputting this message (and when).  It also means that the "script:" and "exit status:" lines would not be needed, as their information is contained in the default formatting added by `log_Err`.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,45 @@
+# Description:
+#
+#   Function for printing debugging information on caught errors.
+#   It makes use of error trapping. To reset the trap, add
+#   end of the script using "trap - EXIT".
```
coalsont: This should probably say `trap - ERR`

global/scripts/debug.shlib
mharms: I asked @tbbrown about this once, and he made a reasonable case for why the error messages should at least go to both stdout and stderr -- namely, people are more likely to look at the end of stdout.

mharms:
I would suggest that any environment variables that we introduce as part of this be prefixed with HCP.  i.e., `HCP_USETRAP` rather than `usetrap` and `HCP_EXIT_ON_ERROR` rather than `exit_on_error`

global/scripts/debug.shlib
coalsont: I edited our log_Err* functions so that they always put something in both stderr and stdout.  This will result in some redundant messages when running existing scripts with parameters missing in a terminal (no log redirection), because currently log_Err duplicates the entire message (while log_Err_Abort uses a generic message on stdout).  A good middle ground may be to put the detailed info only into stderr, as long as both stderr and stdout indicate that something is wrong.

Another option is to actually have the log_Err functions check if stderr is connected to a tty, and change whether they duplicate messages based on that.

coalsont:
These aren't actually environment variables - sourced functions don't propagate to calling another batch script.  This is more like the log_* or opts_* stuff, so I wouldn't say it should be HCP_*.  These things should probably have a prefix, but debug_ would tie it more directly to "debug.shlib".

I think we only really need the "usetrap" setting, though, to completely bypass the entire `trap ... ERR` function body.  The only time it would be useful to have a script show a message when there is a nonzero exit, but continue anyway, is for a pipeline developer working on making a pipeline fit to enable the trap (for which they can temporarily comment out the exit).  If we have both settings, then the main setting should still completely bypass the entire trap function body (as if the trap weren't set).  The second proposal above would not fulfill this.

global/scripts/debug.shlib
mharms: Checking whether stderr is connected to a tty, and modifying the behavior based on that, would be a neat way to avoid redundant messages to the terminal.

mharms:
I've kind of lost track what @coalsont is proposing as regards what we do with the explicit status checks that we already have at miscellaneous places in the scripts.  Are you suggesting that we just not worry about preserving those messages?  Or that we modify each location to locally set `usetrap='off'`?

sivcek:
@mharms I believe that the proposal was to not perserve the messages and go with the more detailed debugging output.

coalsont:
There are likely to be *some* messages in the pipelines that are more detailed or end-user oriented than "this command on this line failed with this code" (for instance, something like "please check your setting of FSLDIR").  I would like the ERR trap function to check a global variable that bypasses the whole thing, such that if we have such a message that we want to preserve, instead of this new output, we have an easy option to do so.  This can be a case-by-case basis, and most of them will probably not be worth keeping.

Any of our existing status checks that don't have a more user friendly message than this ERR trap output can be replaced (ideally, by enabling the trap and removing the existing test).  However, not all of our scripts run cleanly under `set -e` currently, thus we also need something to act as the equivalent of `set +e` for the ERR trap, and conveniently, the same setting (bypass the whole trap function) could serve that purpose also.

As for the formatting, I proposed that the ERR trap first output one line using the pipelines log_Err function so that it contains the time at which it failed and shows the full command that launched the script (and makes it more obvious that the message is coming from the pipelines - see newopts.shlib for how to source log.shlib such that it can know the full script command), then additional lines with any additional information, using whatever formatting makes sense (and with coloring disabled by default).

global/scripts/debug.shlib
coalsont: I have added this tty check.  I doubt there is a cross-platform way to check for stderr and stdout both being redirected to the same file (mac doesn't have /proc), so that case will have duplicate messages, but avoiding duplicates with no redirection seems good enough.

coalsont:
To be clear, I am not suggesting that we identify what existing error messages should be preserved as part of this pull request.  To summarize, the minimal changes I want before approval are:

1) A global variable (or function) to disable the debug trap entirely (and re-enable it again later)
2) Use a `debug_` prefix on all globals (functions, variables) exposed by debug.shlib - `debug_ceho`, `debug_on_error`, `debug_use_trap`, etc.
3) Move all color codes into conditionals (`echo -n` should be useful for this), and please have them off by default (via variable or function, your choice).

Optional things I'd like, but could be added later:
4) Please use `local` on all variables inside functions that should not interact with the sourcing script - this may be a bit redundant on an error handling function that terminates the script, but is good form/habit
5) Source log.shlib and use `log_Err` for the first line of the error handler output, so that we get something like
```
Tue Jul  2 13:49:27 CDT 2019:ReApplyFixPipeline.sh: While running '../../ICAFIX/ReApplyFixPipeline.sh --matlab-run-mode=1':
Tue Jul  2 13:49:27 CDT 2019:ReApplyFixPipeline.sh: ERROR: Study Folder (--path= or --study-folder=) required
```
at the beginning of the error message (which will also give indications in both stderr and stdout when redirected).
6) Don't use unquoted word splitting to try to separate the executable from its arguments, just treat the entire command as one long string (and also note that the executable name may itself use variables, in particular we have been using `$FSLDIR/bin/<whatever>` all over the place).
7) Use "$*" in `debug_ceho` to act more like `echo`

global/scripts/debug.shlib
sivcek: I changed the expanded line to the suggested snippet.

sivcek: Done

sivcek:
In the change to the pull request the following was done:

1/ `debug_enable_trap()` and `debug_disable_trap()`; `HCP__DEBUG_USETRAP=ON` and `HCP_DEBUG_USETRAP=off`
2/ added debug prefix on all functions and variables
3/ all color coding is conditional and there are functions and variables to turn it on and off
4/ local used on all function specific variables
5/ added Log_Err at the beginning of output currently just outputs `"$command command failed with return code: $3"`
6/ For the expanded output it uses the whole command, if `$command` is not set it tries to get it by command splitting.
7/ `$*` is used in expanded echo functions
8/ added verbose echo functions that only print output when `HCP_VERBOSE` is set to TRUE.

global/scripts/debug.shlib
```diff
@@ -2,44 +2,271 @@
...
+debug_enable_exit()
+{
+    HCP_DEBUG_EXIT_ON_ERROR=ON
+}
+
+# Description:
+#
+#   Function to turn off exit upon error trap
+
+debug_diable_exit()
```
coalsont: typo: disable

global/scripts/debug.shlib
```diff
@@ -2,44 +2,271 @@
...
+#   end of the script using "trap - ERR".
 #
 # Input:
 #
 #   — line number
 #   - command ran
 #   - exit status
 
-trap 'on_error $LINENO "$BASH_COMMAND" $?; exit' ERR
+trap 'debug_on_error $LINENO "$BASH_COMMAND" $?; exit' ERR
```
coalsont: This trap contains a hardcoded `exit`, so the debug_disable_exit function doesn't seem like it can do what is described.

global/scripts/debug.shlib
sivcek: Thanks for catching that!

global/scripts/debug.shlib
sivcek: Removed the exit at trap.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,272 @@
...
+#   Function for printing red verbose output.
+#
+# Input:
+#
+#   — text to be printed
+
+verbose_red_echo()
+{   
+    if [ "$HCP_VERBOSE" == "TRUE" ] ; then
+        if [ "$HCP_DEBUG_COLOR" == "TRUE" ] ; then
```
coalsont: After testing `HCP_VERBOSE`, these could call debug_red_echo, etc, instead of repeating their code.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,272 @@
...
+    if [ ! $HCP_DEBUG_USETRAP == 'OFF' ]; then        
+
+        local me=`basename "$0"`
+        local line
+        local startreport = ""
+        local endreport   = ""
+
+        # --- try to identify command if $command is not set
+
+        if [ -z "$command" ]; then
```
coalsont: When is `command` already set before the trap, and why would we trust it before the `BASH_COMMAND` special variable that gets passed in?  `command` is a generic enough name that it might be used by scripts (and not unset after it is used), which seems like it could make the error reporting message wrong.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,272 @@
...
+            line=($(eval echo "$2"))
+            command=${line[0]}
```
coalsont: After a bit of testing, the bulletproof way to do this should be:

```
function debug_first()
{
    echo "$1"
}
...
        local command=$(eval debug_first "$2")
```

I'm not entirely sure how useful it is to report the executable by itself (`wb_command command failed` and `fslmaths command failed` are very nonspecific), and this can still be tripped up by more involved bash syntax, like `LD_LIBRARY_PATH=/blah foo`, though if desired, the function could loop through the arguments and use a regex to skip the leading variable specifications.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,272 @@
...
+#   Function for printing red debug output.
+#
+# Input:
+#
+#   — text to be printed
+
+debug_red_echo()
+{   
+    if [ "$HCP_DEBUG_COLOR" == "TRUE" ] ; then
+        echo -e "\033[31m $* \033[0m"
```
coalsont: Do you actually want the color echos to add a space to both ends of the message?

global/scripts/debug.shlib
sivcek: Good point. Done.

sivcek: Good point. Done.

global/scripts/debug.shlib
sivcek: My mistake. I incorrectly recalled that it was used. Will remove.

global/scripts/debug.shlib
sivcek: Thanks I added it to the code.

global/scripts/debug.shlib
sivcek: Nope. Correcting now.

global/scripts/debug.shlib
```diff
@@ -138,10 +142,7 @@ debug_on_error()
 
         # --- try to identify command if $command is not set
```
coalsont: This comment is obsolete

sivcek: fixed.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,261 @@
...
+    echo "$1"
+}
+
+debug_on_error()
+{    
+    if [ ! $HCP_DEBUG_USETRAP == 'OFF' ]; then        
+
+        local me=`basename "$0"`
+        local line
+        local startreport = ""
```
coalsont: Bash does not allow any unquoted whitespace before or after the = in an assignment:

```
tim@timsdev:~$ function test() { local a= 4; local b =5; }
tim@timsdev:~$ test
bash: local: `4': not a valid identifier
bash: local: `=5': not a valid identifier
```

coalsont:
Can you do a quick functionality test with an intentionally failing command (with and without a variable in arguments and in the path to the command, like ${FSLDIR}/bin/fslmaths)?  That would have caught the whitespace-in-assignment issue.

sivcek:
@coalsont 
I did through testing and cleaned up reporting. The attached script returns with the captured output.
[test.zip](https://github.com/Washington-University/HCPpipelines/files/3691840/test.zip)
test

global/scripts/debug.shlib
```diff
@@ -0,0 +1,261 @@
+# Description:
+#
+#   Function for printing debugging information on caught errors.
+#   It makes use of error trapping. To reset the trap, add
+#   end of the script using "trap - ERR".
+#
+# Input:
+#
+#   — line number
```
coalsont: This dash appears to be unicode, I'd prefer if scripts were strict ascii.

sivcek: Changed all unicode dashes to regular ascii.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,261 @@
+# Description:
```
coalsont: For shlibs, I have started using a convention of `#!/bin/echo This script should not run directly:` for the first line, so that if someone were to make it executable and run it from a modern shell, it does this:

```
tim@timsdev:/mnt/myelin/tim/pipelinedevel/Pipelines/global/scripts$ chmod +x newopts.shlib 
tim@timsdev:/mnt/myelin/tim/pipelinedevel/Pipelines/global/scripts$ ./newopts.shlib
This script should not be run directly: ./newopts.shlib
```

I think shlibs should not have execute permissions, but I have found some in the pipelines that did, so as an extra layer of safety...

global/scripts/debug.shlib
```diff
@@ -0,0 +1,261 @@
...
+}
+
+
+# Description:
+#
+#   Function for printing red debug output.
+#
+# Input:
+#
+#   — text to be printed
```
coalsont: More strange dashes.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,261 @@
...
+{
+    echo "$1"
+}
+
+debug_on_error()
+{    
+    if [ ! $HCP_DEBUG_USETRAP == 'OFF' ]; then        
+
+        local me=`basename "$0"`
+        local line
```
coalsont: This variable is no longer used.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,261 @@
...
+}
+
+
+# Description:
+#
+#   Function for printing red verbose output.
+#
+# Input:
+#
+#   — text to be printed
```
coalsont: More strange dashes.

global/scripts/debug.shlib
sivcek: Added.

global/scripts/debug.shlib
sivcek: Removed.

global/scripts/debug.shlib
sivcek: These are all already corrected.

coalsont:
So, I should have thought of this earlier, but have you checked that all of the scripts that this PR makes source this file are exit-status clean?  Innocuous things like `octavebin=$(which octave)` to test whether something is on $PATH will throw an exit status, which this debug shlib will then abort by default from.  Matt doesn't put `set -e` in his scripts, so we don't know how much of the code is actually exit status unclean.

I won't be as picky on the other PRs, the main reason I was picky here was because it is a new file intended to be used by all scripts going forward.

sivcek:
We've been using this trap for a while now in our fork and had no issues. The best way to fully test this though will be to run a session through all the Pipelines with all the PRs in place.

global/scripts/debug.shlib
coalsont: You marked this resolved, but I don't yet see a commit that removes it.

sivcek: Agh, I'm sorry, it is late over here after a very long day. I deleted the line but forgot to save the file before the last commit.

sivcek:
@mharms do you have any additional comments or reservations or do you approve the PR as well?

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
...
+
+if [[ "$(type -t log_Err)" != "function" ]]
+then
+    source "$HCPPIPEDIR/global/scripts/log.shlib" "$@"
+fi
+
+
+# Description:
+#
+#   Function to turn on error trappimg
```
mharms: Typo.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
...
+
+debug_enable_trap()
+{
+    HCP_DEBUG_USETRAP=ON
+}
+
+# Description:
+#
+#   Function to turn on off error trappimg
```
mharms: Typo

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
+#!/bin/echo This script should not run directly:
+
+# Description:
+#
+#   Function for printing debugging information on caught errors.
+#   It makes use of error trapping. To reset the trap, add
+#   end of the script using "trap - ERR".
```
mharms: FWIW, I don't understand this comment. Add what to the end of what script?

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
...
+#
+# Input:
+#
+#   - line number
+#   - command ran
+#   - exit status
```
mharms: Can you explicitly, as part of the input usage indicate that:
```
$1 - line number
$2 - command run
$3 - exit status
```

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
...
+        echo "         script: $me"                                            
+        echo "stopped at line: $1"                                    
+        echo "           call: $2"
+        echo "  expanded call: $(eval echo "$2")"                            
+        echo "    exit status: $3"                                        
+        echo -e "---------------------------------------------------${endreport}" 
+        echo                                                               
+        
+        # --- check if ok to exit        
+        if [ ! $HCP_DEBUG_EXIT_ON_ERROR = 'OFF' ]; then
```
mharms: `==` for style consistency with above.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
...
+#   Function to turn on verbose reporting
+
+debug_enable_verbose()
+{
+    HCP_VERBOSE=TRUE
+}
+
+# Description:
+#
+#   Function to turn off output coloring
```
mharms: Incorrect description of following function.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,262 @@
...
+# Input:
+#
+#   - text to be printed
+
+verbose_echo()
+{   
+    if [ "$HCP_VERBOSE" == "TRUE" ] ; then
+        echo "$*"
+    fi
+}
```
mharms: Add newline to end of file.

mharms:
I've suggested some minor tweaks for clarity.  And are the "defaults" such that any already existing messages in a script are preserved?  Other than that, I defer to @coalsont if he thinks this is ready.

global/scripts/debug.shlib
coalsont: To be clear, there is both "on off" and "trappimg" to be fixed here.

global/scripts/debug.shlib
coalsont: The new way to disable it is to run `debug_disable_trap`.  The comment should be rewritten.

global/scripts/debug.shlib
sivcek: corrected

global/scripts/debug.shlib
sivcek: corrected

coalsont:
As far as I can tell, the defaults are to switch to the new debug messages on nonzero exit and abort (before any of our existing nonzero exit messages get a chance to display), with coloring disabled.

global/scripts/debug.shlib
sivcek: Removed the confusing part.

global/scripts/debug.shlib
sivcek: Added.

global/scripts/debug.shlib
sivcek: Done.

global/scripts/debug.shlib
sivcek: Corrected.

global/scripts/debug.shlib
sivcek: Added.

global/scripts/debug.shlib
```diff
@@ -19,10 +18,25 @@ then
     source "$HCPPIPEDIR/global/scripts/log.shlib" "$@"
 fi
 
+if [ -z ${HCP_DEBUG_USETRAP} ] ; then
```
coalsont: We could just call the `debug_enable_trap` functions, so that we don't have nother thing referencing the specific magic variable names.

How do you plan to enable coloring in QuNex, by exporting the variables in advance?  If so, you'll need the current -z tests on the color/verbose settings, but the "use trap" and "abort" settings don't need conditionals, since QuNex will use the same defaults.

coalsont: So, I think the debug_disable_trap and debug_enable_trap should be completely under the control of the scripts themselves, because for a section that uses a custom message, the expected pattern is disable, command+message, enable, and if the command is itself a pipeline script, then with the conditionals and the variable flagged for export, it will inadvertently run the second script in no-trap mode.

On the flip side, if you plan to use export to control the coloring/verbosity, it isn't clear that the scripts should ever call the functions to enable or disable color or verbosity (as that would override the QuNex environment).  Thus, the functions for color and verbosity only seem to be useful if we intend to add coloring and verbose option switches to the pipelines.

coalsont: In other words, I think the default setting for "use trap" and "abort on error" should be unconditional.

sivcek: If the defaults are unconditional, then how can be set to something else? The point is to be able to set the environment to one's preference and not overwrite it by default. 

Yes, the coloring and verbosity enabling and disabling functions are there, if a script would need for some reason to change them. But I am ok in removing them completely so that they do not ever change from the environment setting.

coalsont: Only the defaults for the behavior of the debug trap should be unconditional.  Users should not ever mess with the logical effects of the debug trap, because that changes the results of running the script.  The scripts should be in sole control of these things, otherwise weird things can happen.  It is a bad idea for users to be able to set the script to not exit without editing the script (the fact that the exporting by itself would have additional unintended side effects just makes the situation worse).

Changing whether the scripts print more messages or print them in color doesn't affect where the scripts exit or what outputs they make, so those can be user preferences.  I do not feel strongly about whether the functions to set them should be removed.

sivcek: I see your point. Fixed it.

mharms:
Just so I understand, so what is the sequence of events that will occur as a _default_ when `${HCPPIPEDIR}/global/scripts/debug.shlib` is sourced at the front end of a script and any command therein returns with a non-zero exit status?  And how does this interact with whether `set -e` is enabled?

coalsont:
The error trap runs first.  By default, it does an exit, so nothing else can happen.

If the trap doesn't do an exit, then having `set -e` would cause an exit after the trap finishes.  If `set -e` was not set, and the trap doesn't exit, then it goes to the next line of the script, with `$?` still set to whatever the failed command exited with.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,275 @@
+#!/bin/echo This script should not run directly:
+
+# Description:
+#
+#   Function for printing debugging information on caught errors.
+#   It makes use of error trapping. 
+#
+#   To disable error trapping run `debug_disable_trap`
+#   To disable exit uppon trapped error run `debug_disable_exit`
```
mharms: Typo: Uppon

mharms:
And if I'm reading correctly, there is no global way to turn off the trap, or turn off the default to exit on error?  i.e., You have to run `debug_disable_trap` or `debug_disable_exit` individually within each script for which you wish to modify the default behavior?  If so, can you explain again here (I'm sure it was covered at some point above) why we didn't want to make that behavior controllable via a global environment variable...?

coalsont:
The purpose of disabling the trap is for a *script* that contains a custom error message that we want to keep, or is about to source something like the fix settings.sh that isn't exit-status clean, can turn it off temporarily, just like `set +e`.  Personally, I don't know of a use for disabling the exit while keeping the debug message, but it is similar to the way Matt often runs things, except with additional reporting of nonzero exit status.

I do NOT envision the debug trap as something the user should EVER have control over, other than by editing the script.  If you feel strongly that the user should have control of it *without* editing the script, we would need to come up with something more complicated than what we had, because simply allowing an "export" of the existing variable to avoid the defaults causes unintended side effects.

coalsont:
As for the explanation: I don't think it is desirable for a user setting to change the logic of where the scripts stop, because then when a user has a problem, we have to figure out if a previous pipeline the user ran actually had an error they didn't see, because it didn't stop that earlier pipeline from saying "finished".

coalsont:
So, one way Matt debugs pipelines is copy and paste into the terminal window, and this trap, like set -e, kills the bash process, and therefore the terminal (by default on linux, mac keeps the window open).  One thing we could consider is when setting the default, check whether the shell is interactive, and if so, set the default to disabled.

coalsont:
This should work to test for an interactive shell:

```
if [[ "$-" == *i* ]]
then
    HCP_DEBUG_USETRAP=OFF
    HCP_DEBUG_EXIT_ON_ERROR=OFF
else
    HCP_DEBUG_USETRAP=ON
    HCP_DEBUG_EXIT_ON_ERROR=ON
fi
```

I think it should set both to off for an interactive shell because scripts might use the "disable, do something, enable" pattern for the trap itself, which would override the default.  This way, while the messages would still start popping up after such a section, at least it wouldn't kill the shell.

coalsont:
Alternative ideas for not killing interactive shells:
1) check for being interactive in the trap function itself, and always skip the function body when interactive.
2) have a second setting that does the same as the trap disable, and only that one has a default dependent on being interactive, but doesn't have functions to set it, such that scripts should only use the functions, and if someone wants to turn the trap on in an interactive shell, they have to set that variable manually.

coalsont:
Matt suggested that instead of doing the `exit` when the shell is interactive, displaying a message and then using something like `read` to get input from the terminal, effectively acting as a pause.  This could be just a conditional around the current `exit` line, and the trap-related defaults could remain unconditional.  This seems like the best solution for interactively running the script commands.

So:

```
...
        # --- check if ok to exit        
        if [ ! $HCP_DEBUG_EXIT_ON_ERROR == 'OFF' ]; then
            if [[ "$-" == *i* ]]
            then
                echo "execution paused, press enter to continue..."
                read
            else
                debug_red_echo "===> Aborting execution!"
                echo ""
                exit $3
            fi
        fi
```

glasserm:
"execution paused, press enter to continue..."

to

"execution paused due to error, press enter to continue or ctrl+c to halt..."

mharms:
So, once `debug.shlib` is sourced in a script, it would be good hygiene to remove `set -e` from that script, right?

glasserm:
Yes

global/scripts/debug.shlib
```diff
@@ -0,0 +1,275 @@
...
+# Description:
+#
+#   Function to turn off exit upon error trap
+
+debug_disable_exit()
+{
+    HCP_DEBUG_EXIT_ON_ERROR=OFF
+}
+
+# Description:
```
mharms: Can all this coloring related stuff come later, so that we get the definition of `debug_on_error()` (which is the "core" of this shlib) near the top?

global/scripts/debug.shlib
mharms: Also, can we perhaps expand the comments further, for users looking for a simple understanding of how this all work without delving into the code?  e.g.,
```
#   Function for printing debugging information on caught errors.
#   It makes use of error trapping. 
#
#   Default behavior is that error trapping is enabled, 
#   and the script exits upon trapped error with the same status as the failed command.
#   
#   To disable error trapping altogether run `debug_disable_trap`
#      (Note that if `set -e` is set in the calling script, the script would still exit upon error 
#       through that mechanism).
#   To disable exit upon trapped error run `debug_disable_exit`
#       In this case (trap doesn't exit), execution proceeds to the next line of the script, 
#       with `$?` still set to whatever value the failed command errored with.
```

global/scripts/debug.shlib
```diff
@@ -0,0 +1,275 @@
...
+        
+        # --- output a detailed report if usetrap is not turned off    
+        echo                                                               
+        echo -e "${startreport}===> ERROR: Command returned with nonzero exit code"
+        echo "---------------------------------------------------"
+        echo "         script: $me"                                            
+        echo "stopped at line: $1"                                    
+        echo "           call: $2"
+        echo "  expanded call: $(eval echo "$2")"                            
+        echo "    exit status: $3"
```
mharms: Minor quibble, but we alternatively refer to this as "return code" (in the `log_Err` message), "nonzero exit code" in the detailed report, and "exit status" here.

mharms: BTW: With my suggestions/comments this morning, I believe that I'm truly done with this PR, unless feedback on those suggestions/comments requires a further response.

global/scripts/debug.shlib
sivcek: fixed and expanded.

global/scripts/debug.shlib
sivcek: It uses `debug_red_echo` so that has to be defined first. I put all the rest after it.

global/scripts/debug.shlib
sivcek: I'll set it to `exit code` as it is obvious that it is nonzero.

coalsont:
We could have the debug shlib warn if `set -e` is in effect when it is sourced, but that would only catch the cases where `set -e` is before the sourcing line:

```
if [[ "$-" == *e* ]]
then
    log_Warn "set -e is enabled, but debug.shlib was sourced, which is a replacement"
fi
```

Alternatively (or in addition), we could have debug.shlib do `set +e` (or check and warn and then do `set +e`), both when sourced, and whenever `debug_disable_trap` or `debug_disable_exit` are called.

mharms:
I like including a warning if `set -e` is enabled. I tend to favor letting the script still have the ultimate control over the set -/+e behavior.

global/scripts/debug.shlib
coalsont: Functions do not need to be defined before defining other functions that use them, they only need to be defined before the functions or code that use them are called/run.  The debug_on_err function can come first.

mharms: Thanks @coalsont -- that was exactly the question I was going to ask.

coalsont:
So, my thinking is that if a script uses the debug shlib, and then calls `debug_disable_trap`, it is about to do something that isn't clean to exit status, and thus if `set -e` is enabled, it is a mistake.  Better to not just warn about the mistake, but also fix it, so the pipelines don't crash for users just because we forgot a detail.

The script itself always has final control over `set -e` anyway, it would just mean that if it really wanted that behavior, it would have to do `set -e` after every disable.  Since I don't consider it sane to have a script include the debug shlib, disable it, and use `set -e` instead, I don't consider it problematic to need an extra line every time it switches to this insane state.

coalsont:
Wait, I found a wrinkle: `set -eu` will exit on undefined variable, but `trap <...> ERR; set -u` does NOT exit on undefined variable.  So, in fact, we shouldn't even warn for `set -e` (unless we also checked for `set -u` and suppressed the warning then).

coalsont:
In fact, the interaction with `set -u` is kind of weird:
```
tim@timsdev:~$ trap 'echo trap called' ERR
tim@timsdev:~$ set -u
tim@timsdev:~$ echo $asdf
bash: asdf: unbound variable
tim@timsdev:~$ echo $?
1
tim@timsdev:~$ asdfasdasd
asdfasdasd: command not found
trap called
tim@timsdev:~$ echo $?
127
```
So, despite not calling the ERR trap when complaining about an undefined variable, the exit status was nonzero.  Go figure.

mharms:
I'm not quite sure of the implications of the `set -u` stuff.  I defer to whatever you think is best.

coalsont:
So, to allow both `set -eu` and the debug trap to coexist without four lines for the disable/enable pattern, and to play nice with interactive shells, how about this:

1) when sourcing the debug shlib, if `set -e` is set without `set -u`, give a warning
2) also when sourcing the debug shlib, if the shell is interactive, do `set +e`, no warning
3) when disabling the debug trap or exit, if `set -u` is active, do `set +e`, no warning
4) when enabling the debug trap or exit, if *both* the trap and exit are enabled, and `set -u` is enabled and the shell is not interactive, do `set -e`, no warning

global/scripts/debug.shlib
sivcek: Ok. Will change.

mharms:
Ok, I'm completely lost.  My only suggestion would be to include comments that explain what we are doing and why.  And the comment at the top will have to revised as regards the `set -e` stuff.

sivcek:
@coalsont Can I ask you for the final code you would like me to put in, so that I don't make an error in what you would like specifically to do?

global/scripts/debug.shlib
coalsont: To take this further, we can define a defaults function, which can call the enable functions instead of using the magic variable names and values, and then call it at the end of the script:

```
<top of file, shebang>
<comments>
debug_on_error()
{
...
}

debug_defaults()
{
    debug_enable_trap
    debug_enable_exit
    #the ${<var>:-} pattern exists to work properly under set -eu, mac's old bash doesn't have [[ -v <var> ]]
    if [[ "${HCP_DEBUG_COLOR:-}" == "" ]]
    then
        debug_disable_color
    fi
    if [[ "${HCP_VERBOSE:-}" == "" ]]
    then
        debug_disable_verbose
    fi
}
...
<other functions>
debug_defaults
trap 'debug_on_error "$LINENO" "$BASH_COMMAND" "$?"' ERR
```

coalsont:
Maybe I should edit it directly, would adding a commit to your branch be a problem for you?

global/scripts/debug.shlib
sivcek: Done.

sivcek:
No, go ahead, just work from the latest change and when you're done, I believe I can pull your changes back?

sivcek:
Edits from maintainers are allowed.

coalsont:
Yeah, github will allow me to push directly the the branch for this PR, since you left that option enabled.

sivcek:
Perfect.

sivcek:
Great, thanks! @coalsont Do you mind if I add a space after `#` in comments. It improves readability.

coalsont:
Sure.  I am giving it another once over, but I think it is done now (except for testing).

sivcek:
Ok. Let me know once done and I'll add spaces. Don't mind if you do if you're doing any other changes.

coalsont:
Well, actually - if you could convert the sourcing lines in all the other files to NOT source log.shlib first, and then source debug.shlib like `source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions`, that would be an improvement, as it will give log.shlib more information about the script arguments (and add some space-safety).

coalsont:
Specifically, this makes a difference in log_Err_Abort:

https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/log.shlib#L2
https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/log.shlib#L159

sivcek:
Ok. So I remove all:

```
source ${HCPPIPEDIR}/global/scripts/log.shlib  # Logging related functions
```

and replace them with:

```
source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions
```

coalsont:
Actually, this is in log_Err too, so any multi-call log_Err things would be a problem:

https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/log.shlib#L171

Maybe it should be removed from log_Err because of this multi-line use case - or it could keep track of having logged the full command line and only do it once.

coalsont:
Yes, that is what I meant.  I can edit log.shlib to handle this pattern better:

```
log_Err "line 1"
log_Err "line 2"
log_Err_Abort "line 3"
```

sivcek:
Ok will do.

coalsont:
How about I edit this on your branch before you commit those changes to the sourcing lines?

mharms:
Not sure why we want to remove sourcing of log.shlib, but if we do can we make the comment with debug.shlib be `# Debugging functions; also sources log.shlib`

coalsont:
Wait, your version of log.shlib doesn't have the full command logging, so you must be behind our master somewhat.  What do you suggest?

coalsont:
@mharms removing because it is redundant, but sure, the comment can be updated

coalsont:
So, the mode check shlib PR *is* based off a recent-enough master to have the full command error logging changes, and it makes use of that multiline error code pattern.  Perhaps I should make the edits to log.shlib there.

sivcek:
Will add '; also sources log.shlib"

sivcek:
Added.

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -96,8 +96,8 @@
 set -e
```
coalsont: As long as we are editing all these files, we should also remove `set -e` lines (but leave `set -eu` lines).

sivcek: So we just remove `set -e` and not replace them. However, if there is a `set -eu` line, we leave it?

coalsont: Right.  When I searched, the only script that currently has `set -eu` is RSNregression.sh, which you haven't edited, so you shouldn't have to worry about that case.

mharms:
I thought I approved this previously.  Strangely, it is listing me simultaneously as having "approved these changes" and yet being "requested for review".

I took a quick look, and a see a bunch of `set -e` lines above the `source .../debug.shlib` lines.  I thought that we wanted to remove those `set -e` lines.

DiffusionPreprocessing/DiffPreprocPipeline.sh
sivcek: Done. In a few cases that had set +e / set -e in code I changed it to debug_disable_trap and debug_enable_trap

sivcek:
@mharms It was really late at night and I only got to it today. All set -e are removed now.

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -36,26 +34,19 @@ set -e # If any command exits with non-zero value, this script exits
 usage()
 {
 	cat <<EOF
-
```
coalsont: Why did you remove these blank lines in the usage info?  Please put them back in.

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -36,26 +34,19 @@ set -e # If any command exits with non-zero value, this script exits
 usage()
 {
 	cat <<EOF
-
```
coalsont: Same here, these blank lines were intentional formatting.

ICAFIX/ReApplyFixMultiRunPipeline.sh
sivcek: I did not even know they were taken out. I had to re-merge as the code for FIX in my branch and master was not in sync for some reason.

ICAFIX/ReApplyFixPipeline.sh
sivcek: same here.

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -40,6 +38,7 @@ usage()
 ${g_script_name}: ReApplyFix Pipeline for MultiRun ICA+FIX
 
 This script has two purposes (both in the context of MultiRun FIX):
+
```
coalsont: This blank line appears to have been added by you.

ICAFIX/hcp_fix
```diff
@@ -199,7 +197,7 @@ if [ ! -z "${log_debugOn}" ] ; then
 fi
 
 #grab melodic from $PATH by default, don't hardcode it with respect to $FSLDIR
-#we need to do "if which ..." because the script currently uses set -e
+#we need to do "if which ..." because the script currently uses ERR trapp
```
coalsont: typo: trapp

ICAFIX/ReApplyFixMultiRunPipeline.sh
sivcek: Fixed.

ICAFIX/hcp_fix
sivcek: Fixed

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -93,11 +93,10 @@
 
 # Set up this script such that if any command exits with a non-zero value, the 
 # script itself exits and does not attempt any further processing.
```
mharms: Ok, I hate to suggest further changes on this PR, but for the scripts that had these comments preceding `set -e`
```
# Set up this script such that if any command exits with a non-zero value, the 
# script itself exits and does not attempt any further processing.
```
now that `set -e` is removed, shouldn't those comment lines be removed as well?  Otherwise, those comments don't seem to apply to anything.

ICAFIX/ApplyHandReClassifications.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 # If any command exits with non-zero value, this script exits
```
mharms: Orphaned comment.

DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh
```diff
@@ -65,11 +65,10 @@
 
 # Setup this script such that if any command exits with a non-zero value, the 
 # script itself exits and does not attempt any further processing.
```
mharms: Orphaned comments.

DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
```diff
@@ -65,11 +65,10 @@
 
 # Setup this script such that if any command exits with a non-zero value, the 
 # script itself exits and does not attempt any further processing.
```
mharms: Orphaned comments.

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -64,11 +64,10 @@
 
 # Set up this script such that if any command exits with a non-zero value, the 
 # script itself exits and does not attempt any further processing.
```
mharms: Orphaned comments.

PreFreeSurfer/PreFreeSurferPipeline.sh
```diff
@@ -151,7 +151,6 @@
 
 # Setup this script such that if any command exits with a non-zero value, the
 # script itself exits and does not attempt any further processing.
```
mharms: Orphaned comments.

DiffusionPreprocessing/DiffPreprocPipeline.sh
sivcek: Yes, you are right. It is a pain though ... I'll get to it.

sivcek: Hmmm ... I found just two files with those comments.

ICAFIX/ApplyHandReClassifications.sh
sivcek: One with those.

DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh
sivcek: Three with these.

sivcek:
Let me look for other variations...

global/scripts/debug.shlib
```diff
@@ -0,0 +1,328 @@
...
+        echo "stopped at line: $1"                                    
+        echo "           call: $2"
+        echo "  expanded call: $(eval echo "$2")"                            
+        echo "      exit code: $3"                                        
+        echo -e "---------------------------------------------------${endreport}" 
+        echo                                                               
+        
+        # --- check if ok to exit        
+        if [ ! $HCP_DEBUG_EXIT_ON_ERROR == 'OFF' ]; then
+            if [[ "$-" == *i* ]]
```
mharms: Add comment:
`# Check if interactive shell`

sivcek: Updated.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,328 @@
...
+    then
+        debug_disable_verbose
+    fi
+}
+
+# == Logging functionality
+
+if [[ "$(type -t log_Err)" != "function" ]]
+then
+    source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions; also sources log.shlib
```
mharms: Looks like an over-aggressive find-replace.  Aren't we supposed to source log.shlib here?

sivcek: It was not find-replace. As per @coalsont instructions I replaced all log.shlib sourcing with debugging ones.

mharms: Well, I  believe that this is the one instance of `source log.shlib` that we need to keep (in `debug.shlib` itself).  Otherwise, we have an infinite loop of `debug.shlib` continually sourcing itself.

sivcek: :-) I completely overlooked that it was in the debug.shlib itself. Fixed.

global/scripts/debug.shlib
```diff
@@ -0,0 +1,328 @@
...
+        echo "---------------------------------------------------"
+        echo "         script: $me"                                            
+        echo "stopped at line: $1"                                    
+        echo "           call: $2"
+        echo "  expanded call: $(eval echo "$2")"                            
+        echo "      exit code: $3"                                        
+        echo -e "---------------------------------------------------${endreport}" 
+        echo                                                               
+        
+        # --- Check if interactive shell
```
mharms: The comment `# --- Check if interactive shell` should be associated with the `if [[ "$-" == *i* ]]` line.  Not above the `if [ ! $HCP_DEBUG_EXIT_ON_ERROR == 'OFF' ]; then` line, which implies that the entire `if [ ! $HCP_DEBUG_EXIT_ON_ERROR == 'OFF' ]; then` is about an interactive shell.

sivcek: Changed.

DiffusionPreprocessing/DiffPreprocPipeline.sh
coalsont: By default, debug.shlib does effectively the same thing as set -e, but with more verbosity, so that comment still applies, as long as it is near the debug sourcing.  I don't care if these comments exist.

sivcek: I weeded them out. I'd rather not be putting them in everywhere.

PR #123, opened 2019-07-04
add deletion of unneeded and intermediate files to FIX scripts

coalsont:


glasserm:
As to your question about ${hpStr} vs _hp${hp} these are equivalent.

glasserm:
"${fmrihp}" is not equivalent and represents a combination of the fMRIName and the HighPass string.

coalsont:
I was never asking about whether the fMRI name should be in the file name twice or zero times.  Let me be more verbose: ${fmrihp} vs ${fmri}${hpStr} vs ${fmri}_hp${hp}.

${hpStr} appears to sometimes be the empty string (hp "0" in single-run or something), and I think ${fmrihp} can be equal to ${fmri} under similar circumstances.  This is what I was asking about in the first place.

glasserm:
I guess I have to merge this to edit it.

mharms:
The preferred way to do this is to fetch Tim's branch into your local git, and we iterate on that until we have it ready to go.  This would be easier if, in the future, @coalsont pushes the branch into the Washington-University repo (in which case, I believe that a `git fetch origin` will grab it).

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -415,22 +452,16 @@ main()
 	if [ -z "${9}" ]; then
 		MotionRegression="${G_DEFAULT_MOTION_REGRESSION}"
```
mharms: If using the default setting, `MotionRegression` will be set to `FALSE`, and never gets converted to a boolean (as is neccessary for fix_3_clean).  In the previous code, the conversion to boolean happened after MotionRegression was already defined, regardless of whether we were setting to `G_DEFAULT_MOTION_REGRESSION`

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -415,22 +452,16 @@ main()
...
 	else
-		MotionRegression="${9}"
+		MotionRegression=$(interpret_as_bool "${9}")
+	fi
+
+	local DeleteIntermediates
+	if [ -z "${10}" ]; then
+		DeleteIntermediates="${G_DEFAULT_DELETE_INTERMEDIATES}"
```
mharms: Ditto -- if using the defaults, `DeleteIntermediates` will be `FALSE`, rather than a boolean, which doesn't work with the later code.

ICAFIX/ReApplyFixMultiRunPipeline.sh
coalsont: These should be resolved by f137e43d94bbafbdb7aced00dae753bdbbd2aca8, including in single-run reapply.  The hcp_fix* versions should work as-is, because their "interpret_as_bool" actually outputs strings rather than numbers, because that is what they expected (and I don't want to reengineer their existing boolean handling for this minor change).

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -938,6 +969,15 @@ main()
     # Remove the 'fake-NIFTI' file created in fix_3_clean for high-pass filtering of the CIFTI (if it exists)
 	$FSLDIR/bin/imrm ${concatfmrihp}.ica/Atlas
  
+	#always delete things with too-generic names
+	rm -f ${concatfmrihp}.ica/filtered_func_data.nii.gz ${concatfmrihp}.ica/Atlas.dtseries.nii
+
+	#delete highpass intermediates
+	if [[ "$DeleteIntermediates" == "1" ]]
+	then
+		rm -f ${concatfmri}.nii.gz ${concatfmri}_hp${hp}.nii.gz ${concatfmri}_Atlas.dtseries.nii ${concatfmri}_Atlas_hp${hp}.dtseries.nii
```
mharms: The usage statement says that `--delete-intermediates` will delete the high-pass filtered files, yet here we are deleting the concatenated NIFTI and CIFTI files (prior to HP) as well. That seems a reasonable thing to do if the option is set, but means that we should change the usage statement to be more precise.  Also, if we are deleting the other concatenated files, shouldn't we be deleting the _cleaned_ concatenated files as well, if this option is set?

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -321,14 +344,14 @@ main()
...
 
 	local DeleteIntermediates
 	if [ -z "${9}" ]; then
 		DeleteIntermediates="${G_DEFAULT_DELETE_INTERMEDIATES}"
 	else
-		DeleteIntermediates="${9}"
+		DeleteIntermediates=$(interpret_as_bool "${9}")
 	fi
 
 	# Turn MotionRegression into an appropriate numeric value for fix_3_clean
```
mharms: Can't we delete this section, as long as we put the setting of `MotionRegression` and `DeleteIntermediates` through the `interpret_as_bool` function as well, in the default case?

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -663,7 +686,11 @@ main()
 			$FSLDIR/bin/imrm ${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}${hpStr}
 		fi
 		$FSLDIR/bin/imrm ${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}${hpStr}.ica/filtered_func_data
-        rm ${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}${hpStr}.ica/Atlas_hp_preclean.dtseries.nii
+		#TSC: should these be ${hpStr} or _hp${hp}?
```
mharms: The key distinction is that this script supports the use of hp<0 (no filtering), in which case `${hpStr}` gets set earlier to the empty string.

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -663,7 +686,11 @@ main()
...
+		rm -f "${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}${hpStr}.ica/Atlas_hp_preclean.dtseries.nii" "${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}${hpStr}.nii.gz"
```
mharms: L690 has a bug.  We can delete `Atlas_hp_preclean.dtseries.nii`, but `${fMRIName}${hpStr}.nii.gz` is already deleted above, and importantly, is only deleted there conditional on hp>=0.  Otherwise, if hp<0 applies, we would be deleting the _original_ volume (since `${hpStr}` will be the empty string in that condition).

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -663,7 +686,11 @@ main()
...
+	else
+		#don't leave this file with a hard to interpret name - again, hpStr?
+		mv -f "${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}${hpStr}.ica/Atlas_hp_preclean.dtseries.nii" "${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas_hp${hp}.dtseries.nii"
```
mharms: For the same reasons as above, this needs to be conditional on hp>=0 (otherwise we are creating a file named "_hp${hp}" that wasn't actually filtered.

ICAFIX/hcp_fix
```diff
@@ -119,31 +124,41 @@ fi
...
         ;;
-esac
+    esac
+}
+
+doMotionRegression=$(interpret_as_bool "$3")
+# 5/20/2019: For the time being, do_motion_regression=FALSE doesn't work properly with single-run FIX,
+# so don't allow that mode of operation
+# https://github.com/Washington-University/HCPpipelines/issues/108
+if [[ "$doMotionRegression" == "FALSE" ]]
```
mharms: Shouldn't the check here be against 0, since it was converted to boolean at L142?

coalsont: We can't be consistent with ReApply without rewriting more of the code, which I don't want to do right before a release.  By the way, both hcp_fix* scripts use TRUE/FALSE.

ICAFIX/ReApplyFixPipeline.sh
coalsont: I think Matt saw and fixed this in 4eedd5ed883a2e2972bae49e4abaee5394c43c2d and 5ee0ad8fa9d29added87b5792475bf5c3d00c0df.

ICAFIX/hcp_fix
```diff
@@ -378,4 +393,16 @@ log_Msg "Done renaming files"
 # Remove the 'fake-NIFTI' file created in fix_3_clean for high-pass filtering of the CIFTI (if it exists)
 $FSLDIR/bin/imrm ${fmrihp}.ica/Atlas
 
+#delete things with too-generic names
+rm -f ${fmrihp}.ica/filtered_func_data.nii.gz ${fmrihp}.ica/Atlas.dtseries.nii
+
+#optional cleanup of hp intermediates
+if [[ "$DeleteIntermediates" == "TRUE" ]]
+then
+    rm -f ${fmrihp}.ica/Atlas_hp_preclean.dtseries.nii ${fmri}_hp${HighPass}.nii.gz
```
mharms: This script does not have a `${HighPass}` variable.  It is called `${hp}`, or use `${hpStr}` defined at L367.  Regardless, removal of the .nii.gz needs to be conditioned on hp>=0.

ICAFIX/hcp_fix
```diff
@@ -378,4 +393,16 @@ log_Msg "Done renaming files"
...
+else
+    #even if we don't delete it, don't leave this file with a hard to interpret name
+    mv -f ${fmrihp}.ica/Atlas_hp_preclean.dtseries.nii ${fmri}_Atlas_hp${HighPass}.dtseries.nii
```
mharms: Again, variable is `${hp}` and needs to be conditioned on hp>=0.

ICAFIX/hcp_fix_multi_run
```diff
@@ -771,6 +785,15 @@ log_Msg "Done renaming files"
 # Remove the 'fake-NIFTI' file created in fix_3_clean for high-pass filtering of the CIFTI (if it exists)
 $FSLDIR/bin/imrm ${concatfmrihp}.ica/Atlas
 
+#always delete things with too-generic names
+rm -f ${concatfmrihp}.ica/filtered_func_data.nii.gz ${concatfmrihp}.ica/Atlas.dtseries.nii
+
+#delete highpass intermediates
+if [[ "$DeleteIntermediates" == "TRUE" ]]
+then
+    rm -f ${concatfmri}.nii.gz ${concatfmrihp}.nii.gz ${concatfmri}_Atlas.dtseries.nii ${concatfmri}_Atlas_hp${hp}.dtseries.nii
```
mharms: Ditto this comment https://github.com/Washington-University/HCPpipelines/pull/123/files#r300772789 that I had for this analogous section in ReApplyFixMultiRunPipeline.sh

ICAFIX/hcp_fix
coalsont: I don't know how this got through testing, as I tested with hp 2000 and looked for the file in question after running it, and didn't find it.  Should be fixed by 9aea92921075857b7f6041ef9d2b6b93ee6c8698.

ICAFIX/hcp_fix
coalsont: See 9aea92921075857b7f6041ef9d2b6b93ee6c8698.

coalsont:
Thanks for looking through this.

ICAFIX/ReApplyFixPipeline.sh
coalsont: See 4eedd5ed883a2e2972bae49e4abaee5394c43c2d.

ICAFIX/ReApplyFixPipeline.sh
coalsont: Stylistic choice, I prefer to only assign to any variable once, to make it conceptually easier to track, though this situation is borderline.  However, putting both cases through interpret_as_bool may be a good idea for anyone who edits a default.

mharms:
Happy to help, but @glasserm can we please not do it this way again, in which we have to navigate between the original PR and what are conceptually edits to the same PR made within the master branch itself?  It is much too difficult to keep track of what is going on, and checking whether the "final product" is what it needs to be.  Thx.

ICAFIX/hcp_fix
mharms: Please feel free to resolve any comments that are no longer an issue, based on either @glasserm's edits in the master branch, or your recent additions.

coalsont: I don't know for certain if Matt's (or my) commits fully resolve the issues you raised.  You identified at least one issue in a later commit - I was thinking you should be the one to mark them as resolved when appropriate, as you raised the issues.

ICAFIX/ReApplyFixPipeline.sh
mharms: I'm not sure if my main point got across here.  Now that both cases go through `interpret_as_bool`, why should ReApplyFixPipeline.sh have a section of code after this (L357-380) in which it checks again whether `MotionRegression` and `DeleteIntermediates` are appropriately set to 0 or 1?  We don't have that bit of redundancy in ReApplyFixMultiRunPipeline.

ICAFIX/hcp_fix
mharms: Normally, I wouldn't mind doing that, but in this case it is a royal pain.  Is there a way in GitHub to get a consolidated list of the accumulated changes from prior to when the PR was merged to now (the current master branch)?

ICAFIX/ReApplyFixPipeline.sh
coalsont: Indeed I had missed the lines below the one you attached the comment to, as I only looked at the comment in the conversation tab.  Should be resolved by e9eaaff7abcdeeb565596c76e2e6ccd7cacca818.

ICAFIX/hcp_fix
coalsont: It's not obvious how to do so, but here it is:

https://github.com/Washington-University/HCPpipelines/compare/0f37e01ebec91f711bacf4c04c4abbf55a59af3e...master

What I did to get this is I went to the commit before the PR, clicked "browse files", then clicked "new pull request" (yes, seriously), and then finally, clicked the "change the base" link in the middle of the empty diff.

mharms: That's handy.  I'll take a look at that, or an updated version thereof, once Matt has had a chance to catch up later.

glasserm:
I think everything got fixed by my commits that Mike identified (and I described in the associated e-mail) and the one thing I hadn't identified (undefined HIghPass) was fixed.  I would have rather edited the pull request, but I couldn't seem to do that in the GitHub web interface a it seemed easier to just make the edits rather than describing them in this case.  If someone tells me how to edit a pull request in the GitHub interface, I'm happy to do that in the future.

coalsont:
I'm getting this message while running ReApplyMulti-run on yesterday's run of hcp_fix_multi_run (which had delete intermediates set to true):

```
Fri Jul  5 16:07:49 CDT 2019:ReApplyFixMultiRunPipeline.sh: Run interpreted MATLAB/Octave (octave-cli -q --no-window-system) with command...
Fri Jul  5 16:07:49 CDT 2019:ReApplyFixMultiRunPipeline.sh: addpath('/usr/local/fix'); addpath('/usr/local/fsl/etc/matlab'); addpath('/basedir/Pipelines/global/matlab'); addpath('/basedir/Pipelines/ICAFIX/scripts'); fix_3_clean('.fix',0,0,-1,0);
Image Exception : #63 :: No image files match: filtered_func_data
No image files match: filtered_func_data
TR = [](0x0)
ans = unaggressive cleanup
Fri Jul  5 16:08:31 CDT 2019:ReApplyFixMultiRunPipeline.sh: Done running fix_3_clean
```

Is this expected?

mharms:
@glasserm There is this issue that I raised on hcp_fix: https://github.com/Washington-University/HCPpipelines/commit/5ee0ad8fa9d29added87b5792475bf5c3d00c0df#r34201455.  I don't know about editing a PR in GitHub directly -- perhaps that would have been possible if @coalsont had pushed the branch to the Washington-University repo?  I think that is the key -- we need to push dev branches to the Washington-University repo/remote (since we all have permission to do so), rather than (or at least in addition to) our personal forks, at which point it becomes easy to fetch the branch, check it out locally, edit it, and then push it back up to the repo.

coalsont:
I also disabled edits to my pr branch from maintainers, otherwise it may have been possible anyway.  I will try to ensure that I push PR branches to the main repo.

mharms:
@coalsont Not finding filtered_func_data is a definite problem, and is arising because `--delete-intermediates` resulted in the deletion of ${concatfmrihp}.  But, as with ReApplyFix, if DoVol=0, filtered_func_data is presumably only used to extract the TR by fix_3_clean, which ironically is irrelevant here because we aren't filtering within fix_3_clean in the multi-run case anyway.

The problem is that this conditional
```
	if [[ -f ${ConcatNameNoExt}_Atlas${RegString}_hp${hp}.dtseries.nii ]]; then
		log_Warn "${ConcatNameNoExt}_Atlas${RegString}_hp${hp}.dtseries.nii already exists."
		if (( DoVol && $(${FSLDIR}/bin/imtest "${ConcatNameNoExt}_hp${hp}") )); then
			log_Warn "$($FSLDIR/bin/imglob -extension ${ConcatNameNoExt}_hp${hp}) already exists."
		fi
		log_Warn "Using preceding existing concatenated file(s) for recleaning."
```
wasn't constructed in a way to force all the concat and filtering code to re-run if the `${ConcatNameNoExt}_hp${hp}` file doesn't exist.

So, that needs to be changed in the context of the addition of `--delete-intermediates` to hcp_fix_multi_run, but we'd want to do it in a manner that only regenerates the file in the DoVol=1 condition. In the DoVol=0 condition, we'd presumably just want to use a non-Hp'ed file as as stand-in for filtered_func_data simply so that fix_3_clean won't crash.

glasserm:
We will link to the concatenated cleaned files, as that should always exist.  We won't be deleting that file because downstream may pipelines use it for some things.

mharms:
Actually, looking at fix_3_clean and @coalsont's pasted log output, perhaps this line in fix_3_clean is harmless
```
[grot,TR]=call_fsl('fslval filtered_func_data pixdim4'); TR=str2num(TR)
```
because in the MR-FIX case, hp=-1, so the TR isn't needed for filtering the CIFTI, and the only other place that TR appears is
```
confounds=[];
if domot == 1
  confounds = functionmotionconfounds(TR,hp);
end
```
and `functionmotionconfounds` itself only needs the TR if hp>0.

So, while leaving everything "as is" may not be elegant, and might cause some concern for those that notice the same message that @coalsont did, it appears to not actually be a "definite problem" as I stated originally.

ICAFIX/ReApplyFixMultiRunPipeline.sh
coalsont: I'm not sure what the usage should say, I wasn't involved in deciding what gets deleted.  Feel free to edit it.

glasserm: We should not delete the concatenated cleaned file, as that gets used.

coalsont:
We have fixed this error by using the _clean file when DoVol is 0, see e3b7c434b07243cd3fb33503d20c6b4195dcc483 and c824c5c41febc3ebf026c07a16000681ab7fc7e1.

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: Ok, the concatenated cleaned file is preserved, but I edited the usage statement in what I'll be pushing shortly.

mharms:
Ok, see some tweaks that I made with f2b9015c0f2 and e2cda9b3ce898.  With that, I think my review of what started as PR #123 is done.

glasserm:
I think this is all okay.  Mike do you think we need to test again before a release?

mharms:
If @coalsont is willing to run his test scripts one more time, that would be great.

coalsont:
I am having issues with VPN to the lab.

coalsont:
I figured out the VPN issue.  The current master (fce2ce895f6c6271768d343af3721e67e5d05769) appears to work with the delete intermediates turned on.

Issue #124, opened 2019-07-10
fMRIVolume OneStepResampling.sh is breaking the TR if input data is not in seconds


cmpetty:
When data is passed to OneStepResampling.sh that has TR in ms it is breaking the files afterwards because the time_units are not being accounted for.

You're saving the TR, then re-setting in later with fslmerge, but fslmerge requires TR in seconds.  

We found that our files after this step had TR set to 1000.00 seconds, instead of 1 sc, or 1000.00 ms 

`
########################################## DO WORK ########################################## 

#Save TR for later
TR_vol=`${FSLDIR}/bin/fslval ${InputfMRI} pixdim4 | cut -d " " -f 1`
TR_units=`${FSLDIR}/bin/fslval ${InputfMRI} time_units | cut -d " " -f 1`
NumFrames=`${FSLDIR}/bin/fslval ${InputfMRI} dim4`

#added to account for time units
if [ ${TR_units} = "s" ] ; then
        # this is fine
        continue
elif [ ${TR_units} = "ms" ] ; then 
        TR_vol=$( perl -e "print $TR_vol / 1000.0");
elif [ ${TR_units} = "us" ] ; then
        TR_vol=$( perl -e "print $TR_vol / 1000000.0");
fi
`

`
# Merge together results and restore the TR (saved beforehand)
${FSLDIR}/bin/fslmerge -tr ${OutputfMRI} $FrameMergeSTRING $TR_vol
${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol`

glasserm:
I assume we would use bc instead of adding a perl dependency, but this does seem worth fixing.

cmpetty:
> I assume we would use bc instead of adding a perl dependency, but this does seem worth fixing.

Good point .. i changed mine to 
```
118 if [ ${TR_units} = "s" ] ; then
119         # this is fine
120         continue
121 elif [ ${TR_units} = "ms" ] ; then
122         TR_vol=$( echo "scale=4;${TR_vol} / 1000.0" | bc -l );
123 elif [ ${TR_units} = "us" ] ; then
124         TR_vol=$( echo "scale=4;${TR_vol} / 1000000.0" | bc -l );
125 fi
```

mharms:
@cmpetty Would you be willing to create a PR for this, including finding other instances in the code base where we set the TR as part of `fslmerge`?  Thanks.

[...]

PR #125, opened 2019-07-12
add logic to adjust TR to seconds for fslmerge -tr commands

cmpetty:
added logic for TR in any files that used fslmerge, since it is expecting TR in seconds only

### DIFF
--- a/ICAFIX/ReApplyFixMultiRunPipeline.sh
+++ b/ICAFIX/ReApplyFixMultiRunPipeline.sh
@@ -646,7 +646,19 @@ main()
 		# If ${RegString} is empty, the movement regressors will also automatically get re-filtered.
 		
 		tr=`$FSLDIR/bin/fslval $fmri pixdim4`  #No checking currently that TR is same across runs
-		log_Msg "tr: $tr"
+		TR_units=`${FSLDIR}/bin/fslval $fmri time_units | cut -d " " -f 1` #get the time_units for TR
+		TR_vol=${tr}		
+
+		log_Msg "tr: $tr ${TR_units}"
+
+		if [ ${TR_units} = "s" ] ; then
+                	#this is correct;
+                	break
+        	elif [ ${TR_units} = "ms" ] ; then
+                	TR_vol=$( echo "scale=4;${TR_vol} / 1000.0" | bc -l );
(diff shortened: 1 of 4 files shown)


glasserm:
Hi cmpetty,

Are these changes tested?

Thanks,

Matt.

cmpetty:
@glasserm 

yes to the ones called from GenericfMRIVolumeProcessingPipeline.sh, which should cover OneStepResampling and mcflirt.  Will try ICAFix tomorrow, which isn't something we've been running

mharms:
@cmpetty  Thanks for assembling this, but I think it needs a couple tweaks:

1. Did you test this in the context of the standard situation in which the TR_units are already in seconds?  I don't see how this could work, as the `continue` and `break` constructs are for controlling `for` and `while` loops, and using `continue/break` will thus mess up the intended flow through any surrounding `for/while` loops, or generate an outright error if the new logic is not embedded in a `for/while` loop.  Please use `true` rather than `continue/break`, or depending on our decision on the following point, simply don't check for the `${TR_units} = "s"` condition at all.

1. Are "s", "ms", and "us" the full set of allowed `TR_units` values?  Can `TR_units` sometimes be undefined or empty?  In particular, I'm wondering if we should abort the script via our `log_Err_Abort` function if `TR_units` is not one of those 3 strings?

1. The value of the TR gets used in `hcp_fix_multi_run` and `ReApplyFixMultiRunPipeline.sh` in places beyond just the `fslmerge -tr` calls, and is expected to be in units of seconds in those other usages as well.  Rather than defining a new TR variable in any of the scripts (e.g., `TR_vol`), let's just use the script's natively defined tr variable, and reassign its value if it isn't in seconds.  Thus, any other usages of the TR within the script will be automatically correct.

mharms:
@coalsont   Other than what I just mentioned above, do you see any other necessary changes to this PR, to resolve issue #124?

coalsont:
Personally, I would use `case` instead of `if`/`else`, and there are unnecessary semicolons at the end of the math lines.

Nifti also defines hertz, ppm, and rads as additional units that could be used instead of time, no idea if fslval supports them.  It would be good to at least warn when the fslval output is not recognized as a time unit, but perhaps they should be treated as if they were seconds rather than erroring.

PR #129, opened 2019-09-09
Fix diffusion mask in voxels outside the FOV

MichielCottaar:
The voxels inside the FOV at the first b0 might not be included in some other volumes due to movement. In eddy the signal these voxels, which are not included in all the volumes, are set to zero. In the HCP pipeline the output of eddy is resampled to T1 space. During this resampling, these voxels become non-zero due to the spline interpolation and are included within the final nodif_brain_mask.

This patch fixes this by:
1. Creating a mask which includes only those voxels consistently within the FOV across all volumes as determined by eddy (called fov_mask). This mask is created from the eddy (or eddy_combine) output before resampling. 
2. Transforming this mask to T1-weighted space using trilinear interpolation. This is then thresholded at a value of 1, so that any voxel which are at the edge of the FOV are excluded.
3. Setting to zero any data outside this transformed FOV mask

fix_mask

This image shows the change in the final mask for subject 101309, which was the first subject where I noticed this issue, when looking at DTI fits. Note the very low range of the colour bar when plotting the B0 on the left. This is to visualise the striping caused by the spline interpolation.

### DIFF
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -95,13 +95,13 @@ done
 ${GlobalScripts}/Rotate_bvecs.sh "$DataDirectory"/bvecs "$WorkingDirectory"/diff2str.mat "$T1wOutputDirectory"/bvecs
 cp "$DataDirectory"/bvals "$T1wOutputDirectory"/bvals
 
-DILATE_DISTANCE=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
+DilateDistance=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
 
 #Register diffusion data to T1w space. Account for gradient nonlinearities if requested
 if [ ${GdcorrectionFlag} -eq 1 ]; then
     echo "Correcting Diffusion data for gradient nonlinearities and registering to structural space"
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
-    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DILATE_DISTANCE NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
+    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DilateDistance NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
@@ -113,7 +113,7 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/grad_dev -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/grad_dev  #Mask-out values outside the brain 
 else
     #Register diffusion data to T1w space without considering gradient nonlinearities
-    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/data.nii.gz $DILATE_DISTANCE NEAREST $DataDirectory/data_dilated.nii.gz
+    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/data.nii.gz $DilateDistance NEAREST $DataDirectory/data_dilated.nii.gz
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/data_dilated -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -105,8 +105,7 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
-    ${FSLDIR}/bin/fslroi "$DataDirectory"/warped/data_warped "$DataDirectory"/warped/nodif_warped 0 1
-    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/nodif_warped -abs -bin "$DataDirectory"/warped/fov_mask
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -abs -Tmax -bin "$DataDirectory"/warped/fov_mask
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
 
     #Now register the grad_dev tensor 
@@ -118,7 +117,7 @@ else
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/data_dilated -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
-    ${FSLDIR}/bin/fslmaths "$DataDirectory"/nodif -abs -bin "$DataDirectory"/fov_mask
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/data -abs -Tmax -bin "$DataDirectory"/fov_mask
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
 
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -120,7 +120,7 @@ else
 fi
 
 # only include voxels fully(!) within the field of view for every volume
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 1 -bin "$T1wOutputDirectory"/fov_mask
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 0.999 -bin "$T1wOutputDirectory"/fov_mask
 
 ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp -mas "$T1wOutputDirectory"/fov_mask "$T1wOutputDirectory"/data  #Mask-out data outside the brain
 ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -thr 0 "$T1wOutputDirectory"/data      #Remove negative intensity values (caused by spline interpolation) from final data
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -103,10 +103,11 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
 	# We combine the GDC warp with the diff2str affine, and apply to warped/data_warped (i.e., the post-eddy, pre-GDC data)
 	# so that we can have a one-step resampling following 'eddy'
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
+    # Dilation outside of the field of view to minimise the effect of the hard field of view edge on the interpolation
     ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DilateDistance NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
-    #Create a mask covering voxels within the field of view for all volumes
+    #Create a mask representing voxels within the field of view for all volumes prior to dilation
     ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -abs -Tmin -bin -fillh "$DataDirectory"/warped/fov_mask
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
 
@@ -114,11 +115,12 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/vecreg -i "$DataDirectory"/grad_dev -o "$T1wOutputDirectory"/grad_dev -r "$T1wRestoreImage"_${DiffRes} -t "$WorkingDirectory"/diff2str.mat --interp=spline
     ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/grad_dev -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/grad_dev  #Mask-out values outside the brain 
 else
-    #Register diffusion data to T1w space without considering gradient nonlinearities
+    # Dilation outside of the field of view to minimise the effect of the hard field of view edge on the interpolation
     ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/data.nii.gz $DilateDistance NEAREST $DataDirectory/data_dilated.nii.gz
+    #Register diffusion data to T1w space without considering gradient nonlinearities
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/data_dilated -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
 
-    #Create a mask covering voxels within the field of view for all volumes
+    #Create a mask representing voxels within the field of view for all volumes prior to dilation
     ${FSLDIR}/bin/fslmaths "$DataDirectory"/data -abs -Tmin -bin -fillh "$DataDirectory"/fov_mask
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -103,15 +103,26 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
+    #Create a mask covering voxels within the field of view for all volumes
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -Tmin -bin "$DataDirectory"/warped/fov_mask
+    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
+
     #Now register the grad_dev tensor 
     ${FSLDIR}/bin/vecreg -i "$DataDirectory"/grad_dev -o "$T1wOutputDirectory"/grad_dev -r "$T1wRestoreImage"_${DiffRes} -t "$WorkingDirectory"/diff2str.mat --interp=spline
     ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/grad_dev -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/grad_dev  #Mask-out values outside the brain 
 else
     #Register diffusion data to T1w space without considering gradient nonlinearities
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/data -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
+
+    #Create a mask covering voxels within the field of view for all volumes
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/data -Tmin -bin "$DataDirectory"/fov_mask
+    ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
 
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/data  #Mask-out data outside the brain 
+# only include voxels fully(!) within the field of view for every volume
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 1 -bin "$T1wOutputDirectory"/fov_mask
+
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp -mas "$T1wOutputDirectory"/fov_mask "$T1wOutputDirectory"/data  #Mask-out data outside the brain
 ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -thr 0 "$T1wOutputDirectory"/data      #Remove negative intensity values (caused by spline interpolation) from final data
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/nodif_brain_mask_temp
 
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -104,7 +104,8 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
-    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -abs -Tmin -bin "$DataDirectory"/warped/fov_mask
+    ${FSLDIR}/bin/fslroi "$DataDirectory"/warped/data_warped "$DataDirectory"/warped/nodif_warped
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/nodif_warped -abs -bin "$DataDirectory"/warped/fov_mask
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
 
     #Now register the grad_dev tensor 
@@ -115,7 +116,7 @@ else
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/data -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
-    ${FSLDIR}/bin/fslmaths "$DataDirectory"/data -abs -Tmin -bin "$DataDirectory"/fov_mask
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/nodif -abs -bin "$DataDirectory"/fov_mask
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
 
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -46,8 +46,6 @@ QAImage=`getopt1 "--QAimage" $@`                   # "$14" #Temporary file for s
 # Set default option values
 dof=`defaultopt $dof 6`
 
-echo $T1wOutputDirectory
-
 # Paths for scripts etc (uses variables defined in SetUpHCPPipeline.sh)
 GlobalScripts=${HCPPIPEDIR_Global}
 
@@ -97,7 +95,7 @@ done
 ${GlobalScripts}/Rotate_bvecs.sh "$DataDirectory"/bvecs "$WorkingDirectory"/diff2str.mat "$T1wOutputDirectory"/bvecs
 cp "$DataDirectory"/bvals "$T1wOutputDirectory"/bvals
 
-DILATE_DISTANCE=echo "$DiffRes * 4" | bc  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
+DILATE_DISTANCE=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
 
 #Register diffusion data to T1w space. Account for gradient nonlinearities if requested
 if [ ${GdcorrectionFlag} -eq 1 ]; then
@@ -107,7 +105,7 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
     #Create a mask covering voxels within the field of view for all volumes
-    ${FSLDIR}/bin/fslroi "$DataDirectory"/warped/data_warped "$DataDirectory"/warped/nodif_warped
+    ${FSLDIR}/bin/fslroi "$DataDirectory"/warped/data_warped "$DataDirectory"/warped/nodif_warped 0 1
     ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/nodif_warped -abs -bin "$DataDirectory"/warped/fov_mask
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
 
@@ -136,11 +134,11 @@ ${FSLDIR}/bin/immv "$T1wOutputDirectory"/nodif_brain_mask.nii.gz "$T1wOutputDire
 ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/nodif_brain_mask_old.nii.gz -mas "$T1wOutputDirectory"/temp "$T1wOutputDirectory"/nodif_brain_mask
 
 # Create a simple summary text file of the percentage of spatial coverage of the dMRI data inside the FS-derived brain mask
-NvoxBrainMask=`fslstats  -V "$T1wOutputDirectory"/nodif_brain_mask_old | awk '{print $1}'`
-NvoxFinalMask=`fslstats  -V "$T1wOutputDirectory"/nodif_brain_mask | awk '{print $1}'`
+NvoxBrainMask=`fslstats "$T1wOutputDirectory"/nodif_brain_mask_old -V | awk '{print $1}'`
+NvoxFinalMask=`fslstats "$T1wOutputDirectory"/nodif_brain_mask -V | awk '{print $1}'`
 PctCoverage=`echo "scale=4; 100 * ${NvoxFinalMask} / ${NvoxBrainMask}" | bc -l`
 echo "PctCoverage, NvoxFinalMask, NvoxBrainMask" >| "$T1wOutputDirectory"/nodif_brain_mask.stats.txt
-echo "${PctCoverage}, ${NvoxFinalMask}, ${NvoxBrainMask}" >> fov_mask.stats.txt
+echo "${PctCoverage}, ${NvoxFinalMask}, ${NvoxBrainMask}" >> "$T1wOutputDirectory"/nodif_brain_mask.stats.txt
 
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/temp
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/nodif_brain_mask_old
--- a/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
+++ b/DiffusionPreprocessing/scripts/DiffusionToStructural.sh
@@ -107,21 +107,19 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DilateDistance NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
-    #Create a mask representing voxels within the field of view for all volumes prior to dilation
-    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -abs -Tmin -bin -fillh "$DataDirectory"/warped/fov_mask
+    # Transforms field of view mask to T1-weighted space
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
 
-    #Now register the grad_dev tensor 
+    # Now register the grad_dev tensor
     ${FSLDIR}/bin/vecreg -i "$DataDirectory"/grad_dev -o "$T1wOutputDirectory"/grad_dev -r "$T1wRestoreImage"_${DiffRes} -t "$WorkingDirectory"/diff2str.mat --interp=spline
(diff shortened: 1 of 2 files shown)


coalsont:
If the mask's maximum value is 1 before resampling, thresholding at exactly 1 makes me nervous, as you are expecting the 8 interpolation factors to always sum to exactly 1 and never 0.999999.  Thresholding at 0.999 should be functionally equivalent to what is desired, while allowing plenty of room for floating point rounding error.

Of course, if fsl's thresholding already has a baked-in rounding error margin, or their trilinear interpolation implementation is very careful, and they will never change them, 1 may be fine as-is.  I did include a tiny rounding margin in wb_command -*-math's equality tests, though I'm not sure it would be big enough for the worst cases of multiply and add rounding in some possible implementations of trilinear interpolation.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -103,15 +103,26 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
     ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
 
+    #Create a mask covering voxels within the field of view for all volumes
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -abs -Tmin -bin "$DataDirectory"/warped/fov_mask
```
mharms: Do we want to use the `-abs` option this in `fslmaths` command?  If a value is less than zero, shouldn't that be counted as zero in terms of creating the mask?

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -103,15 +103,26 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
...
+    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
```
mharms: Any particular reason you are using trilinear here, rather than nearest neighbor?

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: In the output of eddy (to my understanding) the meaning of a zero or a very low and possibly negative number is different:
- zero is assigned to voxels outside of the initial FOV
- negative values might arise in voxels with a very low signal
While excluding voxels with a low signal from the diffusion MRI analysis is not a bad idea, here I am trying to adjust an existing brain mask (from the T1-weighted image) to exclude those voxels outside of the FOV for the diffusion data. So, I'm only interested in the voxels where the value is exactly zero.

On a more practical term: I first wrote it without the -abs command. The result is that many voxels with a very low signal (i.e., those outside of the brain) were also excluded from the mask. This included some voxels along the midline, which led to gaps in the nodif_brain_mask without actually improving the nodif_brain_mask in other ways (because the voxels actually outside of the brain were already excluded anyway in the original result from BET on the T1-weighted image).

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: To be more conservative in which voxels to include. The voxels closest to the edge have their signal greatly reduced during the interpolation, because they are affected by the voxels outside of the FOV (with their value of zero). Nearestneighbour might include some of these voxels, while trilinear is a bit more conservative.

However, after thinking it through a bit more: we might want to actually use spline interpolate, because that was used for the data interpolation. So any voxels where the spline interpolation is less than 1, will have its signal reduced due to interpolation of the zeros outside of the FOV.  Instead of excluding all potentially affected voxels, we could do the following:
1. spline interpolate the FOV mask to T1-weighted space
2. Divide the data by the spline interpolated FOV mask (correcting for the signal reduction due to interpolation with the area outside of the FOV)
3. Afterwards we can use Nearestneighbour to exclude those voxels which were outside of the FOV (and whose diffusion MRI data is hence purely based on extrapolation due to the spline).

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
mharms: Ok.  Sounds reasonable.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -103,15 +103,26 @@ if [ ${GdcorrectionFlag} -eq 1 ]; then
...
+    #Create a mask covering voxels within the field of view for all volumes
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/data -abs -Tmin -bin "$DataDirectory"/fov_mask
+    ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
 
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/data  #Mask-out data outside the brain 
+# only include voxels fully(!) within the field of view for every volume
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 0.999 -bin "$T1wOutputDirectory"/fov_mask
+
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp -mas "$T1wOutputDirectory"/fov_mask "$T1wOutputDirectory"/data  #Mask-out data outside the brain
```
mharms: Does masking with nodif_brain_mask_temp (the FS-derived brain mask, dilated 6 times) still serve a purpose if we are masking also with the new fov_mask?

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
mharms: So, if I understand correctly, that approach would be slightly less conservative? (i.e., would preserve a small number of extra voxels around the edges? And the division by the spline interpolated FOV mask "corrects" the signal intensity for the spline interpolation in those edge voxels that get included?)

 Seems like this may be getting more complicated than we need to be, but curious what @coalsont thinks...

mharms:
Can you maybe include an image of `fov_mask` itself?  (I'm assuming that the image above is of the final `nodif_brain_mask`?)

Also, just to clarify, `eddy` only zeros out voxels that moved outside the FOV in that particular volume/frame, right?  (i.e., it doesn't automatically zero out that voxel across all frames as well?  Because if it did, I'm wondering if we could simplify and just use the `Diffusion/data/nodif` file to generate the `fov_mask` and then warp that to T1w space.)

mharms:
And one other thought:  It would be great if we could quantify the fraction of voxels that were zeroed-out due to movement, and put that into a text file to promote a simple QC check.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
coalsont: I would be wary of dividing by a spline interpolated ROI, among other things it could attempt to bring the exterior ringing fringes back in.  It "smells" like something being done wrong, and trying to find a fix that isn't doing it over the right way.

I don't entirely follow the process, but it looks like there may be two separate issues: a brain mask (for which the common advice is to only ever apply it after all your resampling is finished) and movement outside the FoV (which is missing data in the original scan, and thus should be handled differently than brain masking).  I would suggest separating these aspects before considering weird things with division.

MichielCottaar:
Eddy zeros out all the volumes if one of them is out of the FOV. So we should be able to use the nodif to generate the FOV mask.

For reference in this discussion is an image of the FOV mask:

image

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: It still does mask out a large number of voxels outside of the brain, but within the FOV (which I assumed was the purpose of applying the nodf_brain_mask_temp)

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
coalsont: Thinking about it some more, the idea of dividing by a spline interpolated mask has more fundamental problems.  The ringing artifact is *additive*, not multiplicative, to the affected voxels, and its magnitude is proportional to the "cliff height" of the edge that induced it.  However, this edge size is measured locally, not globally, which makes attempting to "correct" it more troublesome.

Really, it seems like the issue is that ideally, we shouldn't be interpolating after masking out an FoV edge from other data.  Spline interpolation can reasonably be written to handle data that is nonzero at the FoV edge, but attempting to get it to ignore an edge that doesn't touch the FoV boundary is very problematic.  Dilation is the approach we have taken in similar situations.

MichielCottaar: I am merely trying to deal with the the FoV issue here. The FoV is in effect a mask, which has been applied by the scanner before any resampling (whether due to eddy or the resampling to T1-weighted space). In eddy this has been dealty with by setting to zero any voxels that leave the FoV in any volume. This creates a sharp boundary in the eddy output data at the edge of the FoV. When resampling to T1-weighted space, this leads to ringing due to the spline interpolation. 

The question of this patch is how to deal with that ringing. I can see three solutions:
1. The aggressive solution is just to identify any voxels that are close enough to the edge of the FoV to have been affected by the sharp boundary and set them all to zero. Given the use of spline interpolation, that would be a large number of voxels.
2. The more moderate approach, which I took here is to only remove the voxels at the actual edge of the FOV which would be greatly affected. This would mean leaving the ringing caused by the resampling across the FoV's edge intact.
3. The more complex approach would be to try to correct for the ringing by dividing the spline interpolation of the FoV mask (as outlined above). However, to be honest this would be a bit of a waste of time as it multiplies all diffusion data in a voxel with the same number, which will not affect any of the diffusion MRI output (except for the S0 image).

The only interaction in this patch with the brain mask, is that I later adjust the nodif_brain_mask to exclude any voxels, where the diffusion data has been set to zero (as we don't want to run dtifit/bedpostX on those voxels).

coalsont: What I was saying is that I don't think 3 is really a solution (because it is not a multiplicative effect, among other things), and I am presenting an option 4: dilate to "invent" some data where the FoV intersects a generous brain mask, to remove the hard edges *before* resampling to T1 space.  This avoids, reduces, or at least moves the ringing artifact so that it has less effect on the real data.  It would still make sense to mask the dilated area back out after the resampling.

Ideally, eddy would output the mask without applying it to the data of the unaffected frames (which would prevent interpolation ringing in frames that didn't have large motion), or even more ideally, we would combine the motion correction and T1-space resamplings into one step somehow (which would prevent the ringing entirely when using a well-written spline interpolation).

However, option 2 sounds tolerable, and I was only arguing against the details of option 3.

MichielCottaar: Sorry, I did not see your latest message before my previous post. Dilation before the spline resampling sounds like a much better idea than trying to correct for the ringing afterwards. I think it's worth adding to this patch and will set it up and test it in the coming days.

coalsont: wb_command's `-volume-dilate` has some options that may be useful for this purpose (specifying ROIs for what data is good and where to dilate into), and with the `WEIGHTED` method it uses a sort of smoothing kernel to generate the dilated values.  However, we have found that the defaults in 1.3.2 and earlier are not all that great, `-exponent 7` should improve the smoothness of the dilated values, and a larger dilation distance could also help with smoothness (specifying a very restricted `-bad-voxel-roi` is recommended for runtime considerations).

The next release will have better defaults, a much faster implementation, and new cutoff logic, but we don't have a timeframe planned for the release yet.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
mharms: Got it.  This is clear now, after seeing the image of the `fov_mask`.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,25 +97,30 @@ done
...
 cp "$DataDirectory"/bvals "$T1wOutputDirectory"/bvals
 
+DILATE_DISTANCE=echo "$DiffRes * 12"  # Extrapolates the diffusion data up to 12 voxels outside of the FOV
+
 #Register diffusion data to T1w space. Account for gradient nonlinearities if requested
 if [ ${GdcorrectionFlag} -eq 1 ]; then
     echo "Correcting Diffusion data for gradient nonlinearities and registering to structural space"
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
-    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+    ${CARET7DIR}/wb_command -volume_dilate $DataDirectory/warped/data_warped $DILATE_DISTANCE NEAREST $DataDirectory/warped/data_warped_dilated
```
coalsont: The command is `-volume-dilate`, no underscores.  I suggest at least trying the weighted method with `-exponent 7` when you test this, for comparison.  However, both methods are somewhat slow in 1.3.2, so I would suggest taking a somewhat-dilated brain mask, finding its intersection with the fov-introduced zeros, and using that as the `-bad-voxel-roi` argument, to speed things up a bit (by not dilating unnecessary locations).

Also, wb_command requires the complete filename, including extension, so add .nii.gz to them.

MichielCottaar: Thanks for catching the typos. The computation speed is indeed a major problem. I tested both versions of the dilation (resulting b0 attached) and it took 1 hour for NEAREST and 5 hours for WEIGHTED with exponent of 7. The weighted version does look a bit better, but it does not seem worth the extra 4 hours of computation time as the effect on the spline interpolation will probably be very small (I have not tested that yet).

I thought $DILATE_DISTANCE would essentially do what you suggest to do with the -bad-voxel-roi, i.e., by setting it to 12 * 1.25 I only compute the dilation for voxels withing 12 voxels of the FOV. I'll test if the run time is shortened by using -bad-voxel-roi.

I'm still trying to figure out how to set the extent of the dilation. There is no hard boundary in how far sharp boundaries are propogated in the spline. 12 voxels is probably overkill, but I'm not sure at all how far down I can safely go. 

b0_dilation

MichielCottaar: I've had a look at the code for -volume-dilate and it appears to me that for this specific application it is very inefficient, because it iterates over all volumes independently. However, in this case the -bad-voxel-roi and -data-roi are the same for all volumes, so once it has determined for a voxel the nearest neighbour or weights in the kernel for a single volume, it could apply it to all volumes.

I guess if we're really worried about runtime we could (1) assign each voxel in -data-roi an index, (2) run -volume-dilate with NEAREST flag, (3) assign diffusion data to each voxel based on its index. That should be a lot faster than running -volume-dilate on the whole diffusion dataset at once, but would not work with the WEIGHTED flag.

I might just be overthinking it and should just compute the dilation for a more limited -bad-voxel-roi as you suggested and ignore the inefficient runtime...

coalsont: The smaller the dilation distance, the faster the command will be.  Even 1 voxel would be an improvement over the previous results, but 4 might be enough get most of the possible benefits.  I suggested intersecting the negation of the FoV with a somewhat dilated brain mask specifically because it should pinpoint only the voxels contributing to this localized ringing effect (and the fewer voxels that -volume-dilate has to fill in, the faster it runs).

The latest code in github is already faster than the 1.3.2 release (it now uses an octtree of only the valid source voxels and uses a kernel cutoff distance based on the closest voxel, 1.3.2 used full-size stencils everywhere and checked against the ROIs repeatedly).  You can try the dev_latest here if you want to compare the timing to 1.3.2:

http://brainvis.wustl.edu/workbench/

Cache locality effects might prevent the "each voxel across all volumes" approach from being noticeably faster, and unfortunately the case without a -bad-voxel-roi or -data-roi can have a different set of valid voxels in every frame, so the code would need to be more complicated to support this optimization (it would be better to save the set of source voxels for each output voxel and still go frame by frame).

If you aren't concerned about the values created by the dilation, you might as well just use fsl's dilation, if it is fast.  Weighted dilation will always be slower, since it takes data from more than one source voxel for every output voxel.  There is also the possibility of -volume-smoothing with the -fix-zeros option (because gaussian smoothing in an orthogonal grid decomposes into 1D smoothings, it is faster - the weighted dilation is trickier because it doesn't use a fixed-size gaussian, but instead a kernel that is invariant to scaling), though that smoothing method assumes that the data can't have real zeros in it.

glasserm:
I'm a bit lost here, why are we dilating?

MichielCottaar:
Because in the diffusion MRI data from eddy there is a sharp edge at the edge of the field of view (FoV). Voxels which in any volume are outside of the FoV are set to zero. The spline interpolation used in the resampling to T1w space is very sensitive to these type of sharp edges, which leads to striping artifacts. Dilation will move such striping artifacts outwards, which minimizes their effect on those voxels fully within the FoV (which we want to keep).

glasserm:
So a few iterations of nearest neighbor should be fine if you will be remasking anyway?

MichielCottaar:
Yeah, that would probably be enough. Nearest neighbour across 4 voxels (using -volume-dilate) took a bit less that 10 minutes, so that seems reasonable.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,21 +97,37 @@ done
...
     ${FSLDIR}/bin/vecreg -i "$DataDirectory"/grad_dev -o "$T1wOutputDirectory"/grad_dev -r "$T1wRestoreImage"_${DiffRes} -t "$WorkingDirectory"/diff2str.mat --interp=spline
     ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/grad_dev -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/grad_dev  #Mask-out values outside the brain 
 else
     #Register diffusion data to T1w space without considering gradient nonlinearities
-    ${FSLDIR}/bin/flirt -in "$DataDirectory"/data -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
+    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/data.nii.gz $DILATE_DISTANCE NEAREST $DataDirectory/data_dilated.nii.gz
+    ${FSLDIR}/bin/flirt -in "$DataDirectory"/data_dilated -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp spline -out "$T1wOutputDirectory"/data
+
+    #Create a mask covering voxels within the field of view for all volumes
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/nodif -abs -bin "$DataDirectory"/fov_mask
```
mharms: I know I suggested using `nodif` as a proxy for the internal fov mask within eddy (which, ideally, would write out the fov mask as an explicit output).  But do we need to be concerned about the possibility of randomly getting a "true zero" within the first frame that wasn't actually part of the `eddy` internal fov mask?

mharms:
@MichielCottaar  See #138 for some files that I'm creating to allow us to quickly assess the spatial coverage issue with the fMRI data.  Ideally, the changes you are working on would incorporate something similar for quickly assessing the coverage in the dMRI data.  I'd like to get that PR wrapped up early next week, and it would be nice if we could get this PR done at the same time as well.

MichielCottaar:
I've rebased onto master, hence all the repeat of the commits.

@mharms Sure, the latest commit should add such a text file.

I'm just running the new pipeline with and without gradient non-linearities (see https://git.fmrib.ox.ac.uk/ndcn0236/mask_striping). If that works, I'm happy for this to be merged.

MichielCottaar:
Ok, this version runs on jalapeno. It produces a summary of the coverage, which for my test subject contains:
```
PctCoverage, NvoxFinalMask, NvoxBrainMask
98.2133, 790150, 804524
```

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: I've addressed this issue by actually checking for the -Tmax after taking the absolute value in the data. So we will only exclude voxels containing only zeroes.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,28 +95,57 @@ done
 ${GlobalScripts}/Rotate_bvecs.sh "$DataDirectory"/bvecs "$WorkingDirectory"/diff2str.mat "$T1wOutputDirectory"/bvecs
 cp "$DataDirectory"/bvals "$T1wOutputDirectory"/bvals
 
+DILATE_DISTANCE=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
```
mharms: Can we call this `DilateDistance` (lower case) so as to not imply that it is some sort of environment variable?

MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,28 +95,57 @@ done
...
 cp "$DataDirectory"/bvals "$T1wOutputDirectory"/bvals
 
+DilateDistance=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
+
 #Register diffusion data to T1w space. Account for gradient nonlinearities if requested
 if [ ${GdcorrectionFlag} -eq 1 ]; then
     echo "Correcting Diffusion data for gradient nonlinearities and registering to structural space"
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
-    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DilateDistance NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
```
mharms: A comment here as we why we are dilating would be nice.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
mharms: Maybe my concern about this possibility was misplaced, since getting a true zero is very unlikely.  The problem with the `-Tmax` approach is what if there are zeros in say half the frames for a given voxel, then we don't want to use that voxel.  I guess that scenario shouldn't happen if `eddy` itself is enforcing a common fov mask across all frames. But I think it is better if we write this code to be robust to whatever `eddy` chooses to do in that regard.  Which would argue for not using `nodif` as a proxy for the internal `eddy` fov mask, and using `-Tmin` instead.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,28 +95,57 @@ done
...
+    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+
+    #Create a mask covering voxels within the field of view for all volumes
+    ${FSLDIR}/bin/fslmaths "$DataDirectory"/warped/data_warped -abs -Tmax -bin "$DataDirectory"/warped/fov_mask
```
mharms: Per my other comment, I'm thinking we should use `-bin -Tmin` instead.  (Or perhaps `-abs -bin -Tmin` if negative values are considered "legitimate").

mharms: That's also a better match to the purpose stated in the comments.  In the general case (without knowing the internals of how `eddy` operates), `-Tmax` does not generate a mask within the field of view "for all volumes".

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,28 +95,57 @@ done
...
+
 #Register diffusion data to T1w space. Account for gradient nonlinearities if requested
 if [ ${GdcorrectionFlag} -eq 1 ]; then
     echo "Correcting Diffusion data for gradient nonlinearities and registering to structural space"
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
-    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DilateDistance NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
+    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+
+    #Create a mask covering voxels within the field of view for all volumes
```
mharms: Can we tweak this comment to say:
`#Create a mask representing voxels within the field of view for all volumes prior to dilation`

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
mharms: I see that later in the script we remove negative intensity values from `data`, although it isn't clear to me if negative values can be present in the `eddy` output, or if they are only introduced by the spline interpolation inherent to resampling into T1w space in this script itself.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,28 +95,57 @@ done
...
+# Create a simple summary text file of the percentage of spatial coverage of the dMRI data inside the FS-derived brain mask
+NvoxBrainMask=`fslstats "$T1wOutputDirectory"/nodif_brain_mask_old -V | awk '{print $1}'`
+NvoxFinalMask=`fslstats "$T1wOutputDirectory"/nodif_brain_mask -V | awk '{print $1}'`
+PctCoverage=`echo "scale=4; 100 * ${NvoxFinalMask} / ${NvoxBrainMask}" | bc -l`
+echo "PctCoverage, NvoxFinalMask, NvoxBrainMask" >| "$T1wOutputDirectory"/nodif_brain_mask.stats.txt
+echo "${PctCoverage}, ${NvoxFinalMask}, ${NvoxBrainMask}" >> "$T1wOutputDirectory"/nodif_brain_mask.stats.txt
+
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/temp
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/nodif_brain_mask_old
+${FSLDIR}/bin/imrm "$T1wOutputDirectory"/fov_mask
```
mharms: I'd suggest that we keep `fov_mask`, just because it is a bit of a pain to generate.

MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Negative values are legitimate I think. Most importantly, if we start removing negative values this will create a lot of holes in the field of view mask and hence in the final nodif_brain_mask, which would be a problem for many analyses. To enforce the lack of hole, I actually added a -fillh. For any valid field of view mask this should not make a difference (and indeed it does not for the test subject). However, it should prevent voxels with a value that just happens to be exactly zero within the field of view from creating holes in the final mask.

Negative values are already there in the output file from eddy (before spline interpolation), but not the input file. So, I guess they arise in the internal interpolation in the eddy script.

glasserm: Why do we think negative values are legitimate?  Won't they create problems for the various model fitting we do with the diffusion data later?

MichielCottaar: First of all, I don't think this patch should address negative values as it is unrelated with the field of view. 

Continuing this tangent anyway: For Gaussian noise models, the negative values should not be a real problem, as long as they are not too low. However, Rician noise models or the linear fitting of log(S) in dtifit do require positive values. Both of these cases would however also struggle with signal being perfectly 0, so setting negative values to 0 does not really help the diffusion modelling.

In FSL in the case of dtifit the negative values are just set to 1 (i.e., logS=0). I'm not sure how bedpostX deals with them.

glasserm: ok sounds good

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
coalsont: `-Tmin` means if even a single frame has an exact zero there, it is considered bad, so it seems worse than the original simpler tactic of looking at just the first frame.

If we want a more robust solution, what if we thresholded to find all exact zeros, counted the number of frames a voxel is zero with a sum operation, and then used a second threshold (say, at least 10% of frames) on that?

However, if we can in fact model the FoV as a mask directly without depending on exact zeros in the output, that would resolve all such concerns.

coalsont: To be clear, unless we think eddy is likely to get changed, I don't think we should be coding for what it might be changed to do, so if it does change to per-volume masking, we should expect to have to rewrite this code anyway (and the correct thing to do then would probably be separate ROIs per frame).

mharms:
`-abs -Tmin -bin`, with a `-fillh` thrown in as a back-up to ensure no holes, seems reasonable to me.

However, while looking through the code, I thought of three other things:
1. We are leaving the same sort of striping problem present in the GDC case in the `Diffusion/data` space data.  Seems like we should derive the actual Diffusion space fov_mask in `eddy_postproc.sh` and apply it as part of that script.  (Which also makes sense conceptually, because the derivation of the Diffusion space fov_mask doesn't have anything to do with the conversion to T1w structural space per se).  Then, in `DiffusionToStructural.sh` we just resample the fov_mask into the `"$T1wOutputDirectory"`.
2. In the no GDC condition, determining the fov_mask in `DiffusionToStructural.sh` is problematic, due to the `fslmaths ${datadir}/data -thr 0 ${datadir}/data` operation at the end of `eddy_postproc.sh`.  This is another reason for moving the determination of fov_mask into `eddy_postproc.sh` (so that the fov_mask can be determined accurately in the no GDC condition).
3. Separate from above, but related, once we apply the fov_mask in Diffusion space, I would argue that we should comment out (with a date provided) the `fslmaths ${datadir}/data -thr 0 ${datadir}/data` operation in `eddy_postproc.sh`, so that any negative values from `eddy` itself are preserved and not artifically truncated at zero.

mharms:
Expanding on my point 3 above, I see that we also do the same `fslmaths -thr 0` operation toward the end of `DiffusionToStructural.sh`.  So maybe that particular operation we should leave alone in both `eddy_postproc.sh` and `DiffusionToStructural.sh` for consistency considerations.  We just need to be careful about deriving any masks after that operation.

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -87,13 +87,31 @@ else
 fi
 #fi
 
+
+#Create a mask representing voxels within the field of view for all volumes prior to dilation
+${FSLDIR}/bin/fslmaths ${datadir}/data -abs -Tmin -bin -fillh ${datadir}/fov_mask
```
mharms: Maybe add a comment here about why we are using `-abs`?
e.g.,
`# 'eddy' can return negative values in some low signal locations, so use -abs for determining the fov mask`

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -87,13 +87,31 @@ else
...
+
+    # Dilation outside of the field of view to minimise the effect of the hard field of view edge on the interpolation
+	  DiffRes=`${FSLDIR}/bin/fslval ${datadir}/data pixdim1`
+    DilateDistance=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
+    ${CARET7DIR}/wb_command -volume-dilate ${datadir}/warped/data_warped.nii.gz $DilateDistance NEAREST ${datadir}/warped/data_warped_dilated.nii.gz
+
+    # apply gradient distortion correction
+    ${globalscriptsdir}/GradientDistortionUnwarp.sh --workingdir="${datadir}" --coeffs="${GdCoeffs}" --in="${datadir}/data_warped_dilated" --out="${datadir}/data" --owarp="${datadir}/fullWarp"
+
+    # Transform field of view mask (using conservative trilinear interolation with high threshold)
```
mharms: Typo on "interolation"

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -87,13 +87,31 @@ else
...
+    ${FSLDIR}/bin/immv ${datadir}/fov_mask ${datadir}/fov_mask_warped
+    ${FSLDIR}/bin/applywarp --rel --interp=trilinear -i ${datadir}/fov_mask_warped -r ${datadir}/fov_mask_warped -w ${datadir}/fullWarp -o ${datadir}/fov_mask
+    ${FSLDIR}/bin/fslmaths ${datadir}/fov_mask -thr 0.999 -bin ${datadir}/fov_mask
+    # mask out any data outside the field of view
+    ${FSLDIR}/bin/fslmaths ${datadir}/data -mas ${datadir}/fov_mask ${datadir}/data
```
mharms: I guess that technically it shouldn't be necessary to apply the fov_mask to the data in the `$GdCoeffs=NONE` condition, but for robustness of code I'd move the final masking operation to outside of the `$GdCoeffs` conditional.

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -102,10 +120,12 @@ if [ ! $GdCoeffs = "NONE" ] ; then
...
-    
+    ${FSLDIR}/bin/imrm ${datadir}/data_warped_dilated
+
     #Keep the original warped data and warp fields
     mkdir -p ${datadir}/warped
     ${FSLDIR}/bin/immv ${datadir}/data_warped ${datadir}/warped
+    ${FSLDIR}/bin/immv ${datadir}/fov_mask_warped ${datadir}/warped
     ${FSLDIR}/bin/immv ${datadir}/fullWarp ${datadir}/warped
     ${FSLDIR}/bin/immv ${datadir}/fullWarp_abs ${datadir}/warped
 fi
```
mharms: With the application of the fov_mask, we shouldn't have any negative values due to the (GDC) spline interpolation, so can we update the following comment to the following:
`#Remove negative intensity values (from eddy) from final data`

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,31 +95,59 @@ done
 ${GlobalScripts}/Rotate_bvecs.sh "$DataDirectory"/bvecs "$WorkingDirectory"/diff2str.mat "$T1wOutputDirectory"/bvecs
 cp "$DataDirectory"/bvals "$T1wOutputDirectory"/bvals
 
+DilateDistance=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
+
 #Register diffusion data to T1w space. Account for gradient nonlinearities if requested
```
mharms: Can we update this comment:
`#Register diffusion data to T1w space. Account for gradient nonlinearities (using one-step resampling) if requested`

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,31 +95,59 @@ done
...
     ${FSLDIR}/bin/convertwarp --rel --relout --warp1="$DataDirectory"/warped/fullWarp --postmat="$WorkingDirectory"/diff2str.mat --ref="$T1wRestoreImage"_${DiffRes} --out="$WorkingDirectory"/grad_unwarp_diff2str
-    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+    # Dilation outside of the field of view to minimise the effect of the hard field of view edge on the interpolation
+    ${CARET7DIR}/wb_command -volume-dilate $DataDirectory/warped/data_warped.nii.gz $DilateDistance NEAREST $DataDirectory/warped/data_warped_dilated.nii.gz
+    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/data_warped_dilated -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=spline -o "$T1wOutputDirectory"/data
+
+    # Transforms field of view mask to T1-weighted space
+    ${FSLDIR}/bin/applywarp --rel -i "$DataDirectory"/warped/fov_mask -r "$T1wRestoreImage"_${DiffRes} -w "$WorkingDirectory"/grad_unwarp_diff2str --interp=trilinear -o "$T1wOutputDirectory"/fov_mask
```
mharms: Can we add an additional comment here:
`# (Be sure to use the fov_mask derived prior to application of GDC)`

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,31 +95,59 @@ done
...
+
+    # Transforms field of view mask to T1-weighted space
+    ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
 
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/data  #Mask-out data outside the brain 
+# only include voxels fully(!) within the field of view for every volume
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 0.999 -bin "$T1wOutputDirectory"/fov_mask
+
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp -mas "$T1wOutputDirectory"/fov_mask "$T1wOutputDirectory"/data  #Mask-out data outside the brain
```
mharms: Update comment at end of line:
`#Mask-out data outside the brain and outside the fov`

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,31 +95,59 @@ done
...
+echo "PctCoverage, NvoxFinalMask, NvoxBrainMask" >| "$T1wOutputDirectory"/nodif_brain_mask.stats.txt
+echo "${PctCoverage}, ${NvoxFinalMask}, ${NvoxBrainMask}" >> "$T1wOutputDirectory"/nodif_brain_mask.stats.txt
+
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/temp
 ${FSLDIR}/bin/imrm "$T1wOutputDirectory"/nodif_brain_mask_old
+if [ ${GdcorrectionFlag} -eq 1 ]; then
+    ${FSLDIR}/bin/imrm $DataDirectory/warped/data_warped_dilated
+else
+    ${FSLDIR}/bin/imrm $DataDirectory/data_dilated
+fi
```
mharms: If we wanted, these `imrm`'s could each be moved inside the appropriate `if/else` clauses of the already existing conditional.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -97,31 +95,59 @@ done
...
+    # Transforms field of view mask to T1-weighted space
+    ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
 fi
 
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp "$T1wOutputDirectory"/data  #Mask-out data outside the brain 
+# only include voxels fully(!) within the field of view for every volume
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 0.999 -bin "$T1wOutputDirectory"/fov_mask
+
+${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp -mas "$T1wOutputDirectory"/fov_mask "$T1wOutputDirectory"/data  #Mask-out data outside the brain
 ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -thr 0 "$T1wOutputDirectory"/data      #Remove negative intensity values (caused by spline interpolation) from final data
```
mharms: Update comment at end of this line to match the same suggested change in `eddy_postproc`:
`#Remove negative intensity values (from eddy) from final data`

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: I've moved them.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: Done

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
```diff
@@ -121,13 +123,15 @@ else
...
     ${FSLDIR}/bin/flirt -in "$DataDirectory"/fov_mask -ref "$T1wRestoreImage"_${DiffRes} -applyxfm -init "$WorkingDirectory"/diff2str.mat -interp trilinear -out "$T1wOutputDirectory"/fov_mask
+    ${FSLDIR}/bin/imrm $DataDirectory/data_dilated
 fi
 
 # only include voxels fully(!) within the field of view for every volume
 ${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/fov_mask -thr 0.999 -bin "$T1wOutputDirectory"/fov_mask
 
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -mas "$T1wOutputDirectory"/nodif_brain_mask_temp -mas "$T1wOutputDirectory"/fov_mask "$T1wOutputDirectory"/data  #Mask-out data outside the brain
-${FSLDIR}/bin/fslmaths "$T1wOutputDirectory"/data -thr 0 "$T1wOutputDirectory"/data      #Remove negative intensity values (caused by spline interpolation) from final data
+# Mask out data outside the brain adn outside the fov
```
mharms: Typo "adn"

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -87,13 +87,30 @@ else
...
     echo "Correcting for gradient nonlinearities"
 	# Note: "data_warped" is eddy-current and suspectibility distortion corrected (via 'eddy'), but prior to gradient distortion correction
 	# i.e., "data_posteddy_preGDC" would be another way to think of it
     ${FSLDIR}/bin/immv ${datadir}/data ${datadir}/data_warped
-    ${globalscriptsdir}/GradientDistortionUnwarp.sh --workingdir="${datadir}" --coeffs="${GdCoeffs}" --in="${datadir}/data_warped" --out="${datadir}/data" --owarp="${datadir}/fullWarp"
+
+    # Dilation outside of the field of view to minimise the effect of the hard field of view edge on the interpolation
+	  DiffRes=`${FSLDIR}/bin/fslval ${datadir}/data pixdim1`
+    DilateDistance=`echo "$DiffRes * 4" | bc`  # Extrapolates the diffusion data up to 4 voxels outside of the FOV
+    ${CARET7DIR}/wb_command -volume-dilate ${datadir}/warped/data_warped.nii.gz $DilateDistance NEAREST ${datadir}/warped/data_warped_dilated.nii.gz
```
mharms: Here's the bug that I was referring to.  (I guess I forgot to actually "Add" the comment at the specific line).  Namely, in `eddy_postproc.sh` the "warped" directory doesn't exist at this point in the script.  This line should be:
`${CARET7DIR}/wb_command -volume-dilate ${datadir}/data_warped.nii.gz $DilateDistance NEAREST ${datadir}/data_warped_dilated.nii.gz`

MichielCottaar:
With these latest fixes both the GDC and no GDC conditions have produced T1-aligned diffusion data with the FoV mask applied.

DiffusionPreprocessing/scripts/DiffusionToStructural.sh
MichielCottaar: fixed

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: fixed

PR #134, opened 2019-09-27
FreeSurfer Pipeline related pull request

sivcek:
### 1/ Addition of optional `—mppversion`  parameter
The parameter has the same functionality as in the PreFreeSurfer pipeline. And is related to the MPP version checking, which in this case only checks for the presence of T2w image.

### 2/ Ability to process w/o T2w image
If T2w image is set to NONE and legacy MPP version is selected then:
* `-T2` option is excluded from `recon_all_cmd`
* `-T2pial` is replaced with `-noT2pial`  in `recon_all_cmd`
* Making T2w to T1w registration available in FSL format section is skipped
* `make_t2w_hires_nifti_file` and `make_t1wxtw2_qc_file` is skipped

### 3/ Cosmetic changes
Switch from tabs to spaces and remove trailing line spaces

### 1/ Cosmetic changes
Switching from tabs to spaces, removing trailing spaces in lines

### 2/ Skipping operations that depend on T2w image
When T2wImage is set to NONE, operations that depend on the presence of T2wImage are skipped.:
* creation of normalized T2w image
* generation of pial first surface with T2 adjustment
* generation of pial second pass surface with T2 adjustment

### 3/ Resolving a bug with computing c_ras.mat
We identified a bug in a code that was triggered in some cases and replaced it with alternative more robust implementation (lines 104-112)

### 1/ Cosmetic changes
Switching from tabs to spaces, removing trailing spaces in lines

### 2/ Skipping operations that depend on T2w image 
When T2wImage is set to NONE the following is skipped:
* computing T2 to T1 transform

### 3/ The use of ceho is suppressed until debugging library is pulled in

### DIFF
--- a/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
+++ b/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
@@ -78,13 +78,23 @@ log_Msg "seed_cmd_appendix: ${seed_cmd_appendix}"
 
 
 # ------------------------------------------------------------------------------
-#  Check MMP Mode
+#  Compliance check
 # ------------------------------------------------------------------------------
 
 MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
 MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
 
-check_mpp_compliance "FreeSurfer53HCP"
+# -- T2w image
+
+if [ "${T2wInputImages}" = "NONE" ]; then
+  ComplianceMsg+=" --t2=NONE"
+  Compliance="LegacyStyleData"
+fi
+
+check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
+
 
 # ------------------------------------------------------------------------------
 #  Show Environment Variables
--- a/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
+++ b/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
@@ -19,7 +19,7 @@ set -e
 
 source $HCPPIPEDIR/global/scripts/log.shlib  # Logging related functions
 source $HCPPIPEDIR/global/scripts/opts.shlib # Command line option functions
-source ${HCPPIPEDIR}/global/scripts/mppmodecheck.shlib  # Functions to check the validity of MPP mode specification
+source ${HCPPIPEDIR}/global/scripts/processingmodecheck.shlib
 
 ########################################## SUPPORT FUNCTIONS ########################################## 
 
@@ -81,8 +81,8 @@ log_Msg "seed_cmd_appendix: ${seed_cmd_appendix}"
 #  Compliance check
 # ------------------------------------------------------------------------------
 
-MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
-MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+ProcessingMode=`opts_GetOpt1 "--processing-mode" $@`
+ProcessingMode=`opts_DefaultOpt $ProcessingMode "HCPStyleData"`
 Compliance="HCPStyleData"
 ComplianceMsg=""
 
@@ -93,7 +93,7 @@ if [ "${T2wInputImages}" = "NONE" ]; then
   Compliance="LegacyStyleData"
 fi
 
-check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
+check_mode_compliance "${ProcessingMoce}" "${Compliance}" "${ComplianceMsg}"
 
 
 # ------------------------------------------------------------------------------
--- a/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
+++ b/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
@@ -19,6 +19,7 @@ set -e
 
 source $HCPPIPEDIR/global/scripts/log.shlib  # Logging related functions
 source $HCPPIPEDIR/global/scripts/opts.shlib # Command line option functions
+source ${HCPPIPEDIR}/global/scripts/mppmodecheck.shlib  # Functions to check the validity of MPP mode specification
 
 ########################################## SUPPORT FUNCTIONS ########################################## 
 
@@ -74,6 +75,16 @@ else
 	seed_cmd_appendix="-norandomness -rng-seed ${recon_all_seed}"
 fi
 log_Msg "seed_cmd_appendix: ${seed_cmd_appendix}"
+
+
+# ------------------------------------------------------------------------------
+#  Check MMP Mode
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+
+check_mpp_compliance "FreeSurfer53HCP"
 
 # ------------------------------------------------------------------------------
 #  Show Environment Variables
--- a/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
+++ b/FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
@@ -93,7 +93,7 @@ if [ "${T2wInputImages}" = "NONE" ]; then
   Compliance="LegacyStyleData"
 fi
 
-check_mode_compliance "${ProcessingMoce}" "${Compliance}" "${ComplianceMsg}"
+check_mode_compliance "${ProcessingMode}" "${Compliance}" "${ComplianceMsg}"
 
 
 # ------------------------------------------------------------------------------
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -205,7 +205,7 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
                  e.g., [--existing-subject --extra-reconall-arg=-show-edits --no-conf2hires]
 
   [--pipeline-mode={HCPStyleData, LegacyStyleData}]
-      Indicates, which variant of MPP to use. "HCPStyleData" (the default) follows the processing steps 
+      Indicates, which processing mode to use. "HCPStyleData" (the default) follows the processing steps 
          described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion. "LegacyStyleData" 
          allows additional processing functionality and use of some acquisitions that do not conform 
          to 'HCP-Style' expectations; i.e., in this case missing high-resolution T2w image 
@@ -379,7 +379,7 @@ get_options()
 	fi
 
 
-	# NOTE: Check for T2w image has moved upwards, as missing T2w is allowed in LegacyStyleData MPP mode.
+	# NOTE: Check for T2w image has moved upwards, as missing T2w is allowed in LegacyStyleData processing mode.
 
 
 	# show optional parameters if specified
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -204,7 +204,7 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
             (ii) you want to be able to run some flag in recon-all, without also regenerating the surfaces.
                  e.g., [--existing-subject --extra-reconall-arg=-show-edits --no-conf2hires]
 
-  [--pipeline-mode={HCPStyleData, LegacyStyleData}]
+  [--processing-mode={HCPStyleData, LegacyStyleData}]
       Indicates, which processing mode to use. "HCPStyleData" (the default) follows the processing steps 
          described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion. "LegacyStyleData" 
          allows additional processing functionality and use of some acquisitions that do not conform 
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -646,8 +646,6 @@ main()
 	# accomplished using --extra-reconall-arg=-noT2pial
 	if [ "${T2wImage}" != "NONE" ]; then
 		recon_all_cmd+=" -T2pial"
-	else
-		recon_all_cmd+=" -noT2pial"
 	fi
 
 	recon_all_cmd+=" -openmp ${num_cores}"
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -215,7 +215,7 @@ PARAMETERs can also be specified positionally as:
   ${g_script_name} <path to subject directory> <subject ID> <path to T1 image> <path to T1w brain mask> <path to T2w image> [<recon-all seed value>]
 
   Note that the positional approach to specifying parameters does NOT support the 
-      --existing-subject, --extra-reconall-arg, --no-conf2hires, and --pipeline-mode options.
+      --existing-subject, --extra-reconall-arg, --no-conf2hires, and --processing-mode options.
   The positional approach should be considered deprecated, and may be removed in a future version.
 
 EOF
@@ -300,8 +300,8 @@ get_options()
 				p_extra_reconall_args+="${extra_reconall_arg} "
 				index=$(( index + 1 ))
 				;;
-			--pipeline-mode=*)
-				p_pipelinemode=${argument#*=}
+			--processing-mode=*)
+				p_processing_mode=${argument#*=}
 				index=$(( index + 1 ))
 				;;
 			--no-conf2hires)
@@ -322,7 +322,7 @@ get_options()
 	#  Compliance check
 	# ------------------------------------------------------------------------------
 	
-	ProcessingMoce=${p_pipelinemode:-HCPStyleData}	
+	ProcessingMode=${p_processing_mode:-HCPStyleData}	
     Compliance="HCPStyleData"
     ComplianceMsg=""
 
@@ -336,7 +336,7 @@ get_options()
         p_t2w_image="NONE"
     fi
 
-    check_mode_compliance "${ProcessingMoce}" "${Compliance}" "${ComplianceMsg}"
+    check_mode_compliance "${ProcessingMode}" "${Compliance}" "${ComplianceMsg}"
 
 	# ------------------------------------------------------------------------------
 	#  check required parameters
@@ -395,8 +395,8 @@ get_options()
 	if [ ! -z "${p_conf2hires}" ]; then
 		log_Msg "Include -conf2hires flag in recon-all: ${p_conf2hires}"
 	fi
-	if [ ! -z "${p_pipelinemode}" ] ; then
-  		log_Msg "Set --pipeline-mode to: ${p_pipelinemode}"
+	if [ ! -z "${p_processing_mode}" ] ; then
+  		log_Msg "Set --processing-mode to: ${p_processing_mode}"
 	fi
 
 	if [ ${error_count} -gt 0 ]; then
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -208,7 +208,7 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
       Indicates, which variant of MPP to use. "HCPStyleData" (the default) follows the processing steps 
          described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion. "LegacyStyleData" 
          allows additional processing functionality and use of some acquisitions that do not conform 
-         to 'HCP-Style' expectations in this case missing high-resolution T2w image 
+         to 'HCP-Style' expectations; i.e., in this case missing high-resolution T2w image 
 
 PARAMETERs can also be specified positionally as:
 
@@ -364,6 +364,10 @@ get_options()
 	else
 		log_Msg "T1w Brain: ${p_t1w_brain}"
 	fi
+
+
+	# NOTE: Check for T2w image is moved to check_mpp_compliance(), as missing T2w is allowed in LegacyStyleData MPP mode.
+
 
 	# show optional parameters if specified
 	if [ ! -z "${p_seed}" ]; then
@@ -641,9 +645,9 @@ main()
 		recon_all_cmd+=" -emregmask ${T1wImageBrain}"
 	fi
 
-	# By default, refine pial surfaces using T2.
-	# If for some reason the -T2pial flag needs to be excluded from recon-all, this can be
-	# accomplished using --extra-reconall-arg=-noT2pial
+	# By default, refine pial surfaces using T2 (if T2w image provided).
+	# If for some other reason the -T2pial flag needs to be excluded from recon-all, 
+	# this can be accomplished using --extra-reconall-arg=-noT2pial
 	if [ "${T2wImage}" != "NONE" ]; then
 		recon_all_cmd+=" -T2pial"
 	fi
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -319,11 +319,24 @@ get_options()
 	local error_count=0
 
 	# ------------------------------------------------------------------------------
-	#  Check MMP Mode
+	#  Compliance check
 	# ------------------------------------------------------------------------------
 	
-	MPPMode=${p_mppmode:-HCPStyleData}
-	check_mpp_compliance "FreeSurfer"
+	MPPMode=${p_mppmode:-HCPStyleData}	
+    Compliance="HCPStyleData"
+    ComplianceMsg=""
+
+     # -- T2w image
+
+    if [ -z "${p_t2w_image}" ] || [ "${p_t2w_image}" = "NONE" ]; then
+        if [ -z "${p_existing_subject}" ]; then
+            ComplianceMsg+=" --t2w-image= or --t2= not present or set to NONE"
+            Compliance="LegacyStyleData"
+        fi
+        p_t2w_image="NONE"
+    fi
+
+    check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
 
 	# ------------------------------------------------------------------------------
 	#  check required parameters
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -204,7 +204,7 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
             (ii) you want to be able to run some flag in recon-all, without also regenerating the surfaces.
                  e.g., [--existing-subject --extra-reconall-arg=-show-edits --no-conf2hires]
 
-  [--mpp-mode={HCPStyleData, LegacyStyleData}]
+  [--pipeline-mode={HCPStyleData, LegacyStyleData}]
       Indicates, which variant of MPP to use. "HCPStyleData" (the default) follows the processing steps 
          described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion. "LegacyStyleData" 
          allows additional processing functionality and use of some acquisitions that do not conform 
@@ -215,7 +215,7 @@ PARAMETERs can also be specified positionally as:
   ${g_script_name} <path to subject directory> <subject ID> <path to T1 image> <path to T1w brain mask> <path to T2w image> [<recon-all seed value>]
 
   Note that the positional approach to specifying parameters does NOT support the 
-      --existing-subject, --extra-reconall-arg, --no-conf2hires, and --mpp-mode options.
+      --existing-subject, --extra-reconall-arg, --no-conf2hires, and --pipeline-mode options.
   The positional approach should be considered deprecated, and may be removed in a future version.
 
 EOF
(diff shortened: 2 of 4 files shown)


FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -625,14 +686,18 @@ main()
...
 		recon_all_cmd+=" -emregmask ${T1wImageBrain}"
 	fi
 
 	# By default, refine pial surfaces using T2.
 	# If for some reason the -T2pial flag needs to be excluded from recon-all, this can be
 	# accomplished using --extra-reconall-arg=-noT2pial
-	recon_all_cmd+=" -T2pial"
+	if [ "${T2wImage}" != "NONE" ]; then
+		recon_all_cmd+=" -T2pial"
+	else
```
mharms: The `else` condition here is not necessary (serves no purpose).

glasserm:
Sure, but they already implemented it, so why take it out?

mharms:
Because I'd rather we not inadvertently break something in the FS 5.3 related code.

FreeSurfer/FreeSurferPipeline.sh
sivcek: I've taken it out.

sivcek:
@mharms we've used v5.3.0-HCP extensively in the past and it has been throughly tested. It would be good to support the "LegacyStyleData" in it, so I have added the MPP check there as well.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -204,12 +204,18 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
             (ii) you want to be able to run some flag in recon-all, without also regenerating the surfaces.
                  e.g., [--existing-subject --extra-reconall-arg=-show-edits --no-conf2hires]
 
+  [--mpp-mode={HCPStyleData, LegacyStyleData}]
+      Indicates, which variant of MPP to use. "HCPStyleData" (the default) follows the processing steps 
+         described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion. "LegacyStyleData" 
+         allows additional processing functionality and use of some acquisitions that do not conform 
+         to 'HCP-Style' expectations in this case missing high-resolution T2w image
```
mharms: Awkward phrasing. Better would be: `to 'HCP-Style' expectations; i.e., in this case, missing high-resolution T2w image`

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -345,17 +365,6 @@ get_options()
 		log_Msg "T1w Brain: ${p_t1w_brain}"
 	fi
```
mharms: Would be helpful to maintain a comment here that the check for `{p_t2w_image}` has been moved to the `check_mpp_compliance` function.  That's the problem with the adopted approach.  We have now spun off some relevant code into a function that is easily missed...

Likewise, include a comment for any other scripts that had an existing check that was moved off to `check_mpp_compliance`

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -625,14 +637,16 @@ main()
...
 		recon_all_cmd+=" -all"
 		recon_all_cmd+=" -i ${zero_threshold_T1wImage}"
-		recon_all_cmd+=" -T2 ${T2wImage}"
+		[ "${T2wImage}" != "NONE" ] && recon_all_cmd+=" -T2 ${T2wImage}"
 		recon_all_cmd+=" -emregmask ${T1wImageBrain}"
 	fi
 
 	# By default, refine pial surfaces using T2.
 	# If for some reason the -T2pial flag needs to be excluded from recon-all, this can be
 	# accomplished using --extra-reconall-arg=-noT2pial
```
mharms: Update comments:
```
# By default, refine pial surfaces using T2 (if T2w image provided).
# If for some other reason the -T2pial flag needs to be excluded from recon-all, 
# this can be accomplished using --extra-reconall-arg=-noT2pial
```

mharms:
For documentation purposes, what was the nature of the bug "in some cases" in the computation of c_ras.mat, and under what situations does it manifest?

FreeSurfer/FreeSurferPipeline.sh
sivcek: ok

FreeSurfer/FreeSurferPipeline.sh
sivcek: Added comment. No other script is strucured in the way that would require such a comment.

FreeSurfer/FreeSurferPipeline.sh
sivcek: Updated.

sivcek:
@mharms The problem was when the path to `brain.finalsurfs.mgz` included any of `c_r`, `c_a`, or `c_s`. In those cases it would get through the `grep` statement and try to cut the first line of the output. This happened to us in a huge data processing push.

sivcek:
> The allowance of T2wImage=NONE interacts with the code we have in place regarding FS edits and the `if [ "${existing_subject}" != "TRUE" ]; then` section therefore needs attention. This will also interact with both PreFreeSurfer and PostFreeSurfer, because the file that we are using as a proxy for whether PostFreeSurfer has been run previously is the warp field encoded in `OrigT2wToT1w.nii.gz`

Any suggestions how to deal with that?

mharms:
Per my suggestion on the PreFS PR, I suggest we get PostFS into better shape first, dealing with the issues that Matt raised (i.e., we can't just cut out CreateMyelinMaps entirely, because some other important stuff happens in that script besides just the myelin maps).  Unfortunately, CreateMyelinMaps is a very dense script, and not well-commented at this particular point in time.  Perhaps if either @glasserm or @coalsont could go through that script and split it into clear commented "sub-sections", we could separate out the non-myelin-mapping code into its own separate sub-script called by PostFreeSurferPipeline.sh

sivcek:
> The allowance of T2wImage=NONE interacts with the code we have in place regarding FS edits and the `if [ "${existing_subject}" != "TRUE" ]; then` section therefore needs attention. This will also interact with both PreFreeSurfer and PostFreeSurfer, because the file that we are using as a proxy for whether PostFreeSurfer has been run previously is the warp field encoded in `OrigT2wToT1w.nii.gz`

Would it be ok to instead use another proxy for checking whether PostFreeSurfer has been run. E.g. the presence of `MNINonLinear/*.spec` file?

mharms:
Regarding the test for whether PostFS has been previously run, I think we can simply change the following 2 lines:
`local OutputOrigT2wToT1w="OrigT2w2T1w"  #Needs to match name used in PostFreeSurfer (N.B. "OrigT2" here refers to the T2w/T2w.nii.gz file; NOT FreeSurfer's "orig" space)`
to
`local OutputOrigT1wToT1w="OrigT1w2T1w"  #Needs to match name used in PostFreeSurfer (N.B. "OrigT1" here refers to the T1w/T1w.nii.gz file; NOT FreeSurfer's "orig" space)`

and 
`	if [ `imtest ${SubjectDIR}/xfms/${OutputOrigT2wToT1w}` = 1 ]; then`
to 
` 	if [ `imtest ${SubjectDIR}/xfms/${OutputOrigT1wToT1w}` = 1 ]; then`

sivcek:
I implemented the suggested change in the test for existing PostFreeSurfer run.

sivcek:
@mharms does this code satisfy your concern. @coalsont are you ok with the rest of the code?

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -345,16 +378,9 @@ get_options()
...
-			log_Err "T2w Image (--t2w-image= or --t2=) required"
-			error_count=$(( error_count + 1 ))
-		else
-			p_t2w_image="NONE"  # Need something assigned as a placeholder for positional parameters
-		fi
-	else
-		log_Msg "T2w Image: ${p_t2w_image}"
-	fi
+
+	# NOTE: Check for T2w image is moved to check_mpp_compliance(), as missing T2w is allowed in LegacyStyleData MPP mode.
```
coalsont: This comment should probably say "moved upwards", as the check of the t2 variable is not inside a different file anymore.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -369,6 +395,9 @@ get_options()
 	if [ ! -z "${p_conf2hires}" ]; then
 		log_Msg "Include -conf2hires flag in recon-all: ${p_conf2hires}"
 	fi
+	if [ "${p_t2w_image}" = "NONE" ] ; then
+  		Compliance="legacy"
```
coalsont: This appears to set the wrong value, and seems like it will never get used as it is after the checking call.  However, this variable appears to already be checked above.

FreeSurfer/scripts/FreeSurferHiresPial.sh
```diff
@@ -75,13 +76,22 @@ cp --preserve=timestamps $SubjectDIR/$SubjectID/surf/rh.pial $SubjectDIR/$Subjec
...
+  echo "---> Generating pial first pass surfaces with T2 adjustment."
+  # Generate pial surfaces with T2 adjustment, still first pass.
+  # Same 4 files created as above, but have $outputSuffix1 ("postT2.pass1") appended through use of -output flag.
+  # Use -T2 flag (rather than -T2dura), since -T2 is the flag used within recon-all script of FS 5.3 (but -T2 and -T2dura generate same results)
+  # For mris_make_surface with correct arguments #Could go from 3 to 2 potentially...
+  outputSuffix1=".postT2.pass1"
+  mris_make_surfaces -nsigma_above 2 -nsigma_below 3 -aseg aseg.hires -filled filled.hires -wm wm.hires -mgz -sdir $SubjectDIR -orig white.deformed -nowhite -orig_white white.deformed -orig_pial pial -T2 "$mridir"/T2w_hires.norm -T1 T1w_hires.norm -output $outputSuffix1 $SubjectID lh
+  mris_make_surfaces -nsigma_above 2 -nsigma_below 3 -aseg aseg.hires -filled filled.hires -wm wm.hires -mgz -sdir $SubjectDIR -orig white.deformed -nowhite -orig_white white.deformed -orig_pial pial -T2 "$mridir"/T2w_hires.norm -T1 T1w_hires.norm -output $outputSuffix1 $SubjectID rh
+else
+  #ceho "---> No T2w image, skipping generation of pial first pass surfaces with T2 adjustment."
```
coalsont: `ceho` doesn't even exist in the current state of the debug shlib.  I suggest not adding commented out commands.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -204,12 +204,18 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
             (ii) you want to be able to run some flag in recon-all, without also regenerating the surfaces.
                  e.g., [--existing-subject --extra-reconall-arg=-show-edits --no-conf2hires]
 
+  [--mpp-mode={HCPStyleData, LegacyStyleData}]
+      Indicates, which variant of MPP to use. "HCPStyleData" (the default) follows the processing steps
```
mharms: @coalsont Do you think we should spell out what "MPP" is as part of the usage statement?
i.e., "Indicates which variant of MPP (minimal preprocessing) to use."

FreeSurfer/FreeSurferPipeline.sh
mharms: However, for consistency with the other "optional parameters", we should check whether `p_mppmode` is defined (which it will be, because we set its default above) and then message its value with a `log_Msg` statement.

sivcek: I left the check for p_mppmode so that it will ony be reported when actually set.

FreeSurfer/scripts/FreeSurferHiresPial.sh
sivcek: I have staged things for after the debug.shlib PR is accepted. I've now changed it to `verbose_red_echo`

sivcek:
@mharms I believe the only open issue is the question regarding MPP in the help to @coalsont

FreeSurfer/FreeSurferPipeline.sh
coalsont: I seem to recall there was some discussion about retconning it into "multimodal processing pipelines" or something.

I have mixed feelings about that acronym (we have yet to use it within the pipelines, this option would be the first, but we have used the acronym MMP before for something entirely different), and if I were writing only the usage statement, I'd probably sidestep the issue by saying something like `Indicate whether the HCP acquisition and processing guidelines should be treated as requirements`.

sivcek: Ok. I'll ask you for a consensus here, as this would need to be changed in multiple PRs and I would not like to go back and forth on it. @mharms @coalsont @glasserm please provide a final guideline.

mharms: I like that, but it of course begs the question of why the flag is called `--mpp-mode`, which is why I broached the question earlier (via email) if `--pipeline-mode` wouldn't in fact be a (much) better name... :)  I feel that if the flag is going to be called `--mpp-mode` then we really need to explain what "MPP" actually stands for.

sivcek: I have added that explanation to PreFS:

```
  [--mpp-mode=(HCPStyleData|          Which variant of Minimal Preprocessing Pipelines (MPP) to use. "HCPStyleData" 
               LegacyStyleData)]      (the default) follows the processing stepsdescribed in Glasser et al. (2013) 
                                      and requires 'HCP-Style' data acquistion. "LegacyStyleData" allows additional 
```

coalsont: I think --pipeline-mode is better than --mpp-mode, as the tool name doesn't seem like something that belongs in an option.  Another option might be --processing-mode as more specific and less redundant, as "pipeline" is still a name for the overall thing that is being used, and is therefore somewhat redundant (e.g., I didn't make a `wb_command -workbench-version` option, it is just `wb_command -version`).

mharms: I like `--processing-mode=(HCPStyleData|LegacyStyleData)`.  @glasserm @aanticev Is that acceptable to you?  If so, I think we finally have consensus.  [We either need to explain what "MPP" refers to in Usage statements, or simply drop "mpp" from the argument name, which is the approach that both Tim and I favor].

coalsont: In case he doesn't explicitly send an email or make a comment, @glasserm told me in person he is fine with `--processing-mode` (he did not specifically mention the values it accepts, but I think he is okay with the ones you mentioned).

glasserm: [--processing-mode=(HCPStyleData|LegacyStyleData)] works for me.

mharms: So, if we go that route, presumably we call the new shlib `processingmodecheck.shlib` (rather than `mppmodecheck.shlib`), and the function therein (and called by the scripts) `check_mode_compliance()` (or alternatively `processing_mode_compliance()`), and the internal variable used within the scripts `ProcessingMode` (rather than `MPPMode`)?  That would seem to tie everything together nicely and consistently.  Any objections to those (or better suggestions)? @coalsont @glasserm

glasserm: Sounds good.

coalsont: Sounds reasonable, though I prefer that the function name still starts with "check".

sivcek: So:
-> processingmodecheck.shlib 
-> check_mode_compliance() 
-> ProcessingMode (rather than MPPMode)

I'll implement that throughout

coalsont: Also remove "MPP" from the help info.  You can start from my earlier suggestion of `Indicate whether the HCP acquisition and processing guidelines should be treated as requirements` and rephrase/expand to include the default and accepted values.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -345,16 +378,9 @@ get_options()
...
-			log_Err "T2w Image (--t2w-image= or --t2=) required"
-			error_count=$(( error_count + 1 ))
-		else
-			p_t2w_image="NONE"  # Need something assigned as a placeholder for positional parameters
-		fi
-	else
-		log_Msg "T2w Image: ${p_t2w_image}"
-	fi
+
+	# NOTE: Check for T2w image has moved upwards, as missing T2w is allowed in LegacyStyleData MPP mode.
```
coalsont: It would also be appreciated if you could look for comments that say "MPP" and change them.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -308,7 +318,30 @@ get_options()
...
+
+    if [ -z "${p_t2w_image}" ] || [ "${p_t2w_image}" = "NONE" ]; then
+        if [ -z "${p_existing_subject}" ]; then
+            ComplianceMsg+=" --t2w-image= or --t2= not present or set to NONE"
+            Compliance="LegacyStyleData"
+        fi
+        p_t2w_image="NONE"
+    fi
+
+    check_mode_compliance "${ProcessingMoce}" "${Compliance}" "${ComplianceMsg}"
```
mharms: Typo.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -308,7 +318,30 @@ get_options()
 
 	local error_count=0
 
-	# check required parameters
+	# ------------------------------------------------------------------------------
+	#  Compliance check
+	# ------------------------------------------------------------------------------
+	
+	ProcessingMoce=${p_pipelinemode:-HCPStyleData}
```
mharms: Typo.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -204,12 +204,18 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
             (ii) you want to be able to run some flag in recon-all, without also regenerating the surfaces.
                  e.g., [--existing-subject --extra-reconall-arg=-show-edits --no-conf2hires]
 
+  [--pipeline-mode={HCPStyleData, LegacyStyleData}]
```
mharms: Should be `--processing-mode`

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -204,12 +204,18 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
...
+         allows additional processing functionality and use of some acquisitions that do not conform 
+         to 'HCP-Style' expectations; i.e., in this case missing high-resolution T2w image 
+
 PARAMETERs can also be specified positionally as:
 
   ${g_script_name} <path to subject directory> <subject ID> <path to T1 image> <path to T1w brain mask> <path to T2w image> [<recon-all seed value>]
 
   Note that the positional approach to specifying parameters does NOT support the 
-      --existing-subject, --extra-reconall-arg, and --no-conf2hires options.
+      --existing-subject, --extra-reconall-arg, --no-conf2hires, and --pipeline-mode options.
```
mharms: Should be `--processing-mode`

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -294,6 +300,10 @@ get_options()
 				p_extra_reconall_args+="${extra_reconall_arg} "
 				index=$(( index + 1 ))
 				;;
+			--pipeline-mode=*)
+				p_pipelinemode=${argument#*=}
```
mharms: Change to `--processing-mode` and `p_processing_mode`

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -369,6 +395,9 @@ get_options()
 	if [ ! -z "${p_conf2hires}" ]; then
 		log_Msg "Include -conf2hires flag in recon-all: ${p_conf2hires}"
 	fi
+	if [ ! -z "${p_pipelinemode}" ] ; then
```
mharms: p_processing_mode

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -369,6 +395,9 @@ get_options()
...
+  		log_Msg "Set --pipeline-mode to: ${p_pipelinemode}"
```
mharms: --procesing-mode and p_pipeline_mode

FreeSurfer/FreeSurferPipeline.sh
mharms: Sorry, meant to say use `p_processing_mode` as the internal variable name.

mharms: Let me just give the intended line to avoid any confusion:
`ProcessingMode=${p_processing_mode:-HCPStyleData}	`

FreeSurfer/FreeSurferPipeline.sh
mharms: Sorry, mistyped that as well.  The proposed lines would be:
```
if [ ! -z "${p_processing_mode}" ] ; then
  		log_Msg "Set --processing-mode to: ${p_processing_mode}"
```

FreeSurfer/FreeSurferPipeline.sh
sivcek: I took out MPP from help and comments. I did not know we are changing the wording @glasserm provided and @mharms polished. If we are changing them, can I please get the final agreed upon wording so that there's no back and forth.

FreeSurfer/FreeSurferPipeline.sh
sivcek: Did so.

FreeSurfer/FreeSurferPipeline.sh
sivcek: Corrected.

FreeSurfer/FreeSurferPipeline.sh
sivcek: fixed.

FreeSurfer/FreeSurferPipeline.sh
sivcek: fixed.

FreeSurfer/FreeSurferPipeline.sh
sivcek: fixed.

FreeSurfer/FreeSurferPipeline.sh
sivcek: fixed.

FreeSurfer/FreeSurferPipeline.sh
sivcek: fixed.

FreeSurfer/FreeSurferPipeline-v5.3.0-HCP.sh
```diff
@@ -75,6 +76,26 @@ else
...
+ComplianceMsg=""
+
+# -- T2w image
+
+if [ "${T2wInputImages}" = "NONE" ]; then
+  ComplianceMsg+=" --t2=NONE"
+  Compliance="LegacyStyleData"
+fi
+
+check_mode_compliance "${ProcessingMoce}" "${Compliance}" "${ComplianceMsg}"
```
mharms: Typo.

sivcek: Ahh ... where did this come from ... corrected.

FreeSurfer/FreeSurferPipeline.sh
sivcek: Done.

FreeSurfer/FreeSurferPipeline.sh
mharms: This isn't fixed. Still showing as `--pipeline-mode`.

sivcek: It was in the latest push. Probably did it at the same time.

FreeSurfer/FreeSurferPipeline.sh
mharms: How about:
```
Controls whether the HCP acquisition and processing guidelines should be treated as requirements.
"HCPStyleData" (the default) follows the processing steps described in Glasser et al. (2013) 
   and requires 'HCP-Style' data acquistion. 
"LegacyStyleData" allows additional processing functionality and use of some acquisitions
   that do not conform to 'HCP-Style' expectations.
   i.e., in this case no high-resolution T2w image acquisition
```
Note that I intentionally lead with the mode names as the start of new lines, and used indentation purposely to improve readability.

sivcek: Ok. Changed it here and in PreFS.

FreeSurfer/scripts/FreeSurferHiresPial.sh
```diff
@@ -75,13 +76,20 @@ cp --preserve=timestamps $SubjectDIR/$SubjectID/surf/rh.pial $SubjectDIR/$Subjec
...
+  mris_make_surfaces -nsigma_above 2 -nsigma_below 3 -aseg aseg.hires -filled filled.hires -wm wm.hires -mgz -sdir $SubjectDIR -orig white.deformed -nowhite -orig_white white.deformed -orig_pial pial -T2 "$mridir"/T2w_hires.norm -T1 T1w_hires.norm -output $outputSuffix1 $SubjectID lh
+  mris_make_surfaces -nsigma_above 2 -nsigma_below 3 -aseg aseg.hires -filled filled.hires -wm wm.hires -mgz -sdir $SubjectDIR -orig white.deformed -nowhite -orig_white white.deformed -orig_pial pial -T2 "$mridir"/T2w_hires.norm -T1 T1w_hires.norm -output $outputSuffix1 $SubjectID rh
+else
+  verbose_red_echo "---> No T2w image, skipping generation of pial first pass surfaces with T2 adjustment."
+  outputSuffix1=""
+fi
+
 
 # Bring pial surfaces out of highres space into the 1 mm (FS conformed) space
 # [Note that $regII, although named 1mm2hires.dat, actually maps from the hires space into the FS conformed space].
```
mharms: Are you sure that setting `outputSuffix1=""` in the case of no T2 in the old FS5.3 version code isn't breaking something given that the immediately following `mri_surf2surf` command operates on `pial${outputSuffix1}`, which will then resolve to just `pial`, which is itself is a pretty elemental FS file... ?

FreeSurfer/FreeSurferPipeline.sh
coalsont: Looks reasonable, though I still don't like the seeming redundancy in "i.e., in this case" or that it is a sentence fragment after a period.  How about "In this script, it allows not having a high-resolution T2w image."?

FreeSurfer/scripts/FreeSurferHiresPial.sh
sivcek: @mharms, based on my read, the first pass w/o T2w adjustments generates `lh/rh.pial` surfaces. These are then copied for debugging purposes to `.preT2.pass1`. What else could `mri_surf2surf` work with?

FreeSurfer/FreeSurferPipeline.sh
sivcek: Changed.

FreeSurfer/scripts/FreeSurferHiresPial.sh
mharms: So, in the no T2 case, `--sval-xyz` and `--tval` are both set to 'pial', although I guess that is fine as long they are distinct copies within memory of `mri_surf2surf` (which I would assume to be the case...)  (i.e., that it reads in `--sval-xyz` entirely before writing out `--tval`).

sivcek: I can change that to the stored copy.

sivcek: Just to be sure.

sivcek: Changed.

FreeSurfer/scripts/FreeSurferHiresPial.sh
```diff
@@ -168,12 +179,19 @@ cp --preserve=timestamps $SubjectDIR/$SubjectID/surf/rh.pial $SubjectDIR/$Subjec
...
+  verbose_red_echo "---> Generating pial second pass surfaces with T2 adjustment."
+  # Generate pial surfaces with T2 adjustment, second pass.
+  # Same 4 files created as above, but have $outputSuffix2 ("postT2.pass2") appended through use of -output flag
+  #Could go from 3 to 2 potentially...
+  outputSuffix2=".postT2.pass2"
+  mris_make_surfaces -nsigma_above 2 -nsigma_below 3 -aseg aseg.hires -filled filled.hires -wm wm.hires -mgz -sdir $SubjectDIR -orig white.deformed -nowhite -orig_white white.deformed -orig_pial pial -T2 "$mridir"/T2w_hires.norm -T1 T1w_hires.norm -output $outputSuffix2 $SubjectID lh
+  mris_make_surfaces -nsigma_above 2 -nsigma_below 3 -aseg aseg.hires -filled filled.hires -wm wm.hires -mgz -sdir $SubjectDIR -orig white.deformed -nowhite -orig_white white.deformed -orig_pial pial -T2 "$mridir"/T2w_hires.norm -T1 T1w_hires.norm -output $outputSuffix2 $SubjectID rh
+else
+  verbose_red_echo "---> No T2w image, skipping generation of pial second pass surfaces with T2 adjustment."
+  outputSuffix2=""
```
mharms: For consistency, shouldn't we do something similar then to `outSuffix2`?  e.g.,
`outputSuffix2=preT2.pass2`

sivcek: agreed done.

PR #135, opened 2019-09-27
PostFreeSurfer related pull request

sivcek:
### 1/ Cosmetic changes
Changing tabs to spaces, removal of trailing line spaces 

### 5/ Enabling run w/o T2w image
When T2w image is missing and T2wRestoreImage is set to NONE the following steps are skipped:
* Myelin mapping

### 1/ Cosmetic changes
Changing tabs to spaces, removal of trailing line spaces 

### 4/ Enabling run w/o T2w image
When T2w image is missing and T2wRestoreImage is set to NONE the following steps are skipped:
* Myelin mapping

### 1/ Cosmetic changes
Changing tabs to spaces, removal of trailing line spaces 

### 2/ Fixing a bug with generation of c_ras.mat
The code to generate c_ras.mat failed in some cases so we replaced it with a more robust version.

### 3/ Enabling run w/o T2w image
When T2w image is missing and T2wRestoreImage is set to NONE the following steps are skipped:
* Adding T2w related images to the spec file

### 1/ Cosmetic changes
Changing tabs to spaces, removal of trailing line spaces 

### 2/ Enabling run w/o T2w image
When T2w image is missing and T2wRestoreImage is set to NONE the following steps are skipped:
* Adding T2w related images to the spec file
* `applywarp` to  `AtlasSpaceT2wImage`


glasserm:
We also need to have no T2w checks for the last few loops in CreateMyelinMaps for #Create CIFTI Files and #Add CIFTI Maps to Spec.

glasserm:
Lets also class RegName="FS" into legacy mode as I mentioned in fMRISurface.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -90,6 +90,24 @@ InflateExtraScale=`opts_DefaultOpt $InflateExtraScale 1`
...
+verbose_echo "  --grayordinatesdir: ${GrayordinatesSpaceDIR}"
+verbose_echo "  --grayordinatesres: ${GrayordinatesResolutions}"
+verbose_echo "         --hiresmesh: ${HighResMesh}"
+verbose_echo "        --lowresmesh: ${LowResMeshes}"
+verbose_echo " --subcortgraylabels: ${SubcorticalGrayLabels}"
+verbose_echo "  --freesurferlabels: ${FreeSurferLabels}"
+verbose_echo "     --refmyelinmaps: ${ReferenceMyelinMaps}"
+verbose_echo "           --mcsigma: ${CorrectionSigma}"
+verbose_echo "           --regname: ${RegName}"
+verbose_echo "   --PipelineScripts: ${PipelineScripts}"
```
mharms: Unless you added this, `--PipelineScripts` is not an input argument.  And if it _is_ something that was added, it should be all lower case in the argument name for consistency.

mharms: FWIW, I'm not sure if the longer term plan would be to switch to using `verbose_echo` more, but the other scripts mostly use `log_Msg` for parameter logging.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -140,44 +164,63 @@ for Hemisphere in L R ; do
...
-  ${CARET7DIR}/wb_command -metric-dilate "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap.native.func.gii "$T1wFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".midthickness.native.surf.gii 30 "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap.native.func.gii -nearest
-  ${CARET7DIR}/wb_command -metric-mask "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap.native.func.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap.native.func.gii
+  if [ "${T2wPresent}" = "YES" ] ; then
```
mharms: I would move the preceding `-metric regression` line to the top of its enclosing `for` loop.  Then, if I'm reading correctly, all the code up through L197 [i.e., up to `for STRING in MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func corrThickness@shape ; do`] can occur within a single consolidated `if [ "${T2wPresent}" = "YES" ]` conditional.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -140,44 +164,63 @@ for Hemisphere in L R ; do
...
+        ${CARET7DIR}/wb_command -metric-mask "$AtlasSpaceFolder"/fsaverage_LR"$LowResMesh"k/"$Subject"."$Hemisphere"."$Map"."$LowResMesh"k_fs_LR."$Ext".gii "$AtlasSpaceFolder"/fsaverage_LR"$LowResMesh"k/"$Subject"."$Hemisphere".atlasroi."$LowResMesh"k_fs_LR.shape.gii "$AtlasSpaceFolder"/fsaverage_LR"$LowResMesh"k/"$Subject"."$Hemisphere"."$Map"."$LowResMesh"k_fs_LR."$Ext".gii
+      done
+    done
+  else   # --- when there is no T2w image only work on corrThickness@shape
+    for STRING in corrThickness@shape ; do
```
mharms: Rather than duplicating the following entire block of code, how about something like the following earlier:
```
STRINGList="corrThickness@shape"
if [ "${T2wPresent}" = "YES" ] ; then
  STRINGList+=" MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func"
fi
for STRING in $STRINGList; do
```
Tidy, short and easy to follow the logic of what is going on...

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: @mharms `PipelineScripts` is derived from environment variable `HCPPIPEDIR_PostFS`. I changed the echo to reflect that. 

I've used verbose_echo where there were no `log_Msg` statements before.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Reshuffled.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Excellent!

sivcek:
@glasserm @mharms I believe all the concerns are addressed and the code is significantly simplified.

glasserm:
It looks good to me now

PostFreeSurfer/PostFreeSurferPipeline.sh
coalsont: I don't think we need scripts to log things that are derived in a simple fashion from environment variables, or the values of any of the `HCPPIPEDIR_*` variables either.  If you have a specific reason to log `HCPPIPEDIR`, that might be reasonable.

coalsont: To be clear, I don't think `HCPPIPEDIR_PostFS` should ever be anything other than `${HCPPIPEDIR}/PostFreeSurfer/scripts`, and thus `HCPPIPEDIR` is a more succinct display of the only part of it that can be changed by the user.

mharms: Along those lines, it seems like what we need in the PostFS script is a reporting of the HCPPIPEDIR, the HCPPipelines version, and other tool (FSL, Workbench) versions.  @coalsont can you suggest a script to model that after?  (I believe that you created some functions to simplify the process...?)

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -110,20 +130,22 @@ fslmaths "$OutputMNIT1wImage" -abs "$OutputMNIT1wImage" -odt float
...
 
-applywarp --rel --interp=spline -i "$OrginalT2wImage" -r "$T1wImageBrainMask" -w "$OutputOrigT2wToStandard" -o "$OutputMNIT2wImage"
-fslmaths "$OutputMNIT2wImage" -abs "$OutputMNIT2wImage" -odt float
-fslmaths "$OutputMNIT2wImage" -div "$BiasFieldOutput" "$OutputMNIT2wImageRestore"
-fslmaths "$OutputMNIT2wImageRestore" -mas "$T1wMNIImageBrainMask" "$OutputMNIT2wImageRestoreBrain"
+if [ "${T2wPresent}" = "YES" ] ; then
+  applywarp --rel --interp=spline -i "$OrginalT2wImage" -r "$T1wImageBrainMask" -w "$OutputOrigT2wToStandard" -o "$OutputMNIT2wImage"
+  fslmaths "$OutputMNIT2wImage" -abs "$OutputMNIT2wImage" -odt float
+  fslmaths "$OutputMNIT2wImage" -div "$BiasFieldOutput" "$OutputMNIT2wImageRestore"
+  fslmaths "$OutputMNIT2wImageRestore" -mas "$T1wMNIImageBrainMask" "$OutputMNIT2wImageRestoreBrain"
```
mharms: Ok, this is being picky, but for symmetry with the preceding, I'd conclude the preceding conditional with a `fi` here, and start a new `if [ "${T2wPresent}" = "YES" ] ;` in the `#Create T1w/T2w maps` section, so that the if conditional doesn't "bleed" across the distinct code sections in a weird fashion.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -190,7 +221,13 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
   Mesh=`echo $STRING | cut -d "@" -f 2`
   ROI=`echo $STRING | cut -d "@" -f 3`
-  for STRINGII in MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func corrThickness@shape ; do
+
+  STRINGIIList="corrThickness@shape"
+  if [ "${T2wPresent}" = "YES" ] ; then
+    STRINGIIList+=" MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func"
+  fi
+
+  for STRINGII in $STRINGIIList ; do
```
mharms: Is there really a need to define a `STRINGIIList` variable here?  Isn't it always going to be equal to the previously defined `STRINGList`?    If so, then all we need is the single line:
`for STRINGII in $STRINGList; do`

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -209,11 +246,19 @@ for STRING in "$T1wFolder"/"$NativeFolder"@"$AtlasSpaceFolder"/"$NativeFolder"@n
   FolderI=`echo $STRING | cut -d "@" -f 1`
   FolderII=`echo $STRING | cut -d "@" -f 2`
   Mesh=`echo $STRING | cut -d "@" -f 3`
-  for STRINGII in MyelinMap_BC@dscalar SmoothedMyelinMap_BC@dscalar corrThickness@dscalar ; do
+
+  STRINGIIList="corrThickness@dscalar"
+  if [ "${T2wPresent}" = "YES" ] ; then
+    STRINGIIList+=" MyelinMap_BC@dscalar SmoothedMyelinMap_BC@dscalar"
+  fi
```
mharms: Now here, we do indeed need to define a new list, so this is necessary.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -141,6 +161,27 @@ FreeSurferFolder="$T1wFolder"/"$FreeSurferFolder"
...
 
+
+# ------------------------------------------------------------------------------
+#  Compliance check
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
```
mharms: Not sure how you've structured all the other scripts, but shouldn't the parsing of the `--mpp-mode` value, and the setting of its default value, occur earlier when that is done for all the other variables?  And shouldn't the value of `--mpp-mode` then be included as one of the `verbose_echo` statements?

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -86,21 +98,29 @@ ${CARET7DIR}/wb_command -volume-palette $Jacobian MODE_AUTO_SCALE -interpolate t
 # To detect this situation, and prevent it from occurring, FreeSurferPipeline errors out if the
 # one-step warpfield created here ($OutputOrigT2wToT1w) already exists.]
```
mharms: Please change this comment to the following to reflect the change to checking $OutputOrigT1wToT1w instead in FreeSurferPipeline.
```
# To detect this situation, and prevent FreeSurferPipeline from running after PostFreeSurferPipeline,
# FreeSurferPipeline errors out if the one-step warpfield created here ($OutputOrigT1wToT1w) already exists.]

```

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -86,21 +98,29 @@ ${CARET7DIR}/wb_command -volume-palette $Jacobian MODE_AUTO_SCALE -interpolate t
...
 convertwarp --relout --rel --ref="$T1wImageBrainMask" --premat="$InitialT1wTransform" --warp1="$dcT1wTransform" --out="$OutputOrigT1wToT1w"
-convertwarp --relout --rel --ref="$T1wImageBrainMask" --premat="$InitialT2wTransform" --warp1="$dcT2wTransform" --postmat="$FinalT2wTransform" --out="$OutputOrigT2wToT1w"
+
+if [ "${T2wPresent}" = "YES" ] ; then
+  convertwarp --relout --rel --ref="$T1wImageBrainMask" --premat="$InitialT2wTransform" --warp1="$dcT2wTransform" --postmat="$FinalT2wTransform" --out="$OutputOrigT2wToT1w"
+fi
```
mharms: Let's consolidate this into the `if` conditional immediately following.  (i.e., put the `convertwarp` and `applywarp` commands for the T2w image condition back-to-back).

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -86,21 +98,29 @@ ${CARET7DIR}/wb_command -volume-palette $Jacobian MODE_AUTO_SCALE -interpolate t
...
+  fslmaths "$OutputT2wImageRestore" -mas "$T1wImageBrainMask" "$OutputT2wImageRestoreBrain"
+fi
 
 # Do the same for the equivalents in MNINonLinear space
 convertwarp --relout --rel --ref="$T1wImageBrainMask" --warp1="$OutputOrigT1wToT1w" --warp2="$AtlasTransform" --out="$OutputOrigT1wToStandard"
-convertwarp --relout --rel --ref="$T1wImageBrainMask" --warp1="$OutputOrigT2wToT1w" --warp2="$AtlasTransform" --out="$OutputOrigT2wToStandard"
+
+if [ "${T2wPresent}" = "YES" ] ; then
+  convertwarp --relout --rel --ref="$T1wImageBrainMask" --warp1="$OutputOrigT2wToT1w" --warp2="$AtlasTransform" --out="$OutputOrigT2wToStandard"
+fi
```
mharms: Ditto.  Consolidate with the `applywarp` command for the MNI space.

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: It is placed here because it depends on the definition of `T2wFolder` and I wanted to have everything related to compliance checking in one place. But you are right. Your suggestion makes sense. I'll make the change.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: done

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Done.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Done.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -81,6 +81,9 @@ CorrectionSigma=`opts_GetOpt1 "--mcsigma" $@`
 RegName=`opts_GetOpt1 "--regname" $@`
 InflateExtraScale=`opts_GetOpt1 "--inflatescale" $@`
 
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
```
mharms: Remove blank line preceding setting of MPPMode, and move the setting of its default value down several lines to where the defaults on other parameters are set.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -110,20 +122,23 @@ fslmaths "$OutputMNIT1wImage" -abs "$OutputMNIT1wImage" -odt float
 fslmaths "$OutputMNIT1wImage" -div "$BiasFieldOutput" "$OutputMNIT1wImageRestore"
 fslmaths "$OutputMNIT1wImageRestore" -mas "$T1wMNIImageBrainMask" "$OutputMNIT1wImageRestoreBrain"
 
-applywarp --rel --interp=spline -i "$OrginalT2wImage" -r "$T1wImageBrainMask" -w "$OutputOrigT2wToStandard" -o "$OutputMNIT2wImage"
-fslmaths "$OutputMNIT2wImage" -abs "$OutputMNIT2wImage" -odt float
-fslmaths "$OutputMNIT2wImage" -div "$BiasFieldOutput" "$OutputMNIT2wImageRestore"
-fslmaths "$OutputMNIT2wImageRestore" -mas "$T1wMNIImageBrainMask" "$OutputMNIT2wImageRestoreBrain"
+if [ "${T2wPresent}" = "YES" ] ; then
```
mharms: I'd still prefer for this `if` conditional to terminate after L130, and then start a new conditional after L132 for the "# Create T1w/T2w maps" section.  I know it is a bit redundant, but otherwise, you can get the impression that the "# Create T1w/T2w maps" section is somehow part of the MNI space section, which isn't the case.

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: Ok.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Done.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: done

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Yep, and we don't need to have it within the loop, so I put it before the loop.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: :-)

PostFreeSurfer/PostFreeSurferPipeline.sh
coalsont: Tim Brown may have made such things, I don't think I did.  I see only:

https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/fsl_version.shlib

coalsont: Also, I just noticed that that function puts the result into a variable with 'eval', when the right way to design it is for the function to echo the version number.  Since it is a function, it also shouldn't do an exit (a return, at most), and shouldn't use any `log_*` function since those output to stdout.

So no, there isn't a ready-made good way to report all that.  Just change the verbose echo to report HCPPIPEDIR and call it good enough for now.

mharms: @coalsont @sivcek My goal, once we get all these PRs merged into the master, is that we do some cross-script harmonization, make sure we can easily get usage statements, that all the scripts report the tool and pipeline version in a consistent fashion, etc.  If you see/think of something along those lines, please add it to here https://github.com/Washington-University/HCPpipelines/issues/146 so that we don't lose track of it.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -141,6 +165,29 @@ FreeSurferFolder="$T1wFolder"/"$FreeSurferFolder"
...
+# -- T2w image
+
+if [ `${FSLDIR}/bin/imtest ${T2wFolder}/T2w` -eq 0 ]; then
+    ComplianceMsg+=" T2w image not present"
+    Compliance="LegacyStyleData"
+    T2wRestoreImage="NONE"
+fi
+
+if [ "${RegName}" = "FS" ] ; then
+  logWarn "WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
```
mharms: `log_Warn` and please make sure message is identical to that used in `fMRISurface`

PostFreeSurfer/PostFreeSurferPipeline_1res.sh
```diff
@@ -140,6 +163,30 @@ FreeSurferFolder="$T1wFolder"/"$FreeSurferFolder"
...
+# -- T2w image
+
+if [ `${FSLDIR}/bin/imtest ${T2wFolder}/T2w` -eq 0 ]; then
+    ComplianceMsg+=" T2w image not present"
+    Compliance="LegacyStyleData"
+    T2wRestoreImage="NONE"
+fi
+
+if [ "${RegName}" = "FS" ] ; then
+  logWarn "WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
```
mharms: Ditto on `log_Warn` and matching the message.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -110,20 +122,31 @@ fslmaths "$OutputMNIT1wImage" -abs "$OutputMNIT1wImage" -odt float
...
+  convertwarp --relout --rel --ref="$T1wImageBrainMask" --warp1="$OutputOrigT2wToT1w" --warp2="$AtlasTransform" --out="$OutputOrigT2wToStandard"
+  applywarp --rel --interp=spline -i "$OrginalT2wImage" -r "$T1wImageBrainMask" -w "$OutputOrigT2wToStandard" -o "$OutputMNIT2wImage"
+  fslmaths "$OutputMNIT2wImage" -abs "$OutputMNIT2wImage" -odt float
+  fslmaths "$OutputMNIT2wImage" -div "$BiasFieldOutput" "$OutputMNIT2wImageRestore"
+  fslmaths "$OutputMNIT2wImageRestore" -mas "$T1wMNIImageBrainMask" "$OutputMNIT2wImageRestoreBrain"
+fi
+
+if [ "${T2wPresent}" = "YES" ] ; then
```
mharms: Please move the `# Create T1w/T2w maps` comment to preceding the conditional.

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
```diff
@@ -108,7 +125,13 @@ for STRING in ${STRINGII} ; do
...
   Mesh=`echo $STRING | cut -d "@" -f 2`
   ROI=`echo $STRING | cut -d "@" -f 3`
-  for STRINGII in MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func corrThickness@shape ; do
+
+  STRINGIIList="corrThickness@shape"
+  if [ "${T2wPresent}" = "YES" ] ; then
+    STRINGIIList+=" MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func"
+  fi
+
+  for STRINGII in $STRINGIIList ; do
```
mharms: This hasn't been updated to parallel the treatment in CreateMyelinMaps.sh

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: @mharms Agreed. We have some solutions in Qu|Nex code and I added (and then pulled out) a bit in mppchecking.shlib (now with another name).

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: I did. I also split it in multiple lines.

PostFreeSurfer/PostFreeSurferPipeline_1res.sh
sivcek: Same here.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
sivcek: Ok.

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
sivcek: Ops. Done.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -182,7 +182,14 @@ if [ `${FSLDIR}/bin/imtest ${T2wFolder}/T2w` -eq 0 ]; then
 fi
 
 if [ "${RegName}" = "FS" ] ; then
-  logWarn "WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
+    log_Warn "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated in the"
```
coalsont: log_Warn already prefixes "WARNING" to every line, this is redundant

PostFreeSurfer/PostFreeSurferPipeline_1res.sh
```diff
@@ -180,7 +180,14 @@ if [ `${FSLDIR}/bin/imtest ${T2wFolder}/T2w` -eq 0 ]; then
 fi
 
 if [ "${RegName}" = "FS" ] ; then
-  logWarn "WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
+    log_Warn "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated in the"
```
coalsont: Same, "WARNING" is redundant.

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
```diff
@@ -126,12 +126,7 @@ for STRING in ${STRINGII} ; do
   Mesh=`echo $STRING | cut -d "@" -f 2`
   ROI=`echo $STRING | cut -d "@" -f 3`
 
-  STRINGIIList="corrThickness@shape"
-  if [ "${T2wPresent}" = "YES" ] ; then
-    STRINGIIList+=" MyelinMap@func SmoothedMyelinMap@func MyelinMap_BC@func SmoothedMyelinMap_BC@func"
-  fi
-
-  for STRINGII in $STRINGIIList ; do
+  for STRINGII in $STRINGIList ; do
```
coalsont: Since we are using `STRINGIList` again, unless we are inside the previous `STRINGI` loop, we should probably use `STRINGI` as the loop variable again, otherwise it looks like a bug.

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
```diff
@@ -146,16 +141,17 @@ for LowResMesh in ${LowResMeshes} ; do
 done
 
 #Add CIFTI Maps to Spec Files
+
+STRINGIIList="corrThickness@dscalar"
+if [ "${T2wPresent}" = "YES" ] ; then
+  STRINGIIList+=" MyelinMap_BC@dscalar SmoothedMyelinMap_BC@dscalar"
+fi
+  
 for STRING in ${STRINGII} ; do
```
coalsont: Looks like this should be `STRINGIIList`.

coalsont: Never mind, I didn't look at the context.  I would argue that we shouldn't use `STRINGII` to store a list and then loop over it, as it looks wrong.

coalsont: However, that is our fault, not yours.

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
mharms: I think this is a typo.  There is no `STRINGIList` (or `STRINGI`) variable.  It should be
`  for STRINGII in $STRINGList ; do`

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
mharms: I agree, and I had already previously flagged this exact confusion as item 6 in #146 for some post PR cleanup that is our responsibility.

PostFreeSurfer/PostFreeSurferPipeline.sh
mharms: And as a corollary, if you want to add a couple spaces to lines 2-onward to indent the text a little bit following the 
`<dateTime>; <toolName>: WARNING: `
"prefix", that is fine, but no need to indent by the full width of "WARNING".

PostFreeSurfer/PostFreeSurferPipeline_1res.sh
mharms: Also, update to match the multi-line version.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -230,4 +283,8 @@ argList+="$CorrectionSigma "
 argList+="$RegName "                                  # ${39}
 "$PipelineScripts"/CreateMyelinMaps.sh ${argList}
 
+verbose_green_echo "---> Finished Post FreeSurfer Pipeline"
```
mharms: This is a quibble, but IMO we should refer to the scripts by their actual names (without spaces) -- when you start adding spaces, it can get arbitrary where spaces should go for some of the more convoluted script names.
i.e.,
`verbose_green_echo "---> Finished PostFreeSurferPipeline"`

mharms: Or in this script, where `script_name` is defined, it could simply be:
`verbose_green_echo "---> Finished ${script_name}"`

PostFreeSurfer/scripts/CreateMyelinMaps_1res.sh
sivcek: Fixed.

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: Took out the spaces.

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: Changed.

PostFreeSurfer/PostFreeSurferPipeline_1res.sh
sivcek: Changed

PostFreeSurfer/PostFreeSurferPipeline.sh
sivcek: Changed to using variable.

sivcek:
Default set to MSMSulc

PR #136, opened 2019-09-27
fMRISurface related pull request

sivcek:
### 1/ Cosmetic changes
Changing tabs to spaces, removal of trailing line spaces 

### 1/ Cosmetic changes
Changing tabs to spaces, removal of trailing line spaces 

### 2/ Addition of `--mppversion` parameter and related check / validation code

### DIFF
--- a/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
@@ -73,7 +73,7 @@ log_Msg "RegName: ${RegName}"
 log_Msg "RUN: ${RUN}"
 
 if [ "${RegName}" = "FS" ] ; then
-  log_Warn "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
+  log_Warn "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularization of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
 fi
 
 # Setup PATHS
--- a/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
@@ -73,7 +73,7 @@ log_Msg "RegName: ${RegName}"
 log_Msg "RUN: ${RUN}"
 
 if [ "${RegName}" = "FS" ] ; then
-  logWarn "WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
+  log_Warn "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
 fi
 
 # Setup PATHS
--- a/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
@@ -64,14 +64,17 @@ fi
 RUN=`opts_GetOpt1 "--printcom" $@`  # use ="echo" for just printing everything and not running the commands (default is to run)
 
 # ------------------------------------------------------------------------------
-#  Check MMP Version
+#  Compliance check
 # ------------------------------------------------------------------------------
 
 MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
 MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
-check_mpp_compliance "fMRISurface"
+Compliance="HCPStyleData"
+ComplianceMsg=""
 
-# --- END MPP Version Check
+check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
+
+# --- End compliance check
 
 
 log_Msg "Path: ${Path}"
--- a/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
@@ -1,15 +1,15 @@
-#!/bin/bash 
+#!/bin/bash
 set -e
 
 # Requirements for this script
-#  installed versions of: FSL (version 5.0.6), FreeSurfer (version 5.3.0-HCP) , gradunwarp (HCP version 1.0.2) 
+#  installed versions of: FSL (version 5.0.6), FreeSurfer (version 5.3.0-HCP) , gradunwarp (HCP version 1.0.2)
 #  environment: use SetUpHCPPipeline.sh  (or individually set FSLDIR, FREESURFER_HOME, HCPPIPEDIR, PATH - for gradient_unwarp.py)
 
-########################################## PIPELINE OVERVIEW ########################################## 
+########################################## PIPELINE OVERVIEW ##########################################
 
 # TODO
 
-########################################## OUTPUT DIRECTORIES ########################################## 
+########################################## OUTPUT DIRECTORIES ##########################################
 
 # TODO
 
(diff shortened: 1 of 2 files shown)


fMRISurface/GenericfMRISurfaceProcessingPipeline_1res.sh
```diff
@@ -62,6 +62,49 @@ fi
...
+# ------------------------------------------------------------------------------
+#  Check MMP Version
+# ------------------------------------------------------------------------------
+
+Compliance="hcp"
+MPPVersion=`opts_GetOpt1 "--mppversion" $@`
+MPPVersion=`opts_DefaultOpt $MPPVersion "hcp"`
+
+if [ "${MPPVersion}" = "legacy" ] ; then
+  log_Msg "Legacy Minimal Preprocessing Pipelines: fMRISurface v.XX"
```
coalsont: These checks/messages that are pasted into a bunch of places seem like maybe they should be in a global script or function instead (you could pass in the tool name as a string).

I don't think we really want to start putting hardcoded version numbers in each script, we are unlikely to keep them up to date.  We do currently have a function to show the version of the entire pipelines repo, though I'm not thrilled with it either (because it requires releases to do 2 additional commits: set version to "v4.1.0" or whatever, commit, tag, set version to "Post-v4.1.0", commit).  Unfortunately, since downloading .zip files from github instead of cloning is something we support, I haven't thought of a better solution.

sivcek: Ok. We can put that as a function in a global script and read the version of the pipelines there. What is the function that prints the version?

coalsont: https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/version.shlib

See:
https://github.com/Washington-University/HCPpipelines/blob/f00cdc5d80d3434640a4aeba929824697b5bcd9c/DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh#L71
https://github.com/Washington-University/HCPpipelines/blob/f00cdc5d80d3434640a4aeba929824697b5bcd9c/DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh#L190

Though I advocate always using `"$@"` instead of `$@`.  Some other parts of the code just `cat` the version.txt file, and there is also a show_version script...it is a bit of a mess.

coalsont: Actually, to avoid saying "HCP" in legacy mode, you should probably just `cat "$HCPPIPEDIR"/version.txt` into a variable.

sivcek: Added global mppmodecheck.shlib with version checking.

coalsont:
For clarity: since there is no internal change to this pipeline for legacy data, is the purpose of adding the flag just so that the QuNex framework doesn't have to keep track of which pipelines have optional legacy behavior?

sivcek:
The flag can be omitted it is just consistent now across pipelines and it opens up a possibility of adding changes later on.

sivcek:
Also, if the user sets the "legacy" mppmode for the study, the logs from all the pipelines will be marked appropriately.

coalsont:
The purpose of the flag is to allow the hcp pipelines to be run on legacy datasets that are missing certain inputs.  Thus, I don't think it is expected that we will incrementally add more options controlled by this flag, because the inputs to some pipelines are only the outputs of other pipelines, so they should never need such an option.

A user running the pipelines manually may not edit all the legacy options to enable them in scripts that run without them.  If we want to ensure the log records if the user has used any legacy options for a subject, the main way I can think of is to create a file in the subject directory when a legacy option is used, permanently marking that subject and generating logging messages as such (and we shouldn't worry about users that might try to trick the detection).

sivcek:
I'll just remove it then. It should be obvious from the other logs what was run. And if somebody wants to check them, they can.

glasserm:
In the interest of irritating Grega with back and forth changes (:P) I would actually at this point class RegName="FS" into legacy mode both in PostFreeSurfer and here, as we really want to discourage using FreeSurfer registration now that MSMSulc is widely available and we have shown it to be substantially better in multiple publications.

coalsont:
So, reclassifying FS registration would break the idea of this option being only used for aspects of the input data.  If we say it is always harmful compared to MSMSulc on all data (what about tumor lesions, AVM, etc?), then it would fall into my category of "should not be added to the pipelines at all", however since it has already been available, removing it entirely may be unneccesarily unfriendly.

In my ideal world, if we wanted to change it to be an error by default, we would have two separate options: one to control whether to allow never-recommended legacy features of the *pipeline* (FS reg), and one to allow features of the pipeline that can be appropriate on legacy *data* (no T2 image).  This would draw a clear distinction between "making the best of what you have" and "choosing to do something bad to your data", and produce specific messages for each case.

If we want to have only one option to control these two different categories of behaviors, which by definition is more confusing, then arguably it should use somewhat generic terminology.

coalsont:
To be clear, I think entirely removing the code supporting the FS registration might be an acceptable solution, because if someone wants to match what they did previously on other data, best practice is to use the older release of the pipelines for consistency.

coalsont:
I suppose we could instead add a third option to the existing flag for this situation, something like `"--mpp-mode=HCPStyleData | LegacyStyleData | LegacyPipelineOptions"`, if we specifically consider "using legacy options like FS reg" to be worse than "using legacy data".

glasserm:
Tim's 3 option solution seems reasonable.  I do like the idea of a recommended path for legacy data and further excluding non-recommended options.  Here is a warning for FS registrattion (to be put in both fMRISurface and PostFreeSurfer:

WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc.  Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic).  These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns.  See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details.

mharms:
Usually when something is "deprecated" there is plenty of advance warning, but it continues to work "as is".  The path consistent with that approach would be to add a warning message/statement, but leave RegName=FS functional without needing to invoke a special "mode" to use it.

glasserm:
Given the issues with FS being used in the container for HCP processing despite clear instructions not to, I think we need to be more aggressive and force the user's attention on this one.

mharms:
That would be my preferred approach, and more consistent with the usual path for deprecating a feature.

sivcek:
Ok. Then all should be ready.

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
```diff
@@ -1,24 +1,24 @@
-#!/bin/bash 
+#!/bin/bash
 set -e
 
 # Requirements for this script
-#  installed versions of: FSL (version 5.0.6), FreeSurfer (version 5.3.0-HCP) , gradunwarp (HCP version 1.0.2) 
+#  installed versions of: FSL (version 5.0.6), FreeSurfer (version 5.3.0-HCP) , gradunwarp (HCP version 1.0.2)
```
mharms: @coalsont I've always found the wording `# Requirements for this script` confusing in the context of listing specific versions of the tools.  e.g., It could be interpreted as FSL 5.0.6 is the _only_ supported version.  I think we should either change to `# Minimum requirements for this script`, or simply list the tools without specific version numbers (or perhaps even both).

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
```diff
@@ -72,6 +72,10 @@ log_Msg "GrayordinatesResolution: ${GrayordinatesResolution}"
 log_Msg "RegName: ${RegName}"
 log_Msg "RUN: ${RUN}"
 
+if [ "${RegName}" = "FS" ] ; then
+  logWarn "WARNING: FreeSurfer registration is deprecated in the HCP Pipelines as it results in poorer cross-subject functional and cortical areal alignment relative to MSMSulc. Additionally, FreeSurfer registration results in dramatically higher surface distortion (both isotropic and anisotropic). These things occur because FreeSurfer's registration has too little regularizattion of folding patterns that are imperfectly correlated with function and cortical areas, resulting in overfitting of folding patterns. See Robinson et al 2014, 2018 Neuroimage, and Coalson et al 2018 PNAS for more details."
```
mharms: It's `log_Warn`.  Typo in "regularizattion".   I would start it with "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated ...".   Possibly split into individual lines, pending @coalsont's concurrence.

sivcek: I fixed the bug and changed the wording. Wating on whether to break into multiple lines.

sivcek: Split into multiple lines.

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
coalsont: Minimum tool versions might be useful, though we give such requirements for the pipelines as a whole, and updating them in each script would be tedious and error-prone.  Removing the version numbers from the scripts certainly has some appeal as the laziest way forward.

mharms: Captured for us to do as Item 8 of #146.

sivcek: Ok. I guess not an issue here.

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
```diff
@@ -56,8 +56,8 @@ SmoothingFWHM=`opts_GetOpt1 "--smoothingFWHM" $@`  # "${14}"
 GrayordinatesResolution=`opts_GetOpt1 "--grayordinatesres" $@`  # "${14}"
 RegName=`opts_GetOpt1 "--regname" $@`
 
-if [ "${RegName}" = "" ]; then
-    RegName="FS"
+if [ -z "${RegName}" ]; then
+    RegName="MSMSulc"
```
coalsont: Maybe I missed it, but I don't remember a request to change this default.  It may be a good change, it makes some sense with deprecating FS, but it does mean that the same settings and the new script will do something substantially different than before.  @glasserm @mharms?

mharms: Good point...

sivcek: Since FS is deprecated, I assumed that MSMSulc should be the new default. Do I change back to FS?

coalsont: Matt indicated that he thinks the default should be changed.  I am okay with it changing.

Anyone who is surprised by this change, wanting the same processing as an earlier run, should really have just used the same older version of the pipelines as before, so it doesn't seem like it has any major downside.

mharms: And if you were using FS, you probably didn't have MSM set up, so it is likely the new default won't run, and the user would be alerted to the change in that fashion.

glasserm: Yes the default for PostFreeSurfer and fMRISurface should be MSMSulc

sivcek: Changed in both PRs.

PR #137, opened 2019-09-29
mppcheck shlib

sivcek:
Added `mppmodecheck.shlib` to global scripts to be used across pipelines.


sivcek:
Ok. Will correct right away.

sivcek:
Terminology is now changed from standard/extended to HCPStyleData/LegacyStyleData

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+then
+    source "$HCPPIPEDIR/global/scripts/log.shlib" "$@"
+fi
+
+
+# --- Functions to check that a valid pipeline name was passed
+
+check_pipeline_name()
+{
+    for valid_pipeline in PreFreeSurfer FreeSurfer FreeSurfer53HCP PostFreeSurfer fMRIVolume fMRISurface Difussion
```
mharms: Typo: Difussion

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+        log_Msg "T2w Image present: ${p_t2w_image} (HCPStyleData)"
+    fi
+}
+
+
+check_mpp_PostFreeSurfer()
+{
+    # -- T2w image
+
+    if [ ! -e ${T2wFolder}/T2w.nii.gz ]; then
```
mharms: Similar to using `imcp` and `imrm`, we generally try to use `imtest` when checking for existence of a NIFTI (and exclude the .nii.gz extension).

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+      log_Msg "-> T2w image present (HCPStyleData)"
+    fi
+}
+
+
+check_mpp_fMRIVolume()
+{
+    # -- Slice timing correction
+
+    if [ ! "${DoSliceTimeCorrection}" == 'NONE' ]; then
```
mharms: Minor consistency issue:  Earlier used `=`.  Here onward seems to switch to using `==`.

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+      log_Msg "-> Set to run slice timing correction (LegacyStyleData)"
+      Compliance="LegacyStyleData"
+    else
+      log_Msg "-> Not running slice timing correction (HCPStyleData)"
+    fi
+
+    # -- Use of BOLD scout image
+
+    if [ "${fMRIScout}" == 'NONE' ]; then
```
mharms: I'll have to see the changes in `fMRIVolume`, but `--mpp-mode=HCPStyleData` should already the fMRIScout=NONE condition.  So, this would presumably be changing existing pipeline behavior.

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+    local hcpversion=`cat "$HCPPIPEDIR"/version.txt`
+    local Compliance="HCPStyleData"
+    local ispipeline="false"
+
+    # --- check and abort if pipeline name is not known
+
+    check_pipeline_name
+
+    if [ "${ispipeline}" == "false" ]; then
+        log_Err_Abort "${pipeline} is not a valid HCP Pipeline name. Aborting execution!"
```
mharms: FWIW, `log_Err_Abort` already concludes with "ABORTING".  So, putting "Aborting execution!" into the message is redundant.

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+      log_Msg "Checking data compliance with HCPStyleData HCP MPP mode"
+    fi
+
+    check_mpp_${pipeline}
+
+    # -- Final evaluation
+
+    if [ "${MPPMode}" = "LegacyStyleData" ]; then
+      if [ "${Compliance}" = "LegacyStyleData" ]; then
+        log_Msg "Processing will continue using Extended MPP mode."
```
mharms: Replace "Extended" with "LegacyStyleData"

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+
+    if [ "${ispipeline}" == "false" ]; then
+        log_Err_Abort "${pipeline} is not a valid HCP Pipeline name. Aborting execution!"
+    fi
+
+    # --- Initial information
+
+    if [ "${MPPMode}" = "LegacyStyleData" ]; then
+      log_Msg "LegacyStyleData Minimal Preprocessing Pipelines: ${pipeline} ${version}"
+      log_Msg "NOTICE: You are using MPP mode that enables processing of images that do not"
```
mharms: ... using _an_ MPP mode that enables processing of _acquisitions_ that ...
or just
... using a mode that enables processing of _acquisitions_ that ...

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+    do
+        if [ "${valid_pipeline}" = "${pipeline}" ]; then
+            ispipeline="true"
+        fi
+    done
+}
+
+# --- Functions for pipeline specific checks
+
+check_mpp_PreFreeSurfer()
```
coalsont: I was thinking of leaving the logic of what is or is not hcp-compliant in the pipelines, and this file would just contain the standard log, warning, and error messages.  That way, this file wouldn't need to know any details about the pipelines.

mharms: That was what I interpreted your proposal as (just a simple new function added to the log.shlib library), and I like the idea of not needing to go to the `check_mpp_` functions to see the relevant logic.

global/scripts/mppmodecheck.shlib
sivcek: Changed.

coalsont:
To be clear, the solution I am thinking of would look something like this:

setup.sh:
```
...
export MPP_INPUT_MODE=hcp
...
```

PreFreeSurfer.sh:
```
...
compliance=hcp
if [[ "$T2w" == "NONE" ]]
then
    compliance=permissive
fi
...
"${HCPPIPEDIR}/global/scripts/inputModeCheck.sh" "$compliance"
```

inputModeCheck.sh:
```
case "$MPP_INPUT_MODE" in
    ("")
    (hcp)
        if [[ "$1" != "hcp" ]]
        then
            log_Err_Abort "<stuff>"
        fi
        ;;
    (permissive)
        log_Msg "<stuff>"
...
esac
```

This way, the pipeline scripts each contain their own logic about what is hcp-compliant, keeping it close to the other edits for allowing non-hcp-style input, but only one script has to know what mode the user requested, without requiring any new option flags anywhere.

global/scripts/mppmodecheck.shlib
sivcek: Harmonized to `=` only throughout the script.

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,231 @@
...
+      log_Msg "NOTICE: You are using MPP mode that enables processing of images that do not"
+      log_Msg "        conform to the HCP specification as described in Glasser et al. (2013)!"
+      log_Msg "        Be aware that if the HCP requirements are not met, the level of data quality"
+      log_Msg "        can not be guaranteed and the Glasser et al. (2013) paper should not be used"
+      log_Msg "        in support of this workflow. A manuscript with comprehensive evaluation for"
+      log_Msg "        the Extended MPP workflow is in active preparation and should be appropriately"
+      log_Msg "        cited when published."
+      log_Msg "        "
+      log_Msg "Checking available data and processing options"
+    else
```
coalsont: It isn't clear to me that we should print anything if the mode and compliance are both hcp.

global/scripts/mppmodecheck.shlib
sivcek: I'll remove the check then.

global/scripts/mppmodecheck.shlib
sivcek: Removed.

global/scripts/mppmodecheck.shlib
sivcek: Corrected.

global/scripts/mppmodecheck.shlib
sivcek: Changed.

global/scripts/mppmodecheck.shlib
sivcek: The check starts with:
> Checking data compliance with HCPStyleData HCP MPP mode

Wouldn't it make sense to report that the check was successful?

global/scripts/mppmodecheck.shlib
sivcek: I was not sure about the initial suggestion. Then it made sense that all the criteria for what is and what is not HCPStyleData or LegacyStyleData are in one file, so that it is easy to harmonize the checks across multiple scripts. 

I can restructure the code. But before I do, let's discuss pro's and con's. @mharms @glasserm, do you have a preference?

coalsont:
To add explanatory messages to the outline I gave (if we don't want them to show up unless there is a problem, which is my preference), we could do something like:

PreFreeSurfer.sh:
```
...
compliance=hcp
complianceMsg=""
if [[ "$T2w" == "NONE" ]]
then
    compliance=permissive
    complianceMsg+=" T2w=NONE"
fi
...
"${HCPPIPEDIR}/global/scripts/inputModeCheck.sh" "$compliance" "$complianceMsg"
```

inputModeCheck.sh:
```
...
            log_Err_Abort "hcp-style input mode was requested, but the following settings are not compliant:$2.  Fix these settings, or if you do not have the required data, put 'export MPP_INPUT_MODE=permissive' in your setup script."
...
```

The main issue with doing the tests this way is if someone needs to run some pipelines in permissive mode, but can run others in hcp mode.

sivcek:
The idea of explicitly stating, what mode the pipeline was run in, is that there is a clear record of it and just checking the start of the script it is easy to see whether the HCP requirements were met or not.

sivcek:
Also, we do want to support the choice of mode to be a parameter in a parameter file or provided when running a Qu|Nex command, so having it set as a global variable is problematic.

coalsont:
If the script runs at all, either the hcp requirements are met (my preference, with no message), or the legacy mode was requested (which should still produce a message at the beginning).  My point was that if everything is hcp-compliant, we don't really need a message about every single check done for hcp compliance.

The outline I gave can be adapted to instead use a parameter switch per-script, just add the requested mode as another parameter to inputModeCheck.sh.

global/scripts/mppmodecheck.shlib
coalsont: The design rule I am thinking of is that the fewer files know the internal details of other files, the better (if we swapped out the argument parsing code in a pipeline, variable names could change, breaking the current sourced functions that use these internal details).  As long as we need separate code per pipeline for checking each pipeline's compliance, that code should be in the same file as the pipeline it checks (especially if it uses internal variable names).

My suggestion to consolidate the long identical messages was simply that if we wanted to tweak the wording, editing the identical wording in 4+ files would have been inefficient and error-prone.

coalsont:
Actually, in order for it to abort the parent script (without tripping the ERR trap), it does still need to be sourced instead of called, but it should still be just as minimal.

global/scripts/mppmodecheck.shlib
mharms: I agree with @coalsont's concern here, and prefer his suggested approach.

global/scripts/mppmodecheck.shlib
coalsont: I would also remove the message stating that it is starting checking anything, and really all messages except either reporting failure, or activation of legacy mode (and also report only the non-compliant parameters, or that legacy mode is active with full hcp compliance).

My personal style is that messages can be long if necessary, but if a message doesn't say something that deserves someone's attention, then it shouldn't exist.

sivcek:
I'm about to put the checking back to individual scripts. I would though like to know the consensus on whether the log should or should not contain the information that the script was run in the hcp-compliant mode. @mharms @glasserm

sivcek:
The minimal mode checking script is now in place and used in PreFreeSurfer. If this is ok, I'll change the other PRs accordingly. For a number of other changes a consesus needs to be reached so that I don't change things back and forth.

global/scripts/mppmodecheck.shlib
```diff
@@ -189,29 +28,19 @@ check_mpp_compliance()
...
     fi
 
-    check_mpp_${pipeline}
-
     # -- Final evaluation
 
     if [ "${MPPMode}" = "LegacyStyleData" ]; then
-      if [ "${Compliance}" = "LegacyStyleData" ]; then
-        log_Msg "Processing will continue using LegacyStyleData MPP mode."
+      if [ -z "${ComplianceMsg}" ]; then
```
coalsont: As I commented on a commit, do not use the existence of a message in the logic of whether to abort.  Message formatting is secondary, and should never replace the logic parameters.

coalsont: For reference: https://github.com/Washington-University/HCPpipelines/commit/2c4be159ccb44db54a0506b4d677b17b3ad87e1a#diff-85cc56a06f774b7c676d1160cade6f3bR402

mharms:
I'm fine with @coalsont's preference that we don't need to report that things were run in "hcp-compliant" mode, since that is the default anyway.

coalsont:
While I'm at it, in addition to misusing the message argument for the logic test, the name "mppcheck.shlib" doesn't really relate to its specific purpose.  So, lets start another naming argument: I'll start with `input_mode.shlib`.  @mharms?

mharms:
If this is just going to  basically be relatively simple messaging related function, why not put it into `log.shlib`, and name the function something like `log_Msg_MppCompliance`?

coalsont:
I don't particularly like putting it in the log.shlib, which is complicated already, and doesn't yet contain anything that is this single-purpose - log_Msg is very generic and gets reused everywhere, even the log_File_Must_Exist function could be used for many things, but these messages serve only one purpose - at the start of a pipeline, to indicate if things are not hcp-compliant.  I don't think editing or understanding the messages would be improved by burying it among the logging code, and I don't think the logging shlib would be improved by including these singular-purpose messages.

As for convenience, the input checking shlib already sources the logging shlib, so you never need to source both.

global/scripts/mppmodecheck.shlib
sivcek: implemented

global/scripts/mppmodecheck.shlib
sivcek: Done.

global/scripts/mppmodecheck.shlib
sivcek: Changed.

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,52 @@
...
+    if [ "${MPPMode}" = "LegacyStyleData" ]; then
+      log_Msg "LegacyStyleData mode requested."
+      log_Msg "NOTICE: You are using a mode that enables processing of acquisitions that do not"
+      log_Msg "        conform to the HCP specification as described in Glasser et al. (2013)!"
+      log_Msg "        Be aware that if the HCP requirements are not met, the level of data quality"
+      log_Msg "        can not be guaranteed and the Glasser et al. (2013) paper should not be used"
+      log_Msg "        in support of this workflow. A manuscript with comprehensive evaluation for"
+      log_Msg "        the Extended MPP workflow is in active preparation and should be appropriately"
+      log_Msg "        cited when published."
+      log_Msg "        "
```
mharms: @glasserm Are you good with this "NOTICE" re the LegacyStyleData mode?  I'm worried it might "scare" off people from citing the 2013 paper at all, even if using just one "small" new feature.  (This plays into the broader issue of exactly which "mode" all the new features are going to be assigned to).

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,52 @@
...
+
+    if [ "${MPPMode}" = "LegacyStyleData" ]; then
+      if [ "${Compliance}" = "HCPStyleData" ]; then
+        log_Warn "All conditions for the use of HCPStyleData HCP MPP are met. Consider using HCPStyleData HCP MPP mode instead of LegacyStyleData MPP mode."        
+      else
+        log_Msg "The following LegacyStyleData settings were requested:${ComplianceMsg}"
+      fi
+    else
+      if [ "${Compliance}" = "LegacyStyleData" ]; then
+        log_Err_Abort "HCPStyleData HCP MPP mode was requested, but the following settings are not compliant:${ComplianceMsg}. Fix these settings or consider requesting 'LegacyStyleData' mode."
```
mharms: Similarly, does anyone prefer the following:
`log_Err_Abort "--mpp-mode=HCPStyleData was requested, but the following settings are not compliant:${ComplianceMsg}. Fix these settings or consider requesting 'LegacyStyleData' mode instead."`

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,52 @@
...
+      log_Msg "        the Extended MPP workflow is in active preparation and should be appropriately"
+      log_Msg "        cited when published."
+      log_Msg "        "
+    fi
+
+    # -- Final evaluation
+
+    if [ "${MPPMode}" = "LegacyStyleData" ]; then
+      if [ "${Compliance}" = "HCPStyleData" ]; then
+        log_Warn "All conditions for the use of HCPStyleData HCP MPP are met. Consider using HCPStyleData HCP MPP mode instead of LegacyStyleData MPP mode."
```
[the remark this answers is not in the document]
sivcek: I like it. It is more specific and provides clear instructions.

global/scripts/mppmodecheck.shlib
sivcek: Same here. I implemented both.

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,56 @@
+# Functions for checking compliance with selected MPP mode for HCP Pipelines
```
coalsont: Similar to the debugging shlib, I'd suggest the first line be `#!/bin/echo This script should not run directly:`

global/scripts/mppmodecheck.shlib
```diff
@@ -0,0 +1,56 @@
...
+    # -- Final evaluation
+
+    if [ "${MPPMode}" = "LegacyStyleData" ]; then
+      if [ "${Compliance}" = "HCPStyleData" ]; then
+        log_Warn "All conditions for the use of --mpp-mode=HCPStyleData are met. Consider using that mode (or equivalently, simply leave --mpp-mode empty) instead of --mpp-mode=LegacyStyleData."
+      else
+        log_Msg "The following LegacyStyleData settings were requested:${ComplianceMsg}"
+      fi
+      if [ ! -z "${ComplianceWarn}" ] ; then
+        log_Warn "${ComplianceWarn}"
```
coalsont: I'm guessing this is for the warnings of the nature "freesurfer registration is not as good as MSMSulc, please consider changing"?  Putting it here means that if the script will abort because of a compliance check, it won't print the warning about the requested option.

What happens if there are multiple options that we want warnings on in one script, how do you plan to separate them?  I think just putting the log_Warn directly into the pipeline script is the simplest solution.

@glasserm if someone asks for FS registration and doesn't specify legacy mode, do you want it to show the warning about *why* we recommend not using FS, in addition to aborting?

global/scripts/mppmodecheck.shlib
sivcek: added.

global/scripts/mppmodecheck.shlib
sivcek: I was looking at ways to combine warnings, but found nothing robust. If you have a suggestion, how to combine and pass them in a single variable and then split and loop through them, it would be most welcome!

coalsont: My suggestion is to never combine them in the first place, it is acceptable, and possibly preferable, to have these warnings unconditional (except for whether the setting was specified, of course), and in the pipeline scripts themselves.

Bash does not have a clean way of passing an array of strings as an argument, as far as I am aware.  Since these are sourced functions, and thus running in the "workspace" of the pipeline script, there are tricks that can be used, but I think it is better to just put the log_Warn directly into the relevant argument check in the main script.

For reference only, within the context of one script and its functions, a global array variable is clean for storing arrays of strings:

```
myarray=()
...
myarray+=("somevar is $somevar")
myarray+=("the quick brown fox")
```

sivcek: Ok. I'll pull the warnings into individual scripts.

global/scripts/mppmodecheck.shlib
glasserm: I am okay with the wording, assuming we only put things that trigger it as those we don't want people confusing with the main HCP Pipelines.

global/scripts/mppmodecheck.shlib
glasserm: Yes lets print both the notice and the warning.

global/scripts/processingmodecheck.shlib
```diff
@@ -0,0 +1,54 @@
...
+    # --- Initial information
+
+    if [ "${ProcessingMode}" = "LegacyStyleData" ]; then
+      log_Msg "LegacyStyleData mode requested."
+      log_Msg "NOTICE: You are using a mode that enables processing of acquisitions that do not"
+      log_Msg "        conform to the HCP specification as described in Glasser et al. (2013)!"
+      log_Msg "        Be aware that if the HCP requirements are not met, the level of data quality"
+      log_Msg "        can not be guaranteed and the Glasser et al. (2013) paper should not be used"
+      log_Msg "        in support of this workflow. A manuscript with comprehensive evaluation for"
+      log_Msg "        the Extended MPP workflow is in active preparation and should be appropriately"
```
coalsont: We should consider removing "MPP" here too, perhaps say "legacy data workflow" instead?

mharms: Absolutely!  Don't want to let "MPP" slip in given all the work we've done to avoid it.

global/scripts/processingmodecheck.shlib
```diff
@@ -0,0 +1,54 @@
+#!/bin/echo This script should not run directly:
+
+# Functions for checking compliance with selected MPP mode for HCP Pipelines
```
mharms: "MPP" snuck in here as well.

global/scripts/processingmodecheck.shlib
```diff
@@ -0,0 +1,54 @@
...
+      log_Msg "        the Extended MPP workflow is in active preparation and should be appropriately"
+      log_Msg "        cited when published."
+      log_Msg "        "
+    fi
+
+    # -- Final evaluation
+
+    if [ "${ProcessingMode}" = "LegacyStyleData" ]; then
+      if [ "${Compliance}" = "HCPStyleData" ]; then
+        log_Warn "All conditions for the use of --processing-mode=HCPStyleData are met. Consider using that mode (or equivalently, simply leave --processing-mode empty) instead of --processing-mode=LegacyStyleData."
```
mharms: Candidate for splitting into two lines.

global/scripts/processingmodecheck.shlib
```diff
@@ -0,0 +1,54 @@
...
+      else
+        log_Msg "The following LegacyStyleData settings were requested:${ComplianceMsg}"
+      fi
+    else
+      if [ "${Compliance}" = "LegacyStyleData" ]; then
+        log_Err_Abort "--processing-mode=HCPStyleData was requested, but the following settings are not compliant:${ComplianceMsg}. Fix these settings or consider requesting 'LegacyStyleData' mode instead."
```
mharms: Candidate for 3 lines, which could be implemented as:
```
log_Err "--processing-mode=HCPStyleData was requested, but the following settings are not compliant:"
log_Err "${ComplianceMsg}"
log_Err_Abort "Fix these settings or consider requesting 'LegacyStyleData' mode instead."
```

global/scripts/processingmodecheck.shlib
coalsont: I'm not so sure about this one, it doesn't seem great for it to appear like:

```
Wed Oct  9 16:14:40 CDT 2019:PreFreeSurfer.sh: WARNING: All conditions for the use of ...
Wed Oct  9 16:14:40 CDT 2019:PreFreeSurfer.sh: WARNING: Consider using that mode...
```

Repeating the date and "WARNING" is a bit less clear.  If you wanted to repeat the date, then the second could be a log_Msg, but I'd rather not repeat the date or introduce any artificial separation between these clearly related sentences, so I vote for leaving this one as-is.

coalsont: If you really want a manual line break, it is possible, though a bit ugly, to do something like this:

```
...
            log_Warn "All conditions for the use of --processing-mode=HCPStyleData are met.
Consider using that mode (or equivalently, simply leave --processing-mode empty) instead of --processing-mode=LegacyStyleData."
```

I still vote leave it as-is.

coalsont: Ah, I see the above blocks of log_Msg are doing some manual width control and will repeat the date and tool name on each line - I'm not sure if that kind of manual line splitting is how we really want to do long log messages, but I can live with it for now.

global/scripts/processingmodecheck.shlib
sivcek: It's like weeds ... so difficult to root out!

global/scripts/processingmodecheck.shlib
sivcek: pulled it out.

global/scripts/processingmodecheck.shlib
sivcek: Ok. Did it the same here.

global/scripts/processingmodecheck.shlib
sivcek: done

coalsont:
@sivcek is it okay with you if I edit our log.shlib on this branch, to make the multi-call `log_Err` formatting work better (to not repeat the full command a bunch of times)?

coalsont:
I have made this change, it is pretty simple, so I could redo it later if you need to force-push over it to resolve something.

sivcek:
@coalsont I think the commit is ok.

PR #139, opened 2019-10-03
fMRIVolume Related Pull Request — Slice timing correction

sivcek:
Addition of slice timing correction functionality.


fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,43 @@ then
...
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+DoSliceTimeCorrection=`opts_GetOpt1 "--doslicetime" $@`              # Whether to do slicetime correction (TRUE), FALSE to omit
+SliceCorrectionDirection=`opts_GetOpt1 "--slicetimedir" $@`          # direction: empty for up, --down for down
+SliceCorrectionInterleaved=`opts_GetOpt1 "--slicetimeodd" $@`        # interleaved: empty for not, --odd for interleaved
```
coalsont: There are more details than interleaving or not, and "not odd" doesn't mean "not interleaved".  The simple options act sort of like TV interlacing: progressive scan (1, 2, 3, 4...) or 2 interleaved passes, which can start with odd (1, 3, 5, 7...2, 4, 6...), or start with even (2, 4, 6...1, 3, 5...).

If the scanner sequence and the nifti orientation are different, though, you can get weirder things like reversed order (131, 129, 127...3, 1, 130, 128...), or the slice direction could be the first or second dimension, rather than the third (this is more likely in NHP scanned chest-down, chin-up or such).

Even more involved are multiband acquisitions, where you have sets of widely-spaced slices acquired at the same time, and I have no idea what the order of the sets usually is, the simplest case would be something like ((1 31 61 91), (2 32 62 92)...).

Obviously, the options don't need to support all this (I imagine the tool used doesn't support particularly many variations), but we should at least be careful with our option names, to make sure we are using the correct terminology, so that we can add additional options later if they become necessary.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,43 @@ then
...
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+
+# -- Slice timing correction
+
+if [ "${DoSliceTimeCorrection}" = 'TRUE' ]; then
```
coalsont: Do we have reason to believe that slice timing correction damages good data?  If not, do we really want to restrict it to legacy data mode?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -295,6 +332,16 @@ if [ ! -e "$fMRIFolder" ] ; then
 fi
 cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
 
+# --- Do slice time correction if indicated
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction"
+    #  use FSL's fslreorient2std for reorienting the image to match the approximate orientation of the standard template images MNI152
+    #  this makes the single-band processing more robust to registration problems
+    fslreorient2std "$fMRIFolder"/"$OrigTCSName".nii.gz "$fMRIFolder"/"$OrigTCSName"_prestcor.nii.gz
```
coalsont: It took me a while to parse "prestcor" to pre-st-cor, perhaps add some underscores or capitals.  Alternatively, use something like `_orig_reorient` to describe what was done, rather than what is expected to be done next.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: My understanding was that once the TR gets lower than 1s, slice timing correction (and the issue of its interaction with motion correction) can be avoided. My suggestion for this PR would be to support the "simple" slice timing options and limit it to legacy, single band data. I would let @mharms and @glasserm chime in on whether we'd want to extend that to support more complex multi-band options, or rather switch to emerging possibilities of doing simultaneout slice-timing and motion-correction.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: In our experience with legacy data, direction (up or down) and interleaved or not covered the situations we ran into. Regarding "odd", the parameter refers directly to FSL `slicetimer`:

```
--odd	use interleaved acquisition
```

I agree that it does not cover all possibilities, but we wanted to match with the tool used. The extended option that the `slicetimer` does offer is to provide a file with explicit slice timing information. This would expand the options significantly, but would require also additional coding. I'd suggest we mark it down as future development.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Noted.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: In this case, I would say that just because FSL's option names are bad, doesn't mean ours should be bad.

As I said, I am not suggesting that we implement support for anything more complicated now, I was explaining why using terms like "odd" when we mean "interleave" is a bad idea for future work.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: That was not my question - do we have reason to believe that adding this type of slice timing correction to processing of good data is actually harmful?  If not, then we may not need to categorize it as only allowed for legacy data.  Granted, the slice timing options don't currently support multiband data, and I don't know whether single-band data can meet our guidelines, so maybe it is moot.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Agreed. I changed the parameter name to `--interleavedslices` with options TRUE/FALSE, TRUE being the default.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: ... and done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: Slice timing correction as presently implemented is a legacy technique.

sivcek:
Added the warning.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: The default with the previous option name appeared to be non-interleaved, while now the default is interleaved.  I would generally expect "not specified" to mean "false" for this type of option naming, so this is looks a little confusing to me.  Do you have a reason for interleaving to be the default?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -247,11 +255,10 @@ ComplianceMsg=""
 # -- Slice timing correction
 
 if [ "${DoSliceTimeCorrection}" = 'TRUE' ]; then
-  ComplianceMsg+=" --doslicetime=TRUE"
+  ComplianceMsg+=" --doslicetime=TRUE WARNING: This legacy option of slice timing correction is performed before motion correction (as is typically done in legacy-style brain imaging) and thus assumes that the brain is motionless. Errors in temporal interpolation will occur in the presence of head motion and may also disrupt data quality measures as shown in Power et al 2017 PLOS One 'Temporal interpolation alters motion in fMRI scans: Magnitudes and consequences for artifact detection.' Slice timing correction and motion correction would ideally be performed simultaneously; however, this is not currently supported by any major software tool. HCP-Style fast TR fMRI data acquisitions (TR<=1s) avoid the need for slice timing correction, provide major advantages for fMRI denoising, and are recommended."
```
coalsont: Ditto, the warning should be a separate log_Warn call, not added to the string.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Overall interleaved acquisition is far more common if not almost exclusive, that's the reason I think it better should be default.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok. That makes the code much more complicated though.

sivcek: Implemented here and in the shlib.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: How does this tool account/allow for the fact that Siemens interleaved acquisitions in the context of an even number of slices start with the _second_ slice (counting from 1)?  That's not your problem per se, but we do want to try to help users avoid doing something completely wrong.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -295,6 +342,21 @@ if [ ! -e "$fMRIFolder" ] ; then
 fi
 cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
 
+# --- Do slice time correction if indicated
```
mharms: Can we change two lines prior to:
`${FSLDIR}/bin/imcp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -295,6 +342,21 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction"
+    #  use FSL's fslreorient2std for reorienting the image to match the approximate orientation of the standard template images MNI152
+    #  this makes the single-band processing more robust to registration problems
```
mharms: This won't work.  You can't reorient the slice acquisition axis without also accounting for that as part of STC.  I see that `slicetimer` has an option (`-d, --direction`) to set the direction of slice acquisition, but if you want to use that, you will need to add code that ascertains what is the axes of the slice acquisition after `fslreorient2std` has been applied.  (If you have previously processed any non-axially acquired data using this approach, then the STC applied to that data is completely wrong).

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -295,6 +342,21 @@ if [ ! -e "$fMRIFolder" ] ; then
 fi
 cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
 
+# --- Do slice time correction if indicated
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction"
+    #  use FSL's fslreorient2std for reorienting the image to match the approximate orientation of the standard template images MNI152
+    #  this makes the single-band processing more robust to registration problems    
+    fslreorient2std "$fMRIFolder"/"$OrigTCSName".nii.gz "$fMRIFolder"/"$OrigTCSName"_orig_reorient.nii.gz
```
mharms: Looks like this particular script is one that uses the `${FSLDIR}/bin` convention for calling FSL tools, so we should be consistent with that.  Also, don't need the .nii.gz extensions.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -295,6 +342,21 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+    rm "$fMRIFolder"/"$OrigTCSName".nii.gz
```
mharms: `${FSLDIR}/bin/imrm "$fMRIFolder"/"$OrigTCSName"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: My suggestion at this point would be to add a warning and then in the next iteration offer possibility of using a slice timing file.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: OK.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Sure, if the number of slices is even, and interleaved is selected, we should probably have a `log_Warn` message about differences between vendors as to how the data is acquired.  But to craft the warning in an informative way, we need to know what `slicetimer` assumes as its default.  (After all, since Oxford is a Siemens shop, they may have written `slicetimer` to work correctly with the goofy Siemens convention).  Do you know what `slicetimer` is assuming in that regard?  We should also add some in-script comments on this particular issue.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Is it ok if we do reorient after slice timing correction?

mharms: As regards STC itself, yes, but not in the broader context of `fMRIVolume` overall, where:
1) changing the orientation changes the UnwarpDir.
2) I'm not immediately sure of the implications on the SEFMs either.

This is a complicated issue if you want the reorientation to be part of the pipeline code itself.

sivcek: Hmm. What if we currently limit it to only those data that do not have any DC method available?

sivcek: As this is quite common intersection in legacy datasets with SB bolds with no DC, so it would cover most cases where fslreorient might help.

mharms: I _think_ that would prevent an inappropriate application, but I'd like @coalsont and @glasserm to chime in that I'm not missing some other potential caveat.  Reorienting your data is a tricky issue, because as you can see, numerous things are framed around the orientation of the data.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: @mharms one other option is to ask for all the parameters to be passed to  `slicetimer` in a comma or pipe separted script e.g.: "--down|--odd|-d|2", which we split up and pass to `slicetimer`. The users then have all the `slicetimer` options available and can provide them in a single parameter. They just need to look up the `slicetimer` manual.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: Per the comments on the mppcheck shlib, the log_Warn call should be here, not in the check function.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: @mharms I don't actually like fsl's conventions of not having the extension on the argument, which is the only reason they had to make tools like imrm, so I would suggest easing up on this particular type of correction.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: I think the resampling into standard space uses an isoxform on the MNI T1 atlas image to generate the reference, so the output will still be standard orientation without an explicit reorient.

What is the reason for reorienting closer to the input stage?

coalsont: Oh, right, it's FSL with their ANALYZE-type coordinate system that ignores the nifti orientation and origin info.  So, the initialization for the registration might be at completely the wrong angle.

I wonder how this is handled in the NHP modifications.

mharms: I don't think we do any reorientation for the _fMRI_ data.  We do for the T1w and T2w inputs, in PreFS, and then we comment in `T2wToT1wDistortionCorrectAndReg` that `--unwarpdir` is "post fslreorient2std".  But I think we've always just assumed that the fMRI data would be axially acquired (which is by far and away the most common).

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: I know you don't like that FSL convention, but given that is the predominant style that we use for working with FSL data, I find it silly to have isolated lines using 'rm', 'cp', etc, surrounded by a preponderance of lines using 'imrm', 'imcp', etc.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: If it is axially acquired, then dimension 3 won't become dimension 1 or 2 when reoriented to standard, the worst that could happen is it is flipped.

coalsont: Chad and Takuya said they hand-reoriented (sometimes needing to relabel the axes) the NHP inputs before using them, so the NHP pipelines never need to deal with non-axial data.

glasserm: It is probably okay, though we should think about how this would interact with fieldmap-less distortion correction if it is implemented.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Can I assume the consensus is we limit it to the data w/o available DC correction method?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: In Qu|Nex we mostly use either `|` or `,`. And I'd prefer staying with those. If comma is the intersection then let's go with that.

mharms: Well, to my knowledge, we don't use either `|` or `,`, so there really isn't any "intersection".  I'd wait for @coalsont to chime in on this one before you do anything.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: For now, I think it is either that, or you eliminate the `fslreorient2std` part entirely.  Regardless, people need to be aware of the reorienting, so I'd include a prominent comment at that section of the code, and also issue a `log_Warn` on the issue.

sivcek: Did that. @mharms can you please check the note and warning and suggest changes if necessary.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: @coalsont would comma or pipe be acceptable to you?

coalsont: The pipelines have only used @ for lists with only one delimiter, so there is definitely an established convention, and breaking it would look inconsistent.  There is a trick to splitting on a delimiter with space-safety, it works like this:

https://github.com/Washington-University/HCPpipelines/blob/5d82838467c2e4a1183598b09d2b00fcc7c8ffbf/global/scripts/RSNregression.sh#L104

That said, the use of @ was started by Matt, and I don't have any real attachment to any specific character.

If we wanted things to be simple for users, at the cost of space-safety (for filename options), we could have the arguments separated by spaces instead (so the user would need to use quotes), and then not quote the expansion of the variable in the slicetimer call.  This would need some testing and possible edits to make sure the quoted spaces make it through the existing argument parsing code intact.

coalsont: To be clear: I feel that we should use @ if we use a delimiter, for consistency.  There is no inherent reason for using @ rather than another character, it is merely that it is what we use everywhere else.

sivcek: We never use `@` in  Qu|Nex, but we already do parsing in Qu|Nex to create `@` separated strings, so a parameter to Qu|Nex can conform to Qu|Nex convention and we then convert it to HCP pipelines convention when passed to HCP pipelines. Unfortunately, there are too many intermediate steps for a number of reasons already that we can not use space as a delimiter.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -24,6 +24,8 @@ fi
 
 source $HCPPIPEDIR/global/scripts/log.shlib  # Logging related functions
 source $HCPPIPEDIR/global/scripts/opts.shlib # Command line option functions
+source ${HCPPIPEDIR}/global/scripts/processingmodecheck.shlib  # Check processing mode requirements
```
mharms: Can we put $HCPPIPDIR in brackets for the preceding two lines for consistency.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,53 @@ then
...
+Compliance="HCPStyleData"
+ComplianceMsg=""
+ComplianceWarn=""
+
+# -- Slice timing correction
+
+if [ "${DoSliceTimeCorrection}" = 'TRUE' ]; then
+  ComplianceMsg+=" --doslicetime=TRUE"
+  Compliance="LegacyStyleData"
+  log_Warn "WARNING: This legacy option of slice timing correction is performed before motion correction (as is typically done in legacy-style brain imaging) and thus assumes that the brain is motionless. Errors in temporal interpolation will occur in the presence of head motion and may also disrupt data quality measures as shown in Power et al 2017 PLOS One 'Temporal interpolation alters motion in fMRI scans: Magnitudes and consequences for artifact detection.' Slice timing correction and motion correction would ideally be performed simultaneously; however, this is not currently supported by any major software tool. HCP-Style fast TR fMRI data acquisitions (TR<=1s) avoid the need for slice timing correction, provide major advantages for fMRI denoising, and are recommended."
```
mharms: For the same reasons as articulated previously, referring to something as a "legacy option" is highly ambiguous.  I would instead say: "This LegacyStyleData option of slice timing correction ..."  
(And make some change to the commented section).

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Comparing the diffs, looks l may have put my comment in the wrong place.
The line was previously:
`cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz`
so the replacement would be:
`${FSLDIR}/bin/imcp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
   log_Msg "mkdir ${fMRIFolder}"
   mkdir "$fMRIFolder"
 fi
-cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
+${FSLDIR}/bin/imcp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName"
+
+# --- Do slice time correction if indicated
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction"
```
mharms: `log_Msg "Running slice timing correction using FSL's 'slicetimer' tool"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,51 @@ then
...
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+DoSliceTimeCorrection=`opts_GetOpt1 "--doslicetime" $@`                 # Whether to do slicetime correction (TRUE), FALSE to omit
+SliceTimerCorrectionParameters=`opts_GetOpt1 "--slicetimerparams" $@`   # A '@' separated list of FSL slicetimer options. Please see FSL slicetimer documentation for details.
+                                                                        # Verbose (-v) is already turned on. TR (-r) is specified using --tr parameter.
```
mharms: Perhaps provide a simple example as an illustration:
e.g., 
`--slicetimerparams=--odd@--ocustom=<CustomInterleaveFile>`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,51 @@ then
...
+TR=`opts_GetOpt1 "--tr" $@`                                             # TR of the timeseries
```
mharms: I believe in all our other scripts in which we need to know the TR, we extract it from the NIFTI directly using something like:
```
TR=`${FSLDIR}/bin/fslval ${fMRIName} pixdim4`
```
So, I'd suggest we do the same within the `$DoSliceTimeCorrection` conditional.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: Since this option could contain a filename, and filenames could have whitespace in them, please change the code to:
```
SliceTimerCorrectionParameters=$(opts_GetOpt1 "--slicetimerparams" "$@")
```
Ideally, all instances of `$@` should be quoted as `"$@"`, and `$()` is the more modern style of command substitution that allows better parsing and more obvious reading of the line.  However, you can leave old lines of code as-is, we aren't expecting you to clean up our messes.

coalsont: Also, if there is a default for this option, it may need special care, as the old option parsing code used a bad hack to set defaults, which prevent proper quoting.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
...
   mkdir "$fMRIFolder"
 fi
-cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
+${FSLDIR}/bin/imcp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName"
+
+# --- Do slice time correction if indicated
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction"
+    SliceTimerCorrectionParameters=`echo ${SliceTimerCorrectionParameters} | sed 's/@/ /g'` 
+    ${FSLDIR}/bin/slicetimer -i "$fMRIFolder"/"$OrigTCSName" -o "$fMRIFolder"/"$OrigTCSName"_orig_stc -r ${TR} -v ${SliceTimerCorrectionParameters}
```
mharms: OrigTCSName is already defined as `OrigTCSName="${NameOffMRI}_orig"`
Given that, I see no need to name the output of `slicetimer` as `"$OrigTCSName"_orig_stc`.  The `_stc` suffix alone suffices.

coalsont: Please use the code pattern I pointed to for this, and try to avoid overwriting a variable's value, as it makes debugging harder:

```
IFS='@' read -a SliceTimerCorrectionParametersArray <<< "$SliceTimerCorrectionParameters"
${FSLDIR}/bin/slicetimer -i "$fMRIFolder"/"$OrigTCSName" -o "$fMRIFolder"/"$OrigTCSName"_orig_stc -r ${TR} -v "${SliceTimerCorrectionParametersArray[@]}"
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+    
+    if [ "${DistortionCorrection}" = "NONE"] ; then {
```
mharms: Errant bracket at end of this line.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+      #  use FSL's fslreorient2std for reorienting the image to match the approximate orientation of the standard template images MNI152
+      #  this makes the single-band processing more robust to registration problems.
+      #  NOTE: Take into account that BOLD images might have a different spatial orientation!
```
mharms: I'd like for the comment here to capture more of our discussion as to why this is only allowed if DistortionCorrection=NONE.  e.g., I propose the following as a replacment:
```
# Processing is more robust to registration problems if the fMRI is in the same orientation as the
# standard template (MNI152) images, which can be accomplished using FSL's `fslreorient2std`.
# HOWEVER, if you reorient, other parameters (such as UnwarpDir) need to be adjusted accordingly.
# Rather than deal with those complications here, we limit reorienting to DistortionCorrection=NONE condition.
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+
+      log_Warn "Performing fslreorient2std! Please take that into account when using volume BOLD images in further analyses!"
+      ${FSLDIR}/bin/fslreorient2std "$fMRIFolder"/"$OrigTCSName"_orig_stc "$fMRIFolder"/"$OrigTCSName"
```
mharms: Probably >95% of fMRI is axially acquired, and thus will already be in LAS or RAS orientation.  Thus, doing this indiscriminately is not necessary.  Here is some code you can use to assess the orientation and proceed accordingly:

```
xorient=`$FSLDIR/bin/fslval "$fMRIFolder"/"$OrigTCSName" qform_xorient | tr -d ' '`
yorient=`$FSLDIR/bin/fslval "$fMRIFolder"/"$OrigTCSName" qform_yorient | tr -d ' '`
zorient=`$FSLDIR/bin/fslval "$fMRIFolder"/"$OrigTCSName" qform_zorient | tr -d ' '`

log_Msg "$fMRIFolder/$OrigTCSName: xorient=${xorient}, yorient=${yorient}, zorient=${zorient}"

if [[ "$xorient" != "Right-to-Left" && "$xorient" != "Left-to-Right" || \
      "$yorient" != "Posterior-to-Anterior" || \
      "$zorient" != "Inferior-to-Superior" ]] ; then
reorient=TRUE
else 
reorient=FALSE
fi

```

mharms: Even better flow:
Start with my proposed comments
Use code to set the reorient value (TRUE or FALSE)
Then have a single conditional: `if [[ $reorient = "TRUE" && $DistortionCorrection = "NONE" ]] ; then`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+    fi
```
mharms: You don't have any code that copies the _orig_stc file to "$fMRIFolder"/"$OrigTCSName" if reorienting is skipped.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,13 +340,29 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+fi
 
 #Create fake "Scout" if it doesn't exist
 if [ $fMRIScout = "NONE" ] ; then
   ${RUN} ${FSLDIR}/bin/fslroi "$fMRIFolder"/"$OrigTCSName" "$fMRIFolder"/"$OrigScoutName" 0 1
 else
-  cp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName".nii.gz
+  ${FSLDIR}/bin/imcp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName".nii.gz
```
mharms: Drop the .nii.gz extension.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: That is already a part of the debug PR, but sure. I changed it.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: mhm

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Thanks, I changed it. No defaults for this one. Onus on the user to set it correctly.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done.

sivcek: Changed it. I'm just weary whether it works on all systems and BASH versions.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ops. Fixed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok. I hope I got it right.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: added

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Dropped.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: We already use it elsewhere:

https://github.com/Washington-University/HCPpipelines/blob/master/ICAFIX/ReApplyFixMultiRunPipeline.sh#L681

mac has the oldest bash version I know of for any non-ancient system, and it works there (though it is missing associative arrays and some nice things like `[[ -v <var ]]`).

coalsont: Better example: https://github.com/Washington-University/HCPpipelines/blob/master/DeDriftAndResample/DeDriftAndResamplePipeline.sh#L469

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Fixed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,51 @@ then
...
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+DoSliceTimeCorrection=`opts_GetOpt1 "--doslicetime" $@`                  # Whether to do slicetime correction (TRUE), FALSE to omit
+SliceTimerCorrectionParameters=$(opts_GetOpt1 "--slicetimerparams" "$@") # A '@' separated list of FSL slicetimer options. Please see FSL slicetimer documentation for details.
+                                                                         # Verbose (-v) is already turned on. TR (-r) is specified using --tr parameter.
```
mharms: `--tr` flag was removed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,15 +340,61 @@ if [ ! -e "$fMRIFolder" ] ; then
   log_Msg "mkdir ${fMRIFolder}"
   mkdir "$fMRIFolder"
 fi
-cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
+${FSLDIR}/bin/imcp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName"
+
+# --- Do slice time correction if indicated
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction using FSL's 'slicetimer' tool"
+    TR=`${FSLDIR}/bin/fslval ${fMRIName} pixdim4`
```
mharms: When I suggested using `fsval` the command I provided was just an example.  In this script `${fMRIName}` is not even a variable.
Use `"$fMRIFolder"/"$OrigTCSName"` instead.
Also, we should probably have a `log_Msg` reporting the TR.
`log_Msg "TR: ${TR}"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,15 +340,61 @@ if [ ! -e "$fMRIFolder" ] ; then
...
-  cp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName".nii.gz
+  ${FSLDIR}/bin/imcp "$fMRIScout" "$fMRIFolder"/"$OrigScoutName"
 fi
 
+if [ $DistortionCorrection = "NONE" ] ; then
+    # Processing is more robust to registration problems if the fMRI is in the same orientation as the
+    # standard template (MNI152) images, which can be accomplished using FSL's `fslreorient2std`.
+    # HOWEVER, if you reorient, other parameters (such as UnwarpDir) need to be adjusted accordingly.
+    # Rather than deal with those complications here, we limit reorienting to DistortionCorrection=NONE condition.
+    xorient=`$FSLDIR/bin/fslval "$fMRIFolder"/"$OrigTCSName" qform_xorient | tr -d ' '`
```
mharms: Add newline, then following comment
`#First though, detect if reorienting is even necessary`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Changed to

```
# Verbose (-v) is already turned on. TR is read from the file itself
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Added comment.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -229,7 +229,7 @@ fi
 
 DoSliceTimeCorrection=`opts_GetOpt1 "--doslicetime" $@`                  # Whether to do slicetime correction (TRUE), FALSE to omit
 SliceTimerCorrectionParameters=$(opts_GetOpt1 "--slicetimerparams" "$@") # A '@' separated list of FSL slicetimer options. Please see FSL slicetimer documentation for details.
-                                                                         # Verbose (-v) is already turned on. TR (-r) is specified using --tr parameter.
+                                                                         # Verbose (-v) is already turned on. TR is read from the file itself.
```
mharms: "file" is too generic.  Can we say: "TR is read from 'pixdim4' of the input NIFTI itself".

sivcek: Updated here and in the joint PR.

sivcek:

PR #140, opened 2019-10-03
fMRIVolume Related Pull Request — Use a specific BOLD for reference

sivcek:
Added ability to specify another BOLD image as reference. In that case the reference BOLD image is the target for motion correction and the distortion correction computed for the reference BOLD image is reused instead of recomputed. 

This option has shown to provide better between-bold registration for lower resolution single-band images. It also speeds up computation time as distortion correction need not be recomputed for each bold, but is computed just for the reference bold, usually the first BOLD in a series of acquisitions, and reused. It does require the reference BOLD processing to be finished beforehand.

### DIFF
--- a/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
+++ b/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
@@ -220,6 +220,23 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Compliance check
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+
+check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
+
+
+# -- End compliance check
+
+
+
 # Setup PATHS
 PipelineScripts=${HCPPIPEDIR_fMRIVol}
 GlobalScripts=${HCPPIPEDIR_Global}
--- a/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
+++ b/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
@@ -225,6 +225,11 @@ fi
 #  Legacy Style Data Options
 # ------------------------------------------------------------------------------
 
+fMRIReference=`opts_GetOpt1 "--fmriref" $@`                          # reference BOLD run to use for movement correction target and to copy atlas registration from (or NONE; default)
+
+# Defaults
+fMRIReference=${fMRIReference:-NONE}
+
 
 # ------------------------------------------------------------------------------
 #  Compliance check
@@ -235,7 +240,12 @@ MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
 Compliance="HCPStyleData"
 ComplianceMsg=""
 
+# -- Use of BOLD reference
 
+if [ ! "${DoSliceTimeCorrection}" = 'NONE' ]; then
+  ComplianceMsg+=" --t2=NONE"
+  Compliance="LegacyStyleData"
+fi
 
 check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
 
@@ -310,6 +320,10 @@ T1wFolder="$Path"/"$Subject"/"$T1wFolder"
 AtlasSpaceFolder="$Path"/"$Subject"/"$AtlasSpaceFolder"
 ResultsFolder="$AtlasSpaceFolder"/"$ResultsFolder"/"$NameOffMRI"
 
+if [ ! $fMRIReference = "NONE" ] ; then
+    fMRIReference="$Path"/"$Subject"/"$fMRIReference"
+fi
+
 mkdir -p ${T1wFolder}/Results/${NameOffMRI}
 
 if [ ! -e "$fMRIFolder" ] ; then
@@ -365,6 +379,15 @@ fi
 
 log_Msg "mkdir -p ${fMRIFolder}/MotionCorrection"
 mkdir -p "$fMRIFolder"/MotionCorrection
+
+if [ ! $fMRIReference = "NONE" ] ; then
+    log_Msg "Using reference image from ${fMRIReference}"
+    reference="$fMRIReference"/"$ScoutName"_gdc
+else
+    reference="$fMRIFolder"/"$ScoutName"_gdc
+    ReferenceReg="NONE"  # only do nonlinear or linear registration for target that is not of the same bold run
+fi
+
 ${RUN} "$PipelineScripts"/MotionCorrection.sh \
        "$fMRIFolder"/MotionCorrection \
        "$fMRIFolder"/"$NameOffMRI"_gdc \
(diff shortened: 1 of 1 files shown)


fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,39 @@ then
...
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+
+# -- Use of BOLD reference
+
+if [ ! "${DoSliceTimeCorrection}" = 'NONE' ]; then
```
coalsont: This check appears to reference the wrong variable (one from a different PR entirely).  The message also references the t2w option.

sivcek: Indeed an error. Corrected now.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,39 @@ then
...
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+
+# -- Use of BOLD reference
+
+if [ ! "${fMRIReference}" = 'NONE' ]; then
```
mharms: Nitpicking, but can we use double quotes around NONE, for consistency with the rest of the code?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,39 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+fMRIReference=`opts_GetOpt1 "--fmriref" $@`                          # reference BOLD name run to use for movement correction target and to copy atlas registration from (or NONE; default)
```
mharms: Unclear to me, is `--fmriref` allowed to be a whole run?  Is it expected to be just a single frame?  If the former, should there be an 'fslroi' operation at some stage to select just a single frame?  (As we've learned, between FSL 5.x and 6.x, the default behavior of some FSL tools changed considerably when using 4D files as inputs, so we had to become more careful about the distinction that we were in the past).

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -353,46 +399,61 @@ ${RUN} "$PipelineScripts"/MotionCorrection.sh \
...
+         --subjectfolder=${SubjectFolder} \
+         --biascorrection=${BiasCorrection} \
+         --usejacobian=${UseJacobian}
+else
+    log_Msg "linking EPI distorsion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${fMRIFolder}/${DCFolderName} ] ; then
+        log_Msg "     ... removing stale link (or preexisiting files)"
+        rm -r ${fMRIFolder}/${DCFolderName}
+    fi
+    ln -s -f ${fMRIReference}/${DCFolderName} ${fMRIFolder}/${DCFolderName}
```
mharms: Is -f adding anything important here, given that you just removed `${fMRIFolder}/${DCFolderName}`?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -353,46 +399,61 @@ ${RUN} "$PipelineScripts"/MotionCorrection.sh \
...
+         --method=${DistortionCorrection} \
+         --topupconfig=${TopupConfig} \
+         --ojacobian=${fMRIFolder}/${JacobianOut} \
+         --dof=${dof} \
+         --fmriname=${NameOffMRI} \
+         --subjectfolder=${SubjectFolder} \
+         --biascorrection=${BiasCorrection} \
+         --usejacobian=${UseJacobian}
+else
+    log_Msg "linking EPI distorsion correction and T1 registration from ${fMRIReference}"
```
mharms: Typo: distorsion

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,6 +379,15 @@ fi
 
 log_Msg "mkdir -p ${fMRIFolder}/MotionCorrection"
 mkdir -p "$fMRIFolder"/MotionCorrection
+
+if [ ! $fMRIReference = "NONE" ] ; then
+    log_Msg "Using reference image from ${fMRIReference}"
+    reference="$fMRIReference"/"$ScoutName"_gdc
+else
+    reference="$fMRIFolder"/"$ScoutName"_gdc
+    ReferenceReg="NONE"  # only do nonlinear or linear registration for target that is not of the same bold run
```
mharms: Where do these `reference` and `ReferenceReg` variables come into play?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -287,6 +320,10 @@ T1wFolder="$Path"/"$Subject"/"$T1wFolder"
 AtlasSpaceFolder="$Path"/"$Subject"/"$AtlasSpaceFolder"
 ResultsFolder="$AtlasSpaceFolder"/"$ResultsFolder"/"$NameOffMRI"
 
+if [ ! $fMRIReference = "NONE" ] ; then
```
mharms: When this condition is true, shouldn't it trigger other downstream changes in terms of the "Scout" treatment?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -353,46 +399,61 @@ ${RUN} "$PipelineScripts"/MotionCorrection.sh \
...
+        log_Msg "     ... removing stale link (or preexisiting files)"
+        rm -r ${fMRIFolder}/${DCFolderName}
+    fi
+    ln -s -f ${fMRIReference}/${DCFolderName} ${fMRIFolder}/${DCFolderName}
+ 
+    WD=${fMRIReference}/${DCFolderName}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/${RegOutput}
+    ${FSLDIR}/bin/imcp ${WD}/fMRI2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
+    ${FSLDIR}/bin/imcp ${WD}/Jacobian2T1w ${fMRIFolder}/${JacobianOut}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${QAImage} ${fMRIFolder}/${QAImage}
```
mharms: At what point did the fMRI run being processed actually get registered to the `${fMRIReference}`?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok. Done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: `--fmriref` is actually the name of the BOLD that is to be used as the reference. The actual reference is then `$fMRIReference"/"$ScoutName"_gdc`. Which is either the scout for that BOLD or the "generated" scout, which is a mean across the BOLD timeseries after motion correction.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Indeed, not needed. Removed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Corrected.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: The idea is to use the actual BOLD scout for other functionality and use the reference scout exclusively as a registration target.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Aaaaa. This is the line I missed when splitting the changes into separate PRs. It is in the MotionCorrection.sh which is now:

```
${RUN} "$PipelineScripts"/MotionCorrection.sh \
       "$fMRIFolder"/MotionCorrection \
       "$fMRIFolder"/"$NameOffMRI"_gdc \
       "$reference" \
       "$fMRIFolder"/"$NameOffMRI"_mc \
       "$fMRIFolder"/"$MovementRegressor" \
       "$fMRIFolder"/"$MotionMatrixFolder" \
       "$MotionMatrixPrefix" \
       "$MotionCorrectionType"
```

Thank you for the sharp eye!!!

sivcek:
@mharms please see the latest version with the missing piece.

sivcek:
I created a combined refBold and nonlin PR to make things clearer.

mharms:
Great. I was going to suggest exactly that.

mharms:
@glasserm This PR is deprecated and replaced by #147.

PR #138, opened 2019-10-01
Add assessment of spatial coverage within both fMRIVolume and fMRISurface

mharms:
This PR is intended to address #132.  This functionality is a necessary addition before we start HCP-D/A processing.

Some minor readability improvements and comment additions are included for the affected scripts as well.

PR #141, opened 2019-10-03
fMRIVolume Related Pull Request — Use of a specific BOLD mask

sivcek:
Adds the (optional) ability to specify the mask to use in the final Intensity Normalisation step. The options are:

T1 - the default option
BOLD - a mask created by bet on the mean bold image across the timeseries
DILATED - use the MNI152_T1_2mm_brain_mask_dil.nii.gz mask
NONE - do not use a mask

In our experience using a wider mask than T1 (DILATED or NONE) helped in checking the quality of bold registration, as it was possible to see if the BOLD image was expanded beyond the borders defined by the T1w mask.


fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,36 @@ then
...
+# ------------------------------------------------------------------------------
+#  Compliance check
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+
+if [ ! "${Mask}" = 'T1' ]; then
```
coalsont: I think the purpose of the mask is more to reduce file sizes by setting non-brain to exact 0s, so I'm not sure this needs to be restricted to legacy data mode.

coalsont: Oh right, for intensity normalization...depending on how we do that, it may be important to use a tight mask.

sivcek: I'm not sure if there is a way to compute the intensity normalization using a tighter mask and then apply it on a wider one.

sivcek:
@glasserm "expand" is not the right term. With lack of DC, expanded mask allows to see how badly aligned the image is.

glasserm:
Let's implement as I suggested above

mharms:
The purpose as stated in the PR was to "help in checking the quality of bold registration".  If that's the goal, I think there are better, more explicit ways we can check the registration than to expand the mask (which seems like a crude solution) -- e.g., we can make sure that we make available a non-masked version of the registered Scout, so that you can easily visualize what is occurring "beyond the borders defined by the T1w mask".  Also, @sivcek, this is going to interact with PR #138 which I'm hoping to merge into the master tomorrow.

glasserm:
Well I also agree with this.

sivcek:
@glasserm @mharms The value goes beyond just QC checking. If somebody want's to stay for some reason in volume-land, they might want to make sure that the BOLD was not cut and would still proceed with suboptimal registration, that would be prohibitive for surface analysis, but might be usable for volume-based analyses. If it's ok with you. I'll wait for @mharms to pull in his PR and I'll see how to re-introduce the functionality to satisfy previous comments.

mharms:
The PR was merged last night, so check out the current master branch, and proceed from there.

sivcek:
Ok.

glasserm:
If we are keeping it, we have the default tight brain mask, a dilated version of this mask, and no mask as reasonable options.

sivcek:
I have now added the following options:

* FINAL (the default)
* T1 (for T1w image based mask)
* DILATEDT1 (for dilated T1w based mask)
* NONE (for no mask)

I've also implemented checks for valid options.

sivcek:
FINAL is the mask created by code that @mharms added.

glasserm:
So final is a combination of the T1w mask and the FOV mask, right?

glasserm:
Okay, lets call it T1_fMRI_FOV

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,39 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+UseMask=`opts_GetOpt1 "--usemask" $@`                                   # what mask to use for the final bold (T1_fMRI_FOV combination of T1w and FOV masks: the default, T1: T1w image based mask, DILATEDT1: dilated T1w image based mask, NONE: none)
```
mharms: `--usemask` seems more in style with one of our TRUE/FALSE/YES/NO arguments.  Would `--maskmode` or `--maskstyle` be a better choice?

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -132,11 +155,28 @@ PctCoverage=`echo "scale=4; 100 * ${NvoxFinalMask} / ${NvoxBrainMask}" | bc -l`
 echo "PctCoverage, NvoxFinalMask, NvoxBrainMask" >| ${FinalMask}.stats.txt
 echo "${PctCoverage}, ${NvoxFinalMask}, ${NvoxBrainMask}" >> ${FinalMask}.stats.txt
 
+# If a user requested, use a T1 mas, a dilated T1 mask or no mask instedad when generating final BOLD timeseries
+
+if [ "${UseMask}" = "DILATEDT1" ] ; then
+    # -- create a dilated version of T1 mask
+    ${FSLDIR}/bin/fslmaths ${BrainMask} -dilF ${BrainMask}_dil
```
mharms: Is a single round of dilation enough to accomplish your purposes with the `DILATEDT1` mask?  Seems like you'd want more than that...

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I'll change it to `--boldmask` as that's what it is mask for BOLD.

fMRIVolume/scripts/IntensityNormalization.sh
sivcek: The options now will be 

* `T1_fMRI_FOV`: combined T1w brain mask and FOV masks (the default), 
* `T1_DILATED_fMRI_FOV`: a once dilated T1w brain based mask combined with fMRI FOV
* `T1_DILATED2_fMRI_FOV`: a twice dilated T1w brain based mask combined with fMRI FOV, 
* `fMRI_FOV`: a fMRI FOV mask

sivcek: I have also changed the flow of the code so that the `FinalMask` is the mask specified. And that the statistics are reported both on the FOV of the T1w brain as well as the actual mask used, with the mask used also listed in the ${FinalMask}.stats.txt

sivcek:
@mharms @glasserm I hope that your requests are now satisfied.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,43 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+BOLDMask=`opts_GetOpt1 "--boldmask" $@`                                   # Specifies what mask to use for the final bold:
+                                                                          #   T1_fMRI_FOV: combined T1w brain mask and FOV masks (the default),
```
mharms: Quibble, but can we say "fMRI FOV" in each case, for consistency.
e.g.,
`# #   T1_fMRI_FOV: combined T1w brain mask and fMRI FOV (the default),`

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -35,18 +36,23 @@ Usage() {
   echo "             [--inscout=<input name for scout image (pre-sat EPI)>]"
   echo "             [--oscout=<output name for normalized scout image>]"
   echo "             [--workingdir=<working dir>]"
+  echo "             [--boldmask=<what mask to use for generating final BOLD timeseries:"
+  echo "                         T1_fMRI_FOV (mask based on T1 and voxels available at all timepoints i.e. FOV - the default)"
```
mharms: `T1_fMRI_FOV (default): mask based on T1 and voxels available at all timepoints (i.e., the fMRI FOV)`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Sure. I like your quibbles.

fMRIVolume/scripts/IntensityNormalization.sh
sivcek: Done.

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -117,20 +140,49 @@ echo "PWD = `pwd`" >> $WD/log.txt
...
-${FSLDIR}/bin/fslmaths ${BrainMask} -bin -mas ${InputfMRI}_mask ${FinalMask}
+T1FOVMask=`${FSLDIR}/bin/remove_ext ${InputfMRI}`_T1FOVmask
+${FSLDIR}/bin/fslmaths ${BrainMask} -bin -mas ${InputfMRI}_mask ${T1FOVMask}
+
+# Use the requested BOLD mask 
+
+if [ "${BOLDMask}" = "T1_fMRI_FOV" ] ; then
+    # FinalMask is a combination of the FS-derived brainmask, and the spatial coverage mask that captures the
+    # voxels that have data available at *ALL* time points (derived in OneStepResampling)
+    ${FSLDIR}/bin/fslmaths ${BrainMask} -bin -mas ${InputfMRI}_mask ${FinalMask}
```
mharms: Since this operation has already now been done above, perhaps the following here instead:
`${FSLDIR}/bin/imcp ${T1FOVMask} ${FinalMask}`

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -117,20 +140,49 @@ echo "PWD = `pwd`" >> $WD/log.txt
...
+
+elif [ "${BOLDMask}" = "T1_DILATED2x_fMRI_FOV" ] ; then
+    # FinalMask is a combination of twice dilated FS-derived brainmask, and the spatial coverage mask that captures the
+    # voxels that have data available at *ALL* time points (derived in OneStepResampling)
+    ${FSLDIR}/bin/fslmaths ${BrainMask} -bin -dilF -dilF -mas ${InputfMRI}_mask ${FinalMask}
+
+elif [ "${BOLDMask}" = "fMRI_FOV" ] ; then
+    # FinalMask is the spatial coverage mask that captures the
+    # voxels that have data available at *ALL* time points (derived in OneStepResampling)
+    {FSLDIR}/bin/imcp ${InputfMRI}_mask ${FinalMask}
```
mharms: Need `${FSLDIR}`

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -117,20 +140,49 @@ echo "PWD = `pwd`" >> $WD/log.txt
...
+
+elif
+    # No valid BOLDMask option was specified
+    log_Err_Abort "Specified BOLD mask to use (--boldmask=${BOLDMask}) is invalid!"
```
mharms: More direct statement:
`log_Err_Abort "Invalid entry for specified BOLD mask (--boldmask=${BOLDMask})"`

fMRIVolume/scripts/IntensityNormalization.sh
sivcek: Was contemplating the same, but thought that if we keep it separate, then it is clearer and less prone to bugs if anything changes. I've changed it to `imcp` now.

fMRIVolume/scripts/IntensityNormalization.sh
sivcek: Ops. A typo.

fMRIVolume/scripts/IntensityNormalization.sh
sivcek: changed.

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -117,20 +140,49 @@ echo "PWD = `pwd`" >> $WD/log.txt
...
+
+# Create a simple summary text file of the percentage of spatial coverage of the fMRI data inside the FS-derived brain mask and the actual mask used
 NvoxBrainMask=`fslstats ${BrainMask} -V | awk '{print $1}'`
+NvoxT1FOVMask=`fslstats ${T1FOVMask} -V | awk '{print $1}'`
 NvoxFinalMask=`fslstats ${FinalMask} -V | awk '{print $1}'`
-PctCoverage=`echo "scale=4; 100 * ${NvoxFinalMask} / ${NvoxBrainMask}" | bc -l`
-echo "PctCoverage, NvoxFinalMask, NvoxBrainMask" >| ${FinalMask}.stats.txt
-echo "${PctCoverage}, ${NvoxFinalMask}, ${NvoxBrainMask}" >> ${FinalMask}.stats.txt
+PctBrainCoverage=`echo "scale=4; 100 * ${NvoxT1FOVMask} / ${NvoxBrainMask}" | bc -l`
+PctMaskCoverage=`echo "scale=4; 100 * ${NvoxFinalMask} / ${NvoxT1FOVMask}" | bc -l`
```
mharms: It isn't clear to me what you are trying to convey via `PctMaskCoverage`.  Since `T1FOVMask` is the most "restrictive" mask, `FinalMask` will either be equal to `T1FOVMask` or a superset of voxels that include `T1FOVMask` -- i.e., `PctMaskCoverage` >= 100%.   Is that the intent?

sivcek: Yes, indeed. It will show how larger the mask is from the T1w based one.

fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -117,20 +140,49 @@ echo "PWD = `pwd`" >> $WD/log.txt
...
+echo "Mask, PctBrainCoverage, PctMaskCoverage, NvoxT1FOVMask, NvoxBrainMask, NvoxFinalMask" >| ${FinalMask}.stats.txt
```
mharms: Maybe just call the first column BOLDMask to tie back to the argument name.

sivcek: Done!

sivcek:
@mharms, @coalsont I'll combine this PR with the rest fMRIVolume ones once also the reference BOLD PR is done.

sivcek:

PR #142, opened 2019-10-03
fMRIVolume Related Pull Request — Option to specify what to use for initial BOLD registration

sivcek:
This PR provides an optional `--preregister` parameter that provides the option to use `flirt` rather than `epi_reg` for the initial registration of undistorted BOLD scout image to T1w image.

In our experience this has worked better for some legacy data.


fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,39 @@ then
...
+# ------------------------------------------------------------------------------
+
+MPPMode=`opts_GetOpt1 "--mpp-mode" $@`
+MPPMode=`opts_DefaultOpt $MPPMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+
+# -- Preregister code
+
+if [ ! "${Preregister}" = 'epi_reg' ]; then
```
coalsont: As long as both registration methods are doing the same type of registration, having an additional option that sometimes works better doesn't sound like it needs to be restricted to only legacy data mode.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -389,7 +392,14 @@ case $DistortionCorrection in
...
         # this is just an initial registration, refined later in this script, but it is actually pretty good
         log_Msg "register undistorted scout image to T1w"
-        ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+
+        if [ $Preregister = "epi_reg" ] ; then
+          log_Msg "... running epi_reg (dof ${dof})"
+          ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+        elif [ $Preregister = "flirt" ] ; then
+          log_Msg "... running flirt"
+          ${FSLDIR}/bin/flirt -in ${WD}/${ScoutInputFile}_undistorted -ref ${WD}/${T1wBrainImageFile} -out ${WD}/${ScoutInputFile}_undistorted2T1w_init -omat ${WD}/${ScoutInputFile}_undistorted2T1w_init.mat -dof 6
```
coalsont: Why does this hardcode `-dof 6` instead of using `-dof ${dof}`?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I'm ok to have it open. Again @mharms and @glasserm need to chime in so that I don't go back and forth changing the code.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Good point! I set it to use the parameter. It might have been a remnant of past code where the parameter might have been fixed.

sivcek:
Preregister code check taken out.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -79,6 +79,7 @@ Usage() {
   echo "                 don't do bias correction"
   echo ""
   echo "             --usejacobian=<\"true\" or \"false\">"
+  echo "             --preregister=<epi_reg (default) or flirt>"
```
mharms: I mildly favor changing this to `--preregistertool` or `--preregistercmd`.  (`--preregister` by itself implies more of a TRUE/FALSE/YES/NO flag to me).
Also, while we are at it, please change the earlier 
`echo "             --method=<method used for readout distortion correction>"`
line to
`echo "             --method=<method used for susceptibility distortion correction>"`

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -207,6 +209,7 @@ WD=`defaultopt $WD ${RegOutput}.wdir`
 dof=`defaultopt $dof 6`
 GlobalScripts=${HCPPIPEDIR_Global}
 TopupConfig=`defaultopt $TopupConfig ${HCPPIPEDIR_Config}/b02b0.cnf`
+Preregister=${Preregister:-epi_reg}
```
mharms: Let's set the defaults in a consistent fashion, which here seems to be using `defaultopt`

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -389,7 +392,14 @@ case $DistortionCorrection in
...
         log_Msg "register undistorted scout image to T1w"
-        ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+
+        if [ $Preregister = "epi_reg" ] ; then
+          log_Msg "... running epi_reg (dof ${dof})"
+          ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+        elif [ $Preregister = "flirt" ] ; then
+          log_Msg "... running flirt"
+          ${FSLDIR}/bin/flirt -in ${WD}/${ScoutInputFile}_undistorted -ref ${WD}/${T1wBrainImageFile} -out ${WD}/${ScoutInputFile}_undistorted2T1w_init -omat ${WD}/${ScoutInputFile}_undistorted2T1w_init.mat -dof ${dof}
+        fi
```
mharms: SInce `epi_reg` and `flirt` are the only two supported modes, we should have an `else` clause here with a `log_Err_Abort` if it isn't one of those two values.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: I agree that there doesn't seem a need to restrict this to LegacyStyleData mode.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Done and done.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Ok.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Ok.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok. Check taken out.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -20,7 +20,7 @@ SPIN_ECHO_METHOD_OPT="TOPUP"
 #  Load Function Libraries
 # --------------------------------------------------------------------------------
 
-source $HCPPIPEDIR_Global/log.shlib # Logging related functions
+source ${HCPPIPEDIR_Global}/log.shlib           # Logging related functions
```
mharms: For simplicity and consistency let's source as 
`source $HCPPIPEDIR/global/scripts/log.shlib` (rather than using the obscure `HCPPIPEDIR_Global` variable.
(Although, sounds like this may be getting replaced anyway by a source of the debug.shlib).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -161,6 +162,7 @@ NameOffMRI=`getopt1 "--fmriname" $@`
 SubjectFolder=`getopt1 "--subjectfolder" $@`
 BiasCorrection=`getopt1 "--biascorrection" $@`
 UseJacobian=`getopt1 "--usejacobian" $@`
+PreregisterTool=`getopt1 "--preregistertool" $@`
 
 if [[ -n $HCPPIPEDEBUG ]]
 then
```
mharms: Another hidden "debugging" type feature.  Harmless I guess, but not consistently implemented.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -389,7 +392,16 @@ case $DistortionCorrection in
...
-        ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+
+        if [ $PreregisterTool = "epi_reg" ] ; then
+          log_Msg "... running epi_reg (dof ${dof})"
+          ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+        elif [ $PreregisterTool = "flirt" ] ; then
+          log_Msg "... running flirt"
+          ${FSLDIR}/bin/flirt -in ${WD}/${ScoutInputFile}_undistorted -ref ${WD}/${T1wBrainImageFile} -out ${WD}/${ScoutInputFile}_undistorted2T1w_init -omat ${WD}/${ScoutInputFile}_undistorted2T1w_init.mat -dof ${dof}
+        else
+          log_Err_Abort "No valid preregister tool specified. Please specify --prepregistertool as either 'epi_reg' or 'flirt'."
```
mharms: Typo: `--prepreregistertool`
FWIW, I would simplify and just say:
`log_Err_Abort "--preregistertool=${PreregisterTool} is not a valid setting."`

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Yes. Replaced by debug.shlib.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Another for cleanup. That was there before.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Changed.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -209,7 +209,7 @@ WD=`defaultopt $WD ${RegOutput}.wdir`
 dof=`defaultopt $dof 6`
 GlobalScripts=${HCPPIPEDIR_Global}
 TopupConfig=`defaultopt $TopupConfig ${HCPPIPEDIR_Config}/b02b0.cnf`
-Preregister=${Preregister:-epi_reg}
+PreregisterTool=${PreregisterTool:-epi_reg}
```
coalsont: Maybe this should also use `defaultopt` for consistency?

I'm not really sure which style I prefer of these 2, and the `defaultopt` implementation is not great (can't handle quoted spaces even by adding quotes, because it abuses unquoted bash argument parsing rather than using a conditional).

As a side note, `newopts.shlib` approaches this in the other direction, by setting the default values before checking what options are specified, so specifying the empty string doesn't result in using the default.

sivcek:

PR #144, opened 2019-10-03
fMRIVolume Related Pull Request — No Distortion Correction Available

sivcek:
Allows specifying BOLD DC method as NONE and so running fMRIVolume when no distortion correction method is available.


fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -432,10 +437,56 @@ case $DistortionCorrection in
...
+            log_Msg "---> No distortion correction"
+
+            log_Msg "---> FAKE WarpField"
+            ### faked warp image with three frames
+            ${FSLDIR}/bin/fslmaths ${T1wImage} -mul 0 ${WD}/WarpField.nii.gz
+            ${FSLDIR}/bin/fslmerge -t ${WD}/WarpField.nii.gz ${WD}/WarpField.nii.gz ${WD}/WarpField.nii.gz ${WD}/WarpField.nii.gz
+
+            log_Msg "---> FAKE Jacobian Volume"
+            ###Jacobian Volume FAKED for Regular Fieldmaps (all ones) ###
+            ${FSLDIR}/bin/fslmaths ${T1wImage} -abs -add 1 -bin ${WD}/Jacobian.nii.gz
```
coalsont: I think `-mul 0 -add 1` would be clearer.

sivcek: Agreed. I changed it and also changed cp to imcp a few lines lower.

sivcek:
Added warning.

coalsont:
The wording of the end of that warning is weird, as it seems to imply that we expect fieldmap-less correction to be better than the gold standard.

How about something like this for the last sentence: `At a future time, we may be able to add support for fieldmap-less distortion correction; however, no extant software package or pipelines have successfully demonstrated clear improvement over no distortion correction, in the direction towards gold standard fieldmap-based correction.`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -236,7 +245,7 @@ Compliance="HCPStyleData"
 ComplianceMsg=""
 
 if [ "${DistortionCorrection}" = 'NONE' ]; then
-  ComplianceMsg+=" --dcmethod=NONE"
+  ComplianceMsg+=" --dcmethod=NONE WARNING: The fMRIVolume pipeline is being run without appropriate distortion correction of the fMRI image. This is NOT RECOMMENDED under normal circumstances. We will attempt 6 DOF FreeSurfer BBR registration of the distorted fMRI to the undistorted T1w image. Distorted portions of the fMRI data will not align with the cortical ribbon. In HCP data 30% of the cortical surface will be misaligned by at least half cortical thickness and 10% of the cortical surface will be completely misaligned by a full cortical thickness. At a future time, we may be able to add support for fieldmap-less distortion correction; however, no extant software package or pipelines have successfully demonstrated clear improvement from fieldmap-less distortion correction when compared with a gold standard fieldmap-based correction."
```
coalsont: I believe we want this warning to ALWAYS be displayed when `DistortionCorrection` is `NONE`, so it should be a separate log_Warn right here, not combined with the message string.  The only drawback is that this will put this warning before the abort message.

glasserm:
I am okay with the rewording

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I changed it so that it is displayed separately and only when legacy mode is chosen.

mharms:
What is the justification for the wording that `however, no extant software package or pipelines have successfully demonstrated clear improvement over no distortion correction`.  By my read, the fMRIPrep paper showed that "fieldmap-less SDC" is clearly better than no distortion correction whatsoever.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,44 @@ then
...
+ProcessingMode=`opts_GetOpt1 "--processing-mode" $@`
+ProcessingMode=`opts_DefaultOpt $ProcessingMode "HCPStyleData"`
+Compliance="HCPStyleData"
+ComplianceMsg=""
+ComplianceWarn=""
+
+if [ "${DistortionCorrection}" = 'NONE' ]; then
+  ComplianceMsg+=" --dcmethod=NONE"
+  Compliance="LegacyStyleData"
+  log_Warn "WARNING: The fMRIVolume pipeline is being run without appropriate distortion correction of the fMRI image. This is NOT RECOMMENDED under normal circumstances. We will attempt 6 DOF FreeSurfer BBR registration of the distorted fMRI to the undistorted T1w image. Distorted portions of the fMRI data will not align with the cortical ribbon. In HCP data 30% of the cortical surface will be misaligned by at least half cortical thickness and 10% of the cortical surface will be completely misaligned by a full cortical thickness. At a future time, we may be able to add support for fieldmap-less distortion correction; however, no extant software package or pipelines have successfully demonstrated clear improvement over no distortion correction, in the direction towards gold standard fieldmap-based correction."
```
mharms: For these long warning messages, I wonder if we should format as a bunch of shorter individual `log_Warn` lines, rather one single long line?  The ensuing result would then look something like:
```
<DateTime>: <ToolName>: WARNING: Line1 
<DateTime>: <ToolName>: WARNING: Line2
<DateTime>: <ToolName>: WARNING: Line3
```
as opposed to

`<DateTime>: <ToolName>: WARNING: One really long single line with too much wraparound to resolve by simply widening your viewing window...`

 I favor the former, as it should be both easier to read, and also highlights the "warning" nature by the fact that "WARNING" gets repeated on multiple lines.

Thoughts @coalsont ?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -432,10 +437,56 @@ case $DistortionCorrection in
...
+            ###Jacobian Volume FAKED for Regular Fieldmaps (all ones) ###
+            ${FSLDIR}/bin/fslmaths ${T1wImage} -mul 0 -add 1 -bin ${WD}/Jacobian.nii.gz
+
+            log_Msg "---> FAKE spline interpolated image of scout"
+            # create a spline interpolated image of scout (distortion corrected in same space)
+            ${FSLDIR}/bin/imcp ${ScoutInputName} ${WD}/${ScoutInputFile}_undistorted
+
+            log_Msg "---> REORIENT scout to standard MNI152 template"
+            # use FSL's fslreorient2std for reorienting the image to match the approximate orientation of the standard template images (MNI152)
+            fslreorient2std ${WD}/${ScoutInputFile}_undistorted.nii.gz ${WD}/${ScoutInputFile}_undistorted_fMRI2str.nii.gz
```
mharms: Another `fslreorient2std`.  Is this necessary?  Not sure of the possible unintended downstream consequences of this particular reorienting...

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -432,10 +437,56 @@ case $DistortionCorrection in
...
         ;;
 
+    ${NONE_METHOD_OPT})
+
+            log_Msg "---> No distortion correction"
+
+            log_Msg "---> FAKE WarpField"
+            ### faked warp image with three frames
+            ${FSLDIR}/bin/fslmaths ${T1wImage} -mul 0 ${WD}/WarpField.nii.gz
+            ${FSLDIR}/bin/fslmerge -t ${WD}/WarpField.nii.gz ${WD}/WarpField.nii.gz ${WD}/WarpField.nii.gz ${WD}/WarpField.nii.gz
```
mharms: Why is a fake warpfield even necessary?  If I'm reading correctly, the eventual `convertwarp/applywarp` are within the "No distortion correction" section of the code, so let's just construct the appropriate command without the fake warpfield (which does nothing).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -432,10 +437,56 @@ case $DistortionCorrection in
...
+
+            log_Msg "---> FAKE Jacobian Volume"
+            ###Jacobian Volume FAKED for Regular Fieldmaps (all ones) ###
+            ${FSLDIR}/bin/fslmaths ${T1wImage} -mul 0 -add 1 -bin ${WD}/Jacobian.nii.gz
```
mharms: Don't see why a fake jacobian volume is necessary either.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -432,10 +437,56 @@ case $DistortionCorrection in
...
+
+            log_Msg "---> FAKE spline interpolated image of scout"
+            # create a spline interpolated image of scout (distortion corrected in same space)
+            ${FSLDIR}/bin/imcp ${ScoutInputName} ${WD}/${ScoutInputFile}_undistorted
```
mharms: And I don't see why we need to create a `${ScoutInputName}_undistorted` file either (which is in fact _distorted_).  It feels to me that this code is unnecessarily being driven by an attempt to "mimic" the SEFM correction code (which leads potentially to terrible confusion), rather than simply computing the necessary registration as its own de novo code, and then applying that registration.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -432,10 +437,56 @@ case $DistortionCorrection in
...
+
+            log_Msg "---> REORIENT scout to standard MNI152 template"
+            # use FSL's fslreorient2std for reorienting the image to match the approximate orientation of the standard template images (MNI152)
+            fslreorient2std ${WD}/${ScoutInputFile}_undistorted.nii.gz ${WD}/${ScoutInputFile}_undistorted_fMRI2str.nii.gz
+            rm ${WD}/${ScoutInputFile}_undistorted.nii.gz
+            mv ${WD}/${ScoutInputFile}_undistorted_fMRI2str.nii.gz ${WD}/${ScoutInputFile}_undistorted.nii.gz
+
+            log_Msg "---> register scout image to T1w"
+            # register undistorted scout image to T1w
```
mharms: Consistent with my above comments, this comment `# register undistorted scout image to T1w` is apparently just a copy/paste from the SEFM correction case, and in fact the comment is completely wrong in this context. (The file may have "_undistorted" in its file name, but it is in fact the original (distorted) scout.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: My preference would be to have:

```
<DateTime>: <ToolName>: WARNING: Line1 
<DateTime>: <ToolName>:          Line2
<DateTime>: <ToolName>:          Line3
```

I am not sure, how to do that programatically, but I can split it up manually. How long should each warning line w/o the `<DateTime>: <ToolName>:` be?

mharms: Let's wait for @coalsont's opinion before you do anything.  I'm not dogmatic about the line length (e.g., < 80 characters!), but I don't like it when the lines get so long that you can't prevent wraparound within a "reasonable" window width (say around 150-160 characters??)

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: I agree, this is a mess right now. Let's do both BOLD and Scout reorient in a single place only if DC method is NONE. The key question is where. I think It needs to be before motion correction (as the reference used can be another BOLD that already had `fslreorient2std` done.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: It was there in case of any depencency later on in the code. If there is none, it can be skipped.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Same reason. To make sure it is there if it is aa dependency later on in the code.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Also, FWIW, we already have other scripts that take the approach of just breaking long messages into multiple `log_Warn` lines.  Also, an advantage of that approach is that if we ever decided for example that `log_Warn` should report to stderr as well, you would still get the complete message.

sivcek:
The reason for the specific implementation was to not break anything later on in the code (e.g. in `OneStepResampling.sh`). If nothing depends on the specific "fake" files, then the code can be simplified and rewritten.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: A quick search seems to indicate that the usage of the `Warpfield` file is restricted to the SEFM correction section.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: The Jacobian is a bit more complicated, as I see that a jacobian of some sort is an "output" of the DistortionCorrection script, and thus might indeed be expected to exist as part of some downstream scripts.

mharms:
What I would suggest is that we attempt to write a "clean" implementation, and only create "fake" (misnamed) files if absolutely necessary.  Do you plan on testing each of these LegacyStyleData features on your end, once we have all the PRs done, right?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: It looks like the `${ScoutInputName}_undistorted*` nomenclature is baked into the downstream FreeSurfer BBR.  My suggestion would be that we write "clean" code for the initial registration, and then at the end the "NONE" subsection we create any additional necessary files for the downstream code, with a very clear comment on why we are creating them.  And then possibly delete those fake/misnamed files within `${WD}` at the end of the DistortionCorrection script.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: The date/tool part of the log messages is already quite long, so I don't really favor repeating it.  However, it is not clear that there is an easy solution that I would favor.  Maybe if I added the word-aware text wrapping with indenting from newopts.shlib to the logging shlib, this could be handled cleanly, but that isn't in scope for these PRs.

In other words: for now, whatever.

On a tangential note, I found it was easy to integrate testing for terminal width into the newopts.shlib word wrapping (though for redirection to files it may report a width of 80).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
coalsont: For any files that are just copies, you might consider making symlinks instead, so that it ives a hint that they are fake.

glasserm:
We couldn't replicate the fMRIPrep results when we tried for the ciftify paper.  I don't remember the details now, but I expect that the fMRIPrep paper was not as careful as we were to show that fMRI cortex was actually better situated in the cortical ribbon.

mharms:
Regardless, I think the proposed statement is too provocative in light of the clear claims made in the fMRIPrep paper, since this isn't the place for a scientific debate on that claim.  How about we simply end it with the more neutral:

`At a future time, we may be able to add support for fieldmap-less distortion correction which may be an improvement over no distortion correction, but is unlikely to be as good as fieldmap-based correction (the current "gold-standard").`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: So, since there is no easy solution, and since we already use multiple consecutive `log_Warn` lines in other places, that is the approach that I suggest we adopt here as well.

glasserm:
I don't agree, we don't name anyone's software explicitly here and make it clear that this is a hard and unsolved problem.

mharms:
That statement makes it appear that we are unaware of the fMRIPrep paper, and the claims therein.  I just don't see why it is necessary, or what it adds relative to what I proposed as a replacement.

glasserm:
Our position is that this problem is unsolved (else we would provide the solution instead of the no fieldmap option).  I’ll note that when I met with the fMRIPrep folks at OHBM and mentioned the non-replication, they agreed that more work was needed to get this right and we discussed collaborating on this.  Here is a potentially clearer version: 

WARNING: The fMRIVolume pipeline is being run without appropriate distortion correction of the fMRI image. This is NOT RECOMMENDED under normal circumstances. We will attempt 6 DOF FreeSurfer BBR registration of the distorted fMRI to the undistorted T1w image. Distorted portions of the fMRI data will not align with the cortical ribbon. In HCP data 30% of the cortical surface will be misaligned by at least half cortical thickness and 10% of the cortical surface will be completely misaligned by a full cortical thickness. At a future time, we may be able to add support for fieldmap-less distortion correction. At this time, however this problem is unsolved and no extant approach has been successfully shown to demonstrate clear improvement according to the accuracy standards of HCP-Style data analysis when compared to gold-standard fieldmap-based correction.

mharms:
That's good.

coalsont:
"At this time, however this problem is unsolved..." is strange comma placement, I would put the comma after "however", or both before and after.  Personally, I would avoid that phrasing and just say "However, this problem is currently unsolved..."

We could work in a "despite ongoing efforts" somewhere to make it more clear we are aware of the research.

glasserm:
Sure those would be okay additions.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Changed the warning.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: I moved the Scout reorient to be done with BOLD reorient in Slice Time Correction PR.

sivcek:
@mharms I tried creating a "clean" implementation with later creation of fake images, however, I got lost in what needs to be where in what form. At this time the code indeed mimics the SE option. All the files are there. I have added the following note:

```
# NOTE: To work with later code a number of fake files are created 
#       with names that do not reflect the true nature of the file
#       this is to be considered an initial fix for datasets that 
#       do not include data that enable distortion correction and
#       will be replaced with a cleaner code. 
```

And I have taken out the `fslreorient2std`, which is now part of the slice time correction PR. I hope this will be satisfactory for a temporary solution. We can clean the code after PRs are integrated and we can free time for housekeeping.

I did also add all the warning changes.

glasserm:
I agree with Mike in principle, but am not sure how long it will take and the amount of downstream incompatibilities this would cause.

mharms:
@sivcek: Are you continuing to work on this one?

sivcek:
@mharms I am waiting for the consensus on whether a clean rewrite is necesary, for which @glasserm shares the same concerns as I do—that it will take a long time and it might cause a number of issues downstream. For that reason I have devoted time to other PRs. My preference would be to not poke the hornet's nest. If you feel strongly that it has to be done and others agree, I'll get to it, but it will take a bit.

mharms:
Well, I share the concern that it might be messy, but I really don't like the idea of having a bunch of files named "_undistorted" that are in fact the exact opposite of that.  Only bad things can come from that rather egregious hack.  It might turn out to be necessary to purposely misname a small number of files, to avoid having to rewrite _large amounts_ of downstream code, but if we know which particular files those are, we can at least mitigate the potential dangers (e.g., delete the misnamed files at the conclusion of `DistortionCorrection...` once we are done with them).

mharms:
BTW, I brought this up, because with basically only one other PR still under work (#147), it seemed like maybe we could make some progress on this one while we figure out the path forward on that one.

sivcek:
@mharms I'm setting up the testing for <https://github.com/Washington-University/HCPpipelines/pull/147>. After that I can review the dependencies in the code in this PR. (probably late tomorrow) and then we can decide on the way forward for both.

coalsont:
As long as the functionality is guarded by the processing mode setting, I don't think it is the end of the world to fake a few more files than is technically necessary, as it should be obvious to the user that distortion correction isn't actually being done.  For files that are copies, using symlinks instead could give a hint that they aren't what they claim to be.

mharms:
Perhaps for the user that actually originally generated the data.  But data gets moved around, uploaded places, new grad students come into the lab and start using the data, etc, etc.

coalsont:
That sounds like a documentation problem, not an implementation one.  We could have the script create a "README_no_distortion_correction.txt" file in the output folder that explains what files are faked and why.

mharms:
That would be better than nothing, but you are still relying on people noticing that that file even exists.  The best solution is to simply avoid the documentation problem via a clean implementation.  At this point, we don't even know yet if a clean implementation that doesn't resort to using misnamed files isn't in fact a viable option.

coalsont:
```
tim@timsdev:/mnt/myelin/tim/pipelinedevel/Pipelines$ find . -type f | xargs grep -l undistorted
./fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
./GenerateSpinEchoBiasFieldPrereqs/GenerateSpinEchoBiasFieldPrereqs.sh
./PreFreeSurfer/PreFreeSurferPipeline.sh
```

So, "undistorted" isn't hardcoded in many places.  Is that the only keyword in filenames you are worried about?

mharms:
I haven't looked through the code to see if there are others, but yes, "_undistorted" is a clear offender that I noticed right away.

sivcek:
I just reviewed the code for nonDC option. 

* fake warpfield is not needed at all,
* fake jacobian is needed both in script and referenced out of it,
* `_undistorted` is mentioned in the script only, with clever use of variables it can be eliminated from use.

It is just too late for me to be able to do it tonight.

sivcek:
I changed the code so that when a DC method is employed the scout has the extension `_undistorted`, when no DC is employed, there is no extension. The code is flexible—I have left in a variable so that instead of no extension `_original` or `_raw` could be used. 

As `_undistorted` is not used outside of the script, there should be no issues downstream.

@mharms does that resolve your concern?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -452,40 +461,36 @@ case $DistortionCorrection in
...
+            cp "${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat" "${WD}/fMRI2str.mat"
 
-            # generate combined warpfields and spline interpolated images + apply bias field correction
+            # generate Scout2T1 warpfield and spline interpolated images + apply bias field correction
             log_Msg "generate combined warpfields and spline interpolated images and apply bias field correction"
-            ${FSLDIR}/bin/convertwarp --relout --rel -r ${T1wImage} --warp1=${WD}/WarpField.nii.gz --postmat=${WD}/${ScoutInputFile}_undistorted2T1w_init.mat -o ${WD}/${ScoutInputFile}_undistorted2T1w_init_warp
-            ${FSLDIR}/bin/applywarp --rel --interp=spline -i ${WD}/Jacobian.nii.gz -r ${T1wImage} --premat=${WD}/${ScoutInputFile}_undistorted2T1w_init.mat -o ${WD}/Jacobian2T1w.nii.gz
+            ${FSLDIR}/bin/convertwarp --relout --rel -r ${T1wImage} --premat=${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat -o ${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init_warp
+            ${FSLDIR}/bin/applywarp --rel --interp=spline -i ${WD}/Jacobian.nii.gz -r ${T1wImage} --premat=${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat -o ${WD}/Jacobian2T1w.nii.gz
```
glasserm: Why the functionality change here

sivcek: One is the use of `ScoutExtension` variable to avoid incorrect naming of the files, the other is the exclusion of fake WarpField. @mharms suggested we don’t need it as it is just 0s as no DC has been computed.

glasserm:
Can you run some tests of SE fieldmap, regular fieldmap, and no fieldmap to ensure nothing is broken by this?

sivcek:
I can run a test w/ and w/o SE, I don’t have data with a field map.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,56 @@ then
...
+                                                       # This is NOT RECOMMENDED under normal circumstances. We will attempt 6 DOF FreeSurfer BBR registration of the 
+                                                       # distorted fMRI to the undistorted T1w image. Distorted portions of the fMRI data will not align with the cortical ribbon. 
+                                                       # In HCP data 30% of the cortical surface will be misaligned by at least half cortical thickness and 10% of the cortical 
+                                                       # surface will be completely misaligned by a full cortical thickness. At a future time, we may be able to add support for 
+                                                       # fieldmap-less distortion correction; however, no extant software package or pipelines have successfully demonstrated 
+                                                       # clear improvement over no distortion correction, in the direction towards gold standard fieldmap-based correction.
```
mharms: The comments here don't match the current `log_Warn` message of what we converged on.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -12,12 +12,14 @@ FIELDMAP_METHOD_OPT="FIELDMAP"
...
 
 ################################################ SUPPORT FUNCTIONS ##################################################
 
 # --------------------------------------------------------------------------------
 #  Load Function Libraries
 # --------------------------------------------------------------------------------
+
 source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions; also sources log.shlib
 
 Usage() {
```
mharms: The parts in the Usage statement that refer to "readout distortion correction", should be "susceptibility distortion correction" instead.  (Currently, L's 49, 56, 59, and 63).  Looks like an incorrect copy/paste at some point in time from PreFreeSurferPipeline.sh.  Also, 'NONE' should probably we added as a listed mode.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -293,52 +305,53 @@ case $DistortionCorrection in
         fi
```
mharms: Please convert preceding `log_Msg; exit 1` to an `log_Err_Abort`

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -430,10 +445,53 @@ case $DistortionCorrection in
...
+            # NOTE: To work with later code a number of fake files are created 
+            #       with names that do not reflect the true nature of the file
+            #       this is to be considered an initial fix for datasets that 
+            #       do not include data that enable distortion correction and
+            #       will be replaced with a cleaner code. 
+
+
+            log_Msg "---> No distortion correction"
+
+            ScoutExtension=""
```
mharms: Weren't you intending to set `ScoutExtension` in this case to something like `orig` or `original`?  That also seems safer (to make sure that we aren't overwriting a file that we don't intend to overwrite).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -430,10 +445,53 @@ case $DistortionCorrection in
...
+                log_Msg "... running epi_reg (dof ${dof})"
+                ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}${ScoutExtension} --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init
+            elif [ $Preregister = "flirt" ] ; then
+                log_Msg "... running flirt"
+                ${FSLDIR}/bin/flirt -in ${WD}/${ScoutInputFile}${ScoutExtension} -ref ${WD}/${T1wBrainImageFile} -out ${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init -omat ${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat -dof 6
+            fi
+
+            #copy the initial registration into the final affine's filename, as it is pretty good
+            #we need something to get between the spaces to compute an initial bias field
+            cp "${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat" "${WD}/fMRI2str.mat"
```
mharms: Is this initial copy of `fMRIstr.mat` necessary?  Does it get used prior to the FS BBR in the 'NONE' case?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -430,10 +445,53 @@ case $DistortionCorrection in
...
+
+            # generate Scout2T1 warpfield and spline interpolated images + apply bias field correction
+            log_Msg "generate combined warpfields and spline interpolated images and apply bias field correction"
```
mharms: Where in here are we applying a "bias field correction"?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -430,10 +445,53 @@ case $DistortionCorrection in
...
+            ${FSLDIR}/bin/convertwarp --relout --rel -r ${T1wImage} --premat=${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat -o ${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init_warp
```
mharms: add a comment preceding this line:
```
# In the NONE condition, we have no distortion Warpfield.  Convert the Scout2T1 registration (affine)
# to its warp field equivalent, since we need "${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init_warp" later
```

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
glasserm: Doesn't look like it

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
glasserm: That only applies to the SE method

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: @mharms  is this correct:

```
# WARNING: The fMRIVolume pipeline is being run without appropriate distortion correction of the fMRI image. 
# This is NOT RECOMMENDED under normal circumstances. We will attempt 6 DOF FreeSurfer BBR registration of 
# the distorted fMRI to the undistorted T1w image. Distorted portions of the fMRI data will not align with 
# the cortical ribbon. In HCP data 30% of the cortical surface will be misaligned by at least half cortical 
# thickness and 10% of the cortical surface will be completely misaligned by a full cortical thickness. 
# At a future time, we may be able to add support for fieldmap-less distortion correction. At this time, 
# however, despite ongoing efforts, this problem is unsolved and no extant approach has been successfully 
# shown to demonstrate clear improvement according to the accuracy standards of HCP-Style data analysis when 
# compared to gold-standard fieldmap-based correction.
```

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Added NONE. Should it say that it is only available for LegacyStyleData processing mode?

```
  echo "               \"${NONE_METHOD_OPT}\""
  echo "                 do not use any susceptibility distortion correction"
```

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: @mharms There are a number of these throughout the script. Do I convert all of them?

mharms: Sure.  Might as well.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: Sure.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: I had it original at first. Then it seemed strange to be named `Scout_gdc_orig`. But I take your point. I'll add orig back in.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Well, if that matches the text of the current `log_Warn`, then yes.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mharms: How about we call it "_nosdc"?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Taken out.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Taken out of the description.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Added.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: I did.

sivcek:
@glasserm @mharms I put all the changes mentioned here in the new PR <https://github.com/Washington-University/HCPpipelines/pull/151> that is being actively tested. Please continue with review on that one.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Added.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: It does.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Ok. `_nosdc` it is.

sivcek:
Joined with other fMRIVolume PRs into <https://github.com/Washington-University/HCPpipelines/pull/156>

PR #145, opened 2019-10-03
fMRIVolume Related Pull Request — Support nonlinear registration to reference

sivcek:
Adds the ability for nonlinear registration of BOLD image to reference. We've noticed that it improves registration of BOLD images to reference image and MNI atlas in some cases of legacy data.


fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,38 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+ReferenceReg=`opts_GetOpt1 "--refreg" $@`                            # How to register bold image to reference? Using 'linear' (default) or also 'nonlinear' registration to movement reference.
```
coalsont: I don't see edits adding this to the help info.  Also, the sub-script also tests for if being `NONE` at one point...

sivcek: In `GenericfMRIVolumeProcessingPipeline.sh`:

```
show_usage() {
    echo "Usage information To Be Written"
    exit 1
}
```

But I added the info to `OneStepResampling.sh`. This parameter is in interaction with the BOLD reference PR. Namely, the parameter is only relevant if another BOLD reference is in use. If it isn't, then `refreg` is set to NONE. See line 338 in PR: <https://github.com/Washington-University/HCPpipelines/pull/140>

```
ReferenceReg="NONE"  # only do nonlinear or linear registration for target that is not of the same bold run
```

sivcek:
Expanded variable description.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +220,40 @@ then
...
+ComplianceMsg=""
+
+# -- BOLD nonlinear registration to reference image
+
+if [ "${ReferenceReg}" = 'nonlinear']; then
+  ComplianceMsg+=" --refreg=${ReferenceReg}"
+  Compliance="LegacyStyleData"
+fi
+
+check_mpp_compliance "${MPPMode}" "${Compliance}" "${ComplianceMsg}"
```
coalsont: FYI, since all of these separate PRs to GenericfMRIVolume add the common `check_mpp_compliance`, initialization of `Compliance`, and option parsing, I expect this will be a problem on merging, and you might need to make a new PR with all of the combined approved edits.

sivcek: I'm happy to merge all fMRIVolume related PRs into a new joint PR once all of them are approved.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,40 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+ReferenceReg=`opts_GetOpt1 "--refreg" $@`                            # In the cases when BOLD image is registered to a specified BOLD reference, this option
```
mharms: From the way the code is set up, it appears that the only thing that is ever checked is whether its value is 'nonlinear'.  i.e., the default of 'linear' doesn't trigger anything.  Given that, I'm wondering if the name of this option shouldn't be `--refregnonlinear` with values of either TRUE or FALSE (default).

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -108,23 +140,28 @@ echo "date: `date`" >> $WD/log.txt
 echo " " >> $WD/log.txt
 
 
-########################################## DO WORK ########################################## 
+########################################## DO WORK ##########################################
+
+if [ `${FSLDIR}/bin/imtest ${BiasField}` -eq 0 ]; then
+  verbose_red_echo "---> Missing Bias Field image, creating a fake one"
+  ${FSLDIR}/bin/fslmaths ${T1wImage} -abs -add 1 -bin ${BiasField}
```
mharms: I think the way we usually accomplish this is `-mul 0 -add 1`

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -30,17 +30,18 @@ Usage() {
   echo "             --scoutgdcin=<input scout gradient nonlinearity distortion corrected image (EPI pre-sat)>"
   echo "             --oscout=<output transformed + distortion corrected scout image>"
   echo "             --ojacobian=<output transformed + distortion corrected Jacobian image>"
+  echo "             --refreg=<whether to do 'linear', 'nonlinear' or no ('NONE') additional registration to movement reference image>"
```
mharms: The framing of this usage statement, with 3 values gives the impression that 'linear' is actually doing something.  But it appears that 'linear' and 'NONE' are the same.  Thence my suggestion that we consider using `--refregnonlinear=(TRUE|FALSE)` instead.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -179,25 +216,44 @@ for ((k=0; k < $NumFrames; k++)); do
...
-  FrameMergeSTRING="${FrameMergeSTRING}${WD}/postvols/vol${vnum}.nii.gz " 
+  FrameMergeSTRING="${FrameMergeSTRING}${WD}/postvols/vol${vnum}.nii.gz "
   FrameMergeSTRINGII="${FrameMergeSTRINGII}${WD}/postvols/vol${vnum}_mask.nii.gz "
 
   #Do Basic Cleanup
   rm ${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_gdc_warp.nii.gz
   rm ${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_all_warp.nii.gz
 done
+
+ceho "---> Merging results"
```
mharms: `ceho` not supported.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -179,25 +216,44 @@ for ((k=0; k < $NumFrames; k++)); do
...
 # Merge together results and restore the TR (saved beforehand)
 ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI} $FrameMergeSTRING $TR_vol
 ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
 fslmaths ${OutputfMRI}_mask -Tmin ${OutputfMRI}_mask
 
+ceho "---> Combining transformations"
```
mharms: Ditto.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -179,25 +216,44 @@ for ((k=0; k < $NumFrames; k++)); do
...
+  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${GradientDistortionField} --warp2=${MotionMatrixFolder}/mc2ref_warp.nii.gz --out=${GradientDistortionField}_nmc
+  GradientDistortionField=${GradientDistortionField}_nmc
+fi
+
+# Add linear transform to reference if needed
+if [ "$ReferenceReg" == "NONE" ]; then
+  lintoref=""
+else
+  lintoref="--premat=${MotionMatrixFolder}/${MotionMatrixPrefix}0000"
+fi
```
mharms: I would name the variable `prematStr`, set it to empty prior to the preceding conditional, and then move the setting of the non-empty version to within the preceding `if [ "$ReferenceReg" == "nonlinear" ]; then` conditional.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -179,25 +216,44 @@ for ((k=0; k < $NumFrames; k++)); do
...
+
+${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} ${lintoref} --warp1=${GradientDistortionField} --warp2=${OutputTransform} --out=${WD}/Scout_gdc_MNI_warp.nii.gz
```
mharms: Something about this doesn't seem right.  GDC is the very first warp done, so I don't see how any affine transform can be a `--premat`

glasserm:
I think the convertwarp commands are okay.

fMRIVolume/scripts/OneStepResampling.sh
mharms: @glasserm How can a `--premat` operation applied prior to any warp involving GDC be the correct thing?

glasserm:
I agree that it doesn't make sense to have an affine before gradient distortion correction.  Also in general there shouldn't be a need to have an affine for resampling the scout image if that is what was used as the registration target for motion correction.  Going back to the original code I wrote, I don't see that.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: That messes up OneStepResampling, when NONE is also an option if no other bold is used for reference.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: Fixed.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: No, NONE is used in the case when there is no external reference to register to. This is linked to PR <https://github.com/Washington-University/HCPpipelines/pull/140>

fMRIVolume/scripts/OneStepResampling.sh
sivcek: Agh difficult to weed. I'll check for them again in the combined fMRIVolume PR.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: Ok.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: If we have an external reference (see PR <https://github.com/Washington-University/HCPpipelines/pull/140>), we're first doing linear (and then also nonlinear if specified and prepared in line 245) registration to the external reference BOLD.

sivcek:
@mharms @glasserm The problem is that PR <https://github.com/Washington-University/HCPpipelines/pull/145> and PR <https://github.com/Washington-University/HCPpipelines/pull/140> shouldn't be split as the confusing part does linear and then (if set) also nonlinear transform to external BOLD if one is specified. And none of these if external BOLD reference is not used.

sivcek:
I created a combined redBold and nonlin PR to make things clearer.

mharms:
@glasserm This PR is deprecated, and has been replaced with #147.

PR #133, opened 2019-09-27
PreFreeSurfer Pipeline related pull request

sivcek:
### 1/ Cosmetic changes
Changed tabs to spaces amd removed spaces at end of lines

### 2/ Added optional `--mppversion` parameter
The parameter has the followinig possible values:
* hcp -> enable only strict HCP pipelines functionality (the default if the parameter is not specified)
* legacy -> enable extended pipelines functionality to work with legacy datasets

A check is performed based on this parameter. If ‘hcp’ is specified then the script ends with an error if:
* T2w input image is specified as NONE
* a custom brain / brain mask is used

The check is performed and logged before any processing is started, with explicit warning that the use does not conform to Glasser et al. (2013) paper when ‘legacy’ option is used.

### 3/ Added optional `--custombrain` parameter
The parameter has the following possible values:
* NONE -> do not use a custom brain (the default if the parameter is not provided)
* CUSTOM -> Use a custom brain image
* MASK -> Use a custom mask

Both CUSTOM and MASK options are provided for cases when  `PreFreeSurfer` has been run once before and the preparation of the structural images (e.g. brain masking) have not resulted in correct structural brain images. If either of these options is specifed, all the steps until “Performance of  Atlas Registration” are skipped.

If MASK is specified, then the user had to have placed a custom brain mask in a file  `${T1wFolder}/custom_acpc_dc_restore_mask` , which is then used  to mask the `${T1wFolder}/${T1wImage}_acpc_dc_restore` image to get the `${T1wFolder}/${T1wImage}_acpc_dc_restore_brain`  before the processing resumes with Atlas Registration.

If CUSTOM is specified, then the user had to have (re)placed the desired images and the processing resumes with Atlas Registration directly. 

### 4/ Enabled possibility to specify T2w input image as NONE
If T2w input image is specified as none, (and ‘legacy’  mppversion) is specified then:
* Gradient Nonlinearity Correction, Scan Averaging, ACPC alignment, and BrainExtraction steps are skipped for (nonexistent) T2w image
* Regular Bias Field Correction is skipped and (if so specified by the `--t1biascorrect` parameter) Bias Filed Correction is applied to T1w image only
* Creating one-step resampled version of T2w image is skipped

### 5/ Added optional `--t1biascorrect` parameter,
The parameter has the following possible values:
* YES -> do perform bias correction on T1w image only, the default and used when the parameter is not specified
* NONE -> do not perform bias correction on T1w image only

[...]

PR #147, opened 2019-10-10
fMRIVolume related pull request — combined use of BOLD refrence and optional nonlinear registration to it

sivcek:
This is a PR that combines using external BOLD as motion correction reference and adding an option to register to it using linear and also nonlinear registration.


fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,43 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+fMRIReference=`opts_GetOpt1 "--fmriref" $@`                          # reference BOLD name run to use for movement correction target and to copy atlas registration from (or NONE; default)
```
mharms: Probably need to be clearer as to how this differs from `--fmriscout`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,43 @@ then
...
+ReferenceReg=`opts_GetOpt1 "--refreg" $@`                            # In the cases when BOLD image is registered to a specified BOLD reference, this option
```
mharms: Maybe call this `--fmrirefreg`, to clearly tie it to the `--fmrireg` flag.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +385,82 @@ fi
 
 log_Msg "mkdir -p ${fMRIFolder}/MotionCorrection"
 mkdir -p "$fMRIFolder"/MotionCorrection
+
+if [ ! $fMRIReference = "NONE" ] ; then
+    log_Msg "Using reference image from ${fMRIReference}"
+    reference="$fMRIReference"/"$ScoutName"_gdc
```
mharms: Should there be a check here that `${reference}` actually exists?
Also, should there be a check that `${fMRIReference}` and `${fMRIFolder}` aren't in fact the same thing?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +385,82 @@ fi
...
+         --ojacobian=${fMRIFolder}/${JacobianOut} \
+         --dof=${dof} \
+         --fmriname=${NameOffMRI} \
+         --subjectfolder=${SubjectFolder} \
+         --biascorrection=${BiasCorrection} \
+         --usejacobian=${UseJacobian}
+else
+    log_Msg "linking EPI distortion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${fMRIFolder}/${DCFolderName} ] ; then
+        log_Msg "     ... removing stale link (or preexisiting files)"
```
mharms: An already existing folder shouldn't really happen, unless someone is running on top of existing data, in which case this seems more appropriate for a `log_Warn` to me.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -180,9 +217,13 @@ for ((k=0; k < $NumFrames; k++)); do
   fi
   prevmatrix="${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}"
 
-  # Combine GCD with motion correction
-  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/prevols/vol${vnum}.nii.gz --warp1=${GradientDistortionField} --postmat=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum} --out=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_gdc_warp.nii.gz
-  # Add in the warp to MNI152
+  ###Add stuff for RMS###
```
mharms: The comment here `###Add stuff for RMS###` should be eliminated, and the comments that I had added
`# Combine GCD with motion correction`
and
`# Add in the warp to MNI152`
restored to the appropriate locations.

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,25 @@ case "$MotionCorrectionType" in
     ;;
 esac
 
+
+# Run nonlinear registration if needed
+
+if [ "${ReferenceReg}" == "nonlinear" ] ; then
+  verbose_echo " ... computing nonlinear transform to reference"
+  verbose_echo "     ... generating bold average"
+  ${FSLDIR}/bin/fslmaths ${WorkingDirectory}/${OutputfMRIBasename} -Tmean ${WorkingDirectory}/${OutputfMRIBasename}_avg
```
mharms: Add comment as to why using the mean here, rather than just extracting the first frame and using that.

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,25 @@ case "$MotionCorrectionType" in
...
+
+  verbose_echo "     ... running fnirt: fnirt --in=${WorkingDirectory}/${OutputfMRIBasename}_avg --ref=${Scout} --iout=${WorkingDirectory}/${OutputfMRIBasename}_avg_nonlin --cout=${WorkingDirectory}/mc2ref_warp"
+  ${FSLDIR}/bin/fnirt --in=${WorkingDirectory}/${OutputfMRIBasename}_avg --ref=${Scout} --iout=${WorkingDirectory}/${OutputfMRIBasename}_avg_nonlin --cout=${WorkingDirectory}/mc2ref_warp
```
mharms: I noticed that this `fnirt` call isn't using either `--aff` or `--regmask`...

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,25 @@ case "$MotionCorrectionType" in
     ;;
 esac
 
+
+# Run nonlinear registration if needed
```
mharms: Let's provide more context here, to help users trying to follow the code:
```
# If registering across runs, perform nonlinear registration if requested.
# (If using linear registration, don't need to do anything extra here, since
# linear registration is handled implicitly via the motion correction).
```

sivcek:
So ...

In GenericfMRIVolumeProcessingPipeline.sh`, line 231-237. It reads whether there is an external reference and it reads how it should register to it. I've now expanded this part of the code, so that if `fMRIReference` is `NONE`, `ReferenceReg` is `NONE` as well (as we are not registering to an external reference. If `ReferenceReg` is not `NONE`, the default is `linear`. (I have now take this test from what was line `399`. 

In `MotionCorrection.sh`, line `64`  if `ReferenceReg` is `nonlinear` it computes a non-linear registration of mean motion crorrected timeseries to the external reference after the motion correction to that reference was done. The both average image and bold are registered and the warp is saved as `mc2ref_warp`

If external reference is used, the DC warp from that BOLD is copied over 

In `OneStepResampling`. If nonlinear was specified, then the nonlinear warp is added to each BOLD frame (line `222`)

Finally, scout output needs to be computed. Here he take the actual scout from this BOLD, not the reference bold. So here it makes a difference whether there was no external Reference (NONE), and no additional step is needed, if there was an external MC target, then the nonlinear warp to it needs to be added. And if there was no nonlinear registration, we still need to register the scout of the current BOLD to the scout of the reference BOLD. And for this if the `refreg` is not 'NONE' but rather either 'linear' or `nonlinear` a registration of frame 0000 to external scout needs to be added as a `--premat` to the warp that generates the final Scout output.

So it is true that we never explicitly test for `linear` but if we would have just two options then we would not be able to differentiate in the last step whether a linear alignment of current scout to refrerence scout is needed or not. 

Does this make it clear?

In the latest commit I have included the changes in the latest master.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done (see below)

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Used `fmrirefreg` and `fMRIReferenceReg`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Added this check as well.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Changed to Warn.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: Fixed with pulling in the latest changes.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: Added a comment.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: There is no affine transform needed, as the timeseries has already been registered linearly to the external scout, so this is an additional non-linear registration. But I have added the reference scout mask.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: Added.

glasserm:
This still does not address the issue of putting an affine matrix before gradient distortion correction on Line 265.

fMRIVolume/scripts/MotionCorrection.sh
glasserm: I guess the preferred FSL way of using FNIRT is to specify the original image and an affine rather than an affinely registered image.  I don't know how much this matters though.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -280,13 +324,31 @@ case "$BiasCorrection" in
...
 
+# set reference and check if external reference (if one is specified) exists 
+
+if [ ! $fMRIReference = "NONE" ] ; then  
+  fMRIReferencePath="$Path"/"$Subject"/"$fMRIReference"
+  log_Msg "Using reference image from ${fMRIReferencePath}"
+  fMRIReferenceImage="$fMRIReferencePath"/"$ScoutName"_gdc
+
+  if [ "$fMRIReferencePath" = "$fMRIFolder" ] ; then
+    log_Err_Abort "Specified BOLD Reference is the same as the current BOLD!"
```
mharms: Elaborate slightly:
`log_Err_Abort "Specified BOLD reference (--fmriref) is the same as the current BOLD (--fmriname)!"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -280,13 +324,31 @@ case "$BiasCorrection" in
...
+  fi
+
+  if [ `${FSLDIR}/bin/imtest reference` -eq 0 ] ; then
+    log_Err_Abort "Specified BOLD Reference ${reference} does not exist!"
```
mharms: I believe you mean to test on `$fMRIReferenceImage` here?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +404,74 @@ fi
...
+         --subjectfolder=${SubjectFolder} \
+         --biascorrection=${BiasCorrection} \
+         --usejacobian=${UseJacobian}
+else
+    log_Msg "linking EPI distortion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${fMRIFolder}/${DCFolderName} ] ; then
+        log_Warn "     ... removing stale link (or preexisiting files)"
+        rm -r ${fMRIFolder}/${DCFolderName}
+    fi
+    ln -s ${fMRIReference}/${DCFolderName} ${fMRIFolder}/${DCFolderName}
```
mharms: Quibbling, but we defined `DCFolder=${fMRIFolder}/${DCFolderName}` back at L421, so why aren't we using that variable where appropriate here?

mharms: Do we really want to create a symlinked _folder_?  That doesn't seem right to me

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +404,74 @@ fi
...
+ 
+    WD=${fMRIReference}/${DCFolderName}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/${RegOutput}
+    ${FSLDIR}/bin/imcp ${WD}/fMRI2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
+    ${FSLDIR}/bin/imcp ${WD}/Jacobian2T1w ${fMRIFolder}/${JacobianOut}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${QAImage} ${fMRIFolder}/${QAImage}
```
mharms: Should we be using `imln` here, rather than `imcp`, to give a nod to using symlinks?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +404,74 @@ fi
...
+    if [ -e ${fMRIFolder}/${DCFolderName} ] ; then
+        log_Warn "     ... removing stale link (or preexisiting files)"
+        rm -r ${fMRIFolder}/${DCFolderName}
+    fi
+    ln -s ${fMRIReference}/${DCFolderName} ${fMRIFolder}/${DCFolderName}
+ 
+    WD=${fMRIReference}/${DCFolderName}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/${RegOutput}
+    ${FSLDIR}/bin/imcp ${WD}/fMRI2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
+    ${FSLDIR}/bin/imcp ${WD}/Jacobian2T1w ${fMRIFolder}/${JacobianOut}
```
mharms: Is there any particular reason we didn't do it this way (to use less obscure copies of certain files outside of the low-level WD's from the ref run)?
e.g.,
```
    ${FSLDIR}/bin/imcp ${T1wFolder}/xfms/${fMRIReference}2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
    ${FSLDIR}/bin/imcp ${fMRIReferencePath}/${JacobianOut} ${fMRIFolder}/.
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +404,74 @@ fi
...
+else
+    log_Msg "linking EPI distortion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${fMRIFolder}/${DCFolderName} ] ; then
+        log_Warn "     ... removing stale link (or preexisiting files)"
+        rm -r ${fMRIFolder}/${DCFolderName}
+    fi
+    ln -s ${fMRIReference}/${DCFolderName} ${fMRIFolder}/${DCFolderName}
+ 
+    WD=${fMRIReference}/${DCFolderName}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/${RegOutput}
```
mharms: So, ${RegOutput} is named "Scout2T1w".  Is that going to lead to ambiguity regarding which particular "Scout" it is referring to?

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,28 @@ case "$MotionCorrectionType" in
...
+# (If using linear registration, don't need to do anything extra here, since
+# linear registration is handled implicitly via the motion correction).
+
+if [ "${fMRIReferenceReg}" == "nonlinear" ] ; then
+  verbose_echo " ... computing nonlinear transform to reference"
+  verbose_echo "     ... generating bold average"
+  # Generating a mean image to increase SNR to noise ratio when registering to scout.
+  ${FSLDIR}/bin/fslmaths ${WorkingDirectory}/${OutputfMRIBasename} -Tmean ${WorkingDirectory}/${OutputfMRIBasename}_avg
+
+  verbose_echo "     ... running fnirt: fnirt --in=${WorkingDirectory}/${OutputfMRIBasename}_avg --ref=${Scout} --refmask=${Scout}_mask --iout=${WorkingDirectory}/${OutputfMRIBasename}_avg_nonlin --cout=${WorkingDirectory}/mc2ref_warp"
```
mharms: Does `${Scout}_mask` actually exist?  I'm not seeing where it would have been created...
And can we call the warp `postmc2fmriref_warp` just to make it a little more descriptive?

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,28 @@ case "$MotionCorrectionType" in
     ;;
 esac
 
+
+# If registering across runs, perform nonlinear registration if requested.
+# (If using linear registration, don't need to do anything extra here, since
+# linear registration is handled implicitly via the motion correction).
+
```
mharms: Further extend comment:
```
# Note that if registering across runs, the "$Scout" input to MotionCorrection will
# be the *reference* scout image (by construction in GenericfMRIVolume).
```

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -108,23 +137,28 @@ echo "date: `date`" >> $WD/log.txt
 echo " " >> $WD/log.txt
 
 
-########################################## DO WORK ########################################## 
+########################################## DO WORK ##########################################
+
+if [ `${FSLDIR}/bin/imtest ${BiasField}` -eq 0 ]; then
+  verbose_red_echo "---> Missing Bias Field image, creating a fake one"
+  ${FSLDIR}/bin/fslmaths ${T1wImage} -mul 0 -add 1 ${BiasField}
+fi
```
mharms: I guess this is fine, but under what situation would the BiasField be missing in the first place?

coalsont:
> If external reference is used, the DC warp from that BOLD is copied over

So wait, you are expecting that the external reference is still distorted, with comparable size/direction of distortion, but has had distortion correction run?  This probably deserves a more explanatory comment, at least.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Yes

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: But nothing should be written back into that folder. No code downstream or anywhere else outside the `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh`, which is skipped at this point, writes to it.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I am vary of using symlinks for crucial files. I'd rather leave cp here unless there is a strong reason to move to links.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: That's fine with me. Changed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I thought that the file is used downstream and it has to keep that name, but it apparently is not, so I changed it to: 

```
${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/Reference${RegOutput}
```

So that it is clear that it is the reference scout.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: Well, `${Scout}_mask` does exist, but it is only generated in `IntensityNormalization.sh`. We did not use `--refmask` before. I checked, if existed in the final file structure and put it in because of your previous comment why we are not using `-refmat`. It worked fine before w/o it. Is there a strong reason we should?

I did change the warp name to the suggested one.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: Added.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: This is a remnant of no-biasfield option when T2w image is missing. Now we made biasfield obligatory even if ony T1w image is present, so I can take this part out.

sivcek: That was in the PreFreeSurferPipeline.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: done

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -220,6 +222,48 @@ then
 	log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+
+# ------------------------------------------------------------------------------
+#  Legacy Style Data Options
+# ------------------------------------------------------------------------------
+
+fMRIReference=`opts_GetOpt1 "--fmriref" $@`                          # reference BOLD name run to use for movement correction target and to copy atlas registration from (or NONE; default)
```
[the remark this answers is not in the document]
sivcek: done.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: @glasserm To use the `--aff` we would need to then first compute the motion corrected version of bold to its first image or scout and then compute an affine transform to reference scout and then (optionally) also add the nonlinear transform. Would this additional two step path improve FNIRT solution compared to working with BOLD that has been directly motion registered to reference Scout?

sivcek:
@coalsont 
Let's say, we have run BOLD_A fully through `fMRIVolume`. All the files are there including the DistorsionCorrection warp that was applied to the BOLD_A after the gradient distortion was computed. So we have `_gdc` for BOLD_A and SCOUT_A, and we have the `fMRI2str.nii.gz` warp that is used to register the `_gdc` version of BOLD_A and SCOUT_A to standard. This warp includes Distortion correction and registration to MNI.

Now, when we get to BOLD_B, we are using the very same files for computing Distortion Correction. so instead of recomputing all the steps in `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh`, all we need to do is register the BOLD_B and SCOUT_B to BOLD_A and SCOUT_A and then apply the rest of the transforms. 

The way it is done is to use SCOUT_A_gdc as a motion correction target for BOLD_B. This registration can be further improved by following up with a nonlinear registration of the mean BOLD_B registered to SCOUT_A to SCOUT_A. At this point, BOLD_B should be as close to SCOUT_A as possible, so the same `fMRI2str.nii.gz` computed and used for BOLD_A can now be used for BOLD_B. 

After running motion correction for BOLD_B_gdc, we have an affine transform for each BOLD_B_gdc frame to SCOUT_A_gdc and (optionally) a nonlinear warp (`postmc2fmriref_warp`) to follow up with to get BOLD_B_gdc_mc best registered to SCOUT_A_gdc.

When we get to `OneStepResampling.sh`, if we have used (optional) nonlinear registration, we need to add `postmc2fmriref_warp` as an additional end warp to each `# Combine GCD with motion correction` step. And after that, we can apply the `fMRI2str.nii.gz` warp computed on SCOUT_A_gdc. 

This is the whole logic of it.

@mharms suggestion was the following comment for `--fmriref`:

```
Reference BOLD run name (i.e., --fmriname from run to be used as *reference*) to use as 
motion correction target and to copy atlas (MNI152) registration from (or NONE; default).
```

@coalsont, how would you expand on that comment to make things even clearer?

sivcek:
I have added the following note to `--fmriref` explanation:

```
# NOTE: The reference BOLD has to have been fully processed using fMRIVolume pipeline, so
# that a distortion correction and atlas (MNI152) registration solution for the reference
# BOLD already exists.
```

sivcek:
**Re: `--premat` issue**
I have reviewed the code in detail. The `--premat` was indeed wrong in a number of ways and essentially not needed. I have removed it fully. It was added to generate the correct final Scout image, however, the Scout image from the currently processed bold is not even the scout that was used in any way. So now, I have just removed the generation of final scout in the `OneStepResampling` and instead the reference BOLD final scout is copied in the current BOLD folder from the refrence BOLD folder. It has to retain the regular name though as it is used downstream in `IntensityNormalization.sh`.

If you think copying the reference scout is misleading, then I can omit the copy and instead generate a new "fake" scout as a mean across the final BOLD timeseries.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -280,13 +328,31 @@ case "$BiasCorrection" in
...
 
+# set reference and check if external reference (if one is specified) exists 
+
+if [ ! $fMRIReference = "NONE" ] ; then  
+  fMRIReferencePath="$Path"/"$Subject"/"$fMRIReference"
+  log_Msg "Using reference image from ${fMRIReferencePath}"
+  fMRIReferenceImage="$fMRIReferencePath"/"$ScoutName"_gdc
+
+  if [ "$fMRIReferencePath" = "$fMRIFolder" ] ; then
+    log_Err_Abort "Specified BOLD reference (--fmriref) is the same as the current BOLD (--fmriname)!"
```
mharms: `log_Err_Abort "Specified BOLD reference (--fmriref=${fMRIReference}) is the same as the current BOLD (--fmriname=${NameOffMRI})!"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -280,13 +328,31 @@ case "$BiasCorrection" in
...
+  fi
+
+  if [ `${FSLDIR}/bin/imtest ${fMRIReferenceImage}` -eq 0 ] ; then
+    log_Err_Abort "Specified BOLD Reference ${fMRIReferenceImage} does not exist!"
```
mharms: `log_Err_Abort "Intended BOLD Reference does not exist (${fMRIReferenceImage})!"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +408,74 @@ fi
...
+         --dof=${dof} \
+         --fmriname=${NameOffMRI} \
+         --subjectfolder=${SubjectFolder} \
+         --biascorrection=${BiasCorrection} \
+         --usejacobian=${UseJacobian}
+else
+    log_Msg "linking EPI distortion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${fMRIFolder}/${DCFolderName} ] ; then
+        log_Warn "     ... removing stale link (or preexisiting files)"
+        rm -r ${fMRIFolder}/${DCFolderName}
```
mharms: We can use ${DCFolder} on these two lines as well.

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,30 @@ case "$MotionCorrectionType" in
...
+# be the *reference* scout image (by construction in GenericfMRIVolume).
+
+if [ "${fMRIReferenceReg}" == "nonlinear" ] ; then
+  verbose_echo " ... computing nonlinear transform to reference"
+  verbose_echo "     ... generating bold average"
+  # Generating a mean image to increase SNR to noise ratio when registering to scout.
+  ${FSLDIR}/bin/fslmaths ${WorkingDirectory}/${OutputfMRIBasename} -Tmean ${WorkingDirectory}/${OutputfMRIBasename}_avg
+
+  verbose_echo "     ... running fnirt: fnirt --in=${WorkingDirectory}/${OutputfMRIBasename}_avg --ref=${Scout} --iout=${WorkingDirectory}/${OutputfMRIBasename}_avg_nonlin --cout=${WorkingDirectory}/postmc2fmriref_warp"
+  ${FSLDIR}/bin/fnirt --in=${WorkingDirectory}/${OutputfMRIBasename}_avg --ref=${Scout} --iout=${WorkingDirectory}/${OutputfMRIBasename}_avg_nonlin --cout=${WorkingDirectory}/postmc2fmriref_warp
```
mharms: If you want to echo the full command call, it would be cleaner, and less prone to error, if you set up the command just once.  e.g.,
```
cmd="${FSLDIR}/bin/fnirt --in=..."
verbose_echo "         ... running ${cmd}"
${cmd}
```

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -44,6 +59,30 @@ case "$MotionCorrectionType" in
...
+
+  verbose_echo "     ... applying warp"
+  ${FSLDIR}/bin/applywarp --rel --interp=spline -i ${WorkingDirectory}/${OutputfMRIBasename} -r ${Scout}  -w ${WorkingDirectory}/postmc2fmriref_warp -o ${WorkingDirectory}/${OutputfMRIBasename}_nonlin
+
+  tmcbold="_nonlin"
```
mharms: define `tmcbold` _prior_ to the `applywarp` call, and then use `$tmcbold` within the `applywarp` call itself (so that changing the definition of `tmcbold` automatically keeps both pieces of code in sync).

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: indeed.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: ok.

fMRIVolume/scripts/MotionCorrection.sh
sivcek: ok.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -181,7 +210,12 @@ for ((k=0; k < $NumFrames; k++)); do
   prevmatrix="${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}"
 
   # Combine GCD with motion correction
-  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/prevols/vol${vnum}.nii.gz --warp1=${GradientDistortionField} --postmat=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum} --out=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_gdc_warp.nii.gz
+  if [ "$fMRIReferenceReg" == "nonlinear" ]; then
+      ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/prevols/vol${vnum}.nii.gz --warp1=${GradientDistortionField} --midmat=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum} --warp2=${MotionMatrixFolder}/postmc2fmriref_warp --out=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_gdc_warp.nii.gz
```
mharms: Add comment that the name of the nonlinear warp field is also hard-coded into MotionCorrection.sh

fMRIVolume/scripts/MotionCorrection.sh
sivcek: ok

fMRIVolume/scripts/OneStepResampling.sh
sivcek: done.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -208,9 +243,17 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
 # (gets applied in IntensityNormalization.sh; so don't change name here without changing it in that script as well).
 fslmaths ${OutputfMRI}_mask -Tmin ${OutputfMRI}_mask
 
+verbose_red_echo "---> Combining transformations"
 # Combine transformations: gradient non-linearity distortion + fMRI_dc to standard
-${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${GradientDistortionField} --warp2=${OutputTransform} --out=${WD}/Scout_gdc_MNI_warp.nii.gz
-${FSLDIR}/bin/applywarp --rel --interp=spline --in=${ScoutInput} -w ${WD}/Scout_gdc_MNI_warp.nii.gz -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -o ${ScoutOutput}
+if [ "$fMRIReferenceReg" == "nonlinear" ]; then
+  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${GradientDistortionField} --warp2=${MotionMatrixFolder}/mc2ref_warp.nii.gz --out=${GradientDistortionField}_nmc
```
mharms: This command still uses the old warp file name.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -208,9 +243,17 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
...
+  GradientDistortionField=${GradientDistortionField}_nmc
+fi
+${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} ${prematStr} --warp1=${GradientDistortionField} --warp2=${OutputTransform} --out=${WD}/Scout_gdc_MNI_warp.nii.gz
```
mharms: This command still includes `${prematStr}`

fMRIVolume/scripts/OneStepResampling.sh
sivcek: Changed it and also checked all other files.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -208,9 +243,17 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
...
+
+if [ "$fMRIReferenceReg" = "NONE" ]; then
+  ${FSLDIR}/bin/applywarp --rel --interp=spline --in=${ScoutInput} -w ${WD}/Scout_gdc_MNI_warp.nii.gz -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -o ${ScoutOutput}
+fi
 
 # Create spline interpolated version of Jacobian  (T1w space, fMRI resolution)
 #${FSLDIR}/bin/applywarp --rel --interp=spline -i ${JacobianIn} -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -w ${StructuralToStandard} -o ${JacobianOut}
```
mharms: I can't tell if the following Jacobian related stuff needs adjustment as well...

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -208,9 +243,17 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
...
-${FSLDIR}/bin/applywarp --rel --interp=spline --in=${ScoutInput} -w ${WD}/Scout_gdc_MNI_warp.nii.gz -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -o ${ScoutOutput}
+if [ "$fMRIReferenceReg" == "nonlinear" ]; then
+  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${GradientDistortionField} --warp2=${MotionMatrixFolder}/mc2ref_warp.nii.gz --out=${GradientDistortionField}_nmc
+  GradientDistortionField=${GradientDistortionField}_nmc
+fi
+${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} ${prematStr} --warp1=${GradientDistortionField} --warp2=${OutputTransform} --out=${WD}/Scout_gdc_MNI_warp.nii.gz
+
+if [ "$fMRIReferenceReg" = "NONE" ]; then
+  ${FSLDIR}/bin/applywarp --rel --interp=spline --in=${ScoutInput} -w ${WD}/Scout_gdc_MNI_warp.nii.gz -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -o ${ScoutOutput}
+fi
```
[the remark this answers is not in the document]
sivcek: Added.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +408,74 @@ fi
...
+         --usejacobian=${UseJacobian}
+else
+    log_Msg "linking EPI distortion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${DCFolder} ] ; then
+        log_Warn "     ... removing stale link (or preexisiting files)"
+        rm -r ${DCFolder}
+    fi
+    ln -s ${fMRIReference}/${DCFolderName} ${DCFolder}
+ 
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/Reference${RegOutput}
```
mharms: Is this `Reference${RegOutput}` file used anywhere?  If not, does it serve a purpose?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +408,74 @@ fi
...
+    ${FSLDIR}/bin/imcp ${T1wFolder}/xfms/${fMRIReference}2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
+    ${FSLDIR}/bin/imcp ${fMRIReferencePath}/${JacobianOut} ${fMRIFolder}/${JacobianOut}
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${QAImage} ${fMRIFolder}/${QAImage}
```
mharms: Likewise, does `${QAImage}` serve a purpose if we aren't running `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased`?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +408,74 @@ fi
...
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${NameOffMRI}_SBRef_nonlin ${fMRIFolder}/${NameOffMRI}_SBRef_nonlin
```
mharms: This doesn't seem right.  `${NameOffMRI}` for the current BOLD run is not going to be an appropriate name for the `${fMRIReference}`.  The following might work, but I'm not sure what you are trying to accomplish here:
`${FSLDIR}/bin/imcp ${fMRIReferencePath}/${fMRIReference}_SBRef_nonlin ${fMRIFolder}/${fMRIReference}_SBRef_nonlin`
But that's not going to work properly with the code in GenericfMRIVolume

fMRIVolume/scripts/OneStepResampling.sh
mharms: My point was that that whole section of Jacobian related code may need modification as well, so that it is still applicable in the context of the additional registration to another run.  (I can't tell based on a relatively quick look).

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +408,74 @@ fi
...
+else
+    log_Msg "linking EPI distortion correction and T1 registration from ${fMRIReference}"
+    if [ -e ${DCFolder} ] ; then
+        log_Warn "     ... removing stale link (or preexisiting files)"
+        rm -r ${DCFolder}
+    fi
+    ln -s ${fMRIReference}/${DCFolderName} ${DCFolder}
+ 
+    ${FSLDIR}/bin/imcp ${fMRIReference}/${RegOutput} ${fMRIFolder}/Reference${RegOutput}
+    ${FSLDIR}/bin/imcp ${T1wFolder}/xfms/${fMRIReference}2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
```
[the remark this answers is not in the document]
sivcek: done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Well, you suggested it would be good to have it with a name that would better inform what it is. And it is copied in the `${fMRIFolder}` from `${DCFolder}` and I guessed there is a reason for it. Why does `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh` save it, if it is of no interest?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: As above. I guessed it is a file that is useful for review. I can omit both.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -208,9 +244,17 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
...
```
mharms: Given that `--oscout` (`$ScoutOutput`) seems to be an expected output of `OneStepResampling` for use in other things in `GenericfMRIVolume`, I'm having a hard time seeing we can get by without it.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: That was the intended copy. I corrected it. It will work as in `OneStepResampling` if Reference is specified, a new `ScoutOutput` is not generated, but copied here. There is a note in `OneStepResampling` in that effect (line `256`). But I'll add one here as well.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: Yes, that's why it is copied earlier from the Reference folder.

sivcek: It's what I tried to explain here: <https://github.com/Washington-University/HCPpipelines/pull/147#issuecomment-542240850>

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Not everything that gets copied necessarily gets used, and it appears that the `--oregim=${fMRIFolder}/${RegOutput}` flag of `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh` is in that category.  IMO, we want to copy the minimal number of files necessary as part of this "reference BOLD" addition.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: That's generated by another flag (`--qaimage=${fMRIFolder}/${QAImage}`) that doesn't seem to extend beyond `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh` into GenericfMRIVolume.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok. I'll remove what is not needed downstream.

sivcek: It seems `${fMRIFolder}/${JacobianOut}` is not needed either.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: removed

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -342,57 +408,77 @@ fi
...
+    ln -s ${fMRIReference}/${DCFolderName} ${DCFolder}
+ 
+    if [ `${FSLDIR}/bin/imtest ${T1wFolder}/xfms/${fMRIReference}2str` -eq 0 ]; then
+      log_Err_Abort "The expected ${T1wFolder}/xfms/${fMRIReference}2str from the reference (${fMRIReference}) does not exist!"    
+    else
+      ${FSLDIR}/bin/imcp ${T1wFolder}/xfms/${fMRIReference}2str ${T1wFolder}/xfms/${fMRI2strOutputTransform}
+    fi
+    
+    # -- Note: we are copying the SBRef as it is the one used in this case and it won't be generated by OneStepResampling
+    ${FSLDIR}/bin/imcp ${fMRIReferencePath}/${fMRIReference}_SBRef_nonlin ${fMRIFolder}/${NameOffMRI}_SBRef_nonlin
```
mharms: Would it be much simpler (and less prone to confusion) if we simply:
1) Stopped pretending that the `--fmriscout` flag was somehow a valid flag in the context of using `--fmriref`?  i.e., If someone uses `--fmriref` AND `fMRIScout != "NONE"`, an error is generated.
2) Just make copies of the necessary "scout" files from the reference BOLD run, which I believe would be `${fMRIReferencePath}/${OrigScoutName}` and `${fMRIReferencePath}/${ScoutName}_gdc` (i.e., what we've defined as `${fMRIReferenceImage}`

Then, wouldn't all the scout related stuff in OneStepResampling (and downstream of that) simply work "as is"?

coalsont:
@sivcek the T1w to MNI warp isn't recomputed per BOLD run, it is taken from the structurals, right?  So, it seems like copying some concatenated warp from the another BOLD is more logic and puts more constraints on the option.  I was expecting the MC reference option to only take an image, really.

How does this interact with not having a fieldmap?  Rigid registration of distorted BOLD frames to undistorted scout sounds like it could have some issues, and the only obvious solution I can think of is to distort the MC reference to match the distorted BOLD.

I guess I'm also not really clear - when would you need to use a different BOLD run as a reference for motion correction?

fMRIVolume/scripts/MotionCorrection.sh
coalsont: Please don't use that whitespace-unsafe construction, if you want to do this, use this array-based pattern:

```
cmd=(${FSLDIR}/bin/fnirt --in="...")
verbose_echo "         ... running fnirt: ${cmd[*]}"
"${cmd[@]}"
```
The `*` in the echo line is to prevent it from splitting the echo string into multiple arguments, the `@` in the actual command line inside `""` causes it to break array elements into arguments while being safe for whitespace.

sivcek:
@coalsont 
Say that there is a session with:
T1w
T2w
SE-AP
SE-PA
BOLD_A
BOLD_B
BOLD_C

In `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh`, BOLD distortion correction is computed based on the same SE-AP/PA pair each time. So, if you process BOLD_A first, the `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh ` will compute a warp for registering the SCOUT_A to T1. You can then use distorted SCOUT_A as a motion correction reference for BOLD_B and apply the combined DC+fMRI-to-T1 warp that was computed to undistort and register distorted Scout_A to T1 based on the SE-AP/PA pair for BOLD_A also distroted SCOUT_A






what is computed for

fMRIVolume/scripts/MotionCorrection.sh
sivcek: ... and we could have just the uncomplicated initial version ... changed now. Again.

coalsont:
@sivcek Ok, so the option explanation should also mention that the reference fMRI must have the same phase encoding direction?  Or does it try to detect that and change whether it uses the distorted scout or undistorted scout?

To me, it seems slightly more obvious to copy only the initial BOLD-specific stuff, and rerun the combination with the MNI warp, rather than copying an already-combined warp.  It also seems like that would be more flexible for incorporating future improvements to fieldmap-less distortion correction.  Combining warps doesn't take much time.

As to your example, if BOLD_A has a good SCOUT_A, why doesn't BOLD_B have a good SCOUT_B?  QC failure, or protocol only acquired one scout per subject, or what?  I thought the first frames of single-band BOLD were basically as good as a scout, but I'm not really a scanner person.

fMRIVolume/scripts/MotionCorrection.sh
coalsont: The initial version wasn't whitespace safe, but could have been just by adding some quotes.  @mharms' version had the advantage of only defining the arguments once, so that editing it doesn't report the wrong command, but adding quotes would never make it whitespace-safe.  My version adds whitespace safety and removes the repetition, and making the initial version whitespace-safe would have needed all of those quotes anyway, and wouldn't resolve the repetition issue.  It is an improvement.

My comment was also intended for @mharms, so that he might avoid suggesting/using the unsafe pattern in the future.  No, we don't expect you to make these scripts completely whitespace-safe, but as long something is being edited, the new code shouldn't use bad patterns.

coalsont:
@sivcek Maybe I was confused by your terminology -  fMRI2str.nii.gz sounds like it should only go to T1w space, which we do NOT call MNI space (sure, it uses an MNI template, but it uses a rigid registration, so to us it is conceptually just an acpc alignment).  When we say MNI, we generally mean MNINonLinear.

Copying the combined warp from the distorted scout to T1w space seems okay if the option is only used when the BOLD distortion is effectively identical to the one being used as reference.  However, regenerating it from its pieces should be quite fast, so simplicity of code and ease of future development should probably be the deciding factors here.

mharms:
@sivcek: I think @coalsont's suggestion that we simplify this further and just go ahead and re-run `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh` is interesting.    The resulting transforms generated by `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh` should be identical if given the otherwise identical inputs (although that would be a point where users could mess up).

If you are too concerned about that possibility (users messing up), I still think the next easiest path forward is to copy reference scouts over, and basically run everything except `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh`.  And then just include the 'nonlinear' registration option stuff that you already have (and get any incidential Jacobian stuff right).

coalsont:
@mharms If I understand correctly, if `DistortionCorrection...` were also passed a new option of the reference fMRI to use, it wouldn't need to count on the user to get things right, and could potentially skip some redundant computation without necessarily having as much complication.

mharms:
@sivcek: Actually, it seems that we could simplify this considerably more by moving the extra 'nonlinear' registration to the reference bold out of MotionCorrection.sh and into OneStepResampling.sh.  It looks like all that would really be needed to make that work would be to add a `--fmrimc="$fMRIFolder"/"$NameOffMRI"_mc` flag to OneStepResampling, so that you have the name of the output of motion correction available.  Under that implemention, I don't think that MotionCorrection.sh would even need to be touched.  @coalsont @glasserm: What do you think of that?

coalsont:
I assume the nonlinear registration is to try to deal with having non-matching distortion between fMRI runs?  It seems like if the distortion is different, then having the nonlinear registration between the fMRIs done before motion correction could allow its use in making the motion correction target more similar to the distortion of the unregistered frames, by applying the inverse warp.  However, if we have no intention to ever do it that way, then simplicity may rule.

mharms:
> @mharms If I understand correctly, if `DistortionCorrection...` were also passed a new option of the reference fMRI to use, it wouldn't need to count on the user to get things right, and could potentially skip some redundant computation without necessarily having as much complication.

Interesting idea, but I'm not quite sure what we would do within `DistortionCorrection...` other than just re-run everything.  Can you elaborate?

mharms:
Also, while it is certainly important that the reference fMRI must have the same phase encoding direction as the run being processed, I don't see a practical way to check that (because we don't have a log/json file containing the parameters of previous pipeline runs).  So, I think a `log_Warn` on that particular assumption is certainly well warranted.

coalsont:
I meant skipping the entire topup/fieldmap case statement, and instead copy the distortion correction warp from the reference fMRI folder, then continue with the rest of registering and combining warps.  That way we wouldn't have to set up running topup or whatever in the new fMRI folder (and we wouldn't need the names of the original fieldmap files).  However, I haven't looked at the details of this, it just seems like a reasonable idea at a general level.

mharms:
@sivcek:  As @coalsont has raised, I think we need to better understand the intended role of the 'nonlinear' option, and whether it relates in any fashion to the no distortion correction PR.

coalsont:
Looks like what I was thinking of would also skip the BBR registration, and basically only do the various `convertwarp` commands in `DistortionCorrection...`.  So, what I was thinking of might be simpler to skip `DistortionCorrection...` entirely and just copy the handful of `convertwarp`s needed, which is fairly close to what the current state is.

fMRIVolume/scripts/MotionCorrection.sh
glasserm: --aff would be whatever linear transform is computed before the nonlinear transform.  It would be atypical to start with a nonlinear transform first.

glasserm:
I'm not really convinced one needs to register different runs to the same scout.  We have never worried about this with HCP data.  Regarding these other questions, it's a bit hard to know, as I haven't analyzed data in this way.  In the end we will want to be accounting for susceptibility by motion interactions, which are the main residual distortions in HCP-Style data.

sivcek:
# Replies to comments:
@coalsont 
> Ok, so the option explanation should also mention that the reference fMRI must have the same phase encoding direction?

Yes, that is an important point and I have added it to the option desciription:

```
Also, the reference BOLD must have been acuired using the same
phase encoding direction, or it can not serve as a valid reference. 
```

@mharms
> So, I think a log_Warn on that particular assumption is certainly well warranted.

Agreed. I have added:

```
  log_Warn "You are using an external reference (--fmriref=${fMRIReference}) for motion registration and"
  log_Warn "  distortion correction and registration to T1w image. Please make sure that the reference BOLD"
  log_Warn "  (--fmriref=${fMRIReference}) and the current bold (--fmriname=${NameOffMRI}) were acquired "
  log_Warn "  using the same acquisition parameters, e.g. phase encoding direction."
```

@coalsont
> As to your example, if BOLD_A has a good SCOUT_A, why doesn’t BOLD_B have a good SCOUT_B? QC failure, or protocol only acquired one scout per subject, or what? I thought the first frames of single-band BOLD were basically as good as a scout, but I’m not really a scanner person.

The problem is that low-resolution SB BOLD has less information than higher-resolution MB so the independent registrations for BOLD_A and BOLD_B  might just so slightly differ that they are not well aligned to each other. That is what we have noticed and present a problem when you e.g. have a task that spans multiple bolds. If BOLD_B is aligned to BOLD_A, then tje between-bold registration is improved.

@glasserm
> I’m not really convinced one needs to register different runs to the same scout.

Please see my previous reply  This is something we have noticed and tested empirically. Registering to the same scout does improve between-bold registration in the case of legacy, single-band data.

@coalsont
> Maybe I was confused by your terminology - fMRI2str.nii.gz sounds like it should only go to T1w space

I might have written it wrong in one of the comments before. Yes, what we want to re-use after registering BOLD_B to BOLD_A is the bold-to-T1-with-distortion-correction warp.

@coalsont
> Copying the combined warp from the distorted scout to T1w space seems okay if the option is only used when the BOLD distortion is effectively identical to the one being used as reference.

In the scenario:

* T1w
* T2w
* SE-AP
* SE-PA
* BOLD_A
* BOLD_B
* BOLD_C

Where all three bolds have the same acquisition parameters, BOLD distortion correction is the same for all three bolds.

@mharms
> I still think the next easiest path forward is to copy reference scouts over, and basically run everything except DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh. And then just include the ‘nonlinear’ registration option stuff that you already have (and get any incidential Jacobian stuff right).

Why would that be easier than the current implementation?

@coalsont
> I assume the nonlinear registration is to try to deal with having non-matching distortion between fMRI runs? It seems like if the distortion is different, then having the nonlinear registration between the fMRIs done before motion correction could allow its use in making the motion correction target more similar to the distortion of the unregistered frames, by applying the inverse warp.

Nonlinear registration is to help with cases that @glasserm mentioned. When between bold movement can slightly change the distortion, but we only have one SE-AP/PA pair available that was acquired before the BOLD_A, with which the distortion correction matches best. 

Taking this into account, I agree with @coalsont that it might be better to first do nonlinear warp and then motion correction. @mharms, we could avoid changes to `MotionCorrection.sh` and so simplify things if we do linear and (optionally) nonlinear registration of the current scout in the `GenericfMRIVolumeProcessingPipeline.sh` before `MotionCorrection.sh`, which can now stay as it was and work with the current Scout. We then combine the relevant warps at the `OneStepRegistration`. The added benefit is that we can also retain the current scout in the `OneStepRegistration`. 

@coalsont, @mharms, @glasserm
For me, this has been a very productive discsussion. If you agree with the above plan, I’ll go ahead and implement it.

fMRIVolume/scripts/MotionCorrection.sh
mharms: But, per your earlier comment, acceptable not to use `--aff` here, given that the data has already been linearly registered through the motion correction, right?

BTW: It has never been clear to me from the FNIRT documentation whether the `--aff` input is the _only_ affine that FNIRT ever uses, or if it is just the "starting guess" (and then iterates to update the affine component as part of iterating over the nonlinear warps).  Several parts of the documentation explicitly refer to it as the "starting guess", but other parts lead me to wonder if the affine component is ever indeed updated.

sivcek: Yes, though see my proposal for a change in how this is done.

mharms:
Lots of stuff here.  In no particular order:
1.  I think we should include some comments in the code regarding the intended purpose of the 'nonlinear' option.  i.e., to potentially help deal with situations in which movement between runs changes the susceptibility field, but a new estimate of the field is not available.  (And NOT to attempt to deal with differing distortions between a BOLD of one polarity and a reference/scout of another polarity, which is likely to go terribly bad).
2. The treatment of the current implementation regarding the scouts is unnecessarily confusing. In principle, the existing code already permits using any desired image as the scout, so we should make use of that capability and handle/name the scouts through the standard "channels" (unless someone has an argument for why that would be a problem).  In this regard, `--frmiref` is nothing but a convenient mechanism to specify a different run to copy the scout from as opposed to providing an actual reference/scout image as the input directly (and why there should be an error if `--fmriref` and `--fmriscout` are jointly set in GenericfMRIVolume).
3. If we do the nonlinear registration prior to MC:
A. We could either (i) warp the BOLD timeseries and MC that to the same Scout, or (ii) invwarp the Scout and use that as the target for the MC of the BOLD timeseries (with only GDC previously applied).  The motion parameter estimates won't be identical, but I'm not sure if one direction is better than the other.  I guess it would at least be faster to invwarp the Scout and use that as the MC target, so that we only need a spline interpolation on a single frame rather than a whole time series...
B. We either need to (i) add a small additional script prior to MotionCorrection, (ii) allow MotionCorrection to continue to be `--fmrirefreg` aware, or (iii) insert the nonlinear registration and resampling code into GenericfMRIVolume itself.  I personally favor either (i) or (ii).
4. It sounds like @coalsont is in agreement that skipping `DistortionCorrection...` entirely, and just copying the minimally necessary files, is the way that we should proceed (i.e., closest to what you already have).

BTW: I'm curious if you have looked at whether using the same scout for across run registration might marginally improve the across run alignment of non-legacy (high spatial resolution) data as well...?

fMRIVolume/scripts/MotionCorrection.sh
mharms: That seems an empirical question, and unless we have a good theoretical reason to favor that approach, introducing a "scout" for the run that is separate from the "reference scout" seems unnecessarily complicated, and would prevent the simplification that I've proposed that we simply treat the `--fmriref` as the only "scout".

coalsont:
Under the specific condition that we are reusing another fMRI's distortion correction outputs, yes, skipping `DistortionCorrection...` and copying/regenerating files seems like the simple solution.  This is a difference between --fmriref and --fmriscout.

However, it sounds like you could get all the effects of --fmriref by using both --fmriscout and pointing the fieldmap options at the applicable files (it would just take a little extra time to do the fieldmap processing again).  To me, using --fmriscout and pointing the fieldmap options to specific files seems less confusing and more direct about what is really going on (and couldn't that also deal with a phase-reversed scan, while --fmriref can't?).

@sivcek Do you see any benefit of --fmriref to the *processing* except saving time/space?  I ask this because it could change the framing of the discussion.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -229,7 +229,8 @@ fMRIReference=`opts_GetOpt1 "--fmriref" $@`                          # Reference
                                                                      # motion correction target and to copy atlas (MNI152) registration from (or NONE; default).
                                                                      # NOTE: The reference BOLD has to have been fully processed using fMRIVolume pipeline, so
                                                                      # that a distortion correction and atlas (MNI152) registration solution for the reference
-                                                                     # BOLD already exists.
+                                                                     # BOLD already exists. Also, the reference BOLD must have been acuired using the same
```
coalsont: typo: acquired

mharms:
@coalsont:  No, the fMRIVolume stuff inherently assumes that the "scout" and BOLD run are acquired with the same polarity and have the same distortions (and probably the spatial resolution and matrix size as well).  But yes, other than the extra 'nonlinear' registration, the `--fmriref` is nothing but a convenient proxy for specifying the `--fmriscout` input that was used already for another scan (and to allow for skipping `DistortionCorrection...`).  I personally think I favor the `--fmriref` convenience, and it think this can be plenty direct once we make it clear that is what is going on).

fMRIVolume/scripts/MotionCorrection.sh
coalsont: @mharms a warpfield can encode an affine, and I doubt fnirt does anything to prevent the warpfield from developing some amount of global affine component, and I really doubt that fnirt makes any modification to any actual set of affine parameters while registering.  Some registration tools basically pretend the data has already been resampled according to the initial input guess (in reality, using the initial affine as a premat when sampling through the current warpfield guess), which means you would need to concatenate the warpfield onto the initial affine to get the correct warp, and I think fnirt might be among them.  Do any of these get at your question?

coalsont:
I don't see why the extra registration option should depend on --fmriref, if the scout is specified as an arbitrary file anyway.

Having two ways to do the same thing is always going to confuse somebody.  Explaining that an option activates two other options (and currently allows a third option), but skips computation, sounds complicated even before trying to write the help info.

Basically, where I currently stand, I don't see a need for the extra complication of having the --fmriref option at all, the registration option can just be uncoupled from it, and people can choose what files to use as scout and fieldmap, same as before.  Simpler code is easier to maintain.

sivcek:
@mharms
1. Agreed
2. That is true, if one provides the reference scout as `—fmriref` then the final results would be the same just w/o saving the processing time (which is not non-negligible).
3. A (ii) seems better to me, the drawback is that in this case the scout is registered to the first frame, which might not be as good as the mean accross motion corrected timeseries.
B I think it depends on A. but I’d favor (iii) and (ii), adding another script seems unnecessary.
4. Skipping would be my preference as it appreciably reduces compute time. 

I think we tried it with MB data Alan was collecting, but as he had some files with reversed phase encoding polarity and we were in a hurry, we did not test it thorougly. I believe there were minor improvements though.

@coalsont
`--fmriscout` would have the same issues of knowing which files are of the same phase encoding (PA) direction. As scouts with a different PA would just be bad references. 
The key difference between using `--fmriscout` vs `--fmrireg` would be 
1. saving of processing time (which in our use proved to be quite substantial)
2. the use of `--fmrireg` would enable showing the effect on the scout of the BOLD in question, which makes it easier to review, but I agree this is probably a minor issue.

Oh, I see additional comments already ...

sivcek:
@coalsont I'm with @mharms. I think that from the user's perspective it is more confusing to provide a different scout but the actual BOLD scout, (which the parameter requests), than providing the actual current BOLD scout and explicitly specifying, what should be used as the reference.

Also, I think that having both could be beneficial for MB recordings as the actual scout is the better fit for the BOLD. It really is recorded at the very onset of the BOLD. So registering current scout to the reference scout would make for a better bridge between the two BOLD recordings.

mharms:
I'm not opposed to that, although it does somewhat bury the critical assumption that we get an identical registration if applied to same scout with the otherwise same inputs.  (I _think_ that is the case currently, but we need to confirm).  Then, if the behavior of one of the tools (epi_reg, topup, bbregister) changed in that regard, the implicit assumption would no longer hold (whereas that wouldn't be even a theoretical concern if we just directly copied the relevant registration files).

coalsont:
Specifying the real BOLD scout and then having the script not actually use it also sounds confusing, so if we went that direction, the scout should be optional.  Or is it used for something else?

mharms:
The terminology here of "current" and "actual" scout is getting confusing.  In _our particular acquisitions_ we happen to have an "SBRef" recon associated with every BOLD scan, which we typically use as the `--fmriscout` input for that scan.  But not all acquisitions have a "SBRef" recon.  More generally, the `--fmriscout` that we input for a scan is whatever we want it to be (or even 'NONE', in which case the first BOLD frame is extracted and named/used as the "Scout").

It seems we need to first address the lingering issue (brought up earlier in the context of the `--aff` discussion) of whether or not we want to support the ability to continue to register to the SBRef associated with the run, _and then as a second step_ affine register that SBRef to the "fMRI Reference" run.  I don't personally see the need for that, and if that is the case, then I don't see any reason why we would allow `--fmriscout` as an argument when `--fmriref` is being used.

mharms:
I would think that in this particular situation that chaining/bridging registrations in the manner that you are proposing might actually just lead to more chances for errors in the registration...

fMRIVolume/scripts/MotionCorrection.sh
glasserm: I think there is a preference in FSL to feed FNIRT the original image and the affine matrix rather than the resampled image and no affine.

mharms:
I was thinking about this a little more, and had the following realization which @sivcek probably already appreciated, but I'm not sure was made explicit for the rest of us.  Namely, we need _something_ to register to the "reference scout" for the 'nonlinear' case -- previously, the mean image of the post-motion corrected time series was used for that.  However, if we are going to move the nonlinear registration to occur prior to MC we need something else to register to the "reference scout", and it doesn't make much sense to use the mean of the _pre_ motion corrected time series.  So, we either need to use the first frame of the pre-MC'ed time series, which will have lower SNR, or we use the typical `--fmriscout` input to provide a 2nd scout that can be used to bridge the two runs in the manner that @sivcek was proposing.  (Or, we stick with doing the `nonlinear` registration _after_ MC, as is currently implemented, and we can continue to use the mean of the post-MC'ed timeseries for registration).

Note that if we went with supporting the bridging approach (reference scout from `--fmriref` and the "usual" scout from `--fmriscout`, the `--fmriscout=NONE` option would presumably result in the first frame of the BOLD gettting extracted as the usual scout, and used as the bridge.  So users would have flexibility regarding the choice of the bridging scout (at the cost of increased complexity).

I continue to be concerned that we have no empirical data that the bridging approach will work as well as the approach that @sivcek used previously on some legacy-style data, or even that it will satisfy the lesser bar of merely working robustly.

glasserm:
This makes sense and these are reasonable concerns.

mharms:
So, just to be clear (if case it wasn't already), notwithstanding my concern about changing the algorithm relative to previously and the potential impact of that, I think I've come around to seeing the value of maintaining the usual `--fmriscout` input/interpretation in the context of adding the `--fmriref` option.

mharms:
One other possible "insight" for framing the discussion:  What is really complicating this whole thing is the interaction between the reference BOLD option and the additional 'nonlinear' registration.  If 'linear' was the only option, it very well might make the most sense (and be most robust) to not worry about bridging between the two scouts, and simply register to the reference scout directly during motion correction (i.e., the current implementation).  It's not until we add the 'nonlinear' option that we need to worry about all these other issues.

sivcek:
@mharms I am in complete agreement about the strategy to movement reguster to the actual scout and then add the linear and (optional) nonlinear registration of the actual scout to the reference scout in OneStepResampling. I think it should work well with legacy data and Siemens MB data. And it would be the clearest for the user. The only thing I am slightly worried about is the situation we are actually in right now in Ljubljana with the Philips scanner. Namely, whereas Philips does the REF scan as part of the bold acquisition (the first volume before it continues with the MB acquisition) with Philips, we have to make it a separate sequence so there is a small but still a delay between the scout is taken and the MB BOLD acquisition. It might (probably?) be nothing to worry about, but should the participant move in between... 

I guess in this case linear registration to scout and nonlinear between bolds should still be the best option. 

If we have an agreement, I'll go ahead and implement it.

coalsont:
@sivcek The subject is expected to have some motion during the BOLD sequence, and I don't think the motion correction assumes the first frame is already aligned with the scout.  So, I am a little lost as to what problem you are suggesting.

Also, based on this, you were only worried about the scout to structural registration being inconsistent when using separate scouts, but weren't worried about the motion correction to a different reference scout having a similar effect?

mharms:
> 
> Also, based on this, you were only worried about the scout to structural registration being inconsistent when using separate scouts, but weren't worried about the motion correction to a different reference scout having a similar effect?

That's exactly the concern that I'm waffling on regarding the bridging approach.  While it might make for a clean implementation, how do we know that adding an extra registration won't in fact lead to the same sort of subtle across-run misalignment that this whole `--fmriref` approach was intended to prevent (and shown to resolve adequately in your previous testing).

glasserm:
I guess I am still having a hard time following this discussion.  I am concerned that if we need to do some of these things, it suggests the fMRI to T1w registration is not stable in this data, which is surprising if we are using BBR.  I think that like Mike says we should have a look at this data and test the data.  I am okay with an alternative method being developed and tested.  After tomorrow I will have more time to review in detail.

mharms:
@sivcek: How important is the extra 'nonlinear' registration to this?  Without that, specifying an alternative run as the reference would be a relatively simple thing.  We could just use the scout from the reference scan as the motion correction target, skip any concern about "bridging" between the two different scans, and be done with it.

sivcek:
@mharms I was trying to find the data on our SB acquisitions where we compared these, but it seems we've removed them already. I can re-run a couple of studies with independent processing, linear and non-linear registration to the first BOLD and then do a formal comparison. This would be using the existing Qu|Nex approach w/o the bridging. We can then decide whether nonlinear provides enough improvement to include it. How does that sound?

mharms:
Well, we basically need some sort of comparison like that for the proposed paper anyway, right?  But what is the time frame here?  We are getting very near to the point in time where we need to wrap up this current round of additions, get some testing done to ensure the pipelines still run after the PRs, get a new pipeline version tagged, build a new container, and start processing the HCP-Lifespan data.  I'm assuming that the testing alone of all the newly added options (to ensure that they simply run to completion) will take a week...

sivcek:
Well, of the three options (independent reference, first reference linear registration, first reference non-linear registration), the first one is already running. I think we should have them completed over the weekend. In the meantime we can work on PreFS PR.

mharms:
?? "PreFS PR" to me would be #133, which has already been merged.

sivcek:
I'm sorry, I meant <https://github.com/Washington-University/HCPpipelines/pull/144#issuecomment-543806085>

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: corrected

sivcek:
Joined with other fMRIVolume related PRs into PR <https://github.com/Washington-University/HCPpipelines/pull/156>. The new PR has the discussed direct motion correction registration to the specified reference.

PR #149, opened 2019-10-16
Add capability to use FLAIR style acquisitions within FreeSurferPipeline.sh

mharms:
This will hopefully deal with the file handling issues of using a FLAIR acquisition within recon-all, but note that per `recon-all.v6.hires`:
`# - Might run with FLAIR instead of T2, but this has not been tested.`

So caveat emptor.

### DIFF
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -171,7 +171,7 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
 
   [--seed=<recon-all seed value>]
 
-  [--flair]
+  [--flair]  (experimental)
       Indicates that recon-all is to be run with the -FLAIR/-FLAIRpial options
       (rather than the -T2/-T2pial options).
       The FLAIR input image itself should still be provided via the '--t2' argument.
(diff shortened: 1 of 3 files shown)


mharms:
One other tidbit to be aware of from inside the `conf2hires` script:
```
    if($usedev == 0) set cmd = (mris_make_surfaces)
    set cmd = ($cmd  -cortex 0 -noaparc\
      -pial pial.rawavg -white NOWRITE -nowhite \
      -orig $srcsurf.rawavg -orig_white white.rawavg -orig_pial wo${MMode}.pial.rawavg \
      -aseg rawavg.aseg.presurf -wm rawavg.wm -filled rawavg.filled -mgz \
      -T1 rawavg.brain.finalsurfs)
    if($DoT2)    set cmd = ($cmd -T2 rawavg.$MMode -nsigma_above 2 -nsigma_below 3)
    if($DoFLAIR) set cmd = ($cmd -T2 rawavg.$MMode -nsigma_above 3 -nsigma_below 3)
```
So, note that the value of `-nsigma_above` differs between the T2 and FLAIR modes in terms of how exactly the pial surface is positioned.

mharms:
I wonder if the $DoFLAIR version of the command isn't supposed to use the `-FLAIR` flag...  I'll check Doug.

sivcek:
Looks good to me.

mharms:
Cool.  Let's not merge this quite yet, until I hear from Doug if there is a typo in the `mris_make_surfaces` command in the `$DoFLAIR` condition in the `conf2hires` helper script.

mharms:
Ok.  Doug confirmed that there was indeed a bug in the `conf2hires` script in the `DoFLAIR` mode, which is now fixed.  This should now be ready to merge.

PR #152, opened 2019-10-23
Tested bugfixes for PreFreeSurferPipeline

sivcek:
These are bugfixes for PreFreeSurferPipeline scripts that have been tested to work on the standard use and new functionalities.

### DIFF
--- a/PreFreeSurfer/PreFreeSurferPipeline.sh
+++ b/PreFreeSurfer/PreFreeSurferPipeline.sh
@@ -540,10 +540,10 @@ if [ "$CustomBrain" = "NONE" ] ; then
       # skip modality if no image
 
       if [ "${TXwInputImages}" = "NONE" ] ; then
-        log_Msg "Skipping Modality: " $TXw " - image not specified."
+        log_Msg "Skipping Modality: $TXw - image not specified."
         continue
       else
-        log_Msg "Processing Modality: " $TXw
+        log_Msg "Processing Modality: $TXw"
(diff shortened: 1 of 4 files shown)


PreFreeSurfer/PreFreeSurferPipeline.sh
```diff
@@ -546,6 +537,15 @@ if [ "$CustomBrain" = "NONE" ] ; then
       fi
       OutputTXwImageSTRING=""
 
+      # skip modality if no image
+
+      if [ "${TXwInputImages}" = "NONE" ] ; then
+        log_Msg "Skipping Modality: " $TXw " - image not specified."
+        continue
+      else
+        log_Msg "Processing Modality: " $TXw
```
mharms: Unrelated, but this just looks weird to me.  Can we put the entire `log_Msg`'s inside a single quotation mark?
e.g.,
```
      if [ "${TXwInputImages}" = "NONE" ] ; then
        log_Msg "Skipping Modality: $TXw - image not specified."
        continue
      else
        log_Msg "Processing Modality: $TXw"
```

coalsont: I missed that - I agree, please change the messages to Mike's suggestion, to make it obvious how spaces get preserved (the existing version will result in double spaces around the value of TXw).

sivcek: Changed.

Issue #146, opened 2019-10-03
Misc tweaks to include as part of current coding sprint


mharms:
1) ~Add reporting of tool and pipeline versions to `fMRIVolume` and `fMRISurface`~
2) ~Fix the name of the `BiasFieldCorrection_sqrtT1wXT1w.sh` script (and the working dir that it creates) to "BiasFieldCorrection_sqrtT1wXT2w.sh"~
3) ~Enable the PreFS, FS, and PostFS scripts to _conveniently_ provide a usage statement (i.e., without checking the environment variables first).~
4) ~Add some comments in TopupProcessingAll.sh about `--usejacobian`.~
5) ~Add a usage statement to `fMRIVolume` (important given all its new features).~
6) ~Clean up reuse of "STRINGII" in CreateMyelinMaps to reduce confusion.~
7) ~Remove `set -e` from scripts that now source `debug.shlib`~
8) ~Harmonize/update the "# Requirements for this script" language at the front end of scripts (remove specific version listed within individual scripts).~
9) ~Update the contents of T1w/Diffusion/eddylogs to include newer files~

mharms:
2. Done
4. Done
7. Moot -- already done.
8. Done

PR #153, opened 2019-10-26
Various cleanup items

mharms:
I've started implementing the "cleanup" items mentioned in #146.
If you foresee any problems, let me know.


mharms:
@glasserm @coalsont @sivcek  Does anyone object if I remove the reliance on scripts of obscure environment variables such as `HCPPIPEDIR_Global` and `HCPPIPEDIR_PreFS`, and instead simply define those variables directly in the scripts that need them. 
e.g.,
```
HCPPIPEDIR_Global=${HCPPIPEDIR}/global/scripts
HCPPIPEDIR_PreFS=${HCPPIPEDIR}/PreFreeSurfer/scripts
```
within `PreFreeSurferPipeline.sh`

glasserm:
Isn't it better that the within pipeline paths are all defined in the central setup script rather than repeatedly in the pipelines themselves?

coalsont:
A larger number of environment variables may not be better than hardcoding small parts of the repository structure.  I've always been lukewarm about those, and have avoided using them.

They would only be a benefit if we renamed some folders (and since these variables are generally unused, we currently wouldn't get that benefit).  It isn't clear that any of the pipelines should really be reaching into a different pipeline's code directory anyway, so having all of them defined globally also doesn't seem quite right.

mharms:
@coalsont's comment highlights the problem:  _Even if_ we wanted to mess with the folder structure/location of the "helper" scripts, we can't rely on changing the environment variables in the setup script to capture all instances of the use of those scripts, since some of the main-level scripts "hard-code" the location anyway.  Given that, the use of these obscure environment variables seems pointless to me, thence my proposal that I simply define them directly in the scripts that use them.  (We could leave them defined in `finish_hcpsetup.shlib`, just as a fail-safe, lest I miss any).

glasserm:
Perhaps it would be worth keeping those variables that are used across multiple scripts?

coalsont:
`_Global` and related might have some merit as globally exported (but the `binaries` folder doesn't exist currently, so that can be deleted).

All the per-pipeline ones should probably not be created by the environment setup at all (by default, the per-pipeline scripts directory should be considered private from other pipelines), and they probably don't need to be exported at any point (even if an internal script calls another internal script, it can just define the variable at the top - if it is common enough, then the main pipeline script could export it instead).

In short, I tend to agree that many of these do not belong in any global file, as that is another way for dependencies to spread more than necessary.  If there are several pipelines that reach into another pipeline's scripts directory, then that would indicate that either those things should be moved to global, or this concept of the pipelines being mostly independent from each other is wrong.

glasserm:
Seems reasonable.

FreeSurfer/scripts/FreeSurferHiresWhite.sh
```diff
@@ -1,6 +1,59 @@
 #!/bin/bash
 
-source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions; also sources log.shlib
+# --------------------------------------------------------------------------------
+#  Usage Description Function
+# --------------------------------------------------------------------------------
+
+script_name=$(basename "${0}")
```
coalsont: Change to `script_name=$(basename -- "${0}")` to not error when pasted into interactive (though it will then say "-bash" is the tool name).  Yes, this edit seems to work on mac, too.

glasserm:
Mike is this ready to merge?

mharms:
Not quite.  Let me get the cleanup of the global environment variables in there...

sivcek:
If you can finish and merge today, I will prepare fMRIVolume full PR tomorrow with the changes integrated.

Examples/Scripts/SetUpHCPPipeline.sh
```diff
@@ -1,12 +1,27 @@
 #!/bin/echo This script should be sourced before calling a pipeline script, and should not be run directly:
 
+SAVEHCPPIPE="${HCPPIPEDIR:-}"
+
+# Set up specific environment variables for the HCP Pipeline
+export HCPPIPEDIR=
+
+#trick to allow certain setups to source an unmodified SetUp... and get some functionality
+if [[ -z "$HCPPIPEDIR" ]]
```
mharms: Isn't this always going to be true, given that 3 lines earlier is `export HCPPIPEDIR=` ?

coalsont: That is the line the user is intended to edit.

mharms: Ah, ok.  But I'd suggest that we consolidate all the variables that the user needs to set into one section at the top (HCPPIPEDIR, MSMBINDIR, MATLAB_COMPILER_RUNTIME) and then follow with all the stuff that the user shouldn't need to mess with, with a comment to that effect.  Right now it is a bit confusing how the `if [[ -z "$HCPPIPEDIR" ]]` stuff is "breaking up" the variables that the user needs to set.

Also, within that conditional, why aren't we using `export HCPPIPEDIR=` rather than just `HCPPIPEDIR=`?

coalsont: Since one of the default values (MSMCONFIGDIR) uses HCPPIPEDIR, they can't all be together if we have this conditional allowing an external value of HCPPIPEDIR to be used without editing.  Either these two variables are separated by some code, or we abandon this feature.

Exporting a variable once sets an internal flag on it.  Redundantly exporting has no effect.  The fact that it can be combined with setting a value is only for convenience.

coalsont: Another possibility is that if the user shouldn't actually have control of MSMCONFIGDIR, that can be moved to finish_hcpsetup.shlib.  However, this would mean that if someone wants to play with registration parameters, they have to make edits inside their repository.

mharms:
Ok, if everyone is ok with the commits that I just pushed, we should probably merge this.  I'll need to do a round of harmonization with the fMRIVolume scripts yet, but waiting for the @sivcek PRs to avoid merge conflicts.

Do we want to tackle the issue of removing the hard-coded paths to FSL and Workbench?  If so, I could perhaps take a stab at that, but I'd start it as a new branch/PR, once we have this PR merged in.

coalsont:
We don't need to change working code.  Renaming all the usage functions to match is more pedantic than I would have cared to change.

Examples/Scripts/SetUpHCPPipeline.sh
mharms: My point was that there is no need for the setting of MSMBINDIR, MATLAB_COMPILER_RUNTIME and FSL_FIXDIR to occur after the `if [[ -z "$HCPPIPEDIR" ]]` conditional, since none of those depend on `HCPPIPEDIR`.  So, I'd move up all the stuff that the user is expected to have to mess with, and then distinguish from the rest of the script with a :
```
# ----------------------
# Should not need to modify anthing beyond here
# ----------------------
```
section.

Now that you mention it, yes I think the setting of `MSMCONFIGDIR` should go into `finish_hcpsetup.shlib`, since very few people will ever need to mess with this, since we are including the config files as part of the repo itself.
If someone wants to mess with their own config files, they can "reset" `MSMCONFIGDIR` at the end of `SetUpHCPPipeline.sh`, or simply reset it outside of `SetUpHCPPipeline.sh` entirely.

My one minor concern about not "reexporting" `HCPPIPEDIR` is that if a user deletes the `export HCPPIPEDIR=` line entirely, then it potentially doesn't get exported at all, right?  Why not just redundantly reexport, since it causes no harm, and protects against users that might not know all the details of how exporting works?

mharms:
I was in the scripts already changing the return status, so I just changed the name of the usage functions for consistency while I was at it.

coalsont:
Also, I think it is fine to merge, we don't need to immediately resolve whether the potential for confusing users is worth the feature of sourcing an unmodified setup file (after all, a similar effect can be accomplished by just making a copy of the setup and editing it once, and then sourcing it in your shell init or manually before trying to run things in the repo).

Examples/Scripts/SetUpHCPPipeline.sh
coalsont: If they delete an export line entirely, they are doing something wrong.  I guess I can see the possibility.

There already is a line saying the user shouldn't need to modify anything below it.

mharms:
Not following.  We are already sourcing an unmodified setup file, so how it is confusing to put something else in there?

coalsont:
No, the still recommended, and generally used, procedure says to make a copy of the setup file and modify it and then source it.  These edits to SetUp... were for a request you made to be able to more quickly test things out of a fresh repo clone.

You yourself were confused by the new SetUp script because of how it interrupted the previously simple block of export lines to edit.

mharms:
I see it now.  I was expecting the "ignore from here onward" to be much higher up actually.  I just pushed a commit with a reorg that I think accomplishes that.  I can revert it if necessary.

coalsont:
Your reorg also makes it call freesurfer's setup script (was previously commented out in the editable section).  I don't know whether that could cause a problem for some users (also, the fsl setup is still commented out).

mharms:
Yeah, it is a bit unparallel because I didn't want to automatically source the fsl setup in the "don't need to edit" section, assuming that if a user already has FSLDIR set in their environment, then the fsl setup script has probably already been sourced.  Would you be happy if I just moved the the sourcing of the FS setup script back above the "don't need to edit section", with a default of it commented out?

coalsont:
That seems safest, but I don't know how much danger there really is.

mharms:
Ok.  This good?  I left out the explicit setting of `FSL_DIR` since FS's own setup scripts inherently already take care of that.

glasserm:
Note that ${FSL_DIR} and ${FSLDIR} are not the same.  I think the former is a FreeSurfer variable.

coalsont:
FS's scripts may guess at FSLDIR, the purpose of FSL_DIR is to tell FS what version we REALLY want.  At least, that is how I remember it the last time I took a deep dive into FS setup.  Again, I would recommend leaving the FSL/FS logic as it was, since we don't hve a need to mess with it.

mharms:
Yes, my point was that FS's own setup scripts already export `FSL_DIR` to be `FSLDIR` if FSL_DIR isn't already defined.

mharms:
@coalsont So, we _have_ told FS what version of FSL we REALLY want, by setting FSLDIR prior to running the FS setup scripts.

coalsont:
As I recall, FS setup can IGNORE what FSLDIR is already set to, but its first guess is always FSL_DIR if it isn't empty.

Please put the FS/FSL logic back the way it was, for safety, so we don't have to take a deep dive into the FS setup script during this "cleanup" PR.

mharms:
I probably should have mentioned that I implemented @coalsont's request, so I believe this PR is ready to go.

PR #155, opened 2019-10-30
ENH: Generate the eddy QC report

MichielCottaar:
This should add the single-subject QC report to the HCP pipelines. I am still testing this locally, but that requires rerunning eddy, so it will probably not be finished before the AHM.


MichielCottaar:
One thing to think about is whether we want the QC report to be in its default location (${subject}/Diffusion/eddy/eddy_unwarped_images.qc/*). It might be nice to move it to a location easier to find (e.g. ${subject}/QC/Diffusion/*).

glasserm:
Is it images or a QC report?  If it is images, why not resample them to T1w space and do that?

MichielCottaar:
It is just a QC report (no NIFTI files). The report does contain pngs of the brain, which would of course not be showing the brain in T1-weighted space, but that should not be an issue. So I think the `eddylogs` is a good place to put the report.

Another issue is: there are some text files with the movement, average residual, etc. per volume. Due to the resampling of two volumes into one after eddy, these can no longer be used as confound regressors (that is true both in the original diffusion space and the T1-weighted space). I don't see an obvious fix for this (and of course having the report itself would still be useful anyway).

mharms:
That said, while `eddy_quad` is more "report" (with png's) than images, perhaps we should be resampling the cnr maps generated by eddy (`--cnr_maps` flag) into the T1w space?  I believe that @MichielCottaar raised as a possible addition previously?

DiffusionPreprocessing/scripts/run_eddy.sh
```diff
@@ -599,6 +599,8 @@ main()
 	eddy_command+="${sep_offs_moveOption} "
 	eddy_command+="${rmsOption} "
 	eddy_command+="${ff_valOption} "
+	eddy_command+="--cnr_maps "
+	eddy_command+="--residuals "
```
mharms: Not sure if we should "hard code" these flags in -- that is kind of what we included the `--extra-eddy-arg` flag for.  I'm amenable to including `--cnr_maps` as a hard-coded default, since that seems relatively harmless, but I don't think we should hard-code in `--residuals`.

MichielCottaar: `--residuals` does indeed produce a large output file. I can see the argument for leaving it as an option. We could also add a `--qc_options` flag to run_eddy.sh, which would add the flags needed to produce a full QC report (just `--cnr_maps` and `--residuals` in this case).

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -17,6 +17,18 @@ globalscriptsdir=${HCPPIPEDIR_Global}
...
 datadir=${workingdir}/data
 
+echo "Generating eddy QC report in ${eddy_dir}/eddy_unwarped_images.qc"
+qc_command="${FSLDIR}/bin/eddy_quad "
+qc_command+="${eddydir}/eddy_unwarped_images "
+qc_command+="-idx ${eddydir}/index.txt "
+qc_command+="-par ${eddydir}/acqparams.txt "
+qc_command+="-m ${eddydir}/nodif_brain_mask.nii.gz "
+qc_command+="-b ${eddydir}/Pos_Neg.bvals "
+qc_command+="-g ${eddydir}/Pos_Neg.bvecs "
```
mharms: What does the `-g` flag do?  I can't tell from the QUAD UserGuide: https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/eddyqc/UsersGuide.  In particular, what if `eddy` isn't run with the `--residuals` option?  Does QUAD then fail if `-g` is specified?

Also, what is the role of the `-s, --slspec` flags in QUAD?  Should we be including that?  And if so, is there an option to read the json instead?

MichielCottaar:
Transforming some of the output maps from eddy to T1-weighted space (in particular the CNR maps) would be easy to do. I agree with Jesper that it is probably not very useful, but we might as well.

DiffusionPreprocessing/scripts/run_eddy.sh
mharms: Personally, I'd be fine with just hard coding in `--cnr_maps` as a flag.  One could argue that should always be generated anyway.  But leaving `--residuals` for specification via `--extra-eddy-arg`

DiffusionPreprocessing/scripts/eddy_postproc.sh
mharms: The plan is to call `eddy` using `--extra-eddy-arg=--json=<jsonfile.json>`.  So since, we will never need to construct the `slspec` file, I guess we wouldn't have whatever additional info the `slspec` approach can approach.  Are we missing anything substantive if we don't have that?

DiffusionPreprocessing/scripts/run_eddy.sh
MichielCottaar: Ok, I'll just remove the `--residuals` flag from this patch

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: I think QUAD just adds a slice2volume summary measure to the tables in the single-subject PDF report. The distribution of this summary measure would then also be included in the group report by SQUAD. So I don't think it would be too crucial to include (my guess would be that this slice2volume metric would correlate quite strongly with the general movement metrics anyway, but I have no evidence for that).

mharms:
One other random thought:  I don't suppose that `eddy --json` generates an output text file of the form that would be appropriate for the `slspec` format?  If so, then obviously we could use that file as the `slspec` file for QUAD.

MichielCottaar:
TBH, it would not be very hard to include a python script to do that conversion from jsom to slspec ourselves (I don't think eddy does so).

MichielCottaar:
Ok, this version works. I've attached the produced QC report for reference.

[qc.pdf](https://github.com/Washington-University/HCPpipelines/files/3810066/qc.pdf)

What I haven't implemented:
- passing in slspec (either from user or converted from json)
- converting CNR maps. There seemed to be a lot of steps to do this properly (first convert to undistorted diffusion space in eddy_postproc, keeping the unwarped version and then converting to Structural space in Diffusion2Structural.sh). It does not seem worth it to me to add so much code for a file that people probably won't use. There are already CNR summary measures in the report.

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -17,6 +17,20 @@ globalscriptsdir=${HCPPIPEDIR_Global}
 eddydir=${workingdir}/eddy
 datadir=${workingdir}/data
 
+echo "Generating eddy QC report in ${workingdir}/qc"
+rm -r ${workingdir}/qc
+qc_command="${FSLDIR}/bin/eddy_quad "
+qc_command+="${eddydir}/eddy_unwarped_images "
+qc_command+="-idx ${eddydir}/index.txt "
+qc_command+="-par ${eddydir}/acqparams.txt "
+qc_command+="-m ${eddydir}/nodif_brain_mask.nii.gz "
```
coalsont: I know none of the rest of the script is whitespace-safe, but would you mind converting this command-building code to the whitespace-safe array pattern anyway?  Note that there needs to be an unquoted space between separate arguments in the same set of parenthesis, `(-idx "${eddydir}/index.txt")`, but separate sets of parenthesis don't require any added space, as the separation is accomplished by being separate array elements, not by literal whitespace.

```
qc_command=("${FSLDIR}/bin/eddy_quad")
qc_command+=("${eddydir}/eddy_unwarped_images")
qc_command+=(-idx "${eddydir}/index.txt")
...
"${qc_command[@]}"
```

DiffusionPreprocessing/scripts/eddy_postproc.sh
mharms: I know you said that the `-g` flag does nothing currently, but for the future (when it does something) shouldn't we be using `${eddydir}/eddy_unwarped_images.eddy_rotated_bvecs` as the argument to `-g`?

Relatedly, in this section of `eddy_postproc.sh`:
```
if [ ${CombineDataFlag} -eq 2 ]; then
	${FSLDIR}/bin/imcp  ${eddydir}/eddy_unwarped_images ${datadir}/data
	cp ${eddydir}/Pos_Neg.bvals ${datadir}/bvals
	cp ${eddydir}/Pos_Neg.bvecs ${datadir}/bvecs
```
shouldn't we instead be doing:
```
if [ ${CombineDataFlag} -eq 2 ]; then
	${FSLDIR}/bin/imcp  ${eddydir}/eddy_unwarped_images ${datadir}/data
	cp ${eddydir}/Pos_Neg.bvals ${datadir}/bvals
	cp ${eddydir}/Pos_Neg.bvecs ${datadir}/bvecs_noRot
        cp ${eddydir}/eddy_unwarped_images.eddy_rotated_bvecs ${datadir}/bvecs
```

mharms:
Now that we've figured out the spline resampling/masking issues, isn't it just a matter of a couple lines in `eddy_postproc.sh` and `DiffusionToStructural.sh` to get the CNR maps into the T1w space?  I actually think they would be quite a nice addition.

glasserm:
I agree with Mike about the CNR maps.

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: Fixed

MichielCottaar:
I'm testing it with the suggested changes again

glasserm:
Thanks Michiel.

MichielCottaar:
Ok, this version appears to work

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -19,17 +19,17 @@ datadir=${workingdir}/data
...
+qc_command+=("${eddydir}/eddy_unwarped_images")
+qc_command+=(-idx "${eddydir}/index.txt")
+qc_command+=(-par "${eddydir}/acqparams.txt")
+qc_command+=(-m "${eddydir}/nodif_brain_mask.nii.gz")
+qc_command+=(-b "${eddydir}/Pos_Neg.bvals")
+qc_command+=(-g "${eddydir}/Pos_Neg.bvecs")
+qc_command+=(-o "${workingdir}/qc")
+qc_command+=(-f "${workingdir}/topup/topup_Pos_Neg_b0_field.nii.gz")
+qc_command+=(-v)
+${qc_command[@]}
```
coalsont: This line should be `"${qc_command[@]}"`, using @ inside double quotes is a special syntx that causes the argument separation to be done by the array elements, rather than by whitespace.

DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
```diff
@@ -353,6 +353,10 @@ main()
 		cp -p ${filename} ${to_location}
 	done
 
+  mkdir -p ${outdirT1w/qc
+  cp ${outdir}/qc/* ${outdirT1w}/qc
+  immv ${outdirT1w}/cnr_maps ${outdirT1w}/qc/cnr_maps
```
mharms: Missing a closing bracket.  And, can we preserve the original time stamp on the files during the copy? And can we call the output directory "QC", just to match the convention that we are using for a "StructuralQC" directory?
i.e., this set of 3 lines then becomes:
```  
mkdir -p ${outdirT1w}/QC
cp -p ${outdir}/QC/* ${outdirT1w}/QC
immv ${outdirT1w}/cnr_maps ${outdirT1w}/QC/cnr_maps
```

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -17,6 +17,20 @@ globalscriptsdir=${HCPPIPEDIR_Global}
 eddydir=${workingdir}/eddy
 datadir=${workingdir}/data
 
+echo "Generating eddy QC report in ${workingdir}/qc"
+if [ -d "{workingdir}/qc" ]; then rm -r ${workingdir}/qc; fi
```
mharms: Change to capital "QC" if you adopt my other suggestion.

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -17,6 +17,20 @@ globalscriptsdir=${HCPPIPEDIR_Global}
...
+qc_command=("${FSLDIR}/bin/eddy_quad")
+qc_command+=("${eddydir}/eddy_unwarped_images")
+qc_command+=(-idx "${eddydir}/index.txt")
+qc_command+=(-par "${eddydir}/acqparams.txt")
+qc_command+=(-m "${eddydir}/nodif_brain_mask.nii.gz")
+qc_command+=(-b "${eddydir}/Pos_Neg.bvals")
+qc_command+=(-g "${eddydir}/eddy_unwarped_images.eddy_rotated_bvecs")
+qc_command+=(-o "${workingdir}/qc")
```
mharms: And would need capital "QC" here as well.

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -95,42 +111,45 @@ ${FSLDIR}/bin/fslmaths ${datadir}/data -abs -Tmin -bin -fillh ${datadir}/fov_mas
 	 
 if [ ! $GdCoeffs = "NONE" ] ; then
     echo "Correcting for gradient nonlinearities"
-	# Note: "data_warped" is eddy-current and suspectibility distortion corrected (via 'eddy'), but prior to gradient distortion correction
-	# i.e., "data_posteddy_preGDC" would be another way to think of it
-    ${FSLDIR}/bin/immv ${datadir}/data ${datadir}/data_warped
+	  # Note: data in the warped directory is eddy-current and suspectibility distortion corrected (via 'eddy'), but prior to gradient distortion correction
+	  # i.e., "data_posteddy_preGDC" would be another way to think of it
+    mkdir -p ${datadir}/warped
```
mharms: If we want to consolidate the code a bit, we could define `warpedDir=${datadir}/warped` and then use `${warpedDir}` throughout the following.

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -95,42 +111,45 @@ ${FSLDIR}/bin/fslmaths ${datadir}/data -abs -Tmin -bin -fillh ${datadir}/fov_mas
...
     # apply gradient distortion correction
-    ${globalscriptsdir}/GradientDistortionUnwarp.sh --workingdir="${datadir}" --coeffs="${GdCoeffs}" --in="${datadir}/data_warped_dilated" --out="${datadir}/data" --owarp="${datadir}/fullWarp"
+    ${globalscriptsdir}/GradientDistortionUnwarp.sh --workingdir="${datadir}" --coeffs="${GdCoeffs}" --in="${datadir}/warped/data_dilated" --out="${datadir}/data" --owarp="${datadir}/fullWarp"
+    ${FSLDIR}/bin/immv ${datadir}/fullWarp ${datadir}/warped
+    ${FSLDIR}/bin/immv ${datadir}/fullWarp_abs ${datadir}/warped
+    ${FSLDIR}/bin/imrm ${datadir}/warped/data_dilated
+
+    # Transform CNR maps
+    ${CARET7DIR}/wb_command -volume-dilate ${datadir}/warped/cnr_maps_warped.nii.gz $DilateDistance NEAREST ${datadir}/warped/cnr_maps_dilated.nii.gz
+    ${FSLDIR}/bin/applywarp --rel --interp=trilinear -i ${datadir}/warped/cnr_maps_dilated -r ${datadir}/warped/cnr_maps_dilated -w ${datadir}/warped/fullWarp -o ${datadir}/cnr_maps
```
mharms: Did you intentionally choose trilinear rather than spline interpolation here?

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: This type of syntax is why I don't like writing in bash...

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: No, that should have been spline interpolation

MichielCottaar:
Indeed, the test will probably not finish today, but I'll let you know if it works.

MichielCottaar:
This version seems to work both with and without gradient-distortion correction

glasserm:
I would merge this when the conflicts are resolved

mharms:
What conflicts?  The PR is reporting that it has no conflicts and merging can be performed automatically.

glasserm:
This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

coalsont:
It conflicts for the rebasing merge mode, which I asked him to use.  If @MichielCottaar can rebase this branch onto master and manually resolve any conflicts, that should fix it.  However, if he has been merging in the latest master at several points, then we may have to use the merge commit mode.

glasserm:
I am letting Tim handle any PR that does not Rebase and merge automatically since he has asked me not to use the other modes.

MichielCottaar:
Sorry, I did use merge onto master. If you prefer, I could just squash all of the history, which would allow a rebase to work. However, that would hide the history anyway, so I think it makes more sense to just use merge commit mode for this branch.

PR #156, opened 2019-11-01
fMRIVolume with all functionality

sivcek:
This is a combined PR for all fMRIVolume functionality. Use of external BOLD reference is based on motion correction step. Scout / BOLDRef is copied from the reference BOLD. The PR was fully tested and includes the changes in the latest master.


coalsont:
Could you give a summary of the testing you have done?  I think the most critical test is to run it in hcp mode and a somewhat in-depth comparison of the results to the existing pipelines release.

sivcek:
@coalsont The test was run on single-band data with high-resolution T1w and T2w images and a SE pair for distortion correction. It was run w/o registration to external reference (which has the same behavior as if the functionality would not exist), w/ linear, and w/ nonlinear registration to external reference. The tests are summarised in Confluence <https://wiki.humanconnectome.org/display/CCFPRIV/Qu%7CNex+Pull+Requests+Functionality+Testing+Progress>. What specific tests would you like me to run and on what data?

coalsont:
I am suggesting running an HCP subject through this branch (with hcp options), and comparing to the results from the pipelines before any of your changes (the v4.0.1 release).  The idea is to verify that we haven't introduced any problems into the hcp-style data stream.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+                                                                         # Verbose (-v) is already turned on. TR is read from 'pixdim4' of the input NIFTI itself.
+                                                                         # e.g. --slicetimerparams="--odd@--ocustom=<CustomInterleaveFile>"
+BOLDMask=`opts_GetOpt1 "--boldmask" $@`                                  # Specifies what mask to use for the final bold:
+                                                                         #   T1_fMRI_FOV: combined T1w brain mask and fMRI FOV masks (the default), 
+                                                                         #   T1_DILATED_fMRI_FOV: a once dilated T1w brain based mask combined with fMRI FOV
+                                                                         #   T1_DILATED2x_fMRI_FOV: a twice dilated T1w brain based mask combined with fMRI FOV, 
+                                                                         #   fMRI_FOV: a fMRI FOV mask
+                                                                         #   NOTE: When using SEBASED bias field correction, the mask will default to T1_fMRI_FOV
```
mharms: This comment is unclear to me.  `T1_fMRI_FOV` is already the default setting.  Are you saying that when using SEBASED bias field correction that the `--boldmask` option is _irrelevant_ and `T1_fMRI_FOV` mode will always be used, regardless of the `--boldmask` setting?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+fMRIReferenceReg=`opts_GetOpt1 "--fmrirefreg" $@`                        # In the cases when BOLD image is registered to a specified BOLD reference, this option 
+                                                                         #   specifies whether to use 'linear' or 'nonlinear' registration to reference BOLD.
+                                                                         #   Default is 'linear'.
+
+
+
+# Defaults
+PreregisterTool=`opts_DefaultOpt $PreregisterTool "epi_reg"`
+DoSliceTimeCorrection=`opts_DefaultOpt $DoSliceTimeCorrection "FALSE"`   # WARNING: This LegacyStyleData option of slice timing correction is performed before motion correction 
+                                                                         #   (as is typically done in legacy-style brain imaging) and thus assumes that the brain is motionless.
```
mharms: I don't know if I would imply that there is consensus on the order issue, given that AFNI and SPM apply STC prior to MC, and FSL does it opposite.  See brief discussion here https://wiki.humanconnectome.org/display/PublicData/HCP+fMRI+slice-timing+acquisition+parameters and the links therein.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+                                                                         #   data quality measures as shown in Power et al 2017 PLOS One "Temporal interpolation alters motion in fMRI
+                                                                         #   scans: Magnitudes and consequences for artifact detection." Slice timing correction and motion correction 
+                                                                         #   would ideally be performed simultaneously; however, this is not currently supported by any major software 
+                                                                         #   tool. HCP-Style fast TR fMRI data acquisitions (TR<=1s) avoid the need for slice timing correction, 
+                                                                         #   provide major advantages for fMRI denoising, and are recommended. 
+                                                                         #   No slice timing correction is done by default.  
+
+# If --dcmethod=NONE                                                     # WARNING: The fMRIVolume pipeline is being run without appropriate distortion correction of the fMRI image.
```
mharms: There doesn't appear to be any parsing of `--dcmethod` anywhere in the script.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+                                                                         #   thickness and 10% of the cortical surface will be completely misaligned by a full cortical thickness. 
+                                                                         #   At a future time, we may be able to add support for fieldmap-less distortion correction. At this time, 
+                                                                         #   however, despite ongoing efforts, this problem is unsolved and no extant approach has been successfully 
+                                                                         #   shown to demonstrate clear improvement according to the accuracy standards of HCP-Style data analysis when 
+                                                                         #   compared to gold-standard fieldmap-based correction.
+
+# Defaults
+
+fMRIReference=`opts_DefaultOpt $fMRIReference "NONE"`
+BOLDMask=`opts_DefaultOpt $BOLDMask "T1_fMRI_FOV"`
```
mharms: Move up to be with the other "Defaults".

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+  fi 
+
+  if [ ! -d "$ReferenceResultsFolder" ] ; then
+    log_Err_Abort "Reference results folder does not exist and can not be used (${ReferenceResultsFolder})!"
+  fi 
+
+  # print warning
+
+  log_Warn "You are using an external reference (--fmriref=${fMRIReference}) for motion registration and"
+  log_Warn "  distortion correction and registration to T1w image. Pleaase consider using this option only"
```
mharms: Typo, "Pleaase ..."

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+  log_Warn "  in cases when only one BOLD Reference image is available or when processing low resolution"
```
mharms: Do we want to provide any sense of why this option might be considered in the case of low spatial resolution data?  (Also, given that resolution can be spatial or temporal, I'd explicitly say "... low spatial resolution").

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,15 +457,82 @@ if [ ! -e "$fMRIFolder" ] ; then
   log_Msg "mkdir ${fMRIFolder}"
   mkdir "$fMRIFolder"
 fi
-cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
+${FSLDIR}/bin/imcp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName"
+
+# --- Do slice time correction if indicated
```
mharms: Can we add to this comment the following:
```
# Note that in the case of STC, $fMRIFolder/$OrigTCSName will NOT be the "original" time-series
# but rather the slice-time corrected version thereof.
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,15 +457,82 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+    log_Msg "Copying Scout from Reference BOLD"
+    ${FSLDIR}/bin/imcp ${fMRIReferencePath}/Scout* ${fMRIFolder}
+
+    for simage in SBRef_nonlin SBRef_nonlin_norm
+    do
+        ${FSLDIR}/bin/imcp ${fMRIReferencePath}/"${fMRIReference}_${simage}" ${fMRIFolder}/"${NameOffMRI}_${simage}"
+    done
+
+    mkdir -p ${ResultsFolder}
+    ${FSLDIR}/bin/imcp ${ReferenceResultsFolder}/"${fMRIReference}_SBRef" ${ResultsFolder}/"${NameOffMRI}_SBRef"
```
mharms: Were we wanting to use symlinks rather than copies here, so that there is an indication of where the data came from?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Never mind.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,15 +457,82 @@ if [ ! -e "$fMRIFolder" ] ; then
   log_Msg "mkdir ${fMRIFolder}"
   mkdir "$fMRIFolder"
 fi
-cp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName".nii.gz
+${FSLDIR}/bin/imcp "$fMRITimeSeries" "$fMRIFolder"/"$OrigTCSName"
+
+# --- Do slice time correction if indicated
+if [ $DoSliceTimeCorrection = "TRUE" ] ; then
+    log_Msg "Running slice timing correction using FSL's 'slicetimer' tool"
+    TR=`${FSLDIR}/bin/fslval "$fMRIFolder"/"$OrigTCSName" pixdim4`
```
mharms: Would suggest modifying the log_Msg's as well:
```
log_Msg "Running slice timing correction using FSL's 'slicetimer' tool ..."
log_Msg "... $fMRIFolder/$OrigTCSName will be a slice-time-corrected version of the original data"
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -293,15 +457,82 @@ if [ ! -e "$fMRIFolder" ] ; then
...
+    if [ $reorient = "TRUE" ] ; then
+      log_Warn "Performing fslreorient2std! Please take that into account when using volume BOLD images in further analyses!"
+
+      # --- reorient BOLD
+      ${FSLDIR}/bin/immv "$fMRIFolder"/"$OrigTCSName" "$fMRIFolder"/"$OrigTCSName"_pre2std
+      ${FSLDIR}/bin/fslreorient2std "$fMRIFolder"/"$OrigTCSName"_pre2std "$fMRIFolder"/"$OrigTCSName"
+      ${FSLDIR}/bin/imrm "$fMRIFolder"/"$OrigTCSName"_pre2std
+
+      # --- reorient SCOUT
+      if [ "$fMRIReference" != "NONE" ]; then
```
mharms: Why wrap the reorienting of the SCOUT in this conditional?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -428,26 +684,44 @@ mkdir -p ${ResultsFolder}
...
+        ${FSLDIR}/bin/applywarp --rel --interp=spline --in=${DCFolder}/PhaseTwo_gdc_dc_unbias -w ${AtlasSpaceFolder}/xfms/${AtlasTransform} -r ${fMRIFolder}/${NameOffMRI}_SBRef_nonlin -o ${ResultsFolder}/${NameOffMRI}_PhaseTwo_gdc_dc
+        ${FSLDIR}/bin/fslmaths ${ResultsFolder}/${NameOffMRI}_PhaseTwo_gdc_dc -mas ${fMRIFolder}/${FreeSurferBrainMask}.${FinalfMRIResolution}.nii.gz ${ResultsFolder}/${NameOffMRI}_PhaseTwo_gdc_dc    
+    else        
+        #as these have been already computed, we can copy them from the reference bold
+        ${FSLDIR}/bin/imcp ${ReferenceResultsFolder}/${fMRIReference}_PhaseOne_gdc_dc ${ResultsFolder}/${NameOffMRI}_PhaseOne_gdc_dc
+        ${FSLDIR}/bin/imcp ${ReferenceResultsFolder}/${fMRIReference}_PhaseOne_gdc_dc ${ResultsFolder}/${NameOffMRI}_PhaseOne_gdc_dc
+        ${FSLDIR}/bin/imcp ${ReferenceResultsFolder}/${fMRIReference}_PhaseTwo_gdc_dc ${ResultsFolder}/${NameOffMRI}_PhaseTwo_gdc_dc
+        ${FSLDIR}/bin/imcp ${ReferenceResultsFolder}/${fMRIReference}_PhaseTwo_gdc_dc ${ResultsFolder}/${NameOffMRI}_PhaseTwo_gdc_dc
```
mharms: Again, should we be using symlinks here?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Yes, that's what I'm trying to say. SEBASED bias field is T1 masked so applying it effectively makes any other mask that is larger than T1 mask irrelevant.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: @mharms this is text as provided by @glasserm. I'm ok to include what is your consensus.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Then we should say that explicitly.  Also, if that is the case, isn't the whole mask option somewhat irrelevant, since SEBASED is the bias field correction mode that people should be using?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Corrected.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I moved them to line 301. I hope that is what you meant.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: I added additional explanation and explicitly referenced low **spatial** resolution.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Added.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: As I wrote before, in our experience symlinks can be an issue in interaction with XNAT uptake. Also, if for any reason the reference BOLD is deleted then the referenced image is lost. And if the reference BOLD is redone the symlink will be misleading. To be on the safe side, I would use copy and not symlinks.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: It was an error, it should be:

```
if [ "$fMRIReference" = "NONE" ]; then
```

As we only want to reorient it when we are not using a copy of the reference scout, which has already been reoriented when processing the reference bold.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Same as above.

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -43,6 +56,32 @@ case "$MotionCorrectionType" in
...
+# If registering across runs, perform nonlinear registration if requested.
+# (If using linear registration, don't need to do anything extra here, since
+# linear registration is handled implicitly via the motion correction).
+# Note that if registering across runs, the "$Scout" input to MotionCorrection will
+# be the *reference* scout image (by construction in GenericfMRIVolume).
+
+if [ "${fMRIReferenceReg}" == "nonlinear" ] ; then
+  verbose_echo " ... computing nonlinear transform to reference"
+  verbose_echo "     ... generating bold average"
+  # Generating a mean image to increase SNR to noise ratio when registering to scout.
```
mharms: "increase SNR to noise ratio" is redundant.  Either "increase SNR", or "increase signal-to-noise ratio"

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: In legacy data SE images are a rarity so for most of them there are no SEBASED bias fields. Also I'd rather make SEBASED bias field nonmasked—if possible—than remove the mask options. I have not had the time and am not familiar with the code enough to enable that. @coalsont, would that be possible?

sivcek: @mharms How would you state that more explicitly?

fMRIVolume/scripts/MotionCorrection.sh
sivcek: done.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -92,10 +94,47 @@ ScoutInput=`getopt1 "--scoutin" $@`  # "${15}"
...
+verbose_echo "             --oscout: ${ScoutOutput}"
+verbose_echo "          --ojacobian: ${JacobianOut}"
+verbose_echo "        --fmrirefpath: ${fMRIReferencePath}"
+verbose_echo "         --fmrirefreg: ${fMRIReferenceReg}"
+verbose_echo " "
 
 BiasFieldFile=`basename "$BiasField"`
 T1wImageFile=`basename $T1wImage`
 FreeSurferBrainMaskFile=`basename "$FreeSurferBrainMask"`
+ScoutName=`basename "$ScoutInputgdc"`
```
mharms: Why was this variable added?  i.e., where does it ever get used?

sivcek: It was used in previous iteration. Not used and removed now.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: Not sure, @glasserm is the one who came up with the idea of the SEBASED bias field.  It basically compares the gradient echo and spin echo images, as I recall.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -209,23 +257,29 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
...
+  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/postmc2struct_warp --warp1=${GradientDistortionField} --warp2=${WD}/postmc2struct_warp -o ${WD}/gdc_dc_warp --jacobian=${WD}/gdc_dc_jacobian
+else
+  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${fMRIToStructuralInput} --warp1=${GradientDistortionField} --warp2=${fMRIToStructuralInput} -o ${WD}/gdc_dc_warp --jacobian=${WD}/gdc_dc_jacobian
+fi
 #but, convertwarp's jacobian is 8 frames - each combination of one-sided differences, so average them
 ${FSLDIR}/bin/fslmaths ${WD}/gdc_dc_jacobian -Tmean ${WD}/gdc_dc_jacobian
 
 #and resample it to MNI space
-${FSLDIR}/bin/applywarp --rel --interp=spline -i ${WD}/gdc_dc_jacobian -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -w ${StructuralToStandard} -o ${JacobianOut}
+${FSLDIR}/bin/applywarp --rel --interp=trilinear -i ${WD}/gdc_dc_jacobian -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -w ${StructuralToStandard} -o ${JacobianOut}
```
mharms: Is the change from spline to trilinear interpolation intentional here?

coalsont: As I recall, there were ringing artifacts in the jacobian when using spline resampling, so this change was intentional.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -358,64 +372,76 @@ case $DistortionCorrection in
...
         # this is just an initial registration, refined later in this script, but it is actually pretty good
         log_Msg "register undistorted scout image to T1w"
-        ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+
+        if [ $PreregisterTool = "epi_reg" ] ; then
+          log_Msg "... running epi_reg (dof ${dof})"
+          ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/${ScoutInputFile}_undistorted --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}_undistorted2T1w_init
+        elif [ $PreregisterTool = "flirt" ] ; then
+          log_Msg "... running flirt"
+          ${FSLDIR}/bin/flirt -in ${WD}/${ScoutInputFile}_undistorted -ref ${WD}/${T1wBrainImageFile} -out ${WD}/${ScoutInputFile}_undistorted2T1w_init -omat ${WD}/${ScoutInputFile}_undistorted2T1w_init.mat -dof ${dof}
```
mharms: Shouldn't the code through here be using `${ScoutExtension}` rather than `_undistorted`, least someone actually change the `ScoutExtension` definition earlier in the code?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: Well there are problems with both approaches (which is why there is no consensus), so the point of the warning is that slice timing correction as currently implemented is problematic.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -430,14 +456,54 @@ case $DistortionCorrection in
...
 
         ;;
 
+    ${NONE_METHOD_OPT})
+
+            # NOTE: To work with later code a number of fake files are created 
+            #       with names that do not reflect the true nature of the file
+            #       this is to be considered an initial fix for datasets that 
+            #       do not include data that enable distortion correction and
+            #       will be replaced with a cleaner code.
```
mharms: If if don't create the proposed "cleaner code" now, it is very unlikely to get done in the future.  I'd still like to see that get done in this PR.  Or, is this comment no longer entirely accurate...?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: The last couple of lines of the script produce mildly dilated bias files.  One could dilate to fill the files with fslmaths -dilall.

fMRIVolume/scripts/OneStepResampling.sh
glasserm: Right

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: I'm confused here.  `BOLDMask` only seems to get used in `IntensityNormalization`.  So what exactly is the relationship of that to the SEBASED bias field correction?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: My point is that it isn't accurate to say "(as is typically done in legacy-style brain imaging)", and we should just drop that parenthetical.  The entire rest of the warning is still fine as is.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: I don't think we've run into a problem in using symlinks in other scripts, but ok.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+  if [ ! -d "$ReferenceResultsFolder" ] ; then
+    log_Err_Abort "Reference results folder does not exist and can not be used (${ReferenceResultsFolder})!"
+  fi 
+
+  # print warning
+
+  log_Warn "You are using an external reference (--fmriref=${fMRIReference}) for motion registration and"
+  log_Warn "  distortion correction and registration to T1w image. Please consider using this option only"
+  log_Warn "  in cases when only one BOLD Reference image is available or when processing low spatial"
+  log_Warn "  resolution legacy BOLD images (as in this cases within BOLD registration might be more robust"
```
mharms: Minor grammatical tweak:
`log_Warn " resolution legacy BOLD images (for which within BOLD registration might be more robust"`

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,171 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+    log_Err_Abort "Reference results folder does not exist and can not be used (${ReferenceResultsFolder})!"
+  fi 
+
+  # print warning
+
+  log_Warn "You are using an external reference (--fmriref=${fMRIReference}) for motion registration and"
+  log_Warn "  distortion correction and registration to T1w image. Please consider using this option only"
+  log_Warn "  in cases when only one BOLD Reference image is available or when processing low spatial"
+  log_Warn "  resolution legacy BOLD images (as in this cases within BOLD registration might be more robust" 
+  log_Warn "  than independent registration to structural images. Please make sure that the reference BOLD"
```
mharms: Missing closing `)`:
`log_Warn "  than independent registration to structural images). Please make sure that the reference BOLD"`

fMRIVolume/scripts/OneStepResampling.sh
mharms: Right.  @sivcek Please add a comment explaining why it is set to trilinear, possibly even including the URL to the GitHub discussion of the issue.  Thx.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: Symlinking across different BOLD directories doesn't sound like a good idea to me, too many potential surprises when reprocessing one run and not another.  The other symlinks I know of in the pipelines don't cross between runs.

fMRIVolume/scripts/OneStepResampling.sh
sivcek: I added:
```
#and resample it to MNI space
# Note that trilinear instead of spline interpolation is used with the purpose to minimize the ringing artefacts that occur 
# with downsampling of the jacobian field and are then propagated to the BOLD image itself.
```
The discussion occured over email and can not be linked to.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Indeed. I have changed all the `_undistorted` with `${ScoutExtension}`.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
sivcek: Indeed the code has been cleaned. The only "Fake" image is the uniform Jacobian that is created here. I have changed the note to:

```
# NOTE: To work with later code a uniform Jacobian is created.
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Done.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Corrected

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Ok. I removed te parenthetical.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: @glasserm would you be ok with me changing the two last lines 87, 88 of `ComputeSpinEchoBiasField.sh` to `-dilall`?

sivcek: @mharms in line 189 in Intensity normalization:
```
${FSLDIR}/bin/fslmaths ${InputfMRI} $biascom $jacobiancom -mas ${FinalMask} -thr 0 -ing 10000 ${OutputfMRI} -odt float
```

`$biascom` expands to `biascom="-div $BiasField"` (line 121). If the BiasFiled is masked, then the mask that was used for the `$BiasField` is propagated to the BOLD image as anything that is outside of the masked `$BiasField` will be set to 0.

glasserm: That would not really have the expected effect as that small dilation is just for a trilinear resampling that is immediately masked again.  One would need to change how that remasking is done here: in fMRIVolume.sh lines 426-452

sivcek: @glasserm Then it might be simplest to just do it at the very end in `IntensityNormalisation.sh` within the `if` check (line 121). In this case the biasfield would be dilated just before being used.

glasserm: Yes you could do some dilation there if needed.

sivcek: Ok. I'll add:

```
biascom=""
if [[ "$BiasField" != "" ]]
then
    ${FSLDIR}/bin/fslmaths ${BiasField} -dilall ${BiasField}_dilated
    biascom="-div ${BiasField}_dilated"
fi
```

sivcek: And I removed the warning note about mask use, which is not necessary anymore.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,166 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+
+  # set reference and check if external reference (if one is specified) exists 
+
+  fMRIReferencePath="$Path"/"$Subject"/"$fMRIReference"
+  log_Msg "Using reference image from ${fMRIReferencePath}"
+  fMRIReferenceImage="$fMRIReferencePath"/"$ScoutName"_gdc
+  fMRIReferenceImageMask="$fMRIReferencePath"/"$ScoutName"_gdc_mask
+  ReferenceResultsFolder="$Path"/"$Subject"/"$AtlasSpaceFolder"/"$ResultsFolder"/"$fMRIReference"
+
+  if [ "$fMRIReferencePath" = "$fMRIFolder" ] ; then
```
mharms: Given where we ended up, we should make clear in the comments above the `--fmriscout` and `--fmriref` are exclusionary options, and I think we should add an abort if both of those are set to something (other than NONE), since in that case the user is clearly confused about the distinction between the two options.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +257,166 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+  log_Warn "  to gold-standard fieldmap-based correction."
+fi
+
+# -- Slice timing correction
+
+if [ "${DoSliceTimeCorrection}" = 'TRUE' ]; then
+  ComplianceMsg+=" --doslicetime=TRUE --slicetimerparams=${SliceTimerCorrectionParameters}"
+  Compliance="LegacyStyleData"
+  log_Warn "This LegacyStyleData option of slice timing correction is performed before motion"
+  log_Warn "  correction (as is typically done in legacy-style brain imaging) and thus assumes that the"
```
mharms: Keep the `log_Warn` message in sync with the changes made to the comments above.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -209,23 +255,31 @@ ${FSLDIR}/bin/fslmerge -tr ${OutputfMRI}_mask $FrameMergeSTRINGII $TR_vol
...
-# Combine transformations: gradient non-linearity distortion + fMRI_dc to standard
-${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${GradientDistortionField} --warp2=${OutputTransform} --out=${WD}/Scout_gdc_MNI_warp.nii.gz
-${FSLDIR}/bin/applywarp --rel --interp=spline --in=${ScoutInput} -w ${WD}/Scout_gdc_MNI_warp.nii.gz -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -o ${ScoutOutput}
+if [ ${fMRIReferencePath} = "NONE" ] ; then
+  # Combine transformations: gradient non-linearity distortion + fMRI_dc to standard
+  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${GradientDistortionField} --warp2=${OutputTransform} --out=${WD}/Scout_gdc_MNI_warp.nii.gz
+  ${FSLDIR}/bin/applywarp --rel --interp=spline --in=${ScoutInput} -w ${WD}/Scout_gdc_MNI_warp.nii.gz -r ${WD}/${T1wImageFile}.${FinalfMRIResolution} -o ${ScoutOutput}
+fi
 
 # Create spline interpolated version of Jacobian  (T1w space, fMRI resolution)
```
mharms: Update comment here to 'trilinear' rather than 'spline'.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: Do you feel strongly about aborting or would it be enough to provide the following warning for the `--fmriref` in the code and then a similar logged warning if both `--fmriref` and `--fmriscout` are provided:

```
# WARNING: This option excludes tbe use of --fmriscout option, as the scout from the specified reference BOLD is used
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
sivcek: done

fMRIVolume/scripts/OneStepResampling.sh
sivcek: done

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Yes, I feel fairly strongly that should be an abort condition.  While the warnings are well and good, the log files are very verbose, and I doubt that most users will ever notice the warnings in them.  If a user specifies both `--fmriref` and `--fmriscout` within the current code, then they are clearly confused as to what those options represent, and I feel that warrants an abort so that they can make sure they are running the code as they intended.

sivcek: Ok. I'll implement it that way then. In Qu|Nex the omission of the `--fmriscout` when generating the call will be handled automatically for the user, though, as what user wants is specified on a higher level. Qu|Nex will specify `--fmriscout` when calling the pipelines for the run used as the reference run and not for the others that then use it as a reference.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mharms: Do we want to keep this `${BiasField}_dilated` file around once we are done with it?

sivcek: Well, it is the record of what was used. But I'll refer the decision to your team. I don't have a clear idea or guidance on what needs to be retained and what not.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -257,6 +261,179 @@ sebasedBiasFieldMNI="$SubjectFolder/$AtlasSpaceFolder/Results/$NameOffMRI/${Name
...
+
+fMRIReferenceReg=`opts_GetOpt1 "--fmrirefreg" $@`                        # In the cases when BOLD image is registered to a specified BOLD reference, this option 
+                                                                         #   specifies whether to use 'linear' or 'nonlinear' registration to reference BOLD.
+                                                                         #   Default is 'linear'.
+
+# Defaults
+PreregisterTool=`opts_DefaultOpt $PreregisterTool "epi_reg"`
+DoSliceTimeCorrection=`opts_DefaultOpt $DoSliceTimeCorrection "FALSE"`   
+fMRIReference=`opts_DefaultOpt $fMRIReference "NONE"`
+fMRIReference=`opts_DefaultOpt $fMRIReference "NONE"`
```
coalsont: This line appears to be duplicated.

sivcek: fixed

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: It would be fine to keep it.

sivcek: If that's the case, then the PR is ready for pulling in.

glasserm:
Mike asked for a usage statement, are we waiting on that?

sivcek:
I thought we'll add it after the PR is pulled in.

PR #157, opened 2019-11-08
Update hcp_fix to support the MotionRegression=FALSE condition

### LINKED ISSUE #108: hcp_fix doesn't apply any hp filter to CIFTI if do_motion_regression=FALSE
Restarted from #107.

**Background**: In the v4.0.0 HCPpipelines version of `ICAFIX/hcp_fix`, we added the capability to control whether the motion parameters were regressed out of the data as part of the fix-cleaning (via the `do_motion_regression` argument).  [Prior to that (e.g., v3.27.0 and earlier), regression of the motion parameters was always applied].

FIX was originally written for volume (NIFTI) data, with the assumption that any high-pass filtering was applied _prior_ to FIX itself.  Thus, FIX originally only needed to worry about applying the same hp filter to the motion regressors, and thus in the `fix` script, the `-h` argument (to specify the filter) was made a subargument to the `-m` flag (that instructs `fix` to apply motion regression).

When the CIFTI format came along, the hp filtering of the CIFTI dtseries data was implemented in the context of `fix_3_clean` (called by the `fix` script to implement the actual cleaning of the data).  But the `-h` argument in `fix` stayed a subargument to the `-m` flag.

**Problem**:  While `fix_3_clean` has the capability to  independently apply motion regression and an hp filter to the CIFTI dtseries, the `fix` script itself does not support that -- unless one requests motion regression, there is no mechanism in `fix` itself for filtering the dtseries.

Consequently, if running the v4.0.0 version of `hcp_fix` with something like the following
`hcp_fix <data> 2000 FALSE`
the volume (NIFTI) data will be filtered by the requested hp=2000 filter (since that is done in `hcp_fix` script itself), but the CIFTI dtseries data will not get filtered at all (due to the setting of `FALSE` for the `do_motion_regression` argument).  Insidiously in this situation, a cleaned CIFTI file will still be generated, and the file name will in fact be labeled `hp2000_clean.dtseries.nii`, even though an hp=2000 filter wasn't actually applied to it!!

**Suggested solution:**  Split `-m` and `-h` into independent arguments in the `fix` script.  (Ideally, in the FIX distribution itself).

mharms:
With the forthcoming FIX version "1.06.12", we can now properly implement HP filtering in the MotionRegression=FALSE condition within single-run FIX.  This PR checks for the FIX version, and calls `fix` appropriately based on the FIX version.

This should resolve #108.

### DIFF
--- a/ICAFIX/hcp_fix
+++ b/ICAFIX/hcp_fix
@@ -69,10 +69,14 @@ fi
 
 #############################################################
 
-# Need at least fix version 1.06.12 for proper handling of the do_motion_regression=FALSE condition
+# Function to return whether FIX version is "NEW" (i.e., 1.06.12 or later).
+# Need at least that fix version for proper handling of the
+# do_motion_regression=FALSE condition (when highpass>=0)
+
 # NOTE:
 #   Don't echo anything in this function other than the last echo
 #   that outputs the return value
+
 determine_old_or_new_fix()
 {
 	# Default to "OLD", and change to "NEW" only if version is 1.06.12 or later
--- a/ICAFIX/hcp_fix
+++ b/ICAFIX/hcp_fix
@@ -69,6 +69,8 @@ fi
 
 #############################################################
 
+### Support Functions
+
 # Function to return whether FIX version is "NEW" (i.e., 1.06.12 or later).
 # Need at least that fix version for proper handling of the
 # do_motion_regression=FALSE condition (when highpass>=0)
@@ -76,7 +78,7 @@ fi
 # NOTE:
 #   Don't echo anything in this function other than the last echo
 #   that outputs the return value
-
+#
 determine_old_or_new_fix()
 {
 	# Default to "OLD", and change to "NEW" only if version is 1.06.12 or later
@@ -118,6 +120,21 @@ determine_old_or_new_fix()
 	fi
 
 	echo ${old_or_new}
+}
+
+function interpret_as_bool()
+{
+    case $(echo "$1" | tr '[:upper:]' '[:lower:]') in
+    (true | yes | 1)
+        echo TRUE
+        ;;
+    (false | no | none | 0)
+        echo FALSE
+        ;;
+    (*)
+        log_Err_Abort "error: '$1' is not valid for this argument, please use TRUE or FALSE"
+        ;;
+    esac
 }
 
 #############################################################
@@ -170,6 +187,10 @@ log_Msg "FSL_FIX_MATLAB_MODE: ${FSL_FIX_MATLAB_MODE}"
 
 #############################################################
 
+## ---------------------------------------------------------------------------
+## Parse and check parameters
+## ---------------------------------------------------------------------------
+
 fmri=$1
 cd `dirname $fmri`
 fmri=`basename $fmri`  # After this, $fmri no longer includes the leading directory components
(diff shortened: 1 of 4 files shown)


ICAFIX/hcp_fix
```diff
@@ -337,12 +401,24 @@ fi
...
+		fix_cmd=("${FSL_FIXDIR}/fix" "${fmrihp}.ica" "${TrainingData}" "${FixThresh}" -h "${hp}")
+
+	elif (( hp < 0 )); then
+		# Ok to proceed regardless of FIX version, since not filtering, so we don't need the filtering of the CIFTI
+		# that occurs in fix_3_clean. Simply don't supply either the -m or -h flags.
+		fix_cmd=("${FSL_FIXDIR}/fix" "${fmrihp}.ica" "${TrainingData}" "${FixThresh}")
+
+	else  # "OLD" and (hp >= 0)
+		# Already aborted in this case earlier, but include again here for completeness
+		log_Err_Abort "Due to a restriction in the FIX argument parsing, do_motion_regression=$3 is not supported for FIX versions prior to 1.06.12"
```
coalsont: We will already abort when using old fix and `hp` is enabled, so we don't need to check old or new again here.  If we test only `hp`, we can issue calls that will work in all supported cases, and even if we didn't correctly abort in advance, the correct call for new fix would simply fail for old fix anyway.

mharms: I did it that way so that we could parallel the `doMotionRegression=TRUE` case and still pass `-h -1` as an argument to fix if it is the "NEW" fix version and hp<0.  Similarly, in the MR-FIX script, we pass `-m -h -1` to fix in the `doMotionRegression=TRUE` case rather than just `-m`.

mharms: i.e., this way keeps the calls to fix identical to what we were using previously (which seems preferable to me), in which case we need to add a call specific to the NEW fix version to allow `-h` without `-m`.  The repeated abort is indeed redundant, but I felt an explanation of what occurs in that situation is needed anyway given that the abort of that condition happens much earlier in the code.

mharms: And, to be entirely technical about it `-m` (by itself) and `-m -h -1` aren't strictly the same thing, since `fix` itself checks for a design.fsf file if `-h` isn't specified, and if it finds a design.fsf it then pulls the filter value out of that.  So, the way I implemented it _guarantees_ 100% backwards compatibility without any potential weird gotchas.

glasserm:
Has this been tested using the new options?

mharms:
Yes, both hp=-1 and hp=2000 run to completion with doMotionRegression=FALSE.

glasserm:
Does hp=0 work and do a detrend?

mharms:
No, that's a completely separate issue.  Support for hp=0 was not added, so the script still aborts if hp=0 is requested.

glasserm:
Is that worth doing too, or a bigger job?

mharms:
It's worth doing, but it is also a bigger job that I'm not going to tackle right now.  We would either need to: (1) Implement an `fsl_glm` detrending of the volume data in the hp=0 condition, or (2) accomplish the filtering using `functionhighpassandvariancenormalize` (i.e., as we do in `hcp_fix_multi_run`).

PR #158, opened 2019-11-13
Don't use motion-corrupted b0's in topup

MichielCottaar:
Here is a proposed piece of code to identify the bad b0's. The new code below extracts all b0's and runs topup (with the configuration file set for quick convergence). After topup each b0 is then scored based on its average square residual from the average b0.

The next step is to decide how to use these b0's in the later analysis. My proposal is to:
1. Identify the acquisition block with the highest quality initial b0. Merge all the data starting with that acquisition block (rather than starting with whatever acquisition block happened to be the first positive one that the user provided).
2. Merge the first b0 of the first block with the best b0 out of the corresponding block with opposite phase encoding. Run topup on just these two. Using two b0's should be fine as we already confirmed they are good quality. The advantage of not including b0's from the other blocks is that we get the susceptibility field appropriate for the first b0. Eddy's susceptibility-by-motion can then take care of any changes in the susceptibility field in the other acquisition blocks.
3. Run eddy as normal
4. Optionally, revert the reordering done during step 1. Personally, I don't feel this is necessary as the bvecs will be different across all subjects anyway (due to different head rotations).

Feedback on this proposal would be very welcome, before I spend too much time trying to implement it.

### DIFF
--- a/DiffusionPreprocessing/DiffPreprocPipeline.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline.sh
@@ -91,14 +91,12 @@
 #
 #~ND~END~
 
-
 # --------------------------------------------------------------------------------
 #  Usage Description Function
 # --------------------------------------------------------------------------------
 
-show_usage()
-{
-	cat << EOF
+show_usage() {
+	cat <<EOF
 
 Perform the steps of the HCP Diffusion Preprocessing Pipeline
 
@@ -244,8 +242,7 @@ EOF
 #  Support Functions
 # --------------------------------------------------------------------------------
 
-get_options()
-{
+get_options() {
 	local arguments=($@)
 
 	# initialize global output variables
@@ -269,127 +266,127 @@ get_options()
 	local numArgs=${#arguments[@]}
 	local argument
 
-	while [ ${index} -lt ${numArgs} ] ; do
+	while [ ${index} -lt ${numArgs} ]; do
 		argument=${arguments[index]}
 
 		case ${argument} in
-			--help)
-				show_usage
-				exit 0
-				;;
-			--version)
-				version_show $@
-				exit 0
-				;;
-			--path=*)
-				StudyFolder=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--subject=*)
-				Subject=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--PEdir=*)
-				PEdir=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--posData=*)
-				PosInputImages=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--negData=*)
-				NegInputImages=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--echospacing=*)
-				echospacing=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--gdcoeffs=*)
-				GdCoeffs=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--dwiname=*)
-				DWIName=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--dof=*)
-				DegreesOfFreedom=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--b0maxbval=*)
-				b0maxbval=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			--printcom=*)
-				runcmd=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-		  --select-best-b0)
-		    SelectBestB0="true"
-				index=$(( index + 1 ))
-				;;
-			--extra-eddy-arg=*)
-				extra_eddy_arg=${argument#*=}
-				extra_eddy_args+=" ${extra_eddy_arg} "
-				index=$(( index + 1 ))
-				;;
-			--combine-data-flag=*)
-				CombineDataFlag=${argument#*=}
-				index=$(( index + 1 ))
-				;;
-			*)
-				show_usage
-				echo "ERROR: Unrecognized Option: ${argument}"
-				exit 1
-				;;
+		--help)
+			show_usage
+			exit 0
+			;;
+		--version)
+			version_show "$@"
+			exit 0
+			;;
+		--path=*)
+			StudyFolder=${argument#*=}
+			index=$((index + 1))
+			;;
+		--subject=*)
+			Subject=${argument#*=}
+			index=$((index + 1))
+			;;
+		--PEdir=*)
+			PEdir=${argument#*=}
+			index=$((index + 1))
+			;;
+		--posData=*)
+			PosInputImages=${argument#*=}
+			index=$((index + 1))
+			;;
+		--negData=*)
+			NegInputImages=${argument#*=}
+			index=$((index + 1))
+			;;
+		--echospacing=*)
+			echospacing=${argument#*=}
+			index=$((index + 1))
+			;;
+		--gdcoeffs=*)
+			GdCoeffs=${argument#*=}
+			index=$((index + 1))
+			;;
+		--dwiname=*)
+			DWIName=${argument#*=}
+			index=$((index + 1))
+			;;
+		--dof=*)
+			DegreesOfFreedom=${argument#*=}
+			index=$((index + 1))
+			;;
+		--b0maxbval=*)
+			b0maxbval=${argument#*=}
+			index=$((index + 1))
+			;;
+		--printcom=*)
+			runcmd=${argument#*=}
+			index=$((index + 1))
+			;;
+		--select-best-b0)
+			SelectBestB0="true"
+			index=$((index + 1))
+			;;
+		--extra-eddy-arg=*)
+			extra_eddy_arg=${argument#*=}
+			extra_eddy_args+=" ${extra_eddy_arg} "
+			index=$((index + 1))
+			;;
+		--combine-data-flag=*)
+			CombineDataFlag=${argument#*=}
+			index=$((index + 1))
+			;;
+		*)
+			show_usage
+			echo "ERROR: Unrecognized Option: ${argument}"
+			exit 1
+			;;
 		esac
 	done
 
 	local error_msgs=""
 
 	# check required parameters
-	if [ -z ${StudyFolder} ] ; then
+	if [ -z ${StudyFolder} ]; then
 		error_msgs+="\nERROR: <study-path> not specified"
 	fi
 
-	if [ -z ${Subject} ] ; then
+	if [ -z ${Subject} ]; then
 		error_msgs+="\nERROR: <subject-id> not specified"
 	fi
 
-	if [ -z ${PEdir} ] ; then
+	if [ -z ${PEdir} ]; then
 		error_msgs+="\nERROR: <phase-encoding-dir> not specified"
 	fi
 
-	if [ -z ${PosInputImages} ] ; then
+	if [ -z ${PosInputImages} ]; then
 		error_msgs+="\nERROR: <positive-phase-encoded-data> not specified"
 	fi
 
-	if [ -z ${NegInputImages} ] ; then
+	if [ -z ${NegInputImages} ]; then
 		error_msgs+="\nERROR: <negative-phase-encoded-data> not specified"
 	fi
 
-	if [ -z ${echospacing} ] ; then
+	if [ -z ${echospacing} ]; then
 		error_msgs+="\nERROR: <echo-spacing> not specified"
 	fi
 
-	if [ -z ${GdCoeffs} ] ; then
+	if [ -z ${GdCoeffs} ]; then
 		error_msgs+="\nERROR: <path-to-gradients-coefficients-file> not specified"
 	fi
 
-	if [ -z ${b0maxbval} ] ; then
+	if [ -z ${b0maxbval} ]; then
 		error_msgs+="\nERROR: <b0-max-bval> not specified"
 	fi
 
-	if [ -z ${DWIName} ] ; then
+	if [ -z ${DWIName} ]; then
 		error_msgs+="\nERROR: <DWIName> not specified"
 	fi
 
-	if [ -z ${CombineDataFlag} ] ; then
+	if [ -z ${CombineDataFlag} ]; then
 		error_msgs+="\nERROR: CombineDataFlag not specified"
 	fi
 
-	if [ ! -z "${error_msgs}" ] ; then
+	if [ ! -z "${error_msgs}" ]; then
 		show_usage
 		echo -e ${error_msgs}
 		echo ""
@@ -410,51 +407,50 @@ get_options()
 	echo "   b0maxbval: ${b0maxbval}"
 	echo "   runcmd: ${runcmd}"
 	echo "   CombineDataFlag: ${CombineDataFlag}"
-	if [ ! -z ${SelectBestB0} ] ; then
-    echo "   SelectBestB0: ${SelectBestB0}"
-  fi
+	if [ ! -z ${SelectBestB0} ]; then
+		echo "   SelectBestB0: ${SelectBestB0}"
+	fi
 	echo "   extra_eddy_args: ${extra_eddy_args}"
 	echo "-- ${g_script_name}: Specified Command-Line Parameters - End --"
 
-	if [ ! -z "${SelectBestB0}" ] ; then
-	  fwhm_set=false
-    if [ ! -z "${extra_eddy_args}" ] ; then
-      for extra_eddy_arg in ${extra_eddy_args} ; do
-        if [[ ${extra_eddy_arg} == "--fwhm"* ]] ; then
-          fwhm_set=true
-        fi
-      done
-    fi
-    if [ ${fwhm_set} == false ] ; then
-      log_Warn "WARNING: Using --select-best-b0 prepends the best b0 to the start of the file passed into eddy"
-      log_Warn "         To ensure eddy succesfully aligns this new first b0 with the actual first volume,"
-      log_Warn "         we recommend to increase the FWHM for the first eddy iterations if using --select-best-b0"
-      log_Warn "         This can be done by setting the --extra_eddy_args=--fwhm=... flag"
-    fi
-  fi
+	if [ ! -z "${SelectBestB0}" ]; then
+		fwhm_set=false
+		if [ ! -z "${extra_eddy_args}" ]; then
+			for extra_eddy_arg in ${extra_eddy_args}; do
+				if [[ ${extra_eddy_arg} == "--fwhm"* ]]; then
+					fwhm_set=true
+				fi
+			done
+		fi
+		if [ ${fwhm_set} == false ]; then
+			log_Warn "WARNING: Using --select-best-b0 prepends the best b0 to the start of the file passed into eddy"
+			log_Warn "         To ensure eddy succesfully aligns this new first b0 with the actual first volume,"
+			log_Warn "         we recommend to increase the FWHM for the first eddy iterations if using --select-best-b0"
+			log_Warn "         This can be done by setting the --extra_eddy_args=--fwhm=... flag"
+		fi
+	fi
 }
 
 #
 # Function Description
 #  Validate necessary scripts exist
 #
-validate_scripts()
-{
+validate_scripts() {
 	local error_msgs=""
 
-	if [ ! -e ${HCPPIPEDIR}/DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh ] ; then
+	if [ ! -e ${HCPPIPEDIR}/DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh ]; then
 		error_msgs+="\nERROR: HCPPIPEDIR/DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh not found"
 	fi
 
-	if [ ! -e ${HCPPIPEDIR}/DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh ] ; then
+	if [ ! -e ${HCPPIPEDIR}/DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh ]; then
 		error_msgs+="\nERROR: HCPPIPEDIR/DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh not found"
 	fi
 
-	if [ ! -e ${HCPPIPEDIR}/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh ] ; then
+	if [ ! -e ${HCPPIPEDIR}/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh ]; then
 		error_msgs+="\nERROR: HCPPIPEDIR/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh not found"
 	fi
 
-	if [ ! -z "${error_msgs}" ] ; then
+	if [ ! -z "${error_msgs}" ]; then
 		show_usage
 		echo -e ${error_msgs}
 		echo ""
@@ -466,8 +462,7 @@ validate_scripts()
 # Function Description
 #  Main processing of script
 #
-main()
-{
+main() {
 	# Get Command Line Options
 	get_options "$@"
 
@@ -486,9 +481,9 @@ main()
 	pre_eddy_cmd+=" --echospacing=${echospacing} "
 	pre_eddy_cmd+=" --b0maxbval=${b0maxbval} "
 	pre_eddy_cmd+=" --printcom=${runcmd} "
-	if [ ! -z "${SelectBestB0}" ] ; then
-	  pre_eddy_cmd+=" --select-best-b0 "
-  fi
+	if [ ! -z "${SelectBestB0}" ]; then
+		pre_eddy_cmd+=" --select-best-b0 "
+	fi
 
 	log_Msg "pre_eddy_cmd: ${pre_eddy_cmd}"
 	${pre_eddy_cmd}
@@ -501,8 +496,8 @@ main()
 	eddy_cmd+=" --dwiname=${DWIName} "
 	eddy_cmd+=" --printcom=${runcmd} "
 
-	if [ ! -z "${extra_eddy_args}" ] ; then
-		for extra_eddy_arg in ${extra_eddy_args} ; do
+	if [ ! -z "${extra_eddy_args}" ]; then
+		for extra_eddy_arg in ${extra_eddy_args}; do
 			eddy_cmd+=" --extra-eddy-arg=${extra_eddy_arg} "
 		done
 	fi
@@ -520,9 +515,9 @@ main()
 	post_eddy_cmd+=" --dof=${DegreesOfFreedom} "
 	post_eddy_cmd+=" --combine-data-flag=${CombineDataFlag} "
 	post_eddy_cmd+=" --printcom=${runcmd} "
-	if [ ! -z "${SelectBestB0}" ] ; then
-	  post_eddy_cmd+=" --select-best-b0 "
-  fi
+	if [ ! -z "${SelectBestB0}" ]; then
+		post_eddy_cmd+=" --select-best-b0 "
+	fi
 
 	log_Msg "post_eddy_cmd: ${post_eddy_cmd}"
 	${post_eddy_cmd}
@@ -544,8 +539,8 @@ g_script_name=$(basename "${0}")
 
 # Allow script to return a Usage statement, before any other output
 if [ "$#" = "0" ]; then
-    show_usage
-    exit 1
+	show_usage
+	exit 1
 fi
 
 # Verify that HCPPIPEDIR Environment variable is set
@@ -555,13 +550,13 @@ if [ -z "${HCPPIPEDIR}" ]; then
 fi
 
 # Load function libraries
-source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
-source ${HCPPIPEDIR}/global/scripts/opts.shlib                 # Command line option functions
-source ${HCPPIPEDIR}/global/scripts/version.shlib	           # version_ functions
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions; also sources log.shlib
+source ${HCPPIPEDIR}/global/scripts/opts.shlib         # Command line option functions
+source ${HCPPIPEDIR}/global/scripts/version.shlib      # version_ functions
 
-opts_ShowVersionIfRequested $@
+opts_ShowVersionIfRequested "$@"
 
-if opts_CheckForHelpRequest $@; then
+if opts_CheckForHelpRequest "$@"; then
 	show_usage
 	exit 0
 fi
@@ -575,5 +570,4 @@ log_Check_Env_Var FSLDIR
 #
 # Invoke the 'main' function to get things started
 #
-main $@
-
+main "$@"
--- a/DiffusionPreprocessing/DiffPreprocPipeline.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline.sh
@@ -426,10 +426,10 @@ get_options()
       done
     fi
     if [ ${fwhm_set} == false ] ; then
-      echo ""
-      echo "WARNING: Please increase the FWHM for the first few iterations if using --select-best-b0"
-      echo "         This can be done by setting the --extra_eddy_args=--fwhm=... flag"
-      echo ""
+      log_Warn "WARNING: Using --select-best-b0 prepends the best b0 to the start of the file passed into eddy"
+      log_Warn "         To ensure eddy succesfully aligns this new first b0 with the actual first volume,"
+      log_Warn "         we recommend to increase the FWHM for the first eddy iterations if using --select-best-b0"
+      log_Warn "         This can be done by setting the --extra_eddy_args=--fwhm=... flag"
     fi
   fi
 }
--- a/DiffusionPreprocessing/DiffPreprocPipeline.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline.sh
@@ -410,13 +410,13 @@ get_options()
 	echo "   b0maxbval: ${b0maxbval}"
 	echo "   runcmd: ${runcmd}"
 	echo "   CombineDataFlag: ${CombineDataFlag}"
-	if [ -z ${SelectBestB0} ] ; then
+	if [ ! -z ${SelectBestB0} ] ; then
     echo "   SelectBestB0: ${SelectBestB0}"
   fi
 	echo "   extra_eddy_args: ${extra_eddy_args}"
 	echo "-- ${g_script_name}: Specified Command-Line Parameters - End --"
 
-	if [ -z "${SelectBestB0}" ] ; then
+	if [ ! -z "${SelectBestB0}" ] ; then
 	  fwhm_set=false
     if [ ! -z "${extra_eddy_args}" ] ; then
       for extra_eddy_arg in ${extra_eddy_args} ; do
@@ -486,7 +486,7 @@ main()
 	pre_eddy_cmd+=" --echospacing=${echospacing} "
 	pre_eddy_cmd+=" --b0maxbval=${b0maxbval} "
 	pre_eddy_cmd+=" --printcom=${runcmd} "
-	if [ -z "${SelectBestB0}" ] ; then
+	if [ ! -z "${SelectBestB0}" ] ; then
 	  pre_eddy_cmd+=" --select-best-b0 "
   fi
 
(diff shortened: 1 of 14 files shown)


glasserm:
I am not following, are we using the first b0 or the best b0 in topup for both phase encoding directions?

MichielCottaar:
Let's distinguish the reference b0, which is the first volume passed on to both topup and eddy, from the other b0. We would choose as reference b0 be the first b0 of some acquisition block (i.e., the first b0 of 98AP, 98PA, 99AP, or 99PA), so that we don't have to reorder the data. My proposal is to select this reference b0 to be the highest-scoring one from these 4 options.

Once we have selected a reference b0, we can select any b0 with the opposite phase encoding to be the other b0. So for this one, I would propose to select the best b0 even if it is not the first.

glasserm:
ok thta makes sense.

mharms:
Thanks @MichielCottaar for kicking this off.
Some comments, in no particular order:

1) You wrote:

>     2\. The advantage of not including b0's from the other blocks is that we get the susceptibility field appropriate for the first b0

That's only really true to the extent that the subject didn't move between the "reference b0" and the 2nd b0 of opposite polarity that gets selected for inclusion in 'topup'.  And that can be a long interval under the proposal.  The optimal situation for SBM would be to preferentially select the b0 from the _end_ of one run, and the beginning of the following run with opposite polarity, assuming that both are of "good" quality.  Does that impact your thinking on which b0's to preferentially select?  Alternatively, given that the proposal isn't really optimal for SBM considerations, would we be better off continuing with the existing approach of supplying multiple b0's to 'topup' as a way to help ensure that the topup field estimate is robust? (i.e., supply all the same b0's to topup that we currently do, except simply remove those that are deemed "bad").

2) We should think beyond the specific HCP-LS acquisition protocol to what would also be optimal for other acquisition protocols -- e.g., protocols that acquire a short scan with a couple b0s of only one polarity, and then just a single dMRI scan of the opposite polarity (e.g., UK-Biobank); or protocols which acquire only two runs.  In either case, if we _do_ go with just a single pair of b0's as input to topup, then limiting the selection of the reference b0 to the "best" one that happens to also be first in the run seems overly restrictive (and would likely result in cases where we still end up with a "bad" reference b0).  Thus, I'd suggest that we don't necessarily limit the reference b0 to being the first frame in the run.  I understand that complicates things a bit, because the reference b0 then need to be added to the front of the input to 'eddy' (and then possibly removed again when done), but it would be manageable right?  (Alternatively, rather than 'eddy' automatically assuming that the first frame is the reference, perhaps an argument could be added that allows one to specify an arbitrary frame as the reference?  Which of course, even if Jesper was willing, has its own negatives, in that we then need a new version of 'eddy' for that particular feature)

3) Are there any licensing issues with adding code originally released with an Apache license 2.0?

MichielCottaar:
1. For SBM it would indeed be ideal to provide topup with b0's that are well aligned with each other (even if they were not acquired close in time). The scorign algorithm implemented here actually gives us this information, so we could identify the b0's which have the least amount of rotation around the x- or y-axis (which are the problematic axes for SBM). We would have to somehow weight the quality of the b0's and the lack of rotation between them to select the best pair of b0's to use. I'm not sure if it is worth it.

On how many b0's to pass to topup: using only two b0's might be indeed a bit too few unless we want to optimise the SBM. However, I think the main advantage of having more b0's was that if one or two of them were corrupted by motion, they would not affect the final estimate too much. When we preselect good b0's, my understanding is that there is really no reason to go beyond 2 or 3 b0's per phase encode direction.

2. Putting another b0 in front of the eddy run would be easy to implement. My main reason to be worried about this is eddy's assumption that the reference b0 and the first diffusion-weighted image are well aligned with each other (see discussion on the usage of the --b0_peas flag in eddy). If we put another b0 as the reference at the beginning, then this assumption will not hold. We could also start diffusion-weighted images around, but this will become very complicated very quickly.

3. My understanding is that Apache license is a very free one and that it is not a problem as long as we satisfy point 4.1-4.3 in the license (https://opensource.org/licenses/Apache-2.0). However, I'm not a lawyer...

coalsont:
The apache license appears to be a permissive license (like BSD or MIT, not copyleft like GPL), and the main reason it is so long is to prevent abuse by patenting code and then copyright licensing it "openly", while keeping the patent private, and later suing people for infringing the patent.  I think all we need to do is include the original license notice (and any related notices if they exist on the original code) on that part of the code and we are covered (we don't need to change any of our licenses).

mharms:
2. Good point about dont_peas, esp. since our plan (at least currently) is to use that option.  But that means that there is essentially an unavoidable “conflict” between using `dont_peas` and using an “optimized" b0 as the reference space.

I feel that we would benefit from Jesper contributing to this discussion.

VosperFlopstop:
Do we really think this is a real problem? For this to be an issue the first b0 would have to be broken in all 4 blocks.
The issue of the second b0 for topup complicates things a little bit, because we now need a last-first pair to both be non-broken. But the it is not _that_ crucial to have the two b0s for topup to be very close in time. What SBM does is to remove bias by estimating how the field changes when the subject moves. That change can be equally well estimated if one knows the field at "zero" position, as if one doesn't know it. So in the case where we don't know the field "perfectly" because there is movement between the two input images to topup it simply means that there will be a very slight residual distortion (compared to for example a hifi structural), but we are still able to remove any MBS related variance just as well.
Mike, you asked about an option to set an arbitrary scan as ref scan (for example the third b0 in a block). There is already such an (hidden) option. I am just a little wary of using it. What I do there is that I set that as reference after each iteration, which definitely goes a very long way. But within each iteration there is still a difference in how things are done. This is due to some very fundamental choices done early on that would not be easy to change at this stage. I might just be a little too careful, but I prefer to let the first volume be the reference. And as you say, if we want to use --dont_peas that needs to be the "actual" first volume.
In brief, I think Michiel's plan to find which of the 4 blocks have the best first b0 is good. If it turns out that the b0 with opposing PE closest in time is broken, then we go to the next closest. Etc.
Finally, just to keep you up to date with what I am currently doing on this project. I have started doing some profiling of eddy for running slice-to-vol, and it has given some surprising results. It looks like I have already been able to shave 30-40% off the execution time with identical results. I will soon send a new executable to Sri for him to test if he sees the same gain there. I get the 30-40% on our k80s, I get nothing on the GeForce GT 750M on my Mac.

mharms:
My point was that we should think beyond the standard HCP-LS protocol.  What about protocols that collect only a pair of scans?  Or those like UK Biobank with only a single dMRI scan?  In those cases, you only have 2 or even just 1 chance to get a "good" b0 that is also at the start of a run.

Also, I don't think we should necessarily let `dont_peas` drive our decision making.  After all, we were (are) on the fence about using it in the first place.  If other considerations, such as using a temporally separated b0 as the reference b0, take on increase importance, then would it be a big loss simply to revert to using PEAS?

VosperFlopstop:
Hi again Mike. I see your point about thinking beyond the HCP-LS project, but I also think we should be careful not to try and solve "every" problem, and in the process end up overcomplicating and delaying the HCP-LS pipelines.
For the more "austere" acquisitions, like for example Biobank, I invariably suggest that people adapt a protocol like [AP-b0x3] [PA-b0x3 dwis b0 dwis b0 ...]. It costs pretty much nothing extra, and it all but guarantees that they will be able to get a good b0 pair to run topup on. I think that there will have to be a limit to how much we bend over backwards to accommodate people who chose to ignore our advice.
As for dropping dont_peas, yes you are right that it wouldn't cost much. The b0<->dwi alignment is much improved in more recent versions of eddy and it is really a toss up as to what is best.

MichielCottaar:
So it seems like we have two options:
1. Adopt the first b0 of one of the runs as the reference b0. This has the disadvantage that if there are only very few runs (as in some non-HCP datasets), there might not be any good ones.
2. Adopt the best b0 overall as the reference b0. This would require dropping the `--dont_peas` flag, because the first dwi will no longer be necessarily aligned with the reference b0.

It sounds like either would work fine. I think both would have a similar complexity to implement. I'm happy to go with the second option, if you still have a preference for it, @mharms?

mharms:
@MichielCottaar: Yes, that is a nice concise summary of the situation. Ultimately I think you, Jesper, Saad, and Stam need to make the decision as to what is "best".  That said, I think I have a decent understanding of the issues and uncertainties involved, and based on my synthesis of the issues, I've come to the point where I favor NOT using `dont_peas`, for the following reasons:
1. @VosperFlopstop indicated it is a "toss up".
2. Using PEAS (i.e., not providing `dont_peas` as a flag) is the "default" for `eddy`, so using PEAS is parsimonious with the default settings.
3. Some of the kids are pretty squirmy, so I'm sure that there will be some for whom there is a substantial movement between the b=0 and first dwi, _even within the same run_.  So, rather than having to check the log files for the estimated movement (between the b=0 and the dwis), identifying "large values" (a somewhat arbitrary decision), and then either flagging those subjects in the database (something likely to be missed by users) or re-running them with PEAS enabled, why not just run everyone with PEAS in the first place?
4. Recall that we ultimately settled on using `dont_peas` for the HCP-YA processing because that was overall a very compliant young-adult population, so we decided that the small extra uncertainty added by PEAS wasn't worth it.  However, HCP-LS is a less compliant population. Further, it sounds like the accuracy of PEAS has improved relative to when we made the decision to use `dont_peas` for HCP-YA.


So, first I think we need to come to agreement on whether or not we are going to use `dont_peas` for the HCP-LS processing.  If we decide that we will NOT, then it _seems_ like that would bias us toward your Option (2), _unless_ the accuracy of PEAS is itself somewhat dependent on the magnitude of the movement between the "first" b=0 and the first dwi.  That's a question for @VosperFlopstop as to whether such an interaction exists.  i.e., if the accuracy of PEAS declines as the size of the estimated "movement" increases, then that would incline us back toward your Option (1).

I guess another consideration, that should at least be on the table, is that with Option (1), a user could still _choose_ to use `dont_peas`, if they wanted to.  Whereas with Option (2), the use of `dont_peas` would be strongly discouraged (and probably should be checked for and outright prohibited).

glasserm:
Sorry if this has already been discussed, but will this impact use of susceptibility with movement interaction correction?

mharms:
No, MBS related variance itself is removed equally well, per https://github.com/Washington-University/HCPpipelines/pull/158#issuecomment-555533227

That said, if the _pair_ of b=0's used for `topup` have a considerable displacement between them, there will be a "slight residual distortion" (relative to the structurals).  But that's the issue of how do we best choose the opposite polarity b=0 to use in `topup`, which is separate from how we choose the b=0 that defines the reference space.

MichielCottaar:
Debugging this code will probably take a while, but in the meantime more comments are welcome.

I went with the strategy of adopting the best b0 overall out of the positive and the negative volumes (i.e., ignoring that they are not at the start of a series). The best b0 of the positive series is prepended to the full dataset, so that topup and eddy have the same reference b0. This reference b0 is removed again after eddy is finished.

MichielCottaar:
Based on the discussion above I decided to just get the best b0, irrespective of when it was acquired. This best b0 is determined in one of two ways:

1. If there are many b0's with a specific phase encoding (more than 4), then the b0's are registered to each other using mcflirt (i.e., no distortion correction). The b0 most similar to the average b0 is considered the best one.

MichielCottaar:
@glasserm @mharms @VosperFlopstop: do any of you have any feedback on this update to the pipeline? Given that we prepend a new b0 to the data given to eddy, it might be prudent to test this on a few more subjects to check whether eddy reliably converges. But it would be good to get some feedback before running larger scale tests...

glasserm:
It sounds good to me.

MichielCottaar:
Replies to @mharms comments below:

>     1. In your approach (1) ([#158 (comment)](https://github.com/Washington-University/HCPpipelines/pull/158#issuecomment-586326388)), if both the "pos" and "neg" polarities have > 4 b0's, then is the "best" one of each polarity determined completely independently of the other polarity?  Why is that the choice, rather than say choosing one polarity, and then the "best" from the other polarity that matches that one?

I'm not sure what your proposal is here, but there is no matching of b0's between the two polarities irrespective of the number of b0's. The goal is to identify undistorted b0's, not b0's with opposite polarities that "match" well. If some polarity has enough b0's, I can safely assume the avearge b0 at that polarity will be undistorted and hence use that as a reference to identify the undistorted b0's. So, there is no real advantage of including the b0's with opposite polarities in that case.

>     2. It appears that the new approach _replaces_ the previous code entirely.  It seems it might be preferable if it was instead implemented as an _option_, so as to preserve backwards compatibility with the previous way of processing the data...  As part of this option, we could check if `dont_peas` is also set, and if so, either generate a "warning" or simply abort outright.

This makes sense to me. I've added to my TODO list to put the old code back in.

>     3. Given that a "bad" b0 may itself badly corrupt the average, what do you think of the idea of iterating the process until the "average" is itself stable?  (This would require having some threshold, itself potentially a settable parameter, of what defines a "bad" match to the average.  Do we have a handle on what that threshold would be, or are we only capable of operating at the other end of selecting the "best"?)

This also sounds reasonable. I'll have a look at the distribution of the similarities to the average b0 to see if we can define a reasonable(~3 sigma) threshold to get rid of distorted b0's that are so different that they would significantly bias the average b0.

mharms:
@MichielCottaar: My point in (1) was the following:  Once we've identified the best "positive" b0, why not pick the best "negative" b0 against that positive b0, rather than against the average of the negative b0's?  The best positive b0 should be uncorrupted, and thus the negative b0 that best matches that should also be uncorrupted, thus achieving that goal.  The rationale for this proposal was that it would tend to also pick the negative b0 that was close in space (i.e., little movement) to the positive (reference) b0.  Admittedly, @VosperFlopstop indicated above that this consideration shouldn't be seen as critical (https://github.com/Washington-University/HCPpipelines/pull/158#issuecomment-555533227), but all else being equal, isn't it still preferable to have a pos/neg pair for `topup` that are both uncorrupted _and_ close in spatial position?

MichielCottaar:
@mharms: Hmm, I don't see a straightforward way to do that. For example, if we have N b0's with Pos polarity, and already found the best b0 with Neg polarity, would we just put all of those together and run `topup`. The resulting average would be completely dominated by the N b0's with Pos polarity. We could also run `topup` N times with just one of the Pos b0's, but in either case this might take a very long time if N is large. And I'm not even sure the result would be actually better. If there is still some minor corruption in the best Neg b0 identified, we might bias our search of the Pos b0's to have the same corruption, which might bias our estimated susceptibility field more than if we had just identified the best Pos b0 independently.

If we want to encourage a similar head position in our final pos/neg pair, I think it would be better to just make this an explicit part of the selection procedure. So keep the rating procedure we have now to identify the uncorrupted images, but when selecting the final pair also take into account that we want a similar head orientation. I'm not sure how to weight these two goals, but it should be possible to come up with some trade-off (i.e., how much more corrupted can the image be to get a better alignment in head position by N degrees). Still I think this would be quite a bit of work with little extra gain...

mharms:
Hmm.  I didn't think through the _practical_ issues/challenges of how one would actually implement my proposal of selecting the "best" (uncorrupted by motion) Pos and Neg b0, while also selecting a pair close in space.  Input from @VosperFlopstop would be nice here again.  i.e., If he signs off that we simply shouldn't worry (and/or that it simply isn't practical) about getting a b0 pair that is close in space, then that's the approach we'll take.

VosperFlopstop:
Hi guys,

sorry for the delay. I think it is sufficient to pick the best b0 for each PE-direction individually. Unless the subject has moved massively, or for some reason has left the scanner, the errors will be very small.

If you were to pursue the strategy of finding the "best pair" I would suggest using the current strategy to pick a set of b0 for each PE-direction that are all "good", then to put all of those b0s into a 4D file that you run topup on with the test_acqp.cnf configuration file. That configuration file runs just the early, low resolution, steps of the full topup and is intended mainly for checking that ones acqparams.txt file is correct before running the full, and more time consuming, topup.

But importantly for your case the movement parameters are well estimated already by these steps, and typically change very little if/when running to the full resolution. So that means you have a way to determine the relative positions of all the b0s in there, and use that to pick a "best pair" of Pos and Neg.

I hope that addresses your questions?

mharms:
@MichielCottaar: In the interest of moving this forward to completion, how would you like to proceed given @VosperFlopstop's latest guidance?

MichielCottaar:
I've implemented the changes you suggested earlier:

1. I've reverted my changes to the defualt FWHM passed on to eddy. I did add a warning message when setting the --select-best-b0 flag and not changing the FWHM (I don't check whether the FWHM is changed in a sensible way).
2. I've added the old code back in. Because the changes are so massive putting both versions into a single file would make the code unreadable. So, I moved it to two separate files with the a new select-best-b0 flag determining which one gets run.
3. I now iterate the computation of the average b0 a few times, so that only undistorted b0's are used to estimate the average b0.
I'm still testing these changes in jalapeno

Now, for the main remaining concern: as Jesper noted, the main worry is that the subject left the scanner or moved massively between the b0's we passing on to topup. In the data I've looked at in the past (mostly young adult, but also some aging/developmental) these massive movements were mostly between the different diffusion acquisition blocks. So, I think we can get most of the advantage of having well aligned pairs of b0's by ensuring that the b0's we used were acquired during the same block. This would be much easier to implement than actually measuring the alignment between all b0's (which would require combining motion estimates output from mcflirt and topup, and somehow include these motion estimates in the distortion score) and this approach should give us nearly all of the benefits of the more complicated approach. If you agree, I would be happy to implement this.

mharms:
By different "diffusion acquisition blocks", are you referring to `dir98_{AP,PA}` as one "block" and `dir99_{AP,PA}` as another "block?  If so, I'm not sure that is worth the additional complication, or even necessarily better.  It probably is indeed true that the largest absolute motion between different b0's will occur between runs in general, but I don't think there is necessarily any reason to assume that there will be less between run movement for dir98_AP/dir98_PA, than there might be between dir98_PA/dir99_AP.   Thus, I guess I'm back to just independently pick the best Pos b0 and best Neg b0, and not trying to also factor in some additional criterion based on the absolute movement between the two "best" exemplars of each separate polarity.

Do you concur?  And if so, is there anything else that needs implementation yet?

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,275 @@
...
+## Merging Files and correct number of slices 
+################################################################################################
+echo "Merging Pos and Neg images"
+${FSLDIR}/bin/fslmerge -t ${rawdir}/Pos `echo ${rawdir}/${basePos}_[0-9]*.nii*`
+${FSLDIR}/bin/fslmerge -t ${rawdir}/Neg `echo ${rawdir}/${baseNeg}_[0-9]*.nii*`
+
+paste -d' ' `echo ${rawdir}/${basePos}*.bval` >${rawdir}/Pos.bval
+paste -d' ' `echo ${rawdir}/${basePos}*.bvec` >${rawdir}/Pos.bvec
+paste -d' ' `echo ${rawdir}/${baseNeg}*.bval` >${rawdir}/Neg.bval
+paste -d' ' `echo ${rawdir}/${baseNeg}*.bvec` >${rawdir}/Neg.bvec
```
mharms, line 212: I see a similar construction was used in the original `basic_preproc.sh`, but given that `${basePos}=Pos`, and `*` matches against no characters, isn't there a danger with this syntax if the script is run on top of already existing results? i.e., if `${rawdir}/Pos.bval` already exists, this will not generate the intended outputs.

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,275 @@
...
+	${FSLDIR}/bin/imrm ${rawdir}/best_Neg_b0
+	${FSLDIR}/bin/immv ${rawdir}/Posn ${rawdir}/Pos
+	${FSLDIR}/bin/immv ${rawdir}/Negn ${rawdir}/Neg
+	${FSLDIR}/bin/immv ${rawdir}/best_Pos_b0n ${rawdir}/best_Pos_b0
+	${FSLDIR}/bin/immv ${rawdir}/best_Neg_b0n ${rawdir}/best_Neg_b0
+fi
+
+echo "Perform final merge"
+${FSLDIR}/bin/fslmerge -t ${rawdir}/Pos_Neg_b0 ${rawdir}/best_Pos_b0 ${rawdir}/best_Neg_b0
+# include Pos_b0 as the first volume of Pos_Neg, so that eddy will use it as reference
```
mharms, line 242:
```suggestion
# include Pos_b0 as the first volume of Pos_Neg, so that eddy will use it as reference
# it will be removed after eddy is completed
```

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,275 @@
...
+${FSLDIR}/bin/fslmerge -t ${rawdir}/Pos_Neg ${rawdir}/best_Pos_b0 ${rawdir}/Pos ${rawdir}/Neg
+echo 0 `paste -d' ' ${rawdir}/Pos.bval ${rawdir}/Neg.bval` >${rawdir}/Pos_Neg.bvals
```
mharms, line 244: Related to above, this is assuming that the "b0" indeed has b=0.

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,275 @@
...
+# This code in this section was adapted from a script written for the developing HCP (https://git.fmrib.ox.ac.uk/matteob/dHCP_neo_dMRI_pipeline_release/blob/master/utils/pickBestB0s.sh)
+# The original script was released under the Apache license 2.0 (https://git.fmrib.ox.ac.uk/matteob/dHCP_neo_dMRI_pipeline_release/blob/master/LICENSE)
+
+# Merge b0's for both phase encoding directions
+for pe_sign in ${basePos} ${baseNeg} ; do
+  merge_command=("${FSLDIR}/bin/fslmerge" -t "${rawdir}/all_${pe_sign}_b0s")
+  for entry in ${rawdir}/${pe_sign}_[0-9]*.nii*
+  do
+    basename=`imglob ${entry}`
+    ${FSLDIR}/bin/select_dwi_vols ${basename} ${basename}.bval ${basename}_b0s 0
```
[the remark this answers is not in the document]
MichielCottaar: select_dwi_vols considers any b-values within delta of 100 a match (this limit is hard-coded). So, it should be fine for our purposes here.

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,275 @@
...
+  N_b0s=`${FSLDIR}/bin/fslval ${rawdir}/all_${pe_sign}_b0s dim4`
+
+  # if there are less than 5 B0's for a specific phase encoding use topup to find the best ones
+  # otherwise simply register the B0's of the same phase encoding to each other
+  if [[ ${N_b0s} -lt 5 ]] ; then
+    N_other=`${FSLDIR}/bin/fslval ${rawdir}/all_${pe_other}_b0s dim4`
+    if [[ ${N_other} -gt 4 ]] ; then N_other=4 ; fi
+    echo "Score all ${N_b0s} ${pe_sign} B0's based on alignment with mean B0 after topup with ${N_other} ${pe_other} B0's"
+
+    # Select sub-set of other B0's to run topup on
```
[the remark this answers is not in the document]
MichielCottaar: It is indeed partly for convenience, but I think this is the right set to select anyway. We only reach this code if there are a small number of b0's (< 5) with a given phase encoding. Usually, in such an acquisition these would be acquired at the beginning of the diffusion set, so the b0's with opposite phase encoding closest in time are likely to be the first ${N_other}

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -455,9 +486,12 @@ main()
 	pre_eddy_cmd+=" --echospacing=${echospacing} "
 	pre_eddy_cmd+=" --b0maxbval=${b0maxbval} "
 	pre_eddy_cmd+=" --printcom=${runcmd} "
+	if [ ! -z "${SelectBestB0}" ] ; then
+	  pre_eddy_cmd+=" --select-best-b0 "
+  fi
 
 	log_Msg "pre_eddy_cmd: ${pre_eddy_cmd}"
-	${pre_eddy_cmd}
+	pre_eddy_job=`fsl_sub -l log -N pre_eddy ${pre_eddy_cmd}`
```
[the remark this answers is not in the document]
MichielCottaar: Oops, I use this for local testing; it was never meant to be included in the commits. I'll remove them.

MichielCottaar:
These changes should address your comments @mharms . Most importantly I removed the fsl_sub and now use the correct b-value and bvec for the b0 I extract in eddy.

I really screwed up the indentations in these files (my editor inserted two spaces instead of the tab characters used in this rest of the code). Rather than fix this all by hand, I had pycharm fix it for me. This led to some other changes as well (see last commit message for more details).

I'm now rerunning the code on jalapeno to see if all these changes have led to any problems.

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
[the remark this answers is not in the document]
mharms: Can you please add a comment to that effect to the code?  Thx.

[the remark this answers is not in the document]
mharms: Although, there is still a disconnect between the behavior of the `select_dwi_vols` script and the `b0maxbval` variable.  It's unfortunate that `select_dwi_vols` wasn't written to allow the user to provide the allowable delta as an argument.  Is there any easy way to fix this within our script?

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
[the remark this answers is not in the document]
mharms: This seems fine.  And the comment I added to the code makes it clear what it is doing.

mharms:
@MichielCottaar  Could you please go through and "Resolve conversation" on all the items where you either accepted my suggestions, or is simply no longer relevant?  Thx.

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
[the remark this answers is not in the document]
MichielCottaar: I'll add a patched version of `select_dwi_vols`, which allows the user to set the allowable delta (and add it into FSL for the next release)

MichielCottaar:
I believe these changes address all of your comments. Let me know if anything else comes up.

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -399,31 +407,61 @@ get_options()
...
+				fi
+				if [[ ${extra_eddy_arg} == "--dont_peas"* ]]; then
+					show_usage
+					echo -e ""
+					echo -e "ERROR: When using --select-best-b0 post-alignment of shells in eddy is required, "
+					echo -e "       as the first b0 could be taken from anywhere within the diffusion data and "
+					echo -e "       hence might not be aligned to the first diffusion-weighted image."
+					echo -e "       Remove either the --extra_eddy_args=--dont_peas flag or the --select-best-b0 flag"
+					echo ""
+					exit 1
```
mharms, line 432:
```suggestion
					log_Err "When using --select-best-b0, post-alignment of shells in eddy is required, "
					log_Err "       as the first b0 could be taken from anywhere within the diffusion data and "
					log_Err "       hence might not be aligned to the first diffusion-weighted image."
					log_Err_Abort "       Remove either the --extra_eddy_args=--dont_peas flag or the --select-best-b0 flag"
```

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -399,31 +407,61 @@ get_options()
...
+				fi
+			done
+		fi
+		if [ ${fwhm_set} == false ]; then
+			log_Warn "WARNING: Using --select-best-b0 prepends the best b0 to the start of the file passed into eddy"
```
mharms, line 437: `log_Warn` already appends with "WARNING:".  (Just as `log_Err` already appends with "ERROR:")
```suggestion
			log_Warn "Using --select-best-b0 prepends the best b0 to the start of the file passed into eddy"
```

DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
```diff
@@ -95,6 +93,9 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
                           output the commands that would be executed instead of
                           actually running them. --printcom=echo is intended to
                           be used for testing purposes
+  [--select_best_b0]
+                          Whether the --select-best-b0 was used in
```
mharms, line 97:
```suggestion
                          Whether the --select-best-b0 flag was used in
```

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -432,36 +456,41 @@ main()
...
+		if [ ${NegVols[${j}]} -ne 0 ]; then
+			${runcmd} echo ${CorrVols} >>${outdir}/rawdata/${baseNeg}_SeriesCorrespVolNum.txt
 		fi
 	done
 
-	if [ ${Paired_flag} -eq 0 ] ; then
+	if [ ${Paired_flag} -eq 0 ]; then
 		log_Msg "Wrong Input! No pairs of phase encoding directions have been found!"
 		log_Msg "At least one pair is needed!"
 		exit 1
```
mharms, line 481: Preceding two lines should use `log_Err` and `log_Err_Abort`, respectively.  Then delete the `exit 1` line.

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,292 @@
...
+	echo "$(($1 % 2))"
+}
+
+rawdir=${workingdir}/rawdata
+topupdir=${workingdir}/topup
+eddydir=${workingdir}/eddy
+
+if [[ ${PEdir} -ne 1 && ${PEdir} -ne 2 ]]; then
+	echo -e "\n ERROR: ${scriptName}: Unrecognized PEdir: ${PEdir}"
+	exit 1
```
mharms, line 26:
```suggestion
	log_Err_Abort "${scriptName}: Unrecognized PEdir: ${PEdir}"
```

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,292 @@
...
+		${FSLDIR}/bin/imrm ${entry}
+	done
+done
+
+# Here we identify the b0's that are least affected by motion artefacts to pass them on to topup.
+# These b0's are identified by being most similar to a reference b0, which is determined in one of two ways:
+# 1. If there are enough b0's with a given phase encoding (>= 5) we adopt the average b0 as a reference
+# 2. If there are fewer b0's than the average b0 might be contaminated by any motion artefacts in one or two b0's.
+#    So in this case we use topup to combine the b0's with the b0's with opposite phase encoding and get a reference b0
+# To further reduce the chance that our reference b0 is contaminated by motion we compute the average multiple times
```
mharms, line 60:
```suggestion
# 1. If there are enough b0's with a given phase encoding direction (>= 5) we adopt the average b0 as a reference.
# 2. If there are fewer b0's, then the average b0 might be contaminated by any motion artefacts in one or two b0's.
#    In that case we use topup to combine the b0's with the b0's with opposite phase encoding to get a reference b0.
# To further reduce the chance that our reference b0 is contaminated by motion we compute the average multiple times,
```

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,292 @@
...
+			scores=($(${FSLDIR}/bin/fslstats -t ${select_b0_dir}/all_b0s_mcf_res -k ${select_b0_dir}/nodif_brain_mask -M))
+			scores_str="${scores[@]}"
+
+			echo "Iteration ${i} in finding best b0 for ${pe_sign}"
+			echo "Current scores: ${scores_str}"
+
+			# Recomputes the average using only the b0's with scores below (median(score) + 2 * mad(score)),
+			# where mad is the median absolute deviation.
+			idx=$(fslpython -c "from numpy import median, where; sc = [float(s) for s in '${scores_str}'.split()]; print(','.join(str(idx) for idx in where(sc < median(abs(sc - median(sc)) * 2 + median(sc)))[0]))")
+			echo "Recomputing average b0 using indices: ${idx}"
```
mharms, line 175:
```suggestion
			echo "Recomputing average b0 using indices (counting from zero): ${idx}"
```

DiffusionPreprocessing/scripts/basic_preproc_best_b0.sh
```diff
@@ -0,0 +1,292 @@
...
+			scores=($(${FSLDIR}/bin/fslstats -t ${select_b0_dir}/topup_b0s_res -k ${select_b0_dir}/nodif_brain_mask -M))
+			scores_str="${scores[@]}"
+
+			echo "Iteration ${i} in finding best b0 for ${pe_sign}"
+			echo "Current scores: ${scores_str}"
+
+			# Recomputes the average using only the b0's with scores below (median(score) + 2 * mad(score)),
+			# where mad is the median absolute deviation.
+			idx=$(fslpython -c "from numpy import median, where; sc = [float(s) for s in '${scores_str}'.split()]; print(','.join(str(idx) for idx in where(sc < median(abs(sc - median(sc)) * 2 + median(sc)))[0]))")
+			echo "Recomputing average b0 using indices: ${idx}"
```
mharms, line 142: Is this particular index counting from zero? If so, let's say that:
```suggestion
			echo "Recomputing average b0 using indices (counting from zero): ${idx}"
```

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -3,13 +3,15 @@ source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@" # Debugging functions; al
...
-GdCoeffs=$2  #Coefficients for gradient nonlinearity distortion correction. If "NONE" this corrections is turned off
-CombineDataFlag=$3   #2 for including in the ouput all volumes uncombined (i.e. output file of eddy)
-                     #1 for including in the ouput and combine only volumes where both LR/RL (or AP/PA) pairs have been acquired
-                     #0 As 1, but also include uncombined single volumes"
+GdCoeffs=$2        #Coefficients for gradient nonlinearity distortion correction. If "NONE" this corrections is turned off
+CombineDataFlag=$3 #2 for including in the ouput all volumes uncombined (i.e. output file of eddy)
+                   #1 for including in the ouput and combine only volumes where both LR/RL (or AP/PA) pairs have been acquired
+                   #0 As 1, but also include uncombined single volumes"
+SelectBestB0=$4 #0 only the actual diffusion data was fed into eddy
+                #1 least distorted b0 was prepended to the eddy input
```
mharms, line 14:
```suggestion
                #1 least distorted b0 was prepended to the eddy input
                # Note: This numeric value is used within the script as a numeric that controls
                # the number of volumes to skip, so it isn't just used as 0/1 "boolean".
```

DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -31,70 +33,80 @@ qc_command+=(-v)
...
+		cp ${eddydir}/eddy_unwarped_images.eddy_rotated_bvecs ${datadir}/bvecs
+	fi
 else
 	echo "JAC resampling has been used. Eddy Output is now combined."
-	PosVols=`wc ${eddydir}/Pos.bval | awk {'print $2'}`
-	NegVols=`wc ${eddydir}/Neg.bval | awk {'print $2'}`    #Split Pos and Neg Volumes
-	${FSLDIR}/bin/fslroi ${eddydir}/eddy_unwarped_images ${eddydir}/eddy_unwarped_Pos 0 ${PosVols}
-	${FSLDIR}/bin/fslroi ${eddydir}/eddy_unwarped_images ${eddydir}/eddy_unwarped_Neg ${PosVols} ${NegVols}
+	PosVols=$(wc ${eddydir}/Pos.bval | awk {'print $2'})
+	NegVols=$(wc ${eddydir}/Neg.bval | awk {'print $2'})                                                         #Split Pos and Neg Volumes
```
mharms, line 60: Seems to have been a lot of space added between the end of the command and the comment (possibly by pycharm)?

global/scripts/select_dwi_vols
```diff
@@ -0,0 +1,81 @@
...
+    echo ""
+    echo "Usage: select_dwi_vols <data> <bvals> <output> <approx_bval> [other options]"
+    echo ""
+    echo " Optional arguments:"
+    echo ""
+    echo " -b <value>    additional bvalues to be extracted (can be used multiple times)"
+    echo " -m            output mean instead of concat"
+    echo " -v            output variance instead of concat"
+    echo " -obv <bvecs>  produce <output>.bval and <output>.bvec"
+    echo " -db <delta_bvals> only include volumes with offset in b-value less than this value (default: 100 s/mm2)"
```
mharms, line 13: Unclear from this description, does the offset only apply for frames less than the `-b` value, or does it apply in either direction?

MichielCottaar: The absolute value of the offset should be smaller than this number. I've clarified this.

MichielCottaar:
This version of the code succesfully ran on jalapeno (or more technically the version a few days ago, but we have only tweaked the comments & error messages since, so it should still be fine).

glasserm:
Looks good.

MichielCottaar:
Are we waiting for anyone else to review this?

glasserm:
I merged it

PR #159, opened 2019-11-15
Regenerate compiled matlab

mharms:
This PR regenerates the compiled matlab under R2017b (same version as used for the current FSL FIX compiled matlab), and simultaneously incorporates the changes to the dimensionality estimation in MR-FIX made in #150 into the compiled matlab.

Successful execution (in compiled mode) has been demonstrated for:

- `prepareICAs` (PostFix.sh)

- `functionhighpassandvariancenormalize` (hcp_fix_multi_run)

- `RestingStateStats` (although the figure generation hasn't survived the change in matlab version very well -- see example png attached below.  For all I know, the figure generation using interpreted matlab may be equally messed up for recent matlab versions).

The two functions related to MSMAll -- `ComputeVN` and `MSMregression` were not tested, so @coalsont, if you are equipped to run a quick test on those in compiled mode, that would be great.

@glasserm: Is there something that I can look for in the log files just to 100% confirm that the changes made to MR-FIX with #150 have made it into the compiled version?

tfMRI_EMOTION_LR_Atlas_1-5_OrigTCS-UnstructNoiseTCS_QC_Summary_Plot


coalsont:
My previous setup for MSMAll was with octave in a container that didn't install any MCR, so I am not in that great of a position to test the compiled version.

mharms:
Let me see if I can use `MSMAllPipelineBatch.sh` as a template to come up with a token example that I can run on a single fMRI run.

glasserm:
1) I recommend we deprecate RestingStateStats at this time and plan to make a markedly revised version later that has a lot of changes including removal of 24 movement regressors (or at least moving to after sICA+FIX regression instead of before), removal of WM and CSF regression stats, addition of tICA stats, and repair of the matlab figures.  RestingStateStats is no longer required to run the extant HCP Pipelines to completion.  

2) I agree MSMAll and DedriftAndResample should be tested on one subject, though I would recommend using all runs of a lifespan subject.  Perhaps Tim has testing data and launcher scripts he could share that would help with this.  

3) The icaDim logs with the recent changes will have evidence of a final golden search with similar dimensionality to the initial golden search (as opposed to ending with a low dimensionality golden search).

coalsont:
The differences between my msmall singularity script and the Examples MSMAll script seem like only the container stuff.  Looks like I pulled my data from login2.nrg.wustl.edu:/data/hcpdb/packages/unzip/1200subject .

mharms:
1. I suggest we leave RSS "as is" until such time that we can create a revised version.
2. I'm running a test of MSMAllPipeline.sh in compiled mode, on a single run.  The purpose here is just to confirm that the compiled matlab works, not to test the execution of MSMAll in the manner in which we intend to process the HCP-LS data.
3.  I need the specifics of exactly what text I should look for in the stdout of the MR-FIX run.

glasserm:
1.  Okay
2. Presumably that is what we would actually use the compiled matlab for though?
3. You should just send me some logfiles.
4. There is a small bug in the icaDim.m function for Out=icaDim(<data>,<?>,<?>,3,<?>); We don't actually use this mode, but the code would not run correctly if called in this way.  I can try to fix it, but that would require a matlab recompile.

glasserm:
Actually I'll retract #4, as I think it would work okay.  I have updated the comment to describe what it would actually do.

mharms:
2. The compiled version of `MSMRegression.m` works as tested.  The purpose here is simply to test that the compiled code runs to completion.  Is there any functionality of `MSMRegression` that is not adequately tested, as regards that specific goal, if run on just a single run?  Note that the compiled version of `ComputeVN` was not tested, since the default code no longer uses that functionality. However, its compile code is quite straightforward, so we have no reason to doubt that it wouldn't work if called upon. Last, going back to an earlier comment, DedriftAndResample does not involve any unique compiled matlab code, so it was not tested.

3. Sent the log file via email.

4. Recompiling the matlab would be easy at this point, but sounds like it isn't necessary.

glasserm:
2. I thought it would be good to test MSMAll as it will be run on lifespan data to ensure that it works for that data.  It most likely will work based on prior testing, but why not confirm this now?

mharms:
Because I don't have any more time to put into this currently.  As far as I'm concerned, we've established that the compiled matlab works under the new MCR.

mharms:
Just circling back to this to add a note that I tested the compiled version of `ComputeVN` and it worked fine.

PR #161, opened 2019-11-22
Further across script harmonization and fMRIVolume usage statement

mharms:
Do we want to include all the LegacyStyleData options in the fMRIVolume usage statement, or just leave it along the lines of what I have now?

### DIFF
--- a/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
+++ b/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
@@ -2,7 +2,7 @@
 
 # Requirements for this script
 #  installed versions of: FSL, FreeSurfer
-#  environment: HCPPIPEDIR, FSLDIR, FREESURFER_HOME
+#  environment: HCPPIPEDIR, FSLDIR, FREESURFER_HOME, HCPPIPEDIR_Global 
 
 # ---------------------------------------------------------------------------
 #  Constants for specification of susceptibility distortion Correction Method
@@ -123,7 +123,7 @@ fi
 log_Check_Env_Var HCPPIPEDIR
 log_Check_Env_Var FSLDIR
 log_Check_Env_Var FREESURFER_HOME
-
+log_Check_Env_Var HCPPIPEDIR_Global
 
 ################################################ SUPPORT FUNCTIONS ##################################################
 
--- a/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
+++ b/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
@@ -125,6 +125,8 @@ log_Check_Env_Var FSLDIR
 log_Check_Env_Var FREESURFER_HOME
 log_Check_Env_Var HCPPIPEDIR_Global
 
+HCPPIPEDIR_fMRIVol=${HCPPIPEDIR}/fMRIVolume/scripts
+
 ################################################ SUPPORT FUNCTIONS ##################################################
 
 # function for parsing options
--- a/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
+++ b/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
@@ -48,24 +48,23 @@ Usage: ${script_name} [options]
   --gdcoeffs=<gradient non-linearity distortion coefficients (Siemens format)>
   [--qaimage=<output name for QA image>]
 
-  --method=<method used for susceptibility distortion correction>
+  --method=<method to use for susceptibility distortion correction (SDC)>
 
         "${FIELDMAP_METHOD_OPT}"
             equivalent to ${SIEMENS_METHOD_OPT} (see below)
 
         "${SIEMENS_METHOD_OPT}"
-             use Siemens specific Gradient Echo Field Maps for
-             susceptibility distortion correction
+             use Siemens specific Gradient Echo Field Maps for SDC
 
         "${SPIN_ECHO_METHOD_OPT}"
-             use Spin Echo Field Maps for susceptibility distortion correction
+             use a pair of Spin Echo EPI images ("Spin Echo Field Maps") acquired with
+             opposing polarity for SDC
 
         "${GENERAL_ELECTRIC_METHOD_OPT}"
-             use General Electric specific Gradient Echo Field Maps
-             for susceptibility distortion correction
+             use General Electric specific Gradient Echo Field Maps for SDC
 
         "${NONE_METHOD_OPT}"
-             do not use any susceptibility distortion correction"
+             do not use any SDC
              NOTE: Only valid when Pipeline is called with --processing-mode=LegacyStyleData
 
   [--topupconfig=<topup config file>]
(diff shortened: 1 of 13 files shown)


coalsont:
If we want people to know that our pipelines have support for something, then it should be in the usage information.  If we plan to make a big deal out of the new features, then it stands to reason that their usage should be documented.  However, given that the WUSTL side isn't enthusiastic about things like slice timing correction, perhaps the things we want to advertise are the no-T2 support, etc.  Personally, I don't particularly like hiding options in general.

On a side note, my new option processing shlib doesn't currently have a provision for hidden options (unless you manually write the entire usage text instead of using the usage helper function), and I would prefer not to do such a thing (it does accept alternate (deprecated) names for options, and the usage doesn't display what those alternate names are unless you add it manually).

sivcek:
I agree with @coalsont. I think that all the options should be printed out in help. It makes sense to clearly denote the Legacy options, but I would not require users to look for them in the code itself. The only drawback I see is that the warnings add quite a bit of text. So if they are included in the usage string, it will get somewhat long for the `fMRIVolume` pipeline.

mharms:
Ok.  @sivcek: would you mind taking a first pass at adding the appropriate usage statement descriptions for the Legacy options?

@glasserm: Can I set `--biascorrection` to have a default of `NONE`?  (I can note that `SEBASED` is recommended in the Usage statement, but I don't want to make that a "default", since it requires TOPUP).

@sivcek: Can we change the test for whether `DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased` gets invoked to be based on the seemingly more logical `DistortionCorrection` parameter rather than `fMRIReference`?   Correspondingly, a check should be added that aborts if `fMRIReference` is used but the `DistortionCorrection` method is not `NONE` (since that is a nonsensical set of parameters).

coalsont:
Having a default that we don't recommend doesn't seem like a good idea (and I'm not sure disabling GDC by default is a good change either).  If nothing else, we could make the parameter(s) mandatory rather than having a default.

coalsont:
I also don't understand the comment about "requires topup" - as I understand it, topup is an integral part of FSL, not a separate dependency, and if FSL isn't installed, basically nothing will work anyway.

mharms:
@sivcek: Related to your comment about the Usage statement getting long, we could potentially offload all the legacy-related info to its own separate usage statement, which could be invoked via something like `--processing-mode-help`

mharms:
@coalsont: Capital `TOPUP` in the Usage statement refers to one of the allowed distortion correction _methods_.  Not the topup tool itself.

coalsont:
If you don't like having the default be the hcp recommendation of TOPUP, then I don't think there should be a default at all.  Given that proper usage requires other arguments to specify fieldmap files, I don't think having a default for this is particularly useful anyway.

I would argue similarly for GDC - if the hcp recommendation is to use it, it shouldn't default to off, and since we don't have a good guess at the correct filename for the parameter, it therefore shouldn't have any default at all.

Defaults are for user convenience, which is a lower priority than doing the right thing with your data - our scripts shouldn't "help" the user automatically do something bad.

coalsont:
Ah, right, bias correction and SDC are separate.  We don't have an obvious recommendation for bias correction when the user doesn't have spin echo fieldmaps (it could even be read as recommending no bias correction), so that documentation may deserve some attention.

However, I stand by the idea that defaults are only a good idea when they follow best practices.  Having bad defaults is worse than having no defaults.

sivcek:
@mharms I can write a draft for the legacy options help. I also like the suggestions for separate help. I would though use it to print out additional information and warnings. So my suggestion is to keep the useage for the options in the main script, but alert the users that these are legacy options and to please read the additional info before using them. So the param could be `--legacy-mode-info`.

Should I stage the info as a PR to this PR?

mharms:
@sivcek: I added you as as Collaborator to my forked repo, so I believe that means you should be able to push to it (and this PR) directly.

glasserm:
GDC default is on, distortion correction is topup, bias correction is sebased.  The defaults should be the recommended stream.

glasserm:
The recommendation if using a standard fieldmap would be to get pre-scan normalized data and set the bias correction to NONE.

sivcek:
@mharms thanks! I'm cloning right now. I'll prepare and push tomorrow. It's getting late on this end of the Atlantic.

mharms:
Ok, I'll revert GDC to having no default (i.e., is a required parameter, even if using `NONE`).  I will also leave `--biascorrection` and `--dcmethod` as having no defaults, since I agree with @coalsont that it doesn't make a lot of sense to have "defaults" that nonetheless require the specification of other parameters in order to complete the "default" setting.

sivcek:
@mharms I've added descriptions for fMRIVolume legacy parameters.

glasserm:
Perhaps there should be a section header grouping those parameters as for legacy data?

sivcek:
It does start with 

```
"LegacyStyleData" MODE OPTIONS:
```

We can add a line to make it more obvious.

mharms:
@sivcek: I made some tweaks to your usage statement additions, including changing the use of "BOLD" to "fMRI" for consistency with the way we talk about the data within the script.  Relatedly, to be consistent in our argument naming, we should probably change `--boldmask` to `--fmrimask` as well.  Do you mind if I make that change?  Were you going to add the proposed `--processing-mode-info` flag, and associated additional info?

sivcek:
@mharms I'm ok with changing `--boldmask` to `--fmrimask` I think it's good to be consistent. I can add the `--processing-mode-info` flag tomorrow morning.

glasserm:
Indeed, the HCP Pipelines are meant to work on non-BOLD fMRI too, like CBV-weighted MION fMRI in non-human primates.

sivcek:
@mharms I've added the info on legacy options. @coalsont I have also added a function to the opts library that checks for existence of a specific flag within the arguments passed. I have tried using `opts_GetOpt1` first, however it returns true in cases like this:

```
if opts_GetOpt1 a b bah cah dah; then 
    echo "yes"
fi
```

The above code will print yes even though a is not a separate parameter. The following code will not:

```
if opts_CheckForFlag a b bah cah dah; then 
    echo "yes" 
fi
```

mharms:
As far as I'm concerned, this is ready to go, and addresses the last remaining items I had listed in #146.

coalsont:
opts_GetOpt1 was never intended to use the exit status to pass information (or else it would trip set -e and debug.shlib constantly), so of course it didn't work that way.  However, it also won't detect any flag that doesn't have an `=` in it.  I plan to only use newopts.shlib for my new scripts, though it also currently requires an `=` on all arguments except the special implicit -h and --help options.  If I did add the ability to hide options from the standard help info for newopts.shlib, I would likely add another implicit option like `--show-hidden-options` or `--help-full` and make the standard help info say how many hidden options exist and how to see them.

I generally don't use exit status of functions to pass boolean values, partly since zero and nonzero are interpreted backwards from virtually any other language.  However, I don't really see the harm in doing so here, since that is the only purpose the function will be used for.

Is `--processing-mode-info` the flag we want to use for this here?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -74,18 +73,19 @@ Usage: ${script_name} [options]
...
         "LEGACY"
-             use the bias field derived from T1w and T2w images, same as pipeline version 3.14.1 or older"
+             use the bias field derived from T1w and T2w images, same as was used in
+             pipeline version 3.14.1 or older. No longer recommended.
 
         "NONE"
              don't do bias correction
 
-  --usejacobian=<"true" or "false">
+  --usejacobian=<TRUE or FALSE>
```
coalsont: `--use-jacobian` does not currently accept anything other than all-lowercase, this usage change is currently wrong.

mharms: Good catch. I could add the same case conversion logic that we have in `GenericfMRIVolume`.  Or just revert the Usage statement to the lower-case version.  Which do you prefer?

coalsont: It is an internal script, it doesn't need as much user-friendliness, but it can't hurt.

mharms: Done.  Forced to lower-case internally within the script, but left as upper case in the Usage statement to match our "general" boolean style.  (Although we aren't entirely consistent about this -- e.g., `--preregistertool` and `--fmrirefreg` are specified in lower case (and lower case only).

mharms:
@sivcek: Should `--preregistertool` be mentioned as part of `--processing-mode-info`, or is everything that we need to say about that already captured in the Usage statement itself?

glasserm:
This looked good to me, though I missed the last comment.  Hopefully that can be addressed in a separate commit.

PR #163, opened 2020-01-23
Update FreeSurfer2CaretConvertAndRegisterNonlinear.sh

glasserm:
Add checking for mismatching atlas subcortical segmentation and subject's brain mask

### DIFF
--- a/PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
+++ b/PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
@@ -214,7 +214,7 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
     [ "${T2wImage}" != "NONE" ] && applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage"."$GrayordinatesResolution".nii.gz
     applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage"."$GrayordinatesResolution".nii.gz
     applywarp --interp=nn -i "$AtlasSpaceFolder"/"$T1wImageBrainMask".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz
-    ${CARET7DIR}/wb_command "((Brainmask-1)*-1)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
+    ${CARET7DIR}/wb_command -volume-math "(1-Brainmask)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
     MissingGrayordinates=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -V | awk '{print $1}'`
     MissingBrainstem=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -l 15.9 -u 16.1 -V | awk '{print $1}'`
     echo "MissingGrayordinates,MissingBrainstem" > "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
--- a/PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
+++ b/PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
@@ -213,12 +213,11 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
     ${CARET7DIR}/wb_command -volume-label-import "$AtlasSpaceFolder"/ROIs/wmparc."$GrayordinatesResolution".nii.gz ${SubcorticalGrayLabels} "$AtlasSpaceFolder"/ROIs/ROIs."$GrayordinatesResolution".nii.gz -discard-others
     [ "${T2wImage}" != "NONE" ] && applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage"."$GrayordinatesResolution".nii.gz
     applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage"."$GrayordinatesResolution".nii.gz
-    applywarp --interp=nn -i "$AtlasSpaceFolder"/brainmask_fs.nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz
-    ${CARET7DIR}/wb_command "((Brainmask-1)*-1)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
+    applywarp --interp=nn -i "$AtlasSpaceFolder"/"$T1wImageBrainMask".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz
+    ${CARET7DIR}/wb_command "((Brainmask-1)*-1)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
     MissingGrayordinates=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -V | awk '{print $1}'`
     echo "MissingGrayordinates" > "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
     echo "${MissingGrayordinates}" >> "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
-    rm "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz
 done
 
 #Loop through left and right hemispheres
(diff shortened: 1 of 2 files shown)


PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -213,6 +213,12 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
...
     [ "${T2wImage}" != "NONE" ] && applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage"."$GrayordinatesResolution".nii.gz
     applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage"."$GrayordinatesResolution".nii.gz
+    applywarp --interp=nn -i "$AtlasSpaceFolder"/brainmask_fs.nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz
+    ${CARET7DIR}/wb_command "((Brainmask-1)*-1)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
```
mharms: Ditto.  Make use of `$T1wImageBrainMask` variable.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -213,6 +213,12 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
...
+    MissingGrayordinates=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -V | awk '{print $1}'`
+    echo "MissingGrayordinates" > "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
+    echo "${MissingGrayordinates}" >> "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
+    rm "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz
```
mharms: I'd suggest keeping `MissingGrayordinates.nii.gz` around so that the location can be quickly assessed.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -213,6 +213,12 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
...
     applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage"."$GrayordinatesResolution".nii.gz
+    applywarp --interp=nn -i "$AtlasSpaceFolder"/brainmask_fs.nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz
+    ${CARET7DIR}/wb_command "((Brainmask-1)*-1)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
+    MissingGrayordinates=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -V | awk '{print $1}'`
+    echo "MissingGrayordinates" > "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
+    echo "${MissingGrayordinates}" >> "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
```
mharms: Could MissingGrayordinates.txt be generated in a way that additional includes lines that quantify the number of missing grayordinates per CIFTI subcortical ROI?  That would be very valuable for quickly distinguishing between missing brainstem vs. missing other structures.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -213,6 +213,12 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
     ${CARET7DIR}/wb_command -volume-label-import "$AtlasSpaceFolder"/ROIs/wmparc."$GrayordinatesResolution".nii.gz ${SubcorticalGrayLabels} "$AtlasSpaceFolder"/ROIs/ROIs."$GrayordinatesResolution".nii.gz -discard-others
     [ "${T2wImage}" != "NONE" ] && applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage"."$GrayordinatesResolution".nii.gz
     applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage"."$GrayordinatesResolution".nii.gz
+    applywarp --interp=nn -i "$AtlasSpaceFolder"/brainmask_fs.nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/brainmask_fs."$GrayordinatesResolution".nii.gz
```
mharms: Opps.  Must not have saved my comment on this line, which was to make use of the defined variable `$T1wImageBrainMask` rather than relying on the hardcoded file name `brainmask_fs`.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
mharms: See comment on preceding line for context:
https://github.com/Washington-University/HCPpipelines/pull/163#discussion_r370356133

glasserm:
We expect small segmentation differences between individual and atlas ROIs, so we don't want to pick those up.  We just want to pick up cases where the entire brain mask is not including the whole CIFTI space.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
glasserm: I don't know an easy way to do it, but am happy if someone else wants to do it.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
glasserm: OK

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
coalsont: To get all structures, convert the brain mask to the cifti voxel space, import to cifti, parcellate with the subcortical labels and SUM, -cifti-convert -to-text.  Could generate the list of names starting with -*-label-export-table.  Stitching them together could get interesting.

If we specifically expect brainstem to be an issue, we could just hardcode its label value into the script to get stats on it alone (or everything except it).

glasserm: I would prefer the latter solution.

glasserm:
Am I merging this into master?

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -213,6 +213,12 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
...
     [ "${T2wImage}" != "NONE" ] && applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT2wImage"."$GrayordinatesResolution".nii.gz
     applywarp --interp=spline -i "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$AtlasSpaceT1wImage"."$GrayordinatesResolution".nii.gz
+    applywarp --interp=nn -i "$AtlasSpaceFolder"/"$T1wImageBrainMask".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz
+    ${CARET7DIR}/wb_command "((Brainmask-1)*-1)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
```
coalsont: This wb_command call has no command switch.

coalsont: Also, consider "(1 - Brainmask)..." for simplicity.

glasserm: Fixed

coalsont: In fact, if you wanted it more bulletproof, you can use "(!Brainmask)...", ! is boolean negation (after implicitly doing ">0") and is guaranteed to produce exact 1 or 0.

It may give better edges to the mask (when resolutions mismatch) by using trilinear and thresholding at 0.5.  Or, you could use a different threshold than 0.5 to make the test more lenient or restrictive.

coalsont: Given the large expected resolution mismatch between structural and 2mm grayordinates, if we really cared about the edges, we would probably want a little smoothing on the brainmask before resampling.  But, I suspect the exact definition of this count isn't really critical, as most problems will result in a large number of missed voxels regardless.

glasserm: I think it is fine as is

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -213,6 +213,12 @@ for GrayordinatesResolution in ${GrayordinatesResolutions} ; do
...
+    applywarp --interp=nn -i "$AtlasSpaceFolder"/"$T1wImageBrainMask".nii.gz -r "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz -o "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz
+    ${CARET7DIR}/wb_command -volume-math "(1-Brainmask)*CIFTIStandardSpace" "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -var Brainmask "$AtlasSpaceFolder"/"$T1wImageBrainMask"."$GrayordinatesResolution".nii.gz -var CIFTIStandardSpace "$AtlasSpaceFolder"/ROIs/Atlas_ROIs."$GrayordinatesResolution".nii.gz
+    MissingGrayordinates=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -V | awk '{print $1}'`
+    MissingBrainstem=`fslstats "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.nii.gz -l 15.9 -u 16.1 -V | awk '{print $1}'`
+    echo "MissingGrayordinates,MissingBrainstem" > "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
+    echo "${MissingGrayordinates},${MissingBrainstem}" >> "$AtlasSpaceFolder"/ROIs/MissingGrayordinates.txt
```
mharms: Maybe set-off L216-L221 with line breaks, and a comment:
`# Report on number of missing subcortical grayordinates in CIFTI standard space`
so that it'll be easy for us to remember which specific set of lines need to be run on the already processed Struct data?

mharms:
Also, @coalsont, do you want (plan) to add to check to the new v.4.1.3 branch that requires use of WB 1.4.2 or later?

coalsont:
If there isn't already a version check that can be updated, I was not planning on adding one.

mharms:
Certain scripts have version checks for FS version, FIX version, and FSL version.  But there isn't any "master" checking script, nor do I recall any script checking for WB version.  Up to now, WB version really hasn't been an issue.  Maybe at some point we could add a library that makes checking for compliance with a particular version easier to do.

PR #164, opened 2020-02-11
Enable Creation of Cleaned Movement Regressors

glasserm:
Regress sICA noise components out of movement regressors to enable their use on sICA+FIX cleaned fMRI data.  The files are present in text and .nii.gz formats and for both the concatinated and single runs

### DIFF
--- a/ICAFIX/hcp_fix_multi_run
+++ b/ICAFIX/hcp_fix_multi_run
@@ -827,8 +827,7 @@ fi
 
 # Convert sICA+FIX cleaned movement regressors to text
 if [ -f ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
-  fslmaths ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -Tmean -mul 0 -add 1 ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz
-  fslmeants -i ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -m ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz -o ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt --showall
+  fslmeants -i ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -o ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt --showall
   cat ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt | tail -$(fslval ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz dim4) > ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt
 fi
 
@@ -873,16 +872,12 @@ for fmri in $fmris ; do
 
   if [ -f ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
     ${Caret7_Command} -volume-merge ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -volume ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -subvolume ${Start} -up-to ${Stop}
-    fslmeants -i ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -m ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz -o $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt --showall
+    fslmeants -i ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -o $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt --showall
     cat $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt | tail -$(fslval ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz dim4) > $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt
   fi
 
 	Start=`echo "${Start} + ${NumTPS}" | bc -l`
 done
-
-if [ -f ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
-  rm ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz
-fi
 
 cd ${DIR}
 
--- a/ICAFIX/hcp_fix_multi_run
+++ b/ICAFIX/hcp_fix_multi_run
@@ -825,6 +825,13 @@ then
     rm -f ${concatfmri}_Atlas.dtseries.nii ${concatfmri}_Atlas_hp${hp}.dtseries.nii
 fi
 
+# Convert sICA+FIX cleaned movement regressors to text
+if [ -f ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
+  fslmaths ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -Tmean -mul 0 -add 1 ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz
+  fslmeants -i ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -m ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz -o ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt --showall
+  cat ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt | tail -$(fslval ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz dim4) > ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt
+fi
+
 ## ---------------------------------------------------------------------------
 ## Split the cleaned volume and CIFTI back into individual runs.
 ## ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


glasserm:
fix_3_clean.m code: 

%%%% compute movement regressors with noise components removed non aggressively
if domot == 0
  confounds = functionmotionconfounds(TR,hp);
  % non-aggressively regress out noise ICA components from movement regressors
  betaconfounds = pinv(ICA,1e-6) * confounds;                              % beta for confounds (good *and* bad)
  confounds = confounds - (ICA(:,DDremove) * betaconfounds(DDremove,:));    % cleanup
  save_avw(reshape(confounds',size(confounds,2),1,1,size(confounds,1)),'mc/prefiltered_func_data_mcf_conf_hp_clean','f',[1 1 1 TR]);
end

glasserm:
Revised fix_3_clean.m code: 

`%%%% compute movement regressors with noise components removed non aggressively
if domot == 0
  confounds = functionmotionconfounds(TR,hp);
  if aggressive == 1
    % aggressively regress out noise ICA components from movement regressors
    betaconfounds = pinv(ICA(:,DDremove),1e-6) * confounds;                              % beta for confounds (bad only)
    confounds = confounds - (ICA(:,DDremove) * betaconfounds);    % cleanup
  else
    % non-aggressively regress out noise ICA components from movement regressors
    betaconfounds = pinv(ICA,1e-6) * confounds;                              % beta for confounds (good *and* bad)
    confounds = confounds - (ICA(:,DDremove) * betaconfounds(DDremove,:));    % cleanup
  end
  save_avw(reshape(confounds',size(confounds,2),1,1,size(confounds,1)),'mc/prefiltered_func_data_mcf_conf_hp_clean','f',[1 1 1 TR]);
end
`

ICAFIX/hcp_fix_multi_run
```diff
@@ -825,6 +825,13 @@ then
     rm -f ${concatfmri}_Atlas.dtseries.nii ${concatfmri}_Atlas_hp${hp}.dtseries.nii
 fi
 
+# Convert sICA+FIX cleaned movement regressors to text
+if [ -f ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
+  fslmaths ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -Tmean -mul 0 -add 1 ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz
```
coalsont: This looks like the "mask" is an all-ones volume (not a brain mask or nonzero variance mask), is this intended, and if so, why is it needed at all?

ICAFIX/hcp_fix_multi_run
```diff
@@ -825,6 +825,13 @@ then
...
+  fslmeants -i ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -m ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz -o ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt --showall
+  cat ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt | tail -$(fslval ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz dim4) > ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt
```
coalsont: FYI, tail can take a filename as argument, and can be specified as "skip n lines" using +.

ICAFIX/hcp_fix_multi_run
```diff
@@ -859,13 +866,24 @@ for fmri in $fmris ; do
...
+  if [ -f ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
+    ${Caret7_Command} -volume-merge ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -volume ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -subvolume ${Start} -up-to ${Stop}
+    fslmeants -i ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -m ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz -o $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt --showall
+    cat $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt | tail -$(fslval ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz dim4) > $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt
+  fi
+
 	Start=`echo "${Start} + ${NumTPS}" | bc -l`
 done
 
+if [ -f ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
```
coalsont: Is there a case where the mask file exists, but the data doesn't?  If you just want to make sure the mask file isn't kept around, just use `rm -f` on the mask, no conditional.

ICAFIX/hcp_fix_multi_run
mharms: Agreed.  Not sure why L830 is there.

ICAFIX/hcp_fix_multi_run
```diff
@@ -859,13 +866,24 @@ for fmri in $fmris ; do
...
+
 	volume_out=${fmriNoExt}_hp${hp}_clean.nii.gz
 	${Caret7_Command} -volume-merge ${volume_out} -volume ${ConcatFolder}/${concatfmrihp}_clean.nii.gz -subvolume ${Start} -up-to ${Stop}
 	fslmaths ${volume_out} -div ${ConcatFolder}/${concatfmrihp}_vn -mul ${fmriNoExt}_hp${hp}_vn -add ${fmriNoExt}_mean ${volume_out}
+
+  if [ -f ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
+    ${Caret7_Command} -volume-merge ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -volume ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -subvolume ${Start} -up-to ${Stop}
+    fslmeants -i ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -m ${ConcatFolder}/${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean_mask.nii.gz -o $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt --showall
+    cat $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt | tail -$(fslval ${fmriNoExt}_hp${hp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz dim4) > $(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt
+  fi
```
mharms: Given that `${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt` has already been created, a seemingly much simpler way to accomplish this would be:
```
outFile=$(dirname $fmri)/Movement_Regressors_hp${hp}_clean.txt
sed -n "${Start},${Stop}p" ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt > ${outFile}
```

glasserm:
Overall many of the above changes fall in the category of probably accomplishing the same thing with different code, which I am not really sure is necessary (as you may see, I generally don't make these kind of comments on pull requests as it seems like it just generates extra work for the person implementing the request requiring more code changes and testing for the same result).  

1.  These data are already detrended as stated in the name.
2.  hcp_fix does not really need this code does it?  This was to allow combined processing of resting state and task data with MR+FIX but differential treatment of movement regressors.  This situation would not arise for SR+FIX.  Honestly, at this point SR+FIX is kind of deprecated in favor of MR+FIX, and I'm not sure it is worth spending a lot of time on it if we don't plan to recommend people use it.  
3.  That is true, but I expect we need to do it this way as it is probably not identical to compute derivatives after regression instead of before.

coalsont:
Confusing code is bad code in that it makes it more challenging to work on the code later, especially if it is someone else that has to work on it.  The suggested changes should be fairly easy to make and test.  Part of the point of code review is to find simpler and less confusing code to do the same thing.

ICAFIX/hcp_fix_multi_run
glasserm: It is not clear that fslmeants will work without a mask or coordinate, that is why there is a mask.

ICAFIX/hcp_fix_multi_run
glasserm: I am comfortable with the existing usage.

ICAFIX/hcp_fix_multi_run
glasserm: The code as implemented will work correctly, I don't want to generate errors by trying to remove a non existent mask.

ICAFIX/hcp_fix_multi_run
glasserm: I wanted the intermediate NIFTI file in place with the current implementation.

mharms:
Code readability and easy interpretability should be valued as well.

1. I don't see `_dt` in the file name.
2. Adding the analog to `hcp_fix` would allow one to do the "protected" motion regression within the context of SR-FIX.  It should be trivial to add, so why not make that option available (e.g., for people processing long single-run data).
3. Computing the backward differences might be fine, but the squared terms would indeed not be the same.

glasserm:
Well I responded to each comment above.  I don't agree with any of them in this case.  Both of you at times don't see that the costs of changing and testing stuff just for the sake of it are real.  I likely also don't always see the costs of code that is quickly written being confusing to others.  We do need to balance these things, but in this case the decisions I made above were well considered.

glasserm:
1.  _hp${Highpass}_clean = detrend
2.  Because everything takes time and this isn't a priority for me.
3.  The user would be able to compute derivatives after the fact or use the ones with regressed out timeseries, so this should work for both users.

mharms:
@glasserm In this case, both @coalsont and myself are indicating that we don't have the right balance.  The usage statement for `fslmeants` clearly indicates that the mask is an optional argument.  So that particular code is completely unnecessary (and outright to confusing to anyone that might be trying to understand why it is there).

ICAFIX/hcp_fix_multi_run
coalsont: Doesn't seem hard to test, and if it worked, it would entirely remove several lines (some of which were initially confusing).  Doesn't crash on an anatomical image, though it does take many seconds on fsl 5:
```
tim@timsdev:/mnt/myelin/tim/voltest$ fslmeants -i testLPI.nii --showall | tail -c 500
63 2.391223 2.394073 2.396367 2.397355 2.397202 2.396351 2.395081 2.393568 2.392322 2.392496 2.395739 2.403646 2.417208 2.436579 2.461160 2.489789 2.520775 2.551845 2.580199 2.602878 2.617405 2.622426 2.618012 2.605447 2.586682 2.563829 2.539038 2.514791 2.494272 2.481347 2.479870 2.492495 2.519450 2.557961 2.602637 2.646768 2.684010 2.709814 2.722121 2.721240 2.709162 2.688713 2.662853 2.634269 2.605276 2.577890 2.553897 2.534770 2.521444 2.513808 2.510787 2.510695 2.511874 2.513183 2.514090
```
It appears to be outputting voxel indices first, followed by the values, as its output word count is exactly 4 times the number of voxels.

glasserm:
Here is the usage of fslmeants: 

fslmeants -i filtered_func_data -o meants.txt -m my_mask
fslmeants -i filtered_func_data -m my_mask
fslmeants -i filtered_func_data -c 24 19 10

As you can see, 
fslmeants -i filtered_func_data --showall
is not one of the given choices.  I interpreted "optional" in this case to mean that either -m or -c was required to be specified.  I could be wrong, but elected to implement what I knew would work, rather than what might not work.

ICAFIX/hcp_fix_multi_run
mharms: Why do we want a NIFTI version of the movement regressors lying around?  If it serves some purpose, can you add a comment explaining what its value is?

ICAFIX/hcp_fix_multi_run
coalsont: `rm -f` will never give an error that a filename doesn't exist:

```
tim@timsdev:~$ rm -f
tim@timsdev:~$ rm -f kahgsdkfjghakdjfgasf
tim@timsdev:~$ 
```

Checking for the existence of a file other than the one you are removing is confusing, and you haven't clarified if that particular logic actually needs to be that way.  Do you ever want a file that has the mask name to be kept past that point in the script?

coalsont:
I tested --showall without -m for you, since you were unwilling.  Unless the output format is different, it seems to have worked.  None of the help examples use --showall, but you were willing to try that option and see what happened...

I don't think changing the `cat ... | tail` is important, but the other lines definitely are more confusing than what they are doing, and I think they either deserve edits to make them simpler, or edits to add comments explaining why the code isn't simpler.

mharms:
Unrelated, but Sri just coincidentally brought this to my attention, as he was attempting to ascertain some timing info from the log files.  

Can we move this bit of code in `hcp_fix_multi_run`
```
#grab melodic from $PATH by default, don't hardcode it with respect to $FSLDIR
#we need to do "if which ..." because the script currently uses ERR trap
if which melodic &> /dev/null
then
    MELODIC=$(which melodic 2> /dev/null)
else
    #if it isn't even in $PATH, fall back on FSLDIR
    MELODIC="${FSLDIR}/bin/melodic"
fi

log_Msg "Running MELODIC located at: $MELODIC"
log_Debug_Msg "Beginning of melodic version log, help, and checksum"
if [ ! -z "${log_debugOn}" ] ; then
	log_Debug_Msg "$MELODIC --version"
	$MELODIC --version
	log_Debug_Msg "$MELODIC --help"
	$MELODIC --help
	log_Debug_Msg "md5sum $MELODIC"
	md5sum $MELODIC
fi
log_Debug_Msg "End of melodic version log, help, and checksum"
```
down to just after here:
```
## ---------------------------------------------------------------------------
## Run melodic on concatenated file
## ---------------------------------------------------------------------------
```
so that the message about "Running MELODIC" (and version info if `log_debugOn` exists) actually occurs right before we run melodic.

thx

glasserm:
I tested it without the mask and it worked, so I removed those lines.  As for this other change, feel free to make it yourself and it does not require a pull request.  I am going to merge this unless there is any last comment this evening.

PR #166, opened 2020-02-24
Bug fix to cleaned movement regressor code, plus minor simplification

mharms:
This fixes a bug in #164 if `ConcatFolder` is not specified as an absolute path. And implements a suggestion that @coalsont had regarding the `tail` command.  It has been tested.

### DIFF
--- a/ICAFIX/hcp_fix_multi_run
+++ b/ICAFIX/hcp_fix_multi_run
@@ -827,8 +827,11 @@ fi
 
 # Convert sICA+FIX cleaned movement regressors to text
 if [ -f ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz ] ; then
-  fslmeants -i ${concatfmrihp}.ica/mc/prefiltered_func_data_mcf_conf_hp_clean.nii.gz -o ${ConcatFolder}/Movement_Regressors_hp${hp}_clean.txt --showall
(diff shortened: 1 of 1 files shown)


coalsont:
The tail would be better without the contortions required because the line of code chooses to overwrite its own input file (I am guessing the previous code may only have worked because the file was smaller than the buffer used by pipes and/or cat).  If fslmeants by default outputs the same thing to stdout as it does with `-o` to a text file, you could pipe it directly to tail instead and avoid the ugliness and extra step.

mharms:
`fslmeants` without the -o unfortunately includes an extra (blank) newline at the end, so no, it isn't the same output.  I think the reason the previous code worked is that it happened to invoke a subshell as part of the `tail` command as well.  FWIW, the `sed` one-liner to extract lines from the _end_ of a file is a complex mess.
So, I think we either `echo "$(tail ... file.txt)" > file.txt` (as currently, which I think is rather self-explanatory in light of the comment that I added), or I can name the first file `file.txt.tmp`, apply `tail -n $nVols file.txt.tmp > file.txt`, and then delete `file.txt.tmp`.

coalsont:
FYI, `$()` deletes trailing newlines.  I am fine with either of those solutions.

mharms:
Are you saying you are fine with it "as proposed"?

coalsont:
Yes, it is okay as proposed.

mharms:
Great.  I think this is ready to merge then.

PR #170, opened 2020-04-02
Enable Single Regression in RSNRegression

glasserm:
Enable a new 'single' mode of temporal regression for RSNRegression to enable applying existing results to a new CIFTI space or single regression in other contexts

### DIFF
--- a/global/scripts/RSNregression.m
+++ b/global/scripts/RSNregression.m
@@ -183,7 +183,7 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
             NODEts=NODEts.cdata';
             betaICA = ((pinv(normalise(NODEts)) * demean(inputConcat')))';
         otherwise
-            error(['unrecognized method: "' Method '", use "weighted," "dual," or "single"']);
+            error(['unrecognized method: "' Method '", use "weighted", "dual", or "single"']);
     end
     
     NODEtsnorm = normalise(NODEts);
--- a/global/scripts/RSNregression.m
+++ b/global/scripts/RSNregression.m
@@ -1,5 +1,5 @@
-function RSNregression(InputFile, InputVNFile, GroupMaps, Method, ParamsFile, VAWeightsName, OutputBeta, varargin)
-    optional = myargparse(varargin, {'VolInputFile' 'VolInputVNFile' 'VolCiftiTemplate' 'OldBias' 'OldVolBias' 'GoodBCFile' 'VolGoodBCFile' 'SpectraParams' 'OutputZ' 'OutputVolBeta' 'OutputVolZ' 'SurfString' 'ScaleFactor' 'WRSmoothingSigma'});
+function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, varargin)
+    optional = myargparse(varargin, {'GroupMaps' 'VAWeightsName' 'VolInputFile' 'VolInputVNFile' 'VolCiftiTemplate' 'OldBias' 'OldVolBias' 'GoodBCFile' 'VolGoodBCFile' 'SpectraParams' 'OutputZ' 'OutputVolBeta' 'OutputVolZ' 'SurfString' 'ScaleFactor' 'WRSmoothingSigma'});
     
     %InputFile - text file containing filenames of timeseries to concatenate
     %InputVNFile - text file containing filenames of the variance maps of each input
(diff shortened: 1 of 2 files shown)


global/scripts/RSNregression.m
```diff
@@ -174,15 +176,21 @@ function RSNregression(InputFile, InputVNFile, GroupMaps, Method, ParamsFile, VA
...
             [betaICA, NODEts] = weightedDualRegression(GroupMapcii.cdata, inputConcat, AreaWeights);
+        case 'single'
+            SpectraArray = textscan(optional.SpectraParams, '%s', 'Delimiter', {'@'});
+            InputSpectraTS = SpectraArray{1}{1};
+            NODEts=ciftiopen(InputSpectraTS, wbcommand);
+            NODEts=NODEts.cdata';
+            betaICA = ((pinv(normalise(NODEts)) * demean(inputConcat')))';
         otherwise
-            error(['unrecognized method: "' Method '", use "weighted" or "dual"']);
+            error(['unrecognized method: "' Method '", use "weighted," "dual," or "single"']);
```
coalsont, line 186: I would put the commas outside the quotes to reduce confusion, this isn't a manuscript.

global/scripts/RSNregression.sh
```diff
@@ -198,18 +210,27 @@ fi
...
         ;;
     (1 | 2)
         #SurfString and WRSmoothingSigma are optional to the matlab function (needed only for weighted), but we will always have it, so always providing it is simpler here
-        matlab_args="'$tempname.input.txt', '$tempname.inputvn.txt', '$GroupMaps', '$Method', '$tempname.params.txt', '$tempfile.91k.dscalar.nii', '$OutBeta', 'SurfString', '$SurfString', 'WRSmoothingSigma', '$WRSmoothingSigma'"
+        matlab_args="'$tempname.input.txt', '$tempname.inputvn.txt', '$Method', '$tempname.params.txt', '$OutBeta', 'SurfString', '$SurfString', 'WRSmoothingSigma', '$WRSmoothingSigma'"
+
+        if [[ ! ${Method} == "single" ]]
+        then
+        matlab_args+=", 'GroupMaps', '$GroupMaps'"
+        matlab_args+=", 'VAWeightsName', '$tempfile.91k.dscalar.nii'"
```
coalsont, line 232: nitpick: indentation

glasserm:
Any other comments?

coalsont:
Personally, I wouldn't have reused SpectraParams to specify an input file in this mode.

glasserm:
Since it is exclusive, already there, and germane it seemed faster.

glasserm:
Since this is testing and working and I want to run HCP Processing with it now, I will go ahead and commit it.

mharms:
We need to be careful with the versioning. `version.txt` needs to be updated to `Post-v4.2.0` as part of this.  Relatedly, the "rc.1" tag from v4.2.0 can be dropped, and `v4.2.0` released.

glasserm:
That will be annoying to change the version with every commit.  Can we do something more automated?

coalsont:
If it needs to work in the zip files offered on the releases page, then not really.

glasserm:
How do we handle this in Connectome Workbench?

coalsont:
Manually there also, since debian doesn't keep the git stuff in the source tree.

mharms:
Ok, I just noticed that this was a change to a .m file.  @glasserm : I've made this point before, but we can't be making changes to matlab code and merging it into the master without simultaneously recompiling the matlab.  We now have a situation in which the compiled matlab is not in sync with the interpreted matlab.  I'll leave it to you to work with Mike Hodge to resolve this.  IMO, it would have been better to leave this change in a branch and use that for your immediate needs.  Then we could deal with the compiled matlab issue on a more leisurely time frame.

glasserm:
I don't think we should support compiled matlab then.

coalsont:
That script doesn't currently support compiled matlab, as it happens, so it technically isn't out of sync.

mharms:
That's not a solution.  The solution is simply to make use of a branch in your processing.

glasserm:
The feature is ready for routine use by everyone.  Unless compiled matlab can be automatically built upon commit (like Connectome Workbench is), it isn't feasible to support.  We have already seen that compiled matlab has major problems with memory usage.  We should be supporting interpreted matlab and a recipe for building Octave without problems + the container.

mharms:
I have no doubt that Octave is going to have a host of problems that we are completely clueless about currently.  Dropping support for compiled matlab at this point in time would likely be a disaster.  Fortunately, as @coalsont pointed out, `RSNRegression` doesn't even have a compiled counterpart at this point in time.  If dropping support for compiled matlab is a goal, then someone is going to have to put in a bunch of time validating all the code in Octave.  Until that is done, changes to matlab scripts that have a compiled version need to be accompanied by an updated compiled binary as well.

glasserm:
That is not feasible without an automated build system.  It is reasonable for releases but not for every commit to master.

mharms:
> The feature is ready for routine use by everyone. 
BTW: I think you considerably overestimate the number of people that are running the matlab stuff in interpreted mode.  Most people using a cluster are probably going to need to be using either compiled matlab or Octave.

mharms:
> That is not feasible without an automated build system. It is reasonable for releases but not for every commit to master.

Yes, it isn't ideal to have to worry about keeping things in sync, but we can't drop compiled matlab at this point in time; not if you want others to be able to use those scripts on their cluster computing systems (or even just without needing a potentially expensive matlab license on their local computer).

glasserm:
I don't really have anything else to say about this.  Things will not be in sync in master except before releases or we need an automated build system.  No one manually compiles software and uploads it to a repo with every commit.

coalsont:
I think the point of the master matlab compile script was so that when we are about ready to release, we can get them all synced then.  Does our documentation give the impression that the master branch is the first version for users to try?

mharms:
> No one manually compiles software and uploads it to a repo with every commit.

FWIW, the binaries really shouldn't be part of the repo anyway, but that's an organizational discussion for another time.  Storing a history of all the binary changes as part of the repo bloats its size.

PR #171, opened 2020-04-06
Change fmri volume deletes

glasserm:
See comments in new code

### DIFF
--- a/fMRIVolume/scripts/IntensityNormalization.sh
+++ b/fMRIVolume/scripts/IntensityNormalization.sh
@@ -227,7 +227,7 @@ if [ X${ScoutInput} != X ] ; then
 fi
 
 #Basic Cleanup
-rm ${InputfMRI}.nii.*
+#rm ${InputfMRI}.nii.* #Don't delete the unmasked spatially corrected unormalized data by default
 
 echo " "
 echo "END: IntensityNormalization"
--- a/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
+++ b/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
@@ -1106,8 +1106,8 @@ ${FSLDIR}/bin/imrm ${fMRIFolder}/${NameOffMRI}_nonlin_norm
 
 #Econ
 #${FSLDIR}/bin/imrm "$fMRIFolder"/"$OrigTCSName"
-#${FSLDIR}/bin/imrm "$fMRIFolder"/"$NameOffMRI"_gdc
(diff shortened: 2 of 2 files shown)


fMRIVolume/scripts/IntensityNormalization.sh
```diff
@@ -227,7 +227,7 @@ if [ X${ScoutInput} != X ] ; then
 fi
 
 #Basic Cleanup
-rm ${InputfMRI}.nii.*
+#rm ${InputfMRI}.nii.* #Don't delete the unmasked spatially corrected unormalized data by default
```
mharms, line 230: Do we really even need to save this?  Can't we assess the quality of the motion correction sufficiently well from the masked `${OutputfMRI}`?

Also, if we do delete it, why is this line written as it is (with the wildcard after the `.nii.`)?  i.e., Why not `${FSLDIR}/bin/imrm ${InputfMRI}`?  What other files are there other than the .nii.gz itself?

glasserm: We did just need this for some testing, which is why I made this set of changes.  The net is to reduce the footprint of the pipelines and we specifically implemented code to ensure that this file remains int32 to reduce its file size.

mharms: What in `OneStepResampling` forces that file to be int32?

glasserm: Have a look at Tim's new pull request.  It was int32 based on dcm2nii and subsequent FSL commands.

coalsont: There is a new, unreleased option to wb_command to set the output datatype, and the new workbench steps in OneStepResample are hardcoded to always output int32 (which is what the input type is in HCP_YA 3T and 7T data).  Yes, FSL instead takes the datatype from a specific input file, which if we wanted workbench to do, would have to be implemented one-at-a-time in the various wb_command volume operations.

Another possibility would be to make an option for output datatype (and scaling?) that takes a filename, instead of a literal type.

mharms: The specifics of that particular line aren't really important if we aren't deleting the file, although I'm still puzzled why it wasn't written using `imrm`, or alternatively as `rm ${InputfMRI}.nii.gz` since other aspects of the code that use `rm` rather than `imrm` use `.nii.gz` as the extension rather than `.nii.*`

glasserm: I'm puzzled too and I even wrote it...

PR #173, opened 2020-04-08
workbench implementation for OneStepResampling.sh

coalsont:
This cannot be used by the public yet (not easily, anyway), as the -volume-resample command is not released, nor is the control of volume output datatype (added to resolve a considerable file size change between applywarp and wb_command output), so it is disabled by default.  Currently, enabling it requires editing the GenericfMRIVolume pipeline.

Once workbench has a new release, it can be made the default, or the fslsplit/loop/fslmerge code can be removed entirely.

### DIFF
--- a/fMRIVolume/scripts/MotionCorrection.sh
+++ b/fMRIVolume/scripts/MotionCorrection.sh
@@ -181,7 +181,7 @@ function DeriveBackwards {
   # TCS becomes an array of the values from column $i in $in (derived from Var)
   TCS=($Var)
   # random is a random file name for temporary output
-  random="$(mktemp --tmpdir MotionCorrectionRandom_XXXXXX.txt)"
+  random="$(tempfiles_create MotionCorrectionRandom_XXXXXX.txt)"
 
   # Cycle through our array of values from column $i
   j=0
@@ -209,7 +209,6 @@ function DeriveBackwards {
   done
   paste -d " " $out $random > ${out}_
   mv ${out}_ ${out}
-  rm $random
 }
 
 # Run the Derive function to generate appropriate regressors from the par file
--- a/fMRIVolume/scripts/MotionCorrection.sh
+++ b/fMRIVolume/scripts/MotionCorrection.sh
@@ -46,7 +46,8 @@ if [ -z "${HCPPIPEDIR}" ]; then
 fi
 
 source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
-source ${HCPPIPEDIR}/global/scripts/opts.shlib                 # Command line option functions
+source "${HCPPIPEDIR}/global/scripts/opts.shlib"               # Command line option functions
+source "${HCPPIPEDIR}/global/scripts/tempfiles.shlib"          # handle temporary files
 
 opts_ShowVersionIfRequested $@
 
--- a/fMRIVolume/scripts/MotionCorrection.sh
+++ b/fMRIVolume/scripts/MotionCorrection.sh
@@ -181,7 +181,7 @@ function DeriveBackwards {
   # TCS becomes an array of the values from column $i in $in (derived from Var)
   TCS=($Var)
   # random is a random file name for temporary output
-  random=$RANDOM
+  random="$(mktemp --tmpdir MotionCorrectionRandom_XXXXXX.txt)"
 
   # Cycle through our array of values from column $i
   j=0
--- a/fMRIVolume/scripts/OneStepResampling.sh
+++ b/fMRIVolume/scripts/OneStepResampling.sh
@@ -138,6 +138,8 @@ fMRIReferenceReg=`getopt1 "--fmrirefreg" $@`  # "${20}"
 #hidden: toggle for unreleased new resampling command, default off
 useWbResample=`getopt1 "--wb-resample" $@`
 useWbResample="${useWbResample:-0}"
+#with wb_command -volume-resample, the warpfields and per-frame motion affines do not need to be combined in advance,
+#and the timeseries can be resampled without splitting into one-frame files, resulting in much less file IO
 
 case "$(echo "$useWbResample" | tr '[:upper:]' '[:lower:]')" in
     (yes | true | 1)
@@ -260,11 +262,9 @@ ${FSLDIR}/bin/imcp ${WD}/${T1wImageFile}.${FinalfMRIResolution} ${fMRIFolder}/${
 ${FSLDIR}/bin/imcp ${WD}/${FreeSurferBrainMaskFile}.${FinalfMRIResolution} ${fMRIFolder}/${FreeSurferBrainMaskFile}.${FinalfMRIResolution}
 ${FSLDIR}/bin/imcp ${WD}/${BiasFieldFile}.${FinalfMRIResolution} ${fMRIFolder}/${BiasFieldFile}.${FinalfMRIResolution}
 
-
 if ((useWbResample))
 then
-    affseries="$(mktemp --tmpdir OneStepResampleAffSeries_XXXXXX.txt)"
-    tempfiles_add "$affseries"
+    affseries="$(tempfiles_create OneStepResampleAffSeries_XXXXXX.txt)"
     
     for ((k=0; k < $NumFrames; k++))
     do
@@ -284,18 +284,19 @@ then
     done
     
     #gdc warp space is input to input, affine is input to input, OutputTransform is input to MNI
-    xfmargs=(-warp "$GradientDistortionField.nii.gz" -fnirt "$InputfMRI"
+    xfmargs=(-warp "$GradientDistortionField".nii.gz -fnirt "$InputfMRI"
              -affine-series "$affseries" -flirt "$InputfMRI" "$InputfMRI"
-             -warp "$OutputTransform.nii.gz" -fnirt "$InputfMRI")
+             -warp "$OutputTransform".nii.gz -fnirt "$InputfMRI")
     
-    wb_command -volume-resample "$InputfMRI" "$WD/$T1wImageFile.$FinalfMRIResolution.nii.gz" CUBIC "$OutputfMRI.nii.gz" "${xfmargs[@]}" -nifti-output-datatype INT32
+    wb_command -volume-resample "$InputfMRI" "$WD/$T1wImageFile.$FinalfMRIResolution".nii.gz CUBIC "$OutputfMRI".nii.gz "${xfmargs[@]}" -nifti-output-datatype INT32
     
     #resample all-ones volume series with enclosing voxel to determine FOV coverage
     #yes, this is the entire length of the timeseries on purpose
-    fovcheck="$(mktemp --tmpdir OneStepResampleFovCheck_XXXXXX.nii.gz)"
-    tempfiles_add "$fovcheck"
+    fovcheck="$(tempfiles_create OneStepResampleFovCheck_XXXXXX.nii.gz)"
     wb_command -volume-math '1' "$fovcheck" -var x "$InputfMRI"
-    wb_command -volume-resample "$fovcheck" "$WD/$T1wImageFile.$FinalfMRIResolution.nii.gz" ENCLOSING_VOXEL "${OutputfMRI}_mask.nii.gz" "${xfmargs[@]}" -nifti-output-datatype INT32
+    #fsl's "nn" interpolation is just as pessimistic about the FoV as non-extrapolating interpolation
+    #so, in wb_command, use trilinear here instead of enclosing voxel to get the same FoV behavior as CUBIC
+    wb_command -volume-resample "$fovcheck" "$WD/$T1wImageFile.$FinalfMRIResolution".nii.gz TRILINEAR "$OutputfMRI"_mask.nii.gz "${xfmargs[@]}" -nifti-output-datatype INT32
 
 else
     mkdir -p ${WD}/prevols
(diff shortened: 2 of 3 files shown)


fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -22,7 +22,6 @@ Usage: ${script_name} [options]
   --t1=<input T1w restored image>
   --fmriresout=<output resolution for images, typically the fmri resolution>
   --fmrifolder=<fMRI processing folder>
-  --atlasspacedir=<output directory for several resampled images>
```
glasserm, line 25: Why was this removed?

coalsont: Because the option doesn't exist.

coalsont:
For reference, the usage of the unreleased -volume-resample:

```$ wb_command -volume-resample 
TRANSFORM AND RESAMPLE A VOLUME FILE
   wb_command -volume-resample
      <volume-in> - volume to resample
      <volume-space> - a volume file in the volume space you want for the
         output
      <method> - the resampling method
      <volume-out> - output - the output volume

      [-affine] - repeatable - add an affine transform
         <affine> - the affine file to use

         [-flirt] - MUST be used if affine is a flirt affine
            <source-volume> - the source volume used when generating the affine
            <target-volume> - the target volume used when generating the affine

      [-affine-series] - repeatable - add an independent affine per-frame
         <affine-series> - text file containing 12 or 16 numbers per line, each
            being a row-major flattened affine

         [-flirt] - MUST be used if affine is a flirt affine
            <source-volume> - the source volume used when generating the affine
            <target-volume> - the target volume used when generating the affine

      [-warp] - repeatable - add a nonlinear warpfield transform
         <warpfield> - the warpfield file

         [-fnirt] - MUST be used if using a fnirt warpfield
            <source-volume> - the source volume used when generating the
               warpfield

      Resample a volume file with an arbitrary list of transformations.  You
      may specify -affine, -warp, and -affine-series multiple times each, and
      they will be used in the order specified.  For instance, for rigid motion
      correction followed by nonlinear atlas registration, specify
      -affine-series first, then -warp.  The recommended methods are CUBIC
      (cubic spline) for most data, and ENCLOSING_VOXEL for label data.  The
      parameter <method> must be one of:

      CUBIC
      ENCLOSING_VOXEL
      TRILINEAR
```

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -244,50 +260,93 @@ ${FSLDIR}/bin/imcp ${WD}/${T1wImageFile}.${FinalfMRIResolution} ${fMRIFolder}/${
...
+        if [ $k -eq 0 ] ; then
+            echo "0" >> ${fMRIFolder}/Movement_RelativeRMS.txt
+        else
+            rmsdiff ${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum} $prevmatrix ${ScoutInputgdc} ${ScoutInputgdc}_mask.nii.gz | tail -n 1 >> ${fMRIFolder}/Movement_RelativeRMS.txt
+        fi
+        prevmatrix="${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}"
+    done
+    
+    #gdc warp space is input to input, affine is input to input, OutputTransform is input to MNI
+    xfmargs=(-warp "$GradientDistortionField.nii.gz" -fnirt "$InputfMRI"
```
glasserm, line 287: I prefer ${}, particularly when relying on non text for the parsing into variable/non-variable.  I was confused reading this line.

coalsont: Another option would be `"$GradientDistortionField".nii.gz`.  Quotes do something useful, braces generally do not.

glasserm: Yes that would be clearer too.

fMRIVolume/scripts/MotionCorrection.sh
```diff
@@ -181,7 +181,7 @@ function DeriveBackwards {
   # TCS becomes an array of the values from column $i in $in (derived from Var)
   TCS=($Var)
   # random is a random file name for temporary output
-  random=$RANDOM
+  random="$(mktemp --tmpdir MotionCorrectionRandom_XXXXXX.txt)"
```
mharms, line 184: `--tmpdir` isn't a supported option on a standard MacOS installation.

mharms: Also, if going this route, should we also make use of the `tempfiles.shlib` in this script as well?

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -135,6 +135,22 @@ JacobianOut=`getopt1 "--ojacobian" $@`  # "${18}"
 fMRIReferencePath=`getopt1 "--fmrirefpath" $@` # "${19}"
 fMRIReferenceReg=`getopt1 "--fmrirefreg" $@`  # "${20}"
 
+#hidden: toggle for unreleased new resampling command, default off
```
mharms, line 138: Maybe add a comment explaining the advantage of the `--wb-resample` approach?

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -244,50 +260,93 @@ ${FSLDIR}/bin/imcp ${WD}/${T1wImageFile}.${FinalfMRIResolution} ${fMRIFolder}/${
...
-  ${FSLDIR}/bin/convertwarp --relout --rel --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --warp1=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_gdc_warp.nii.gz --warp2=${OutputTransform} --out=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_all_warp.nii.gz
-  # Apply one-step warp, using spline interpolation
-  ${FSLDIR}/bin/applywarp --rel --interp=spline --in=${WD}/prevols/vol${vnum}.nii.gz --warp=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_all_warp.nii.gz --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --out=${WD}/postvols/vol${vnum}.nii.gz
 
-  # Generate a mask for keeping track of spatial coverage (use nearest neighbor interpolation here)
-  ${FSLDIR}/bin/fslmaths ${WD}/prevols/vol${vnum}.nii.gz -mul 0 -add 1 ${WD}/prevols/vol${vnum}_mask.nii.gz
-  ${FSLDIR}/bin/applywarp --rel --interp=nn --in=${WD}/prevols/vol${vnum}_mask.nii.gz --warp=${MotionMatrixFolder}/${MotionMatrixPrefix}${vnum}_all_warp.nii.gz --ref=${WD}/${T1wImageFile}.${FinalfMRIResolution} --out=${WD}/postvols/vol${vnum}_mask.nii.gz
+if ((useWbResample))
+then
+    affseries="$(mktemp --tmpdir OneStepResampleAffSeries_XXXXXX.txt)"
```
mharms, line 266: Ditto comment about `--tmpdir` on MacOS.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -244,50 +260,93 @@ ${FSLDIR}/bin/imcp ${WD}/${T1wImageFile}.${FinalfMRIResolution} ${fMRIFolder}/${
...
+    #gdc warp space is input to input, affine is input to input, OutputTransform is input to MNI
+    xfmargs=(-warp "$GradientDistortionField.nii.gz" -fnirt "$InputfMRI"
+             -affine-series "$affseries" -flirt "$InputfMRI" "$InputfMRI"
+             -warp "$OutputTransform.nii.gz" -fnirt "$InputfMRI")
+    
+    wb_command -volume-resample "$InputfMRI" "$WD/$T1wImageFile.$FinalfMRIResolution.nii.gz" CUBIC "$OutputfMRI.nii.gz" "${xfmargs[@]}" -nifti-output-datatype INT32
+    
+    #resample all-ones volume series with enclosing voxel to determine FOV coverage
+    #yes, this is the entire length of the timeseries on purpose
+    fovcheck="$(mktemp --tmpdir OneStepResampleFovCheck_XXXXXX.nii.gz)"
```
mharms, line 295: Ditto `--tmpdir`

glasserm:
${OutputfMRI} is equivalent for processing.  Perhaps Tim can check the other two quickly.

fMRIVolume/scripts/MotionCorrection.sh
coalsont: Hmm...I had already used `--tmpdir` elsewhere, that is unfortunate...

`tempfiles.shlib` only contains functions for cleanup, it doesn't actually generate any temporary file names currently.  If we need some compatibility workaround on Mac, then maybe it should go in there.

coalsont:
The mask from the workbench resampling has an extra partial voxel on its FoV (the difference image has an interrupted line of voxels on the visible parts of the original FoV edge).  I was expecting less of a difference than that, but it is true that our enclosing voxel method is half a voxel more generous than interpolation (which happens to also be 1% of a voxel more generous than the naive approach, on purpose).  Trilinear should actually be much closer to the fsl "nn" result for this case of all-ones input, while still being much faster (mostly by avoiding two thousand convertwarp calls - these could probably be reduced to one thousand by using --midmat in convertwarp).

Movement_AbsoluteRMS.txt and Movement_RelativeRMS.txt are identical.  The _mean files happen to be missing from my testing folder at the moment, but that is because I interrupted the most recent testing run while trying to pin down something else, they should also be identical.

fMRIVolume/scripts/MotionCorrection.sh
coalsont: On the old mac system I can easily test on, mktemp doesn't even allow generating a temporary filename with a desired file extension:

```
$ mktemp foo-XXXXXXX.txt
foo-XXXXXXX.txt
$ mktemp -t foo-XXXXXXX.txt
/tmp/foo-XXXXXXX.txt.laTQHK7E
```

This is pretty disappointing.  I guess I can write a function that uses --tmpdir on everything except mac, and fake it on mac, but this means I have to detect mac systems reliably.

mharms: I know `mktemp` may be elegant, but why not just stick with the existing code of writing to a file named from `$RANDOM`?  We are deleting the file anyway.

coalsont: I edited it because in my test run, that folder was not writable.  Using a folder that is always writable for all temporaries is better practice than ever using the current directory, and using $RANDOM as a mechanism for trying to avoid filename collision is bad when there is already something designed for specifically that purpose.

coalsont:
I have pushed edits to resolve these concerns, currently being tested.  However, it appears the change to trilinear interpolation for the FoV coverage mask didn't change as much as I expected, there is still an extra half voxel or so on parts of the edge.  I'm not really sure what to make of that.

I also hadn't realized it, but on one edge (P), the FSL mask has an extra half voxel or so, rather than the workbench result.  Chasing this down could get tricky, how worried about this are we?

Anecdotal timing info, one run of only OneStepResample, 3T REST1_LR:
new wb_command code:
```
real	12m28.293s
user	27m43.616s
sys	0m14.204s
```
old applywarp code:
```
over an hour so far, looks like probably 4 hours total, will update when it finishes
```
I didn't notice anything indicating problems with nfs bandwidth/latency, but the applywarp time should be checked against prior timings of GenericfMRIVolume for sanity.  The applywarp code could be made somewhat faster by using --midmat in convertwarp, but it would still do fslsplit, fslmerge, and one convertwarp per frame.  The different ref spaces should be fine (just use the final one), since each transform matches its input space to the previous output space.

As an aside, the practical advantage of the wb_command -fnirt and -flirt setup (besides identity transforms actually meaning what they say, and being able to transform surface files) is that it can use transforms that do NOT match every step's output grid to the next input grid (remember the applyisoxfm stuff to make different resolutions play nice? that doesn't matter to workbench).  If we made "world" convention versions of transforms that we planned to use in wb_command, then we wouldn't need to use -flirt and -fnirt arguments (or remember what specific spaces they use).

fMRIVolume/scripts/OneStepResampling.sh
coalsont: I added a comment under the option parsing lines - I expect when workbench is released, we will just remove the applywarp code path.

coalsont:
For clarity, here is a capture of the current mask difference image (new workbench minus old applywarp), red positive, blue negative, black zero:

maskdiff

glasserm:
I think the results are close enough.

fMRIVolume/scripts/OneStepResampling.sh
mharms: At least initially, I would suggest we leave the applywarp code path, but make `useWbResample=1` the default instead.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -284,18 +284,19 @@ then
...
     
     #resample all-ones volume series with enclosing voxel to determine FOV coverage
     #yes, this is the entire length of the timeseries on purpose
-    fovcheck="$(mktemp --tmpdir OneStepResampleFovCheck_XXXXXX.nii.gz)"
-    tempfiles_add "$fovcheck"
+    fovcheck="$(tempfiles_create OneStepResampleFovCheck_XXXXXX.nii.gz)"
     wb_command -volume-math '1' "$fovcheck" -var x "$InputfMRI"
-    wb_command -volume-resample "$fovcheck" "$WD/$T1wImageFile.$FinalfMRIResolution.nii.gz" ENCLOSING_VOXEL "${OutputfMRI}_mask.nii.gz" "${xfmargs[@]}" -nifti-output-datatype INT32
+    #fsl's "nn" interpolation is just as pessimistic about the FoV as non-extrapolating interpolation
+    #so, in wb_command, use trilinear here instead of enclosing voxel to get the same FoV behavior as CUBIC
```
mharms, line 298: Maybe add to the this comment: `(still not identical, but TRILINEAR closer than ENCLOSING_VOXEL)`

PR #174, opened 2020-04-15
Cifti matlab settings and rearrangement

coalsont:
The rearrangement has not been tested at all, do not merge until tested.


ICAFIX/PostFix.sh
```diff
@@ -409,7 +409,7 @@ main()
 
 	# For interpreted modes, make sure that matlab/octave have access to the functions they need
 	export FSL_MATLAB_PATH="${FSLDIR}/etc/matlab"
-	local ML_PATHS="addpath('${FSL_MATLAB_PATH}'); addpath('${MATLAB_GIFTI_LIB}'); addpath('${this_script_dir}/scripts');"
+	local ML_PATHS="addpath('${FSL_MATLAB_PATH}'); addpath('$HCPCIFTIRWDIR'); addpath('${MATLAB_GIFTI_LIB}'); addpath('${this_script_dir}/scripts');"
```
glasserm, line 412: Should probably name the MATLAB_GIFTI_LIB variable consistently with other functions.

coalsont: The other scripts in ICAFIX/ use `'${HCPPIPEDIR}/global/matlab'`, not a variable.  The folder does contain the gifti library, so it isn't entirely inaccurate.  What are you suggesting, specifically?

glasserm: I think Tim brown wrote that before he understood as much.  That code does not use gifti, it is what we reference in the other scripts as '${HCPPIPEDIR}/global/matlab'.

glasserm:
Lets call the new folder 'cifti-legacy'

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -501,10 +501,10 @@ main()
 	# Several additional necessary environment variables (e.g., ${FSL_FIX_CIFTIRW} and ${FSL_FIX_WBC})
 	# are set in ${FSL_FIXDIR}/settings.sh, which is sourced below for interpreted modes.
 	# Note that fix_3_clean.m *appends* ${FSL_FIX_CIFTIRW} to the matlab path (i.e., the ciftiopen.m, ciftisave.m functions).
-	# BUT the CIFTI I/O functions are also now included within '${HCPPIPEDIR}/global/matlab' as well.
-	# AND since 'addpath' *prepends* to the matlab path, the versions in '${HCPPIPEDIR}/global/matlab' will therefore take precedence!
+	# The pipelines CIFTI I/O functions have been moved to a subdirectory, and the SetUp... script selects what to use with HCPCIFTIRWDIR.
+	# AND since 'addpath' *prepends* to the matlab path, the versions in '${HCPPIPEDIR}/global/matlab' will therefore take precedence.
```
mharms, line 505: Comment should be changed to:
```suggestion
	# AND since 'addpath' *prepends* to the matlab path, the versions in '${HCPCIFTIRWDIR}' will therefore take precedence.
```

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -395,8 +395,8 @@ main()
 	# are set in ${FSL_FIXDIR}/settings.sh, which is sourced below for interpreted modes.
 	# Note that the ciftiopen.m, ciftisave.m functions are *appended* to the path through the ${FSL_FIX_CIFTIRW} 
 	# environment variable within fix_3_clean.m itself.
-	# Note that we are NOT adding '${HCPPIPEDIR}/global/matlab' to the path (which also contains the
-	# CIFTI I/O functions as well), so the versions in ${FSL_FIX_CIFTIRW} will be the ones that get used
+	# Note that we are NOT adding '${HCPPIPEDIR}/global/matlab' to the path (which used to contain the
+	# CIFTI I/O functions as well), or HCPCIFTIRWDIR, so the versions in ${FSL_FIX_CIFTIRW} will be the ones that get used
```
mharms, line 399: Should we consider changing this behavior at this time?  i.e., Make the CIFTI I/O in this script use what is specified by `${HCPCIFTIRWDIR}` rather than `${FSL_FIX_CIFTIRW}` (for interpreted mode)?

ICAFIX/ReApplyFixMultiRunPipeline.sh
glasserm: well it all depends on the order of the addpath commands...

ICAFIX/ReApplyFixPipeline.sh
glasserm: I think it would be reasonable to override anything in the FIX settings.

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: Well, `${HCPPIPEDIR}/global/matlab` will have precedence in the path, but it won't contain any CIFTI I/O functions, which is what is being referred to here.  But, I see the potential source of confusion regarding my suggestion, and have edited it appropriately.

ICAFIX/scripts/Compile_MATLAB_code.sh
```diff
@@ -72,7 +59,7 @@ compile_prepareICAs()
 
 	log_Msg "Compiling ${app_name} application"
 	${MATLAB_HOME}/bin/mcc -m -v ${app_name}.m \
-				  -I ${HCPPIPEDIR}/global/matlab/gifti-1.6 \
+				  -I ${HCPCIFTIRWDIR} \
 				  -I ${HCPPIPEDIR}/global/matlab \
```
mharms, line 63: Now that we've eliminated the `@gifti` symlink, we could make the path order for the compilation consistent with the order that we are using in setting the path (via `addpath`) for the interpreted matlab.  That order would be:
```
...
-I ${HCPPIPEDIR}/global/matlab \
-I ${HCPCIFTIRWDIR} \
...
```
i.e., give `${HCPPIPEDIR}/global/matlab` precedence (given that -I appends).

MSMAll/scripts/MSMAll.sh
```diff
@@ -722,7 +722,7 @@ main()
 						mPath="${HCPPIPEDIR}/MSMAll/scripts"
 						mGlobalPath="${HCPPIPEDIR}/global/matlab"
 
-						matlabCode="addpath '$mPath'; addpath '$mGlobalPath'; MSMregression('${inputspatialmaps}','${inputdtseries}','${inputweights}','${outputspatialmaps}','${outputweights}','${Caret7_Command}','${Method}','${Params}','${VN}',${nTPsForSpectra},'${BC}','${VolParams}');"
+						matlabCode="addpath '$mPath'; addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; MSMregression('${inputspatialmaps}','${inputdtseries}','${inputweights}','${outputspatialmaps}','${outputweights}','${Caret7_Command}','${Method}','${Params}','${VN}',${nTPsForSpectra},'${BC}','${VolParams}');"
```
mharms, line 725: If we want a consistent principle on path precedence, we would do the following instead:
```suggestion
						matlabCode="addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; addpath '$mPath';  MSMregression('${inputspatialmaps}','${inputdtseries}','${inputweights}','${outputspatialmaps}','${outputweights}','${Caret7_Command}','${Method}','${Params}','${VN}',${nTPsForSpectra},'${BC}','${VolParams}');"
```

MSMAll/scripts/MSMAll.sh
```diff
@@ -807,7 +807,7 @@ main()
 						mPath="${HCPPIPEDIR}/MSMAll/scripts"
 						mGlobalPath="${HCPPIPEDIR}/global/matlab"
 						
-						matlabCode="addpath '$mPath'; addpath '$mGlobalPath';
+						matlabCode="addpath '$mPath'; addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath';
```
mharms, line 810:
```suggestion
						matlabCode="addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; addpath '$mPath'; 
```

MSMAll/scripts/MSMAll.sh
```diff
@@ -1225,7 +1225,7 @@ main()
 				mPath="${HCPPIPEDIR}/MSMAll/scripts"
 				mGlobalPath="${HCPPIPEDIR}/global/matlab"
 				
-				matlabCode="addpath '$mPath'; addpath '$mGlobalPath';
+				matlabCode="addpath '$mPath'; addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath';
```
mharms, line 1228:
```suggestion
				matlabCode="addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; addpath '$mPath'; 
```

MSMAll/scripts/MSMAll.sh
```diff
@@ -1294,7 +1294,7 @@ main()
...
```
mharms, line 1297:
```suggestion
				matlabCode="addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; addpath '$mPath'; 
```

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -421,7 +421,7 @@ main()
 						mPath="${HCPPIPEDIR}/MSMAll/scripts"
 						mGlobalPath="${HCPPIPEDIR}/global/matlab"
 						
-						matlabCode="addpath '$mPath'; addpath '$mGlobalPath';
+						matlabCode="addpath '$mPath'; addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath';
```
mharms, line 424:
```suggestion
						matlabCode="addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; addpath '$mPath'; 
```

RestingStateStats/RestingStateStats.sh
```diff
@@ -968,7 +968,7 @@ main()
 			mGlobalPath="${HCPPIPEDIR}/global/matlab"
 			mFslPath="${FSLDIR}/etc/matlab"
 
-			matlabCode="addpath '$mPath'; addpath '$mGlobalPath'; addpath '$mFslPath'; RestingStateStats('${motionparameters}',${g_high_pass},${TR},'${ICAs}','${noise}','${CARET7DIR}/wb_command','${dtseries}','${bias}','${RssPrefix}','${g_dlabel_file}','${g_bc_mode}','${g_out_string}','${WM}','${CSF}');"
+			matlabCode="addpath '$mPath'; addpath '$HCPCIFTIRWDIR' addpath '$mGlobalPath'; addpath '$mFslPath'; RestingStateStats('${motionparameters}',${g_high_pass},${TR},'${ICAs}','${noise}','${CARET7DIR}/wb_command','${dtseries}','${bias}','${RssPrefix}','${g_dlabel_file}','${g_bc_mode}','${g_out_string}','${WM}','${CSF}');"
```
mharms, line 971: In addition to path order consistency, existing command missing a semicolon between two `addpath`
```suggestion
			matlabCode="addpath '$mFslPath'; addpath '$HCPCIFTIRWDIR'; addpath '$mGlobalPath'; addpath '$mPath'; RestingStateStats('${motionparameters}',${g_high_pass},${TR},'${ICAs}','${noise}','${CARET7DIR}/wb_command','${dtseries}','${bias}','${RssPrefix}','${g_dlabel_file}','${g_bc_mode}','${g_out_string}','${WM}','${CSF}');"
```

global/matlab/Compile_All_MATLAB_code.sh
```diff
@@ -113,50 +113,28 @@ get_options()
...
-#  Main processing of script.
-# ------------------------------------------------------------------------------
-main()
-{
-	get_options "$@"
-	
-	compile_ICAFIX_code
-	compile_MSMALL_code
-	compile_RestingStateStats_code
-}
```
mharms, line 158: I would prefer that we not make these changes to the "master" compile script. Having three separate `compile_*_code` functions within `main` makes it very convenient to comment out the sub-pipelines that don't need re-compiling (e.g., if we only changed ICAFIX code, and thus only need to recompile that).

global/scripts/RSNregression.sh
```diff
@@ -92,7 +92,7 @@ case "$MatlabMode" in
         log_Err_Abort "unrecognized matlab mode '$MatlabMode', use 0, 1, or 2"
         ;;
 esac
-matlab_paths="addpath('$FSLDIR/etc/matlab'); addpath('$HCPPIPEDIR/global/matlab'); addpath('$this_script_dir');
+matlab_paths="addpath('$FSLDIR/etc/matlab'); addpath('$HCPPIPEDIR/global/matlab'); addpath('$HCPCIFTIRWDIR'); addpath('$this_script_dir');
```
mharms, line 95:
```suggestion
matlab_paths="addpath('$FSLDIR/etc/matlab'); addpath('$HCPCIFTIRWDIR'); addpath('$HCPPIPEDIR/global/matlab'); addpath('$this_script_dir');
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
coalsont: Since FSL_FIX_CIFTIRW isn't mentioned in the file, it would probably be better to say something like "take precedence over the fix settings file".

mharms: That's fine with me for the line in question, although given the convoluted nature by which this all works together, I think it is still valuable to maintain the mentions of  `${FSL_FIX_CIFTIRW}` and `${FSL_FIX_WBC}` earlier in the comment.

global/matlab/Compile_All_MATLAB_code.sh
coalsont: 1) Scripts already exist to compile only a single pipeline's matlab code, and require no editing, and are co-located with that pipeline's matlab files, which should be more convenient than editing anything.
2) A function that calls only one line of processing code needlessly lengthens a script and makes it look more complicated.
3) You can still comment out only one line to prevent the actual compilation step, and I would be perfectly happy removing the `echo`s entirely, but I have never had to read the script output.
4) If the script should officially support not compiling everything, then that suggests it should have options for this, rather than requiring editing.

Given that the only option actually supported is the help info option, and the scripts it calls do their own checking of environment variables anyway, I would instead be tempted to replace the whole thing with a 5-line script:
```
#!/bin/bash
set -eux
"$HCPPIPEDIR"/ICAFIX/scripts/Compile_MATLAB_code.sh
"$HCPPIPEDIR"/MSMAll/scripts/Compile_MATLAB_code.sh
"$HCPPIPEDIR"/RestingStateStats/scripts/Compile_MATLAB_code.sh
```

mharms: I see your points, regarding the originally proposed changes.  I'm not a fan though of going with the entirely minimalist approach (just 3 lines), as I think the Usage statement adds valuable context and information.

ICAFIX/ReApplyFixPipeline.sh
```diff
@@ -395,11 +395,12 @@ main()
 	# are set in ${FSL_FIXDIR}/settings.sh, which is sourced below for interpreted modes.
 	# Note that the ciftiopen.m, ciftisave.m functions are *appended* to the path through the ${FSL_FIX_CIFTIRW} 
 	# environment variable within fix_3_clean.m itself.
-	# Note that we are NOT adding '${HCPPIPEDIR}/global/matlab' to the path (which also contains the
-	# CIFTI I/O functions as well), so the versions in ${FSL_FIX_CIFTIRW} will be the ones that get used
-	# (assuming that the user hasn't further customized their matlab path, e.g., via a startup.m addition).
+	# Note that previously we did not add '${HCPPIPEDIR}/global/matlab' to the path (which used to contain the
+	# CIFTI I/O functions as well), or HCPCIFTIRWDIR.
+	# Since addpath adds to the front of the matlab path, we are now overriding the fix settings for cifti IO,
```
mharms, line 400:
```suggestion
	# Since addpath adds to the front of the matlab path, we are now overriding the fix settings for cifti I/O,
```

ICAFIX/scripts/Compile_MATLAB_code.sh
```diff
@@ -33,26 +33,13 @@
...
-# because the pipeline scripts have come to rely on that convenience for interpreted matlab mode.
-# We COULD delete the @gifti symlink and simultaneously move the actual '@gifti' folder
-# into '${HCPPIPEDIR}/global/matlab', in which case interpreted matlab mode would continue to work,
-# and we could then consolidate the two different "-I" options into one.
-# But, sticking with the symlink for now, since it was already in place.
+# NOTE: while the matlab interpreter is happy to follow symlinks found on the matlab path, the matlab compiler
+# not only does not follow them, it also remembers that they exist and refuses to allow anything with the same
+# name found in a later -I option to be used.  @gifti used to be a symlink, requiring the compilation to
+# include the real location of the @gifti folder, before a folder that contains an @gifti symlink.
+
```
mharms, line 40: Restoring this tidbit, so that the behavior of -I is explicit, and I don't need to translate the implied ordering when I return to this script in the future
```suggestion
# (since the -I option *appends* folders to the search path, in the order listed)

```

coalsont:
I haven't worked out a plan for testing yet.  Maybe we can wait until we are ready to make cifti-matlab the default, and test it then?

glasserm:
I thought the point of this was to enable testing of CIFTI matlab.  RSNRegression works.

coalsont:
It also moves some things around to more permanently resolve some hazards in the prior matlab compiler setup, and edits several additional scripts that use cifti IO.  I guess if we feel the edits are sufficiently similar, we don't have to test all of them.

We could extend this PR to also be where we add a copy of cifti-matlab (once it has stabilized somewhat), since that is what it is preparing for anyway.  Of course, if you want to be testing cifti-matlab against master for a while before we do that, that means we should prioritize merging this and not extend it.

mharms:
My main concern about timing is the potential for merge conflicts to develop, although I don't foresee that being a big problem...

coalsont:
It may be time to do a non-beta release of cifti-matlab and finish this PR.

glasserm:
I would be fine with a CIFTI matlab release.

coalsont:
I have added cifti-matlab and set it as the default in `Examples/Scripts/SetUp...` and modified the LICENSE.md to note the GPL from some of the code (including gifti-1.6 we were already using).  I am talking to the relevant authors about dual-licensing, and will update the licenses depending on the outcome.

mharms:
What are your thoughts about the compiled matlab?  Do we recompile so that the compiled is based on the same `cifti-matlab` library?

glasserm:
Yes we should do that before the next release.

coalsont:
Because the gifti toolbox is staying GPL (derived from SPM), and no released scripts use it (it is just a dependency of `cifti-legacy`), I have moved it inside of cifti-legacy, so that inclusion of `global/matlab` doesn't automatically drag it in (and using `cifti-legacy` doesn't require including `global/matlab` anymore).  If in-progress scripts use `@gifti` directly, perhaps the thing to do is make a second copy in the original `gifti-1.6` folder (and not make a symlink, so the matlab compiler doesn't get confused).

`@xmltree` is now LGPL v3+ and `cifti-matlab` is BSD, so `cifti-matlab` can be compiled together with whatever other code under whatever licenses.  Additionally, `fsl/etc/matlab/read_avw.m`, etc have been released under BSD terms per Steve Smith (via email), so it won't be a problem to compile with GPL code anyway.

glasserm:
I don't think we load GIFTI files in matlab often and with the new toolbox I expect that to be even less.

coalsont:
We should probably do a quick test on the current branch of something that uses cifti, then I think this is finally ready for merge.

glasserm:
I think we should finish this up.  I have been on this branch in my version of the repo for quite some time and would like to switch back to master.

PR #182, opened 2020-07-07
Extracting specific runs from MRFIX during DeDrift...

coalsont:
This still needs to be tested.

### DIFF
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -1000,11 +1000,12 @@ main()
                 mkdir -p "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}"
                 
                 # Using clean_vn.dscalar.nii estimated from the full concat group for the extracted concat group as well.
-                # (i.e., estimate for the variance normalization map is based on the full concat group, not 
+                # (i.e., estimate for the variance normalization map is based on the full concat group, not
                 #  the subset of extracted scans)
                 # The per-run differences in (unstructured) noise variance were removed before concatenation, and when
-                #  going back to "bias corrected" values, it just used this _clean_vn map in reverse (which is close to the
-                #  average of the original _vn maps), so it should be equally applicable to any subset of runs
+                #  going back to "bias corrected" values, it used something conceptually equivalent to this _clean_vn map
+                #  in reverse (which is close to the average of the original _vn maps), so it is equally applicable to
+                #  any subset of runs
                 
                 cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas${regstring}_hp${HighPass}_clean_vn.dscalar.nii" \
                     "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}/${mrFIXExtractConcatNamesArr[$i]}_Atlas${regstring}_hp${HighPass}_clean_vn.dscalar.nii"
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -987,6 +987,10 @@ main()
             then
                 mkdir -p "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}"
                 
+                # Using clean_vn.dscalar.nii estimated from the full concat group for the extracted concat group as well.
+                # (i.e., estimate for the variance normalization map is based on the full concat group, not 
+                # the subset of extracted scans)
+                
                 cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas${regstring}_hp${HighPass}_clean_vn.dscalar.nii" \
                     "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}/${mrFIXExtractConcatNamesArr[$i]}_Atlas${regstring}_hp${HighPass}_clean_vn.dscalar.nii"
                 
@@ -999,6 +1003,7 @@ main()
         
         if ((mrFIXExtractDoVol))
         then
+            # Using clean_vn.nii.gz estimated from the full concat group for the extracted concat group as well.
             cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_hp${HighPass}_clean_vn.nii.gz" \
                 "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}/${mrFIXExtractConcatNamesArr[$i]}_hp${HighPass}_clean_vn.nii.gz"
             
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -74,10 +74,11 @@ PARAMETERs are [ ] = optional; < > = user supplied value
    [--multirun-fix-concat-names=<day1_concat@day2_concat>] @-delimited list of names of the concatenated
        timeseries, only required when using --multirun-fix-names.
    [--multirun-fix-extract-names=<day1run1@day1run2%day2run1@day2run2>] @ and % delimited list of lists of
-       fMRIName strings to extract from each multi-run ICA+FIX group, only required when using
-       --multirun-fix-extract-concat-names.
+       fMRIName strings to extract, one list for each multi-run ICA+FIX group in --multirun-fix-names (use
+       NONE to skip a group), only required when using --multirun-fix-extract-concat-names.
    [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
-       concatenated extracted timeseries.
+       concatenated extracted timeseries, one for each multi-run ICA+FIX group (i.e. name in
+       --multirun-fix-concat-names; use NONE to skip a group).
    [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will 
        have ICA+FIX reapplied to them (could be either rfMRI or tfMRI). Previously known as --rfmri-names.
    [--dont-fix-names=<not@ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that
@@ -448,6 +449,8 @@ main()
 	log_Msg "MyelinMaps: ${MyelinMaps}"
 	log_Msg "mrFIXNames: ${mrFIXNames}"
 	log_Msg "mrFIXConcatNames: ${mrFIXConcatNames}"
+	log_Msg "mrFIXExtractNames: ${mrFIXExtractNames}"
+	log_Msg "mrFIXExtractConcatNames: ${mrFIXExtractConcatNames}"
 	log_Msg "fixNames: ${fixNames}"
 	log_Msg "dontFixNames: ${dontFixNames}"
 	log_Msg "SmoothingFWHM: ${SmoothingFWHM}"
@@ -532,14 +535,14 @@ main()
 	fi
 	log_Msg "After delimiter substitution, mrFIXExtractConcatNamesArr: ${mrFIXExtractConcatNamesArr[@]}"
 	
-	if (( ${#mrFIXExtractNamesArr[@]} != ${#mrFIXConcatExtractNamesArr[@]} ))
+	if (( ${#mrFIXExtractNamesArr[@]} != ${#mrFIXExtractConcatNamesArr[@]} ))
 	then
-		log_Err_Abort "number of MR FIX extract concat names and run groups are different (use NONE for an empty group)"
+		log_Err_Abort "number of MR FIX extract concat names and run groups are different (use NONE to skip a group)"
 	fi
 	
-	if (( ${#mrFIXConcatExtractNamesArr[@]} > 0 && ${#mrFIXConcatNames[@]} != ${#mrFIXConcatExtractNamesArr[@]} ))
+	if (( ${#mrFIXExtractConcatNamesArr[@]} > 0 && ${#mrFIXConcatNames[@]} != ${#mrFIXExtractConcatNamesArr[@]} ))
 	then
-	    log_Err_Abort "number of MR FIX extract groups doesn't match number of MR FIX groups (use NONE for an empty group)"
+	    log_Err_Abort "number of MR FIX extract groups doesn't match number of MR FIX groups (use NONE to skip a group)"
 	fi
 	
 	CorrectionSigma=$(echo "sqrt ( 200 )" | bc -l)
@@ -916,7 +919,7 @@ main()
 	                    --study-folder="$StudyFolder"
 	                    --subject="$Subject"
 	                    --multirun-fix-names="${mrFIXNames[$i]}"
-	                    --csv-out="$StudyFolder/$Subject/MNINonLinear/Results/$ConcatName/${ConcatName}_Runs.csv"
+	                    --csv-out="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Runs.csv"
 	                    --concat-cifti-input="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas_${ConcatRegName}_hp${HighPass}_clean.dtseries.nii")
         
 	    if (( ${#mrFIXExtractConcatNamesArr[@]} > 0 )) && [[ "${mrFIXExtractConcatNamesArr[$i]}" != NONE ]]
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -1002,10 +1002,20 @@ main()
                 # Using clean_vn.dscalar.nii estimated from the full concat group for the extracted concat group as well.
                 # (i.e., estimate for the variance normalization map is based on the full concat group, not
                 #  the subset of extracted scans)
-                # The per-run differences in (unstructured) noise variance were removed before concatenation, and when
-                #  going back to "bias corrected" values, it used something conceptually equivalent to this _clean_vn map
-                #  in reverse (which is close to the average of the original _vn maps), so it is equally applicable to
-                #  any subset of runs
+                
+                # The per-run differences in (unstructured) noise variance were removed before concatenation.
+                # The average of those maps (across runs) was multiplied back into the concatenated time series
+                #  (in 'hcp_fix_multi_run'), so that the entire concatenated time series has a spatial pattern of
+                #  unstructured noise consistent with the average across runs.
+                # Given this manner of constructing the concatenated time series, any subset of runs extracted from
+                #  the full concatenated set should use this same average map for later variance normalization.
+                # We use the "clean_vn" map for this purpose and thus copy it from the full concatenated set to the
+                #  extracted set of runs.
+                # As a final subtlety, note that the "clean_vn" map itself is not identical to the aforementioned average of
+                #  the individual run vn maps, but it is conceptually very similar. In particular, clean_vn is derived within
+                #  FIX itself from the concatenated time series by regressing out all structured signals and using the
+                #  residual to estimate the unstructured noise, whereas the individual run vn maps were computed using
+                #  PCA-based reconstruction of the unstructured noise in 'icaDim.m'.
                 
                 cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas${regstring}_hp${HighPass}_clean_vn.dscalar.nii" \
                     "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}/${mrFIXExtractConcatNamesArr[$i]}_Atlas${regstring}_hp${HighPass}_clean_vn.dscalar.nii"
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -75,7 +75,9 @@ PARAMETERs are [ ] = optional; < > = user supplied value
        timeseries, only required when using --multirun-fix-names.
    [--multirun-fix-extract-names=<day1run1@day1run2%day2run1@day2run2>] @ and % delimited list of lists of
        fMRIName strings to extract, one list for each multi-run ICA+FIX group in --multirun-fix-names (use
-       NONE to skip a group), only required when using --multirun-fix-extract-concat-names.
+       NONE to skip a group), only required when using --multirun-fix-extract-concat-names.  Exists to
+       enable extraction of a subset of the runs in a multi-run ICA+FIX group into a new concatenated
+       series (which is named using --multirun-fix-extract-concat-names).
    [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
        concatenated extracted timeseries, one for each multi-run ICA+FIX group (i.e. name in
        --multirun-fix-concat-names; use NONE to skip a group).
@@ -915,7 +917,7 @@ main()
 	    log_Msg "reapply_mr_fix_cmd: ${reapply_mr_fix_cmd[*]}"
 	    "${reapply_mr_fix_cmd[@]}"
 	    
-	    extract_cmd=("${HCPPIPEDIR}/DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh"
+	    extract_cmd=("${HCPPIPEDIR}/global/scripts/ExtractFromMRFIXConcat.sh"
 	                    --study-folder="$StudyFolder"
 	                    --subject="$Subject"
 	                    --multirun-fix-names="${mrFIXNames[$i]}"
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -73,6 +73,11 @@ PARAMETERs are [ ] = optional; < > = user supplied value
        lists of runs in this option.
    [--multirun-fix-concat-names=<day1_concat@day2_concat>] @-delimited list of names of the concatenated
        timeseries, only required when using --multirun-fix-names.
+   [--multirun-fix-extract-names=<day1run1@day1run2%day2run1@day2run2>] @ and % delimited list of lists of
+       fMRIName strings to extract from each multi-run ICA+FIX group, only required when using
+       --multirun-fix-extract-concat-names.
+   [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
+       concatenated extracted timeseries.
    [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will 
        have ICA+FIX reapplied to them (could be either rfMRI or tfMRI). Previously known as --rfmri-names.
    [--dont-fix-names=<not@ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that
@@ -112,8 +117,10 @@ get_options()
 	p_MyelinMaps=NONE				# @ delimited
 	p_mrFIXConcatNames=NONE			# @ delimited
 	p_mrFIXNames=NONE				# @ and % delimited, % is top-level and must match number of @s in ConcatNames
-	p_rfMRINames=NONE				# @ delimited
-	p_tfMRINames=NONE				# @ delimited
+	p_mrFIXExtractConcatNames=NONE  # @ delimited
+	p_mrFIXExtractNames=NONE        # @ and % delimited, % is top-level and must match number of @s in ExtractConcatNames
+	p_fixNames=NONE                 # @ delimited
+	p_dontFixNames=NONE             # @ delimited
 	unset p_SmoothingFWHM
 	unset p_HighPass
 	unset p_MotionRegression
@@ -184,12 +191,20 @@ get_options()
 				p_mrFIXNames=${argument#*=}
 				index=$(( index + 1 ))
 				;;
+			--multirun-fix-extract-concat-names=*)
+				p_mrFIXExtractConcatNames=${argument#*=}
+				index=$(( index + 1 ))
+				;;
+			--multirun-fix-extract-names=*)
+				p_mrFIXExtractNames=${argument#*=}
+				index=$(( index + 1 ))
+				;;
 			--fix-names=* | --rfmri-names=*)
-				p_rfMRINames=${argument#*=}
+				p_fixNames=${argument#*=}
 				index=$(( index + 1 ))
 				;;
 			--dont-fix-names=* | --tfmri-names=*)
-				p_tfMRINames=${argument#*=}
+				p_dontFixNames=${argument#*=}
 				index=$(( index + 1 ))
 				;;
 			--smoothing-fwhm=*)
@@ -283,8 +298,8 @@ get_options()
 	fi
 
 	log_Msg "list of MRfix scans: ${p_mrFIXNames}"
-	log_Msg "list of fix scans: ${p_rfMRINames}"
-	log_Msg "list of non-fix scans: ${p_tfMRINames}"
+	log_Msg "list of fix scans: ${p_fixNames}"
+	log_Msg "list of non-fix scans: ${p_dontFixNames}"
 
 	if [ -z "${p_SmoothingFWHM}" ]; then
 		log_Err "smoothing value (--smoothing-fwhm=) required"
@@ -396,29 +411,29 @@ main()
 	local ConcatRegName="${7}"
 	local Maps="${8}"
 	local MyelinMaps="${9}"
-	local rfMRINames="${10}"
-	local tfMRINames="${11}"
+	local fixNames="${10}"
+	local dontFixNames="${11}"
 	local mrFIXNames="${12}"
 	local mrFIXConcatNames="${13}"
 	local SmoothingFWHM="${14}"
 	local HighPass="${15}"
 	local MotionRegression="${16}"
-	
 	local MyelinTargetFile="${17}"
+	local InRegName="${18}"
+	local MatlabRunMode="${19}"
+	local mrFIXExtractNames="${20}"
+	local mrFIXExtractConcatNames="${21}"
+
 	if [ "${MyelinTargetFile}" = "NONE" ]; then
 		MyelinTargetFile=""
 	fi
 
-	local InRegName="${18}"
 	if [ "${InRegName}" = "NONE" ]; then
 		InRegName=""
 	fi
 	
-	local MatlabRunMode
-	if [ -z "${19}" ]; then
-		MatlabRunMode=${G_DEFAULT_MATLAB_RUN_MODE}
-	else
-		MatlabRunMode="${19}"
+	if [ -z "$MatlabRunMode" ]; then
+		MatlabRunMode="$G_DEFAULT_MATLAB_RUN_MODE"
 	fi
 
 	# Log values retrieved from positional parameters
@@ -433,8 +448,8 @@ main()
 	log_Msg "MyelinMaps: ${MyelinMaps}"
 	log_Msg "mrFIXNames: ${mrFIXNames}"
 	log_Msg "mrFIXConcatNames: ${mrFIXConcatNames}"
-	log_Msg "rfMRINames: ${rfMRINames}"
-	log_Msg "tfMRINames: ${tfMRINames}"
+	log_Msg "fixNames: ${fixNames}"
+	log_Msg "dontFixNames: ${dontFixNames}"
 	log_Msg "SmoothingFWHM: ${SmoothingFWHM}"
 	log_Msg "HighPass: ${HighPass}"
 	log_Msg "MyelinTargetFile: ${MyelinTargetFile}"
@@ -462,21 +477,21 @@ main()
 	fi
 	log_Msg "After delimiter substitution, MyelinMaps: ${MyelinMaps}"
 
-	if [ "${rfMRINames}" = "NONE" ] ; then
-		rfMRINames=()
+	if [ "${fixNames}" = "NONE" ] ; then
+		fixNames=()
 	else
-		#rfMRINames=`echo "$rfMRINames" | sed s/"@"/" "/g`
-		IFS=@ read -a rfMRINames <<< "${rfMRINames}"
+		#fixNames=`echo "$fixNames" | sed s/"@"/" "/g`
+		IFS=@ read -a fixNames <<< "${fixNames}"
 	fi
-	log_Msg "After delimiter substitution, rfMRINames: ${rfMRINames[@]}"
+	log_Msg "After delimiter substitution, fixNames: ${fixNames[@]}"
 
-	if [ "${tfMRINames}" = "NONE" ] ; then
-		tfMRINames=()
+	if [ "${dontFixNames}" = "NONE" ] ; then
+		dontFixNames=()
 	else
-		#tfMRINames=`echo "$tfMRINames" | sed s/"@"/" "/g`
-		IFS=@ read -a tfMRINames <<< "${tfMRINames}"
+		#dontFixNames=`echo "$dontFixNames" | sed s/"@"/" "/g`
+		IFS=@ read -a dontFixNames <<< "${dontFixNames}"
 	fi
-	log_Msg "After delimiter substitution, tfMRINames: ${tfMRINames[@]}"
+	log_Msg "After delimiter substitution, dontFixNames: ${dontFixNames[@]}"
 
 	if [[ "${mrFIXNames}" == "NONE" ]] ; then
 		mrFIXNames=()
@@ -499,6 +514,32 @@ main()
 	if (( ${#mrFIXNames[@]} != ${#mrFIXConcatNames[@]} ))
 	then
 		log_Err_Abort "number of MR FIX concat names and run groups are different"
+	fi
+	
+	if [[ "${mrFIXExtractNames}" == "NONE" ]] ; then
+		mrFIXExtractNamesArr=()
+	else
+		#two-level list, % and @, parse only one stage here
+		IFS=% read -a mrFIXExtractNamesArr <<< "${mrFIXExtractNames}"
+	fi
+	log_Msg "After delimiter substitution, mrFIXExtractNamesArr: ${mrFIXExtractNamesArr[@]}"
+	
+	if [[ "$mrFIXExtractConcatNames" == "NONE" ]]
+	then
+		mrFIXExtractConcatNamesArr=()
+	else
+		IFS=@ read -a mrFIXExtractConcatNamesArr <<< "$mrFIXExtractConcatNames"
+	fi
+	log_Msg "After delimiter substitution, mrFIXExtractConcatNamesArr: ${mrFIXExtractConcatNamesArr[@]}"
+	
+	if (( ${#mrFIXExtractNamesArr[@]} != ${#mrFIXConcatExtractNamesArr[@]} ))
+	then
+		log_Err_Abort "number of MR FIX extract concat names and run groups are different (use NONE for an empty group)"
+	fi
+	
+	if (( ${#mrFIXConcatExtractNamesArr[@]} > 0 && ${#mrFIXConcatNames[@]} != ${#mrFIXConcatExtractNamesArr[@]} ))
+	then
+	    log_Err_Abort "number of MR FIX extract groups doesn't match number of MR FIX groups (use NONE for an empty group)"
 	fi
 	
 	CorrectionSigma=$(echo "sqrt ( 200 )" | bc -l)
@@ -827,7 +868,7 @@ main()
 
 	# Resample (and resmooth) TS from Native 
 	log_Msg "Resample (and resmooth) TS from Native"
-	for fMRIName in "${rfMRINames[@]}" "${tfMRINames[@]}" "${mrFIXNamesAll[@]}" ; do
+	for fMRIName in "${fixNames[@]}" "${dontFixNames[@]}" "${mrFIXNamesAll[@]}" ; do
 		log_Msg "fMRIName: ${fMRIName}"
 		cp ${ResultsFolder}/${fMRIName}/${fMRIName}_Atlas${InRegName}.dtseries.nii ${ResultsFolder}/${fMRIName}/${fMRIName}_Atlas_${ConcatRegName}.dtseries.nii
 		for Hemisphere in L R ; do
@@ -851,8 +892,8 @@ main()
 
 	# ReApply FIX Cleanup
 	log_Msg "ReApply FIX Cleanup"
-	log_Msg "rfMRINames: ${rfMRINames[@]}"
-	for fMRIName in "${rfMRINames[@]}" ; do
+	log_Msg "fixNames: ${fixNames[@]}"
+	for fMRIName in "${fixNames[@]}" ; do
 		log_Msg "fMRIName: ${fMRIName}"
 		reapply_fix_cmd=("${HCPPIPEDIR}/ICAFIX/ReApplyFixPipeline.sh" --path="${StudyFolder}" --subject="${Subject}" --fmri-name="${fMRIName}" --high-pass="${HighPass}" --reg-name="${ConcatRegName}" --matlab-run-mode="${MatlabRunMode}" --motion-regression="${MotionRegression}")
 		log_Msg "reapply_fix_cmd: ${reapply_fix_cmd[*]}"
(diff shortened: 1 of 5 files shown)


coalsont:
To be clear, the two-level extract list in the DeDrift script is ONLY because the script takes in multiple groups of MR FIX runs - it requires exactly one set of extractions per MR FIX run, if there are any extractions.  Neither DeDrift nor Extract supports doing two or more extractions from one MR FIX run in a single call.

glasserm:
I see. That probably should be made clear in the usage then.

coalsont:
Would this be better:

```
   [--multirun-fix-extract-names=<day1run1@day1run2%day2run1@day2run2>] @ and % delimited list of lists of
       fMRIName strings to extract, one list for each multi-run ICA+FIX group in --multirun-fix-names (use
       NONE to skip a group), only required when using --multirun-fix-extract-concat-names.
   [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
       concatenated extracted timeseries, one for each name in --multirun-fix-concat-names (use NONE to
       skip a group).
```

glasserm:
I edited it slightly.

coalsont:
According to my testing, it appears to work now.  @mharms, do you want to look it over?

DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh
```diff
@@ -0,0 +1,148 @@
...
+runSplits=()
+runIndices=()
+curTimepoints=0
+#convention: one before the first (1-based) index of the run
+runSplits[0]="$curTimepoints"
+#calculate the timepoints where the concatenated switches runs, find which runs are used
+for ((index = 0; index < ${#mrNamesArray[@]}; ++index))
+do
+    fmriName="${mrNamesArray[$index]}"
+    NumTPS=$(wb_command -file-information "$StudyFolder/$Subject/MNINonLinear/Results/$fmriName/${fmriName}_Atlas.dtseries.nii" -only-number-of-maps)
```
coalsont, line 102: Note that this currently hardcodes to look at the MSMSulc uncleaned per-run cifti files to get the length of each run.

coalsont: If this should be improved, I'll need to know which files are likely to be available in the scenarios we want to support.

glasserm: Volume:
$StudyFolder/$Subject/MNINonLinear/Results/$fmriName/${fmriName}.nii.gz

RegName="NONE":
$StudyFolder/$Subject/MNINonLinear/Results/$fmriName/${fmriName}_Atlas.dtseries.nii

RegName="MSMAll" or RegName="${RegName}"
$StudyFolder/$Subject/MNINonLinear/Results/$fmriName/${fmriName}_Atlas_MSMAll.dtseries.nii
$StudyFolder/$Subject/MNINonLinear/Results/$fmriName/${fmriName}_Atlas_${RegName}.dtseries.nii

coalsont: Is it possible someone will have a concatenated volume file but not those per-run volume files?

coalsont:
The extract script also currently expects every fMRI name to have a corresponding file, so the launch scripts would need to check subjects for missing runs.  This doesn't appear to be the case in the example batch script for DeDrift.

DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh
glasserm: No.

glasserm:
Yes.  In the database that is done, and if we were running this outside the database we would do it in the launcher script.

coalsont:
Should there be a note in the DeDrift batch file that it won't work as-is for subjects with incomplete data?

glasserm:
Well it would work fine so long as the specification in MR+FIX and DeDriftAndResample is the same runs.  A given launcher will either need to tune the arguments per subject to adjust for missing runs or only work on subjects with all runs.

coalsont:
My point is that we have a launcher in Examples/Scripts for DeDrift, which does NOT do that, or mention the issue.

glasserm:
Sure that is fine to edit.

coalsont:
The new extract script could check for missing runs and automatically exclude them, which could be more convenient for users running it on its own.

glasserm:
Sure.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -73,6 +73,12 @@ PARAMETERs are [ ] = optional; < > = user supplied value
        lists of runs in this option.
    [--multirun-fix-concat-names=<day1_concat@day2_concat>] @-delimited list of names of the concatenated
        timeseries, only required when using --multirun-fix-names.
+   [--multirun-fix-extract-names=<day1run1@day1run2%day2run1@day2run2>] @ and % delimited list of lists of
+       fMRIName strings to extract, one list for each multi-run ICA+FIX group in --multirun-fix-names (use
+       NONE to skip a group), only required when using --multirun-fix-extract-concat-names.
```
mharms, line 78:
```suggestion
       NONE to skip a group), only required when using --multirun-fix-extract-concat-names.
       Exists to enable extraction of a subset of the runs in a multi-run ICA+FIX group into a new concatenated series.
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -184,12 +192,20 @@ get_options()
...
+				p_mrFIXExtractConcatNames=${argument#*=}
+				index=$(( index + 1 ))
+				;;
+			--multirun-fix-extract-names=*)
+				p_mrFIXExtractNames=${argument#*=}
+				index=$(( index + 1 ))
+				;;
 			--fix-names=* | --rfmri-names=*)
-				p_rfMRINames=${argument#*=}
+				p_fixNames=${argument#*=}
```
mharms, line 204: Nice.  Thanks for updating the outdated internal variable names.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -870,6 +914,26 @@ main()
 	    reapply_mr_fix_cmd=("${HCPPIPEDIR}/ICAFIX/ReApplyFixMultiRunPipeline.sh" --path="${StudyFolder}" --subject="${Subject}" --fmri-names="${mrFIXNames[$i]}" --concat-fmri-name="${mrFIXConcatNames[$i]}" --high-pass="${HighPass}" --reg-name="${ConcatRegName}" --matlab-run-mode="${MatlabRunMode}" --motion-regression="${MotionRegression}")
 	    log_Msg "reapply_mr_fix_cmd: ${reapply_mr_fix_cmd[*]}"
 	    "${reapply_mr_fix_cmd[@]}"
+	    
+	    extract_cmd=("${HCPPIPEDIR}/DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh"
```
mharms, line 918: So, you decided to situate this Extract script in `DeDriftAndResample`, rather than `global`?

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -870,6 +914,26 @@ main()
...
+	                    --multirun-fix-names="${mrFIXNames[$i]}"
+	                    --csv-out="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Runs.csv"
+	                    --concat-cifti-input="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas_${ConcatRegName}_hp${HighPass}_clean.dtseries.nii")
+        
+	    if (( ${#mrFIXExtractConcatNamesArr[@]} > 0 )) && [[ "${mrFIXExtractConcatNamesArr[$i]}" != NONE ]]
+	    then
+	        mkdir -p "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}"
+	        
+	        cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas_${ConcatRegName}_hp${HighPass}_clean_vn.dscalar.nii" \
+	            "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}/${mrFIXExtractConcatNamesArr[$i]}_Atlas_${ConcatRegName}_hp${HighPass}_clean_vn.dscalar.nii"
```
mharms, line 930: If I'm reading this correctly, we are appropriating the `clean_vn.dscalar` from the full concat group to the extracted concat group?  Is that because anything else would be terribly complex?  A comment here about how we are treating/handling the `clean_vn.dscalar` would be helpful.

glasserm: It is fine to use the concat cleaned VN as the differences between the runs were already removed.

DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh
mharms: So, what's the plan here?  Just always get the length of each run from the MSMSulc uncleaned per-run cifti files (as currently implemented)?  I don't have a strong opinion either way.  While it seems like it might be relatively easy to add a `RegName` argument to select between MSMSulc and MSMAll, if we go the route of tailoring to `RegName`, then we should probably tailor to the volume files as well, which would require adding a volume-specific section to retrieving `NumTPSVol`.

glasserm: Ideally we would pull the info from the type of data we are splitting.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: But to be clear, the resulting `clean_vn.dscalar` is not strictly from the subset.  I just think we need a comment here explaining what we are doing, and the rationale for doing it this way.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: It will be relocated to global/scripts.

coalsont:
DeDrift can now generate the extracted volume output, and cifti for a different registration - this seemed like less work than editing hcp_fix_multi_run and ReApply...

glasserm:
Please let me know if there is some test data to look at.

glasserm:
These results look good.  I think this should be merged.

global/scripts/ExtractFromMRFIXConcat.sh
```diff
@@ -72,50 +73,89 @@ then
...
     do
-        if ((index != index2)) && [[ "${mrNamesArray[$index]}" == "${mrNamesArray[$index2]}" ]]
+        if [[ "${mrNamesArray[$index]}" == "${mrNamesArray[$index2]}" ]]
         then
             log_Err_Abort "MR fix names list contains '${mrNamesArray[$index]}' more than once"
         fi
     done
 done
 
+#sanity check for bad names in to use
```
mharms, line 95:
```suggestion
#sanity check for bad names in --multirun-fix-names-to-use
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -60,30 +60,34 @@ PARAMETERs are [ ] = optional; < > = user supplied value
...
+  [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
        concatenated extracted timeseries, one for each multi-run ICA+FIX group (i.e. name in
        --multirun-fix-concat-names; use NONE to skip a group).
-   [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will 
+  [--multirun-fix-extract-extra-regnames=<regname@regname>] extract MR FIX runs for additional surface
+       registrations, often MSMSulc.
+  [--multirun-fix-extract-volume={TRUE, FALSE}] whether to also extract the specified MR FIX runs from the
+       volume data, requires --multirun-fix-extract-concat-names to work, default FALSE.
+  [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will
        have ICA+FIX reapplied to them (could be either rfMRI or tfMRI). Previously known as --rfmri-names.
```
mharms, line 89:
```suggestion
  [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will
       have ICA+FIX reapplied to them (could be either rfMRI or tfMRI). Previously known as --rfmri-names.
       Not relevant if using multirun-fix.
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -60,30 +60,34 @@ PARAMETERs are [ ] = optional; < > = user supplied value
...
-   [--dont-fix-names=<not@ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that
+  [--dont-fix-names=<not@ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that
        will not have ICA+FIX reapplied to them (not recommended, ICA+FIX is recommended for all fMRI
        data). Previously known as --tfmri-names.
```
mharms, line 92:
```suggestion
  [--dont-fix-names=<not@ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that
       will not have ICA+FIX reapplied to them (not recommended, ICA+FIX is recommended for all fMRI
       data). Previously known as --tfmri-names. Not relevant if using multirun-fix.
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -60,30 +60,34 @@ PARAMETERs are [ ] = optional; < > = user supplied value
...
        NONE to skip a group), only required when using --multirun-fix-extract-concat-names.  Exists to
        enable extraction of a subset of the runs in a multi-run ICA+FIX group into a new concatenated
        series (which is named using --multirun-fix-extract-concat-names).
-   [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
+  [--multirun-fix-extract-concat-names=<day1_newconcat@day2_newconcat>] @-delimited list of names for the
        concatenated extracted timeseries, one for each multi-run ICA+FIX group (i.e. name in
        --multirun-fix-concat-names; use NONE to skip a group).
-   [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will 
+  [--multirun-fix-extract-extra-regnames=<regname@regname>] extract MR FIX runs for additional surface
+       registrations, often MSMSulc.
```
mharms, line 85: So, if wanting to apply to MSMSulc, do you use `NONE` as the argument?  `MSMSulc` as the argument? Either?

DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh
coalsont: This now takes an optional regname for the cifti to test for run length, and if the only data output requested is volume, takes the length from volume files.  At present, the regname should be "" for MSMSulc when calling Extract..., as it doesn't have the internal naming convention for that, DeDrift... handles that instead.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: I think it is technically possible to have some runs in a subject that are handled with MR FIX, and others that aren't - it isn't what we recommend, though.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: MSMSulc.  not specifying it, or using NONE, will cause it to not do any surface registration other than that in the --concat-reg-name argument.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -861,15 +954,64 @@ main()
...
+                            --subject="$Subject"
+                            --multirun-fix-names="${mrFIXNames[$i]}"
+                            --csv-out="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Runs.csv"
+                            --concat-cifti-input="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_Atlas${regstring}_hp${HighPass}_clean.dtseries.nii"
+                            --surf-reg-name="$regname")
+            
+            if (( ${#mrFIXExtractConcatNamesArr[@]} > 0 )) && [[ "${mrFIXExtractConcatNamesArr[$i]}" != NONE ]]
+            then
+                mkdir -p "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}"
+
```
mharms, line 989: Still think we should have a comment here:
```suggestion
                # Using clean_vn.dscalar from the full concat group for the extracted concat group as well.
                # (i.e., estimate for the variance normalization map is based on the full concat group, not 
                # the subset of extracted scans)
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -861,15 +954,64 @@ main()
...
+                extract_cmd+=(--multirun-fix-names-to-use="${mrFIXExtractNamesArr[$i]}"
+                              --cifti-out="$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXExtractConcatNamesArr[$i]}/${mrFIXExtractConcatNamesArr[$i]}_Atlas${regstring}_hp${HighPass}_clean.dtseries.nii")
+            fi
+            
+            "${extract_cmd[@]}"
+        done
+        
+        if ((mrFIXExtractDoVol))
+        then
+            cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_hp${HighPass}_clean_vn.nii.gz" \
```
mharms, line 1002:
```suggestion
# Using clean_vn.nii.gz estimated from the full concat group for the extracted concat group as well.
            cp "$StudyFolder/$Subject/MNINonLinear/Results/${mrFIXConcatNames[$i]}/${mrFIXConcatNames[$i]}_hp${HighPass}_clean_vn.nii.gz" \
```

global/scripts/ExtractFromMRFIXConcat.sh
```diff
@@ -0,0 +1,194 @@
...
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+#this function gets called by opts_ParseArguments when --help is specified
+function usage()
+{
+    #header text
+    echo "
+$log_ToolName: does stuff
```
mharms, line 19: Can we provide a basic description of why this tool/script exists?

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: Well then, instead of my initial suggestion, how about "Relevant for runs processed with single-run FIX" instead?

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: Can we make that explicit in the usage, since I believe in other scripts that `NONE` represents `MSMSulc`?

mharms: Now I'm confused.  The code at one point checks `if [[ "$mrFIXExtractExtraRegNames" == NONE ]]`.  I don't see anywhere that it is checking for `MSMSulc` as the string.

Relatedly, what if someone tried to run MSMAll as the `-extra-regnames` and MSMSulc (`NONE`) as the RegName?  Would that work?

coalsont: Other scripts might use MSMSulc by default (and "NONE" is the old convention for "not specified"), and specifically require one and only one surface registration.  I don't know of a script where specifying "NONE" means "MSMSulc", and specifying "MSMSulc" is wrong.  However, the new Extract... script currently breaks from that convention somewhat, in that currently, specifying "MSMSulc" to it actually is wrong, which perhaps should be changed.

This argument is an optional array argument, and defaults to being empty - the old "NONE" convention to act as though it wasn't specified was added only for consistency with the existing conventions of the script.

What about this small change:
```
  [--multirun-fix-extract-extra-regnames=<regname@regname>] extract MR FIX runs for additional surface
       registrations, e.g. --multirun-fix-extract-extra-regnames=MSMSulc
```

glasserm: Both MSMSulc and NONE should probably do the same thing.

mharms: Isn't it the other way around -- using `NONE` for RegName implies "MSMSulc"?  That's how the ReApplyFix*.sh scripts are set up...

coalsont: @mharms https://github.com/Washington-University/HCPpipelines/pull/182/files#diff-29e440b6c98c0b8335ae2638db8ece93R970 is the MSMSulc test.

So, it looks like ReApply... would error if you said `MSMSulc`.  That is unfortunate.  I see two possibilities - we can either treat the relevant option as referring to the concept of registration type (building in handling of the naming quirks), or we can use it to refer only to naming convention (specify empty string to use whatever doesn't name its registration type).  Referring only to the naming convention would pose some issues, as specifying an array containing only the empty string would get tricky to distinguish from "not specified".

Either way, `NONE` should really only be used to say "don't use this argument", and only in old scripts that already use that convention (because they didn't handle the empty string properly).  Thus, `NONE` should do something different than `MSMSulc` for this array argument in this script.

glasserm: ok

coalsont: If the concat reg name was something other than MSMAll, and MSMAll was in the `--multirun-fix-extract-extra-regnames` argument, yes, that would currently work (assuming the concat reg name was valid).

mharms: Can we maybe beef up the usage statement to make clearer the distinction between `--registration-name` and `--concat-reg-name`?

I think we should change the Extract script so that specifying `MSMSulc` is correct.

FWIW, it appears that the TaskAnalysis*.sh, MakeAverageDataset.sh, and RestingStateStats.sh scripts also allow for handling RegName=NONE as MSMSulc.

coalsont: `--registration-name` and `--concat-reg-name` are part of the prior functionality of the script.  I appear to have run my tests specifying `MSMAll` for both, which appears to be the way to disable the dedrifting step.  It looks like the only reason for having this as two arguments is dedrifting, which should **not** be done on subjects' data anyway.

For scripts that require the use of surface data, defaulting to the MSMSulc data when `NONE` specifies "don't use this argument" is tolerable.  That doesn't mean that `NONE` is being interpreted as explicitly specifying MSMSulc.

mharms: Can we just make that a bit more explicit in the usage notes?  Also, do we actually know if specifying `MSMSulc` for `--registration-name` and `--concat-reg-name` would work?  I suspect that might be a situation that the script was never designed to handle...

coalsont: That is just my guess at the reason for those arguments, and Matt has not always agreed with me about dedrifting.

Given the file naming quirks of MSMSulc, I wouldn't expect it to work as the concatregname/regname, I didn't write that part of the code.  I'm not clear on what point you are making, though.

mharms: I was just cycling back to my earlier comment:

> Relatedly, what if someone tried to run MSMAll as the -extra-regnames and MSMSulc (NONE) as the RegName? Would that work?

So while using `MSMAll` as the -extra-regnames argument might be fine, using `MSMSulc` for --registration-name and --concat-reg-name is not.

coalsont: Right, I expect MSMSulc to fail as `--concat-reg-name` regardless of any other arguments, but only because it has a different naming convention that we don't plan to repeat for any future registration.

coalsont: @glasserm, can you confirm my guess or explain `--registration-name` and `--concat-reg-name`?

DeDriftAndResample/scripts/ExtractFromMRFIXConcat.sh
coalsont: Extract... should now accept `MSMSulc`.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Comments proposed by @mharms have been added here, though they do not currently explain the rationale.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: I don't know that this adds much clarity to the option.  Both of these might benefit from rephrasing, like "rfMRI timeseries to resample that were not processed with ICA+FIX or MR FIX".

DeDriftAndResample/DeDriftAndResamplePipeline.sh
glasserm: --registration-name is the name generated by MSMAll
--concat-reg-name is the final name that includes the MSMAll and dedrifting registrations
MSMSulc would not be appropriate for either of these.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
glasserm: The rationale is that run to run diffferences in VN have already been removed, so a cross-run average VN is valid.

mharms: To me, the rationale is "convenience", with the one based on the full concat group being "close enough" -- the fact is that it doesn't reflect the VN that would be computed specifically for the runs being extracted.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: @mharms what do you want the description of `--registration-name` and `--concat-reg-name` to be?  Do you just want it to also convey "not MSMSulc"?

mharms: Making clear that they are "not MSMSulc" would be helpful, but having come this far, I think it would be helpful to make them generally clearer as well.  For example, `--concat-reg-name` is indicated as optional, but it isn't at all clear under what situations its inclusion is actually necessary (which IIUC is anytime you are doing a "resampling" and not the "dedrifting".  Is it the case that `--dedrift-reg-files` and `--concat-reg-name` are mutually exclusive, and one of the two is required?  If so, making that clear would be helpful.

glasserm: --registration-name and --concat-reg-name would always be used.  If they are the same, --dedrift-reg-files can be omitted.

coalsont: Actually, I think ConcatRegName should be mandatory, it is used with no conditional in the script.  The dedrift files are only used when RegName and ConcatRegName are not equal.

mharms: Making that all clear would I think be very helpful.

coalsont: Oh, I see, when ConcatRegName isn't specified, it defaults to RegName.

coalsont: Maybe we should make it an error to specify --concat-reg-name without --dedrift-reg-files, so we can just treat the option as an output naming convention.

mharms: Also, I think it would help if we made clear that it is called `--concat-reg-name` because it represents the concatenation of MSMAll and dedrifting operations...?  (Hence, if you don't need to dedrift, there is no inherent concatenation, and thus `--registration-name` and `--concat-reg-name` will be the same).  Esp. given that the script now contains other `-concat-` variables which are an entirely different sort of concatenation (across runs).

mharms: That would be a lot more intuitive to me.  Currently it is unclear which options are specifying inputs, and which are just controlling output names...

coalsont: Here is what I would edit it to:

```
  [--dedrift-reg-files=</Path/to/File/Left.sphere.surf.gii@/Path/to/File/Right.sphere.surf.gii>]
       Path to the spheres output by the MSMRemoveGroupDrift pipeline. This option should not be used in
       most circumstances.
  [--concat-reg-name=<regname>] Output name for the dedrifted registration (referred to as the
       concatenated registration), requires using --dedrift-reg-files. This option should not be used in
       most circumstances.
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: @glasserm, is this variance normalization sequence right:

1) each run is variance normalized before in-memory concatenation (using icaDim's (unstructured) noise stdev)
2) ICA, etc
3) the in-memory concatenated data is returned to bias-corrected values by undoing variance normalization (using (unstructured-)noise-only variance of the concatenated data) and then written out

coalsont: Wait, step 3 doesn't make sense, the concatenated data is already variance normalized - what does it do, average the VN maps of each run?

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: Ok, I'm back to somewhat confused.  Here is how we actually called DeDrift for an example HCA subject:
```
# Running: /opt/HCP/HCPpipelines/DeDriftAndResample/DeDriftAndResamplePipeline.sh             --path="/home/hodgem/XNAT_BUILD_DIR/hodgem/msmall_Q2_2020_runs/CCF_HCA_STG/MsmAllProcessing.HCA6347674_V1_MR.1589590114.XNAT_PROCESS_DATA/HCA6347674_V1_MR/subjects/HCA6347674_V1_MR/hcp"             --subject="HCA6347674_V1_MR"             --high-res-mesh="164"             --low-res-meshes="32"             --registration-name="MSMAll_InitialReg_2_d40_WRN"             --dedrift-reg-files="/opt/HCP/HCPpipelines/global/templates/MSMAll/DeDriftingGroup.L.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii@/opt/HCP/HCPpipelines/global/templates/MSMAll/DeDriftingGroup.R.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii"             --concat-reg-name="MSMAll"             --maps="sulc@curvature@corrThickness@thickness"             --myelin-maps="MyelinMap@SmoothedMyelinMap"             --multirun-fix-names="rfMRI_REST1_AP@rfMRI_REST1_PA@tfMRI_VISMOTOR_PA@tfMRI_CARIT_PA@tfMRI_FACENAME_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"             --multirun-fix-concat-names="fMRI_CONCAT_ALL"             --fix-names="NONE"             --dont-fix-names="NONE"             --smoothing-fwhm="2"             --high-pass="0"             --matlab-run-mode="0"             --motion-regression="FALSE"             --myelin-target-file="NONE"             --input-reg-name="NONE"
```
Note that we used `--dedrift-reg-files`, which seems inconsistent with the proposed statement that "This option should not be used in most circumstances".

DeDriftAndResample/DeDriftAndResamplePipeline.sh
glasserm: 1. Runwise variance normalization by icaDim
2. Average of runwise variance normalizations
3. Unnormalization by average
4. Group variance normalization (for melodic only)
5. Cleanup of unormalized data and computation of "clean_vn" variance normalization (that is also better behaved for certain analyses).  It is this "clean_vn" that we use going forward and propagate around.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: `--registration-name="MSMAll_InitialReg_2_d40_WRN" --concat-reg-name="MSMAll"` hints at the difference.

Other users will use MSMAll to register directly to the final (already dedrifted) template.  I'm not sure what we did in the group that we generated the template from, but the technically correct thing to do would have been to take the final (already dedrifted) MSMAll template and run another fresh registration to it, just like other users will on their data.  Hopefully this was just a shortcut to save processing time that doesn't result in an appreciable registration difference?

glasserm: This is how we have run things since HCP-YA.  I don't think there is much effect of this for this registration.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: So, the concatenated file is actually variance (stdev, really) normalized concatenated data with the inverse of the average stdev map applied to it, so every single run has been renormalized into an idealized average bias corrected space.  Here is my proposed comment:

```
# The per-run differences in (unstructured) noise variance were removed before concatenation, and when going back to bias corrected values, it just used this average _vn map in reverse, so it should be equally applicable to any subset of runs
```

glasserm: The "clean_vn" and average VN are not the same, but will be similar.  The "clean_vn" is better behaved in edge cases.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Oh wait, in an HCA subject?  That seems unfortunate, I don't think we should have done that...

coalsont: Well, if it uses the same dedrifting files as in any other group, and doesn't recompute drift, I guess that is okay, it mostly implies an extra shift that should have been in the template to begin with.  Using dedrifting options just to match an existing registration (rather than actually *doing* a dedrifting) makes these options much harder to explain.

glasserm:
Are we ready to merge and move on?

coalsont:
No, there are some edits not yet committed, even if we give up on unresolved things.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Assuming that we tell others to run things the way we do, here is a new stab at edited usage info:
```
   --registration-name=<regname> String corresponding to the input registration sphere name 
       (e.g. \${Subject}.\${Hemisphere}.sphere.\${RegName}.native.surf.gii), usually MSMAll_InitialReg...
  [--dedrift-reg-files=</Path/to/File/Left.sphere.surf.gii@/Path/to/File/Right.sphere.surf.gii>]
       Usually the spheres in global/templates/MSMAll/
  [--concat-reg-name=<regname>] String corresponding to the output name for the dedrifted registration
       (referred to as the concatenated registration), usually MSMAll. Requires --dedrift-reg-files
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: How does this look:
```
  [--fix-names=<ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that will
       have single-run ICA+FIX reapplied to them (could be either rfMRI or tfMRI). Do not specify runs
       processed with MR FIX here. Previously known as --rfmri-names.
  [--dont-fix-names=<not@ICA+FIXed@fMRI@Names>] @-delimited fMRIName strings corresponding to maps that
       will not have ICA+FIX reapplied to them (not recommended, MR FIX or at least ICA+FIX is recommended
       for all fMRI data). Previously known as --tfmri-names.
```

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: So, it is still unclear to me, is the reason we are doing it this way because the templates in `global/templates/MSMAll` were not themselves dedrifted?

What remains unclear in the proposed text above is under what situation `--dedrift-reg-files` and `--concat-reg-name` can actually be omitted...

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: So, in the parenthetical for `--dont-fix-names`, it seems that you are equating "ICA+FIX" with _single-run_ ICA+FIX?  I think there is the potential for some confusion yet in how `--dont-fix-names` relates to the single vs. multi-run FIX issue.

I don't suppose that we have any error checking built in that a run isn't listed in both `--multirun-fix-names` and `--dont-fix-names`?

coalsont: We don't explicitly check for a name in more than one list, no.  I can add "single-run" to the description there.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: I suspect that the template functional data was dedrifted, and that the "drift" files are from just a single iteration of re-register and dedrift.  I talked to Matt about how if other studies computed drifts from their registrations and dedrifted, that could make the alignment across studies worse, and how that means that you shouldn't try to dedrift the final iteration of registration out of subjects' registrations, but that may have been after it was already run on most of the YA cohort.

When/if we come up with a better surface registration, I think we should avoid this situation, such that we would then update these descriptions to say "when using MSMAll, do this, otherwise don't specify this option".  No sense in making it harder to understand when there is only one registration to use with this script.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: Everything I'm hearing indicates that the choice here is a matter of convenience.  So, I'd propose something like the following:

` # We are not re-running MR-FIX on the set of runs to be extracted, so a "clean_vn.dscalar" map specifically derived from the set of extracted runs is not available. Instead, as a matter of convenience, we copy (reuse) the clean_vn map derived from the full group of concatenated scans as the VN map to associate with the extracted set of runs. The expectation is that the difference is inconsequential.`

coalsont: Re-running MR FIX on only the extracted runs would have less power to clean up the data, which is important for shorter acquisitions than the HCP YA, so I don't see it as being purely a matter of convenience.  The effect of using the clean_vn files on the *concatenated* existing data should be, except for the better cleanup offered by MR FIX, conceptually the same normalization as using the per-run _vn on the original runs and concatenating (because that is what MR FIX did in the first place, and then this exact clean_vn file was used to generate the written-out concatenated file from that).

To put it pseudo-mathematically:
```
mrfix_concat = concat(clean_runs ./ run_vns) .* repmat(clean_avg_vn)
extracted_normed = subset(mrfix_concat) ./ repmat(clean_avg_vn)
extracted_normed = subset(concat(clean_runs ./ run_vns) .* repmat(clean_avg_vn)) ./ repmat(clean_avg_vn)
extracted_normed = subset(concat(clean_runs ./ run_vns))
```

So, using anything else for the _vn map would be incorrect for taking a subset of runs from the concatenated file.  Basically, every run in the concatenated file is in a made-up bias corrected normalization, and we need to use exactly the same assumptions to get back to VN normalization correctly.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Yes, it would have made things much simpler if the dedrifting was applied only to the template, and then the registrations were run one more time to get the final MSMAll alignment, without applying any extra shift afterwards.  That way, users could just run MSMAll, resample using just that sphere (no dedrift files, no concatenation), and they are done.  Really, in my opinion, this script should never have said anything about dedrifting, as it isn't something that should be done outside of template generation.

As for user choice, if the dedrift files are from only the last iteration of template refinement, then the shifts in them should be small, and choosing not to use them would be largely imperceptible.  Unfortunately, since that isn't what we did for HCP YA, we are stuck with telling users to specify these extra arguments (because even a small systematic bias is a problem).

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: I'm not saying that the clean_vn map is "incorrect" to use, just that it is estimated from the unstructured noise of the full concatenated data during cleanup, and thus by definition can't be identical to what would be obtained if you actually ran MR-FIX on just the set of runs specified for the extraction.  Some comment alerting developers/users to this is all that I'm looking to add, since previously, all "clean_vn" maps were based on the application of FIX to the actual data (either single or multi-run) in question.

In that regard, I don't think your pseudo-equations necessarily capture the situation as I'm thinking about it, which is more like:
```
mrfix_concat = FIX( concat(runs ./ run_vns) .* repmat(avg(run_vns)))
```
 from which `clean_vn` is derived during FIX.
The question is how much does `clean_vn` differ if `runs` is the full set, vs. if you used the extracted set to repeat the actual computation.

glasserm: I don't really think all this is super helpful.  We remove the differences in the VN pattern between runs, so it is unlikely that there would be meaningfully different VN maps from the full concatenation versus just the resting state data.  The VN stuff is complicated and what the user really needs to know is to use the "clean_vn" file and leave the others alone.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: I think a description of all this as part of a README would be very helpful for users attempting to understand dedrifting and how it relates to MSMAll.  I realize that said README might have to be a little bit messy, perhaps with both "optimal implementation" and "actual/current implementation" components, but that's the reality of the current situation.

glasserm: This has nothing to do with this pull request.  It would be fine to revisit the implementation when we overhaul MSMAll later this year.

glasserm:
We need to not keep delaying pull requests for unrelated stuff.  These kind of things might better be tracked as issues against the code.  A pull request's discussion should be focused on the contained functionality and documentation and not used as an excuse to critique everything that can be improved in a given pipeline.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: FWIW, I'm still not sure I fully understand the situation myself... IIUC, the issue is NOT that the MSMAll templates were not dedrifted, but that an extra (unnecessary) round of dedrifting was applied, and now we are "stuck" replicating the application of that extra dedrift round because otherwise a small bias might be present relative to what we've previously done?

mharms:
There is a balance between that and bringing something to a conclusion after you've invested time into it.  If Tim wants to open an Issue referencing this PR that we need better documentation of how dedrifting works and was implemented for HCP-YA and HCP-LS, that is fine with me.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
mharms: I disagree.  This is exactly the sort of somewhat complicated issue that at least deserves mention as a comment in the code to alert developers to the issue, so that it isn't lost entirely.  My proposed earlier comment captures this thought, and doesn't seem like it should be contentious.  I don't see what the objection is to simply using that and moving on.

glasserm: I just review the code when I want to remember how this works.

mharms: I'm not asking us to comment on exactly how the VN actually works (which is indeed complicated) but simply on the issue that the `clean_vn` file that we are associating with the extracted runs is a bit different from how it is "usually" calculated.  My proposed comment again for convenience:

`# We are not re-running MR-FIX on the set of runs to be extracted, so a "clean_vn.dscalar" map specifically derived from the set of extracted runs is not available. Instead, as a matter of convenience, we copy (reuse) the clean_vn map derived from the full group of concatenated scans as the VN map to associate with the extracted set of runs. The expectation is that the difference is inconsequential.`

coalsont: The cleanup regression is independent of the normalization, it is only the ICA that is sensitive to normalization, therefore by conceptually separating them, the cleanup step is associative with multiplication, allowing the simplification I did.

The goal here is *not* to produce what would have happened if MR FIX was run only on the extracted runs, because the cleanup there would be worse.  I *am* saying that it would be *wrong* to derive something from a subset of the full MR FIX run set and use that to do variance normalization of data from the full concatenated file, because the runs in the concatenated file were *not* restored to their pre-vn state.

Maybe it'll be easier to understand from this angle: if you ran MR FIX on just that set of runs, not only would clean_vn be different, but the normalization of its concatenated file would be different than the normalization of the runs extracted from the "MR FIX on everything" version.

coalsont: However, I don't know the details of the generation of the clean_vn file, and it looks like the generation of the concatenated file just uses the average of the _vn files, not clean_vn.

glasserm: The model is different: regressing out all structured signal versus reconstructing the unstructured data.  The maps have a very similar pattern, but the former method seems to be more forgiving for edge case voxels.

coalsont: Actually, not quite that simple - functionhighpass... uses the icaDim unstructured noise estimate for the highpassed concat data when writing the temporary concatenated file before running fix.  I assume fix doesn't change the normalization?

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: I think that is accurate, but Matt is the only one who knows for sure.  What he has said suggests that it is likely the case.

To be clear, in groups other than the HCP YA, it doesn't count as a round of dedrifting at all, because the "extra round" is derived from HCP YA data and not recomputed from the new group - this is intentional, to avoid causing a different group to add bias via a different dedrifting (in particular, this dedrifting is computed relative to MSMSulc, and therefore could be large if recomputed in a different population).

DeDriftAndResample/DeDriftAndResamplePipeline.sh
glasserm: It is recomputed using a different approach for clean_vn.

coalsont: What is the clean_vn reconstructed from, and how?  It can't be the ICA-type "reconstruction", because the unstructured noise by definition can't be reduced in dimensionality, so I don't know what you mean by that word here.

Back on point, the clean_vn should be conceptually the same normalization as the corresponding _vn file, regardless of however strange the raw runs' variance maps were, right?

glasserm: clean_vn: regressing out all structured signals
_vn: reconstructing unstructured noise PCA

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Resolving this conversation by making it an independent issue: https://github.com/Washington-University/HCPpipelines/issues/185

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: PCA representing unstructured noise sounds unstable to me.  It would be nice if the "bias corrected" concatenated files could be made by using the clean_vn method to close off a caveat, but from what I can tell they are written out before the ICA is computed.

coalsont: Here is the comment I have added:
```
# The per-run differences in (unstructured) noise variance were removed before concatenation, and when
#  going back to "bias corrected" values, it used something conceptually equivalent to this _clean_vn map
#  in reverse (which is close to the average of the original _vn maps), so it is equally applicable to
#  any subset of runs
```
What I objected to from @mharms version is it being called "convenience", when in fact what @mharms seemed to be suggesting for the "inconvenient" version would instead be incorrect (because the concatenated MR FIX run was made in such a way that it has one, and only one, correct variance normalization map, which is by definition applicable to every single frame independently).  @mharms, do you object to my version of the comment?

mharms: How about something more like the following, which is longer, but I think fills in some of the details, and avoids the phrase "bias corrected", which I think is confusing.
```
# The per-run differences in (unstructured) noise variance were removed before concatenation.
# The average of those maps (across runs) was multiplied back into the concatenated time series 
# (in 'hcp_fix_multi_run'), so that the entire concatenated time series has a spatial pattern of 
# unstructured noise consistent with the average across runs.
# Given this manner of constructing the concatenated time series, there is a single variance normalization
# map that is applicable to any subset of runs extracted from the full concatenated set.
# We use the "clean_vn" map for this purpose and thus copy it from the full concatenated set to the 
# extracted set of runs.
# As a final subtlety, note that the "clean_vn" map itself is not identical to the aforementioned average of
# the individual run vn maps, but it is conceptually very similar. (In particular, clean_vn is derived within 
# FIX itself from the concatenated time series by regressing out all structured signals and using the 
# residual to estimate the unstructured noise. Whereas the individual run vn maps were computed using
# PCA-based reconstruction of the unstructured noise in 'icaDim.m').
```

glasserm: Seems fine

coalsont: Rearranging one sentence for clarity:
```
# Given this manner of constructing the concatenated time series, any subset of runs extracted from
#  the full concatenated set should use this same average map for later variance normalization.
```

PR #187, opened 2020-08-04
Use shell internel string operation for clearness

Low-power:
Using this internal string operation instead of command substitution will be much more clear and efficient.

### DIFF
--- a/ICAFIX/README.md
+++ b/ICAFIX/README.md
@@ -141,7 +141,7 @@ MIRROR="http://cloud.r-project.org"
 
 for package in $PACKAGES
 do
-    wget "$MIRROR"/src/contrib/Archive/$(echo "$package" | cut -f1 -d_)/"$package".tar.gz || \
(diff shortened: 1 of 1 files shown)


coalsont:
Thanks for your interest.  Honestly, I can't really remember the differences in the `%`, `%%`, `#`, and `##` shell string manipulation, and always have to look them up (and they specify what to remove, not what to keep, which is not as intuitive).  So, I would consider `cut` to be clearer, especially when extracting only one field.  Considering each loop installs an R package, which invokes decompression and then a compiler toolchain, shell efficiency is not much of a practical concern here.

Low-power:
OK, it seems the readability is depending on person; the `cut(1)` command here is actually a little less readable than shell string manipulation syntax for me, but obviously not for everyone.

Low-power:
OK, it seems the readability is depending on person; the `cut(1)` command here is actually a little less readable than shell string manipulation syntax for me, but obviously not true for everyone.

Issue #189, opened 2020-08-13
Propagate a couple flags into DiffPreprocPipeline.sh


mharms:
I believe that all the flags/functionality available in the separate `DiffPreprocPipeline_{PreEddy,Eddy,PostEddy}.sh` scripts are available from the `DiffPreprocPipeline.sh` script (either as same named flags, or via the generic `--extra-eddy-arg` option) EXCEPT for the `--no-gpu` and `--cuda-version` options in `DiffPreprocPipeline_Eddy.sh`.  We should propagate both of those flags into the `DiffPreprocPipeline.sh` script.

mharms:
@MichielCottaar  Would you be willing to do this?  It originates as a request from our QuNex (container) collaborators.  Thx.

PR #193, opened 2020-08-24
Pass on gpu options

MichielCottaar:
Passes on the --no-gpu and --cuda-version flag from DiffPreprocPipeline.sh to DiffPreprocPipeline_Eddy.sh.

Also contains two other minor fixes:
- replaces some tabs with spaces, so that they are not mixed together
- only print no_gpu when running DiffPreprocPipeline_Eddy.sh if no_gpu has actually been set to something

### DIFF
--- a/DiffusionPreprocessing/DiffPreprocPipeline.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline.sh
@@ -175,6 +175,12 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
                           eddy binary, the following sequence will work:
 
                             --extra-eddy-arg=-flag --extra-eddy-arg=value
+  [--no-gpu]              If specified, use the non-GPU-enabled version of eddy.
+                          Defaults to using the GPU-enabled version of eddy.
+  [--cuda-version=X.Y]    If using the GPU-enabled version of eddy, then this
+                          option can be used to specify which eddy_cuda binary
+                          version to use. If specified, FSLDIR/bin/eddy_cudaX.Y
+                          will be used.
 
   [--combine-data-flag=<value>]
                           Specified value is passed as the CombineDataFlag value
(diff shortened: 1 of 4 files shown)


MichielCottaar:
This resolves #189

DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh
```diff
@@ -379,7 +379,9 @@ get_options() {
 	echo "   resamp_value: ${resamp_value}"
 	echo "   ol_nstd_value: ${ol_nstd_value}"
 	echo "   extra_eddy_args: ${extra_eddy_args}"
-	echo "   no_gpu: ${no_gpu}"
+	if [ ! -z ${no_gpu} ]; then
```
coalsont, line 382: I think our convention in other scripts is to print argument values even if they are empty or used the default.

MichielCottaar: Ok, I've adjusted this to print all argument values. To make the printed output more meaningful I set the default value from no_gpu (and SelectBestB0) to false rather than an empty string.

DiffusionPreprocessing/DiffPreprocPipeline_Eddy.sh
```diff
@@ -223,7 +223,7 @@ get_options() {
 	resamp_value=""
 	unset ol_nstd_value
 	extra_eddy_args=""
-	no_gpu=""
+	no_gpu="False"
```
mharms, line 226: Our changing convention on the case across these scripts is annoying (False vs false vs FALSE), but I see that what you did in `DiffPreprocPipeline_Eddy.sh` is consistent with the rest of the script.

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -513,12 +512,12 @@ main() {
...
 	pre_eddy_cmd+=" --b0maxbval=${b0maxbval} "
 	pre_eddy_cmd+=" --printcom=${runcmd} "
-	if [ ! -z "${SelectBestB0}" ]; then
+	if [ "${SelectBestB0}" == "true" ]; then
 		pre_eddy_cmd+=" --select-best-b0 "
 	fi
 
 	log_Msg "pre_eddy_cmd: ${pre_eddy_cmd}"
-	${pre_eddy_cmd}
+	#${pre_eddy_cmd}
```
coalsont, line 520: I'm guessing these were commented out for testing the variable printing, and weren't intended to be committed this way?  See also lines 545 and 562.

MichielCottaar: Oops

glasserm:
Is this ready now @coalsont ?

PR #198, opened 2020-10-26
BUG: fixes b-value & gradient orientation for the selected b0

MichielCottaar:
This fixes a bug I previously introduced in the code base. When extracting the b-value/gradient orientation from the best b0, best_b0_pos_index gives the index out of all b0's, but this index is used on the full bvals/bvecs. This patch creates a bvals/bvecs for only the Pos b0 values and then uses best_b0_pos_index to extract the correct b0.


glasserm:
Seems fine to me.  I presume we need to incorporate this before running diffusion preprocessing for lifespan?

mharms:
Yes.  Maybe we can incorporate this on top of 6c33037, and tag as `v4.3.0-rc.2`?

coalsont:
I will do that.

coalsont:
Not sure why it didn't have the rebase option selected, I can fix that though.

coalsont:
Tag v4.3.0-rc.2 (which does not contain any other changes to master after v4.3.0-rc.1) now exists, and the merge has been modified to be a rebase.

PR #202, opened 2020-12-02
Support CombineDataFlag of 0 when averaging rotated bvecs

MichielCottaar:
It seems like this was never actually supported

### DIFF
--- a/DiffusionPreprocessing/scripts/eddy_postproc.sh
+++ b/DiffusionPreprocessing/scripts/eddy_postproc.sh
@@ -114,7 +114,7 @@ else # Combining across diffusion directions with opposing phase-encoding polari
 
 	# Average Eddy-Rotated bvecs. Get for each direction the two b matrices, average those and then eigendecompose the average b-matrix to get the new bvec and bval.
 	# Also outputs an index file (1-based) with the indices of the input (Pos/Neg) volumes that have been retained in the output
-	${globalscriptsdir}/average_bvecs.py ${eddydir}/Pos.bval ${eddydir}/Pos_rotated.bvec ${eddydir}/Neg.bval ${eddydir}/Neg_rotated.bvec ${datadir}/avg_data ${eddydir}/Pos_SeriesVolNum.txt ${eddydir}/Neg_SeriesVolNum.txt
+	${globalscriptsdir}/average_bvecs.py ${eddydir}/Pos.bval ${eddydir}/Pos_rotated.bvec ${eddydir}/Neg.bval ${eddydir}/Neg_rotated.bvec ${datadir}/avg_data ${CombineDataFlag} ${eddydir}/Pos_SeriesVolNum.txt ${eddydir}/Neg_SeriesVolNum.txt
 
 	mv ${datadir}/avg_data.bval ${datadir}/bvals
 	mv ${datadir}/avg_data.bvec ${datadir}/bvecs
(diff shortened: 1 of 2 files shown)


DiffusionPreprocessing/scripts/eddy_postproc.sh
```diff
@@ -114,7 +114,7 @@ else # Combining across diffusion directions with opposing phase-encoding polari
 
 	# Average Eddy-Rotated bvecs. Get for each direction the two b matrices, average those and then eigendecompose the average b-matrix to get the new bvec and bval.
 	# Also outputs an index file (1-based) with the indices of the input (Pos/Neg) volumes that have been retained in the output
-	${globalscriptsdir}/average_bvecs.py ${eddydir}/Pos.bval ${eddydir}/Pos_rotated.bvec ${eddydir}/Neg.bval ${eddydir}/Neg_rotated.bvec ${datadir}/avg_data ${eddydir}/Pos_SeriesVolNum.txt ${eddydir}/Neg_SeriesVolNum.txt
+	${globalscriptsdir}/average_bvecs.py ${eddydir}/Pos.bval ${eddydir}/Pos_rotated.bvec ${eddydir}/Neg.bval ${eddydir}/Neg_rotated.bvec ${datadir}/avg_data ${CombineDataFlag} ${eddydir}/Pos_SeriesVolNum.txt ${eddydir}/Neg_SeriesVolNum.txt
```
mharms, line 117: Have you tested that this works?  Because based on the definition of `main()` in `average_bvecs.py`, it doesn't appear to me that we've properly provided inputs for `bvalsoutfile`, `bvecsoutfile`, and `indicesoutfile`.

mharms (commented):
It doesn't appear to me that the call to `average_bvecs.py` in `eddy_postproc.sh` is matched in terms of necessary arguments to the definition of `main()`.

DiffusionPreprocessing/scripts/eddy_postproc.sh
MichielCottaar: Yeah, I've tested running this. The usage for the updated average_bvecs.py is:
`usage: average_bvecs bvals1 bvecs1 bvals2 bvecs2 output_basename combine_data_flag [overlap1 overlap2]`

This matches the way I run the script. `bvalsoutfile`, `bvecsoutfile`, and `indicesoutfiles` are defined based on the `output_basename`

global/scripts/average_bvecs.py
```diff
@@ -27,6 +27,7 @@ def main(bvals1file,
          bvalsoutfile,
          bvecsoutfile,
          indicesoutfile,
+         only_combine,
```
mharms, line 30: I'm confused -- this code block here does not appear consistent with `usage: average_bvecs bvals1 bvecs1 bvals2 bvecs2 output_basename combine_data_flag [overlap1 overlap2]`

MichielCottaar: That's because python scripts don't start by calling the main function. The first code to be run is at the very end of the script (after the `if __name__ == "__main__":` on line 293). The code after that actually parses the arguments from the command line and defines `bvalsoutfile`, `bvecsoutfile`, and `indicesoutfiles`. Finally it calls the `main` function to do the actual work.

mharms: I see.  Thanks for the tutorial.

MichielCottaar:
I'm just rerunning the post_eddy script for the complete dataset (i.e., where we have the same amount of data for both phase encoding). I'm not expecting these changes to affect that output, but it would be good to check. I should know by tomorrow.

MichielCottaar:
The testing is done; this can be merged.

PR #203, opened 2020-12-23
Add generation of scene/snapshots for fMRIQC

mharms:
Still need to revise and add the scene template itself, but wanted to get the code up for initial review.

Of note, the bias corrected, normalized, but not brain masked version of the scout is named with a "_notmasked" suffix currently.  "unmasked" was a bit too cryptic in my opinion, although open to other suggestions.


coalsont:
`_nomask` could be a shorter but equivalent specifier, I like shorter when it doesn't sacrifice clarity.  For brainstorming, "with background" is a different way to think about it, but `_withbg` doesn't look as clear, and `background` is fairly long to put in a filename.

I think a few files use `_brain` for the masked version, and while `_nobrain` might be amusing, it is unlikely to be helpful.

mharms:
`_nomask` seems like a good alternative.

PostFreeSurfer/scripts/GenerateStructuralScenes.sh
```diff
@@ -192,19 +169,18 @@ TemplatesFolderDummyStr="TemplatesFolder"
 
 scriptDir=$(pwd)
 
-OutputSceneFolderSubj=$OutputSceneFolder
-mkdir -p $OutputSceneFolderSubj
-relPathToStudy=$(relativePath $OutputSceneFolderSubj $StudyFolder)
+mkdir -p $OutputSceneFolder
```
coalsont, line 172: Looks like I missed putting quotes around some variables here.

glasserm (commented):
The code looks good to me and thanks for making a no mask SBRef.  I would like to look over some test results prior to approving.

mharms:
With the above additions, this should now be operational.  I'm running another test overnight (denovo execution of fMRIVolume and fMRISurface) and can generate a `-zip-scene-file` for sharing tomorrow.

mharms:
The code as it exists in this PR works.  I'll email you the location of the scene file.

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
```diff
@@ -139,25 +160,39 @@ T1wFolder="$Path"/"$Subject"/"$T1wFolder"
...
+# ------------------------------------------------------------------------------
 
-#Subcortical Processing
-log_Msg "Subcortical Processing"
-"$PipelineScripts"/SubcorticalProcessing.sh "$AtlasSpaceFolder" "$ROIFolder" "$FinalfMRIResolution" "$ResultsFolder" "$NameOffMRI" "$SmoothingFWHM" "$GrayordinatesResolution"
+if ((doProcessing)); then
+    #Make fMRI Ribbon
+    #Noisy Voxel Outlier Exclusion
+    #Ribbon-based Volume to Surface mapping and resampling to standard surface
+	log_Msg "Make fMRI Ribbon"
```
coalsont, line 171: This appears to contain tabs, while most lines use spaces for indenting, please change to spaces.

coalsont: There are also sections of the file that use 2 spaces for indenting, while elsewhere we mostly use 4, it would be nice if that was made consistent at 4.

mharms: I'm not working on anything else, so please feel free to fix the tab/spacing issue.

mharms:
@glasserm : It would be great if you can review the scene file.  Would like to get this PR into the master, tag a new 4.3rc, and then have QuNex folks build a new container using that.

glasserm:
> @glasserm : It would be great if you can review the scene file. Would like to get this PR into the master, tag a new 4.3rc, and then have QuNex folks build a new container using that.

I looked but did not find an e-mail from you showing the location of the scene file.

mharms:
@glasserm I sent the email again.

glasserm (approved):

mharms:
@coalsont: Here's the `git diff` on the approach that @hodgem took for the `set -e` issue:
```
 main()
 {
+
        log_Msg "Starting main functionality"
+       set -e
 
        # Retrieve positional parameters
        local StudyFolder="${1}"
@@ -1045,7 +1047,9 @@ main()
         fi
     done
 
+       set +e
        log_Msg "Completing main functionality"
+
 }
```

He confirmed that this does indeed catch the "error" when wb_command got killed due to insufficient memory.  Although your coupling its placement to the debug.shlib line seems cleaner, if it doesn't need to be inside `main()`

mharms:
Just for public documentation purposes, the `set -e` issue mentioned above it completely unrelated to this PR, and relates to something we need to tweak in `DeDriftAndResample`

coalsont:
The `set -e` should be done immediately before sourcing debug.shlib, so that it can disable it for interactive shells, substantially reducing a considerable annoyance when doing bash debugging (unexpectedly losing the entire shell, with all carefully set up script arguments, when `set -e` trips).

PR #204, opened 2021-01-22
Branch to make subject-level summary directory from Level1 or Level2 as needed

gcburgess:
Edited comments in TaskfMRIAnalysis.sh to correct the information about FSL 5.0.6 and
earlier, and to clarify how Level1 and Level2 analyses are launched in serial.


glasserm (dismissed):
Looks fine to me..

mharms (commented):
@gcburgess , You aren't actually wanting to merge this into master yet, are you?

I'm assuming that this is just the first of what will be a series of commits.

gcburgess:
@mharms That's correct. I was assuming that we'd do one final merge into master once the branch was fully complete. I was expecting to push separate commits to the branch as they're completed, so that others can comment while work on further changes occurs in parallel.

mharms:
@gcburgess : If you need/want comments at an "intermediate" stage, let us know. Otherwise, I'd prefer to wait to review until we have a reasonably complete version that you think is close to being actually ready to merge. (That way, we don't need to waste time reviewing changes that subsequently become moot due to some other later commits).

gcburgess:
@mharms @glasserm @coalsont  Please review edits to replace previous usage and argument parsing with newopts.shlib.

FYI, I do plan to make multiple commits, but I would prefer if you review each commit separately. That way, I can take your suggestions into consideration during subsequent commits.

glasserm:
Greg, how did you determine mandatory vs optional arguments?

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -261,15 +250,15 @@ if [ "$LevelTwofMRIName" != "NONE" ]
 then
 	# Combine Data Across Phase Encoding Directions in the Level 2 Analysis
 	log_Msg "RUN_LEVEL2: Combine Data Across Phase Encoding Directions in the Level 2 Analysis"
-	log_Msg "RUN_LEVEL2: Issuing command: ${HCPPIPEDIR_tfMRIAnalysis}/TaskfMRILevel2.sh $Subject $ResultsFolder $DownSampleFolder $LevelOnefMRINames $LevelOnefsfNames $LevelTwofMRIName $LevelTwofsfNames $LowResMesh $FinalSmoothingFWHM $TemporalFilter $VolumeBasedProcessing $RegName $Parcellation"
+	log_Msg "RUN_LEVEL2: Issuing command: ${HCPPIPEDIR_tfMRIAnalysis}/TaskfMRILevel2.sh $Subject $ResultsFolder $DownSampleFolder $LevelOnefMRINames $LevelOnefsfNames $LevelTwofMRIName $LevelTwofsfName $LowResMesh $FinalSmoothingFWHM $TemporalFilter $VolumeBasedProcessing $RegName $Parcellation"
```
coalsont, line 253: Can you explain why `LevelTwofsfNames` variable was changed to singular?  Was this previously a bug, or is the usage of `TaskfMRILevel2.sh` changing?

coalsont: Sorry, didn't read the full commit message.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
 #
 #~ND~END~
 
+set -eu
```
mharms, line 42: Is this just for temporary debugging purposes?  At some point, we hope that the `debug.shlib` library will handle error trapping.  At a minimum, would be good to put this line immediately in front of the sourcing of `debug.shlib`.  See `DeDriftAndResample` for a recent example where we had to do this.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
...
 
-if [ -z "${HCPPIPEDIR}" ]; then
-  echo "${script_name}: ABORTING: HCPPIPEDIR environment variable must be set"
-  exit 1
-fi
+#arguments to opts_Add*: switch, variable to set, name for inside of <> in help text, description, [default value if AddOptional], [compatibility flag, ...]
+#help info for option gets printed like "--foo=<$3> - $4"
+opts_AddMandatory '--study-folder' 'Path' '/path/to/study/folder' "directory containing imaging data for all subjects"
+opts_AddMandatory '--subject' 'Subject' 'SubjectID' ""
+opts_AddMandatory '--lvl1tasks' 'LevelOnefMRINames' 'tfMRI_TaskName_RL@tfMRI_TaskName_LR' "List of subdirectories containing each time series image for the TaskName task. Multiple scan subdirectories should be provided as a single string separated by '@' character." #Assumes these subdirectories are located in the SubjectID/MNINonLinear/Results directory. Also assumes that timeseries image filename begins with this string.
```
mharms, line 88: For the description, I would use a terminology more in keeping with other scripts.  i.e., it's a list of "fMRI scan names" , not "subdirectories" (which could imply that providing a path as part of the name is appropriate).

mharms: Also, I suggest we switch to using AP and PA as the examples (consistent with HCP-LS), not the RL/LR from HCP-YA.

glasserm: Why does that matter?

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
...
+opts_AddOptional '--finalsmoothingFWHM' 'FinalSmoothingFWHM' 'integer' "Default=2mm, which is no additional smoothing above HCP minimal preprocessing pipelines outputs. Total desired smoothing (finalsmoothingFWHM) is reached by applying additional smoothing onto origsmoothingFWHM." '2'
+opts_AddOptional '--confound' 'Confound' 'filename' "Confound matrix text filename (e.g., output of fsl_motion_outliers). Assumes file is located in SubjectID/MNINonLinear/Results/tfMRI_TaskName_Scan. Default='NONE'" 'NONE'
+opts_AddOptional '--highpassfilter' 'TemporalFilter' 'integer' "Value for nonlinear highpass filter (in seconds). Default=200. Use 2000 for linear detrend." '200'
+opts_AddOptional '--vba' 'VolumeBasedProcessing' 'YES/NO' "Default=NO. CAUTION: Only use YES if you want unconstrained volumetric blurring of your data, otherwise set to NO for faster, less biased, and more senstive processing (grayordinates results do not use unconstrained volumetric blurring and are always produced)" 'NO'
+opts_AddOptional '--regname' 'RegName' 'RegName' "Name of surface registration technique. Default=NONE, which will use the default surface registration." 'NONE'
```
mharms, line 99: "default surface registration" is somewhat ambiguous.  I'd say, "which will use the default (MSMSulc) surface registration".

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
...
+opts_AddOptional '--lvl2task' 'LevelTwofMRIName' 'tfMRI_TaskName' "Name of Level2 subdirectory in which all Level2 feat directories are written for TaskName. Default is 'NONE', which means that no Level2 analysis will run." 'NONE'
+opts_AddOptional '--lvl2fsf' 'LevelTwofsfName' 'DesignName_TaskName' "Prefix of design.fsf filename for the Level2 analysis for TaskName. If no value is passed to --lvl2fsf, the value will be set to the same list passed to --lvl2task."
+opts_AddOptional '--lowresmesh' 'LowResMesh' 'integer' "Default=32, which is appropriate for HCP minimal preprocessing pipeline outputs" '32'
+opts_AddOptional '--grayordinatesres' 'GrayordinatesResolution' 'integer' "Default=2mm, which is appropriate for HCP minimal preprocessing pipeline outputs" '2'
```
mharms, line 93: Maybe for this, and the other variables like it, we should do: `'integer (mm)' "Default=2, ...`"

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
...
+opts_AddOptional '--origsmoothingFWHM' 'OriginalSmoothingFWHM' 'integer' "Default=2mm, which is appropriate for HCP minimal preprocessing pipeline outputs" '2'
+opts_AddOptional '--finalsmoothingFWHM' 'FinalSmoothingFWHM' 'integer' "Default=2mm, which is no additional smoothing above HCP minimal preprocessing pipelines outputs. Total desired smoothing (finalsmoothingFWHM) is reached by applying additional smoothing onto origsmoothingFWHM." '2'
+opts_AddOptional '--confound' 'Confound' 'filename' "Confound matrix text filename (e.g., output of fsl_motion_outliers). Assumes file is located in SubjectID/MNINonLinear/Results/tfMRI_TaskName_Scan. Default='NONE'" 'NONE'
```
mharms, line 96: "... assumes file is located in \<SubjectID>/MNINonLinear/Results/\<fMRIName> ..."

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
...
+opts_AddOptional '--highpassfilter' 'TemporalFilter' 'integer' "Value for nonlinear highpass filter (in seconds). Default=200. Use 2000 for linear detrend." '200'
```
mharms, line 97: I wouldn't call 2000 a linear detrend.  Just leave that bit out.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
mharms: It is simply a more pertinent example for our current data/processing, and also how every one is acquiring their data.  (No one is acquiring RL/LR).

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
mharms: Also, would be good to make clearer that this represents _additional_ temporal smoothing to be applied during task processing, esp. since (as I understand it), there is going to be argument added for specifying the hp filter (or filter string) that was already used previously (e.g., ICAFIX).

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -39,56 +39,92 @@
...
+#help info for option gets printed like "--foo=<$3> - $4"
+opts_AddMandatory '--study-folder' 'Path' '/path/to/study/folder' "directory containing imaging data for all subjects"
+opts_AddMandatory '--subject' 'Subject' 'SubjectID' ""
+opts_AddMandatory '--lvl1tasks' 'LevelOnefMRINames' 'tfMRI_TaskName_RL@tfMRI_TaskName_LR' "List of subdirectories containing each time series image for the TaskName task. Multiple scan subdirectories should be provided as a single string separated by '@' character." #Assumes these subdirectories are located in the SubjectID/MNINonLinear/Results directory. Also assumes that timeseries image filename begins with this string.
+opts_AddOptional '--lvl1fsfs' 'LevelOnefsfNames' 'DesignName_TaskName_RL@DesignName_TaskName_LR' "List of prefixes in the design.fsf filenames for each time series image for the TaskName task. Should contain same number of design files as time series images in --lvl1tasks option. (N-th design will be used for N-th time series image.) Separate multiple design files by '@' character. If no value is passed to --lvl1fsfs, the value will be set to the same list passed to --lvl1tasks."
```
mharms, line 89: "List of fsf file names (don't include .fsf extension) for each fMRI scan name..."

mharms (commented):
For some reason, it didn't give me the option to directly create a suggestion that you could accept.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
coalsont: `set -eu` is intentionally part of the newopts template, it catches errors from typoing a variable name or using a variable before it is set, which the debug trap can't catch.  It is at the top of the file, and not too many lines from the `source` lines, unlike the situation in DeDrift where the `source` lines are at the bottom of the file, so I don't expect as much potential here for copy paste trouble.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
coalsont: I would probably put the "mm" into the explanation and leave it saying `[--grayordinates=<integer>]`, but yes, the default should probably say `2`, without "mm" on the end.

coalsont: Actually, isn't the main other choice for grayordinates 1.6?  It probably shouldn't say integer then...

glasserm: greyordinates are not necessarily integer resolution.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: > For the description, I would use a terminology more in keeping with other scripts. i.e., it's a list of "fMRI scan names" , not "subdirectories" (which could imply that providing a path as part of the name is appropriate).

The only concern that I have is that this does refer specifically to the value of the subdirectory name. If we say "list of fMRI scan names", are users going to set the value to *tfMRI_TASK_AP_Atlas_YadaYada.dtseries.nii*?

glasserm: Is ${fMRIName} not informative enough to folks?  fMRI base name?

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: Similar to above, the script is specifically expecting the prefix of the design file naming convention. The user should not enter "prefix_hp200_s4". The other portions of the file name are currently determined by other variables.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: Hmmm... since the variable is ultimately used to specify the name of file in `wb_command -cifti-create-dense-timeseries`, I guess we should technically say that it is a string. If we call it a float, people might enter 2.0 or 1.60 and expect it to work.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
coalsont: The convention I have been using is that something that takes a filename says "file" or something more descriptive like "timeseries", while a string used in naming is "name" or "string".  In the fairly new ExtractFromMRFIXConcat.sh, I used `--multirun-fix-names=<run1@run2...> - list of run names used in MR FIX, in the SAME ORDER, separated by @s`, sidestepping the use of realistic names in the example, but perhaps not as obvious, and made more tolerable by the fact that it is a copy of one of the MR FIX arguments.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
coalsont: A possibility is to say `--grayordinatesres=<2 or 1.6>`, or `--grayordinatesres=<number> - 2 (default) or 1.6 ...`, assuming we don't plan to make another grayordinates resolution soon.  Not future-proof, but fairly clear.

For reference, here is what I did in PostFreeSurfer (mandatory, not optional, but we expect the user to use Examples/Scripts/...Batch.sh to launch it), it may not be perfect, but we also may not need to be so picky:
```
   --grayordinatesres=<number> - usually '2', resolution of grayordinates to
        use
```

glasserm: Other species have other resolutions.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
mharms: Where does the user learn that the `.fsf` are supposed to have `hpX_sY` as part of their file name?

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: Changed example to 'ScanName1@ScanName2'

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: Changed example to 'DesignName1@DesignName2'.

The file naming conventions exist. If we don't show them in the arguments, I guess the user will need to learn the file naming conventions and CinaB directory structure on by familiarizing themselves with the data and the code.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: Changed to:
opts_AddOptional '--grayordinatesres' 'GrayordinatesResolution' 'integer_as_string' "Use value in 'Atlas_ROIs' filename; Default=2mm ...

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
gcburgess: Changing to: 
"Apply *additional* highpass filter (in seconds) to time series and task design. This is above and beyond temporal filter applied during preprocessing. Default=200."

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
coalsont: This should not use the word "integer", 1.6 is a likely value.  "2mm" is not what the default is, the default is "2", and we should not make users think that the argument should contain the characters "mm".

gcburgess: Oops! Yeah, I change it to:
'number_as_string' "Value (in mm) from 'Atlas_ROIs' filename; Default='2'

coalsont: I don't know that "number_as_string" is any clearer to users than "number" (bets on whether someone tries "--grayordinatesres=two"?), and I think it looks awkward.

gcburgess: That's a valid point. I'll put 'number'. If anyone enters '2.0' and it doesn't run, we'll just need to leave it to them to debug why.

gcburgess:
@glasserm @mharms @coalsont
This commit makes improvements to the usage that were requested by review / comments above. It also allows TemporalFilter to be bypassed by setting --highpassfilter=NONE.

Adding the "ProcString" to specify filename of FIX-cleaned time series will come in a separate commit, because it requires many many changes to filenames throughout Level1 and Level2 scripts.

gcburgess (commented):
@glasserm @mharms @coalsont
This commit makes improvements to the usage that were requested by review / comments above. It also allows TemporalFilter to be bypassed by setting --highpassfilter=NONE.

Adding the "ProcString" to specify filename of FIX-cleaned time series will come in a separate commit, because it requires many many changes to filenames throughout Level1 and Level2 scripts.

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
```diff
@@ -374,7 +383,11 @@ if [[ $runParcellated == true || $runDense == true ]]; then
 	rm ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ParcellationString}_FAKENIFTI.nii.gz ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ParcellationString}_FAKENIFTI_mean.nii.gz
 fi
 
-if $runVolume; then
+if [[ "${TemporalFilter}" == "-1" && $runVolume == true]]; then
+	# We drop the "dilMrim" string from the output file name, so as to avoid breaking
```
glasserm, line 387: What is dilMrim?

coalsont: The structure of this seems a little odd, equivalent to:
```
if B and runVolume
...
else
    if runVolume
...
    fi
fi
```
I would have done this:
```
if runVolume
    if B
...
    else
...
    fi
fi
```
Persinally, I avoid elif, as mostly being a construct to substitute for case/switch (which in some languages can't handle strings), but bash case handles strings (and even globbing).

gcburgess: > What is dilMrim?

'dilMrim' is the string in the filename of the volume image after edge dilation. (The volume dilation has been around for ages... There's a long explanation starting at the line '#Add volume dilation'.)

In the current code, the output filename of the fslmaths -bptf call removes the 'dilMrim' string. Therefore, if we skip temporal filtering, we still need to remove that 'dilMrim' string.

(The fact that this string is added in volume smoothing and then removed in temporal smoothing needs to be fixed so that we can skip volume smoothing and temporal smoothing and still end up with the right filename. I'll need to address this at some point in the future.)

mharms: I added that code a long time ago.  Might just be easiest (and less prone to error) to leave it as is, and just have a conditional that checks for absence of temporal smoothing and then renames as necessary.

gcburgess (commented):
@glasserm @coalsont @mharms 

This commit adds the options for --lowpassfilter and --procstring to the usage statement. (I will incorporate the remaining changes to use those variables in a future commit.)

I also reordered the options in the usage, to group together similar options. I'm hoping that we can approve these final changes to the usage, so that we can pass that info to Grega and Jure.

mharms:
`--lowpassfilter` is a new option that you intend to add?  I don't believe that any of the other scripts support low-pass filterting...

glasserm:
It is useful for comparison with things like Wishart Filtering, but the usage statement should be clear that it is not recommended.

gcburgess:
@glasserm @coalsont @mharms 

Added code block to check for missing files and die gently if missing.

TaskfMRIAnalysis/TaskfMRIAnalysis.sh
```diff
@@ -52,9 +52,10 @@ then
     export HCPPIPEDIR="$(dirname -- "$0")/.."
 fi
 
+source "$HCPPIPEDIR/global/scripts/opts.shlib"			# Function for getting HCPpipelines version
```
coalsont, line 55: We should almost certainly not mix opts.shlib and newopts.shlib.  I don't see other changes to this file in this commit that use it, so I don't understand why it was sourced.

As for the comment on the line, the main way to get pipeline version info is now by running `"$HCPPIPEDIR"/show_version`.  It also has a `--short` option.  The other existing methods have been refitted to run that script one way or another.  I have not put a hardcoded `--version` option into newopts.shlib (while `--help` and `-h` are hardcoded), this could be added if it is desired to have all scripts return the overall pipelines version info (which seems slightly silly, as the user can run `show_version` instead of using some other script).

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
```diff
@@ -190,6 +190,46 @@ log_Msg "MAIN: SET_NAME_STRINGS: RegString: ${RegString}"
 log_Msg "MAIN: SET_NAME_STRINGS: ParcellationString: ${ParcellationString}"
 log_Msg "MAIN: SET_NAME_STRINGS: Extension: ${Extension}"
 
+##### CHECK_FILES: Check that necessary inputs exist before trying to use them #####
+# Assemble list of input filenames that need to be checked
+Filenames="";
```
coalsont, line 195: This could be done with whitespace-safety like so:

```
Filenames=()

if <whatever>; then
    Filenames+=("$somefile")
fi
...
missingfiles=()
for name in "${Filenames[@]}"; do
    if ! [[ -e "$name" ]]; then
        missingfiles+=("$name")
    fi
done
if (( ${#missingfiles[@]} > 0 )); then
    log_Err_Abort "Missing necessary input files: ${missingfiles[*]}"
fi
```

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
```diff
@@ -190,6 +190,46 @@ log_Msg "MAIN: SET_NAME_STRINGS: RegString: ${RegString}"
...
+# Need midthickness GIFTI surfaces
+Filenames="$Filenames ${DownSampleFolder}/${Subject}.L.midthickness.${LowResMesh}k_fs_LR.surf.gii ${DownSampleFolder}/${Subject}.R.midthickness.${LowResMesh}k_fs_LR.surf.gii"
+# Need Atlas_ROIs volume file
+Filenames="$Filenames $ROIsFolder/Atlas_ROIs.${GrayordinatesResolution}.nii.gz"
+
+# Now check each file in list
+missingFiles="";
+for Filename in $Filenames; do
+	# if file does not exist, set errMsg
+	[ -e "$Filename" ] || missingFiles="${missingFiles} ${Filename} "
```
coalsont, line 220: `if` is more readable than short circuit `||`.  I also prefer `[[` to `[` because it has better-behaved syntax.

glasserm:
I didn't see the lowpass variable being passed to fslmaths for CIFTI/parcellated or volume.

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
```diff
@@ -336,60 +416,81 @@ if $runVolume ; then
...
+	if $runVolume; then
+		#Add temporal filtering to the output from above
+		log_Msg "MAIN: TEMPORAL_FILTER: Add temporal filtering to NIFTI file"
+		# Temporal filtering is conducted by fslmaths. 
+		# fslmaths -bptf removes timeseries mean (for FSL 5.0.7 onward), which is expected by film_gls. 
+		# So, save the mean to file, then add it back after -bptf.
+		# We drop the "dilMrim" string from the output file name, so as to avoid breaking
+		# any downstream scripts.
+		fslmaths ${SmoothedDilatedResultFile} -Tmean ${SmoothedDilatedResultFile}_mean
+		fslmaths ${SmoothedDilatedResultFile} -bptf ${hp_sigma} -1 \
```
glasserm, line 486: lp_sigma not used in fslmaths command

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
```diff
@@ -336,60 +416,81 @@ if $runVolume ; then
...
+		# Save mean image
+		fslmaths ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ProcSTRING}${ParcellationString}_FAKENIFTI.nii.gz -Tmean ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ProcSTRING}${ParcellationString}_FAKENIFTI_mean.nii.gz
+		# Use fslmaths to apply high pass filter and then add mean back to image
+		fslmaths ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ProcSTRING}${ParcellationString}_FAKENIFTI.nii.gz -bptf ${hp_sigma} -1 \
+		   -add ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ProcSTRING}${ParcellationString}_FAKENIFTI_mean.nii.gz \
+		   ${ResultsFolder}/${LevelOnefMRIName}/${LevelOnefMRIName}_Atlas${SmoothingString}${RegString}${ProcSTRING}${LowPassSTRING}${ParcellationString}_FAKENIFTI.nii.gz
+		# Convert "fake" NIFTI back to CIFTI
```
glasserm, line 472: lp_sigma not used in fslmaths command

glasserm (changes requested):
I found at least two bugs and also it looks like Tim had made some requests.  Please decide if you want to address those or not as well.  It would also be helpful to look over some sample output.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+# ------------------------------------------------------------------------------
+
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+source "$HCPPIPEDIR/global/scripts/opts.shlib"			# Function for getting HCPpipelines version
```
coalsont, line 53: Do not source `opts.shlib` if you are sourcing/using `newopts.shlib`, `opts.shlib` should be considered legacy-only code at this point, and they should not be mixed.  As for the comment that appears to explain why `opts.shlib` is being sourced, run "$HCPPIPEDIR/show_version" if you need to get the pipelines version info.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+if ((pipedirguessed))
+then
+    log_Err_Abort "HCPPIPEDIR is not set, you must first source your edited copy of Examples/Scripts/SetUpHCPPipeline.sh"
+fi
+
+log_Msg "Platform Information Follows: "
+uname -a
+
+opts_ShowVersionIfRequested $@
+${HCPPIPEDIR}/show_version
```
coalsont, line 129: if you run show_version unconditionally, it doesn't make sense to optionally show version another way, and `opts_ShowVersionIfRequested` is a legacy function, please don't use it.  If we really want all pipeline scripts to have a hardcoded `--version` option that shows the overall version of the entire pipelines repo (seems a bit silly to me, which is why I didn't do it yet), that can be added to `newopts.shlib` without needing any code in each script to activate it.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+log_Msg "MAIN: SET_NAME_STRINGS: TemporalFilterString: ${TemporalFilterString}"
+log_Msg "MAIN: SET_NAME_STRINGS: RegString: ${RegString}"
+
+
+# Determine locations of necessary directories (using expected naming convention)
+ResultsFolder="${Path}/${Subject}/MNINonLinear/Results"
+
+# Determine which analyses need to be packaged into summary directory
+# initialize run variables
+runParcellated=false; runVolume=false; runDense=false;
```
coalsont, line 187: These variables don't appear to get tested, merely set and then ignored.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+Analyses=""; ExtensionList=""; ScalarExtensionList="";
+
+# Determine whether to run Parcellated, and set strings used for filenaming
+if [ "${Parcellation}" != "NONE" ] ; then
+	# Run Parcellated Analyses
+	runParcellated=true;
+	ParcellationString="_${Parcellation}"
+	Extension="ptseries.nii"
+	ScalarExtensionList="${ScalarExtensionList}pscalar.nii "
+	Analyses="${Analyses}ParcellatedStats "; # space character at end to separate multiple analyses
```
coalsont, line 197: `Analyses="${Analyses} ParcellatedStats"` or `Analyses+=" ParcellatedStats"` would be a little more readable.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+  echo "Removing previous summary directory from ${SummaryDirectory}..."
+  rm -r "$SummaryDirectory"
+fi
+
+# Create new summary directory
+echo "Creating new summary directory at ${SummaryDirectory}..."
+mkdir -v $SummaryDirectory
+
+# Create log file with arguments from TaskfMRIAnalysis.sh call
+touch ${SummaryDirectory}/TaskfMRIAnalysis.log
```
coalsont, line 234: Touch doesn't delete if the log happens to exist, and `>>` and similar don't care whether the file exists, this line is basically useless.  I don't think we want scripts to be guessing the name of log files other scripts are using, if that is what is going on here, as that would be inconvenient to edit properly.  I would say either pass in the name of the file to concatenate onto (if that is what we want), or name the hardcoded output according to the current (sub)script's name (and keep it separate from other scripts' outputs).

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+
+# Check if summary directory already exists, and remove old summary directory
+if [ -d "$SummaryDirectory" ]; then
+  echo "Removing previous summary directory from ${SummaryDirectory}..."
+  rm -r "$SummaryDirectory"
+fi
+
+# Create new summary directory
+echo "Creating new summary directory at ${SummaryDirectory}..."
+mkdir -v $SummaryDirectory
```
coalsont, line 231: I am puzzled by the mix of quotes and not-quotes in commands.  It is technically more important to use quotes in commands like `mkdir` (allows them to run on folders with spaces, etc in the name) than it is in reporting like `log_Msg` (mainly allows multiple consecutive spaces as-is, and literal use of some shell-special characters), and yet most uses of unquoted variables are in the processing commands where quotes would have more practical benefit, while for `log_Msg` and similar, you virtually always use quotes.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+				echo "${Analysis}: cope${copeCounter} ($Contrast) exists" >> ${SummaryDirectory}/TaskfMRIAnalysis.log
+			fi
+			for File in {mask,cope1,varcope1,tdof_t1} ; do
+				cifti_in=${LevelTwoFEATDir}/${Analysis}/cope${copeCounter}.feat/${File}.${Extension}
+				if [ -e "$cifti_in" ]; then
+					outdir=${SummaryDirectory}/${Analysis}/cope${copeCounter}.feat
+					if [ ! -d "$outdir" ]; then
+						mkdir -pv $outdir;
+					fi
+					cifti_out=$(echo $cifti_in | sed -e "s|$LevelTwoFEATDir|$SummaryDirectory|")
```
coalsont, line 275: `cifti_out=${SummaryDirectory}/${Analysis}/cope${copeCounter}.feat/${File}.${Extension}` would be clearer.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+					# create symlink to summary directory
+					cp -suv $cifti_in $cifti_out
```
coalsont, line 277: Mac `cp` may not have -s or -u, those are probably GNU specific.  `ln -s` is the expected command for making a single symlink.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+			log_Err_Abort "ERROR: Cannot find Contrasts list at ${LevelOneFEATDir}/design.con. Verify that Level1 analysis completed correctly."
+		fi
+
+		# Create mask and tdof files in tmp directory, to copy them into place as needed
+		tmpdir=${SummaryDirectory}/${Analysis}/tmp
+		mkdir -pv $tmpdir
+		# Symlink Level1 dof and create tdof_t1 image file and mask file
+		dof=`cat ${LevelOneFEATDir}/${Analysis}/dof`;
+		if [[ "$Analysis" = "GrayordinatesStats" || "$Analysis" = "ParcellatedStats" ]] ; then
+			${CARET7DIR}/wb_command -cifti-convert -to-nifti ${LevelOneFEATDir}/${Analysis}/res4d.${Extension} $tmpdir/res4d.nii.gz
```
coalsont, line 309: If you are just doing some simple math on these files, I would suggest `-cifti-stats` and `-cifti-math` (and possibly -volume-stats and -volume-math in the non-cifti path to match) instead of -cifti-convert just to put it through `fslmaths`, as staying within wb_command and non-"fake" formats preserves more provenance metadata.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+		else
+			cp -suv ${LevelOneFEATDir}/${Analysis}/res4d.nii.gz $tmpdir/res4d.nii.gz
+		fi
+		fslmaths $tmpdir/res4d.nii.gz -Tstd -bin $tmpdir/mask.nii.gz;
```
coalsont, line 313: Side note, if you want this `-Tstd -bin` to reliably exclude no-variance elements that aren't exactly zero, you should instead check whether the max and min aren't equal, as computing stdev inevitably contains floating point rounding error when run on anything that isn't zeros.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+			cp -suv ${LevelTwoFEATDir}/Contrasts.txt ${SummaryDirectory}
+		fi
+		
+		copeCounter=1;
+		cat ${LevelTwoFEATDir}/Contrasts.txt | while read Contrast ; do
+			# Check if necessary files exist in cope${i}.feat directory (cope1 should be mean of two runs)
+			if [ -e "${LevelTwoFEATDir}/${Analysis}/cope${copeCounter}.feat/cope1.${Extension}" ]; then
+				echo "${Analysis}: cope${copeCounter} ($Contrast) exists" >> ${SummaryDirectory}/TaskfMRIAnalysis.log
+			fi
+			for File in {mask,cope1,varcope1,tdof_t1} ; do
```
coalsont, line 268: This "brace and commas" syntax is actually special bash syntax that just turns into "words with spaces", and isn't necessary:

`for File in mask cope1 varcope1 tdof_t1; do`

Not sure what people think as far as style is concerned.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -0,0 +1,363 @@
...
+				else
+					# file is missing, write cope number and contrast name to log
+					echo "${Analysis}: NOTE ${File}${copeCounter} ($Contrast) Level1 is missing file $shortName" >> ${SummaryDirectory}/TaskfMRIAnalysis.log
+				fi
+			done
+			cp -v $tmpdir/mask.${Extension} ${SummaryDirectory}/${Analysis}/cope${copeCounter}.feat/
+			cp -v $tmpdir/tdof_t1.${Extension} ${SummaryDirectory}/${Analysis}/cope${copeCounter}.feat/
+			copeCounter=$(($copeCounter+1))
+		done
+		rm -rv $tmpdir
```
coalsont, line 361: `-v` is verbose, why do you want removal of the temporaries to be loud?

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: I'm not sure I follow. What do you want me to change?

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: @coalsont I'm not sure what you're suggesting. 
- This file is a log written by this script only. (Was there a reason you thought it was coming from another script?)
- I can get rid of the `touch` command. (I thought I had come across cases where an initial 'touch' was needed before using `>>` . Since I can recreate that situation here, I'm fine removing it.)
- But, I don't know what you mean by "name the hardcoded output according to the current (sub)script's name".  I want the name to be the same as the primary pipeline script, not this subscript. Even though the log is written by this subscript, it is summarizing what the main TaskfMRIAnalysis pipeline did altogether.

coalsont: The fact that you only used `>>` means it will add to whatever happens to already be in the file.  If you don't want any previous content, do `rm -f` to delete it instead of `touch`.  You should not need to `touch` before `>>` in any circumstance I can think of.

A log named `TaskfMRIAnalysis.log` makes me expect it was written by a `TaskfMRIAnalysis.sh` script, which is part of why I wasn't sure what you were doing with `touch` and `>>`.  I haven't looked at the details of what it contains, or what other logs this script makes, but I was confused by the name, so I will argue for renaming it.  In fact, ".log" implies it was generated while running the script it names, so that contributes to the confusion.  Something like `TaskAnalysisSummary.txt` seems more appropriate for "after the fact summarizing of other processing".

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
coalsont: `mkdir -v "$SummaryDirectory"`

Using quotes like this in all commands that use a variable (that is not intended to become multiple arguments) would make it whitespace-safe.  We have at least one pipelines author who doesn't care to do this, but since you are using quotes already in lots of other places (including simple variable assignment, where technically they don't matter), why not use them in the most important places (commands)?

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: This is addressed in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: This is addressed in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: These are removed in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: This is changed in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: Removed the touch commands in next commit. (However, kept the file name the same.)

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: Changed this in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: Changed this in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: Changed in next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: Actually, `TaskAnalysisSummary.txt` sounds good, so I'll change that in the next commit.

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
gcburgess: Changed in next commit

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
gcburgess: lp_sigma added in next commit.

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
gcburgess: lp_sigma added in next commit.

TaskfMRIAnalysis/scripts/TaskfMRILevel1.sh
gcburgess: Removed short circuit in next commit

glasserm (commented):
Does it work?  The changes look good to me.

PR #207, opened 2021-02-08
rework show_version

coalsont:
Initial attempt at new version reporting.  Question: should it error when "hide rc" is requested by environment variable, but the version isn't a candidate?  If qunex is rebuilt for reasons other than a pipelines release, and the real pipelines release tag is used, that could cause errors (unless qunex noticed the tag is a real release somehow and didn't set the variable).

Remaining items to do:
- ~~remove version.txt and edit scripts that use it~~
- ~~edit other version reporting to use `show_version --short`~~


coalsont:
A quick test before and after committing to this branch (I haven't tested it without git available yet):

```
$ ./show_version 
  DIRECTORY: /home/tim/pipelinedevel/HCPpipelines
    PRODUCT: HCP Pipeline Scripts
    VERSION: Post-v4.3.0-MOD-93f192c
     COMMIT: 93f192c3f9cf29f86ac72309bdde0fadc0083c9a
   MODIFIED: YES
$ ./show_version 
  DIRECTORY: /home/tim/pipelinedevel/HCPpipelines
    PRODUCT: HCP Pipeline Scripts
    VERSION: Post-v4.3.0-9b276ed
     COMMIT: 9b276ed82fda1356c1f185b0a35919ab431b00f2
   MODIFIED: no
```

show_version
```diff
@@ -1,40 +1,132 @@
...
+
+candidate=0
+release=0
+
+if [[ -f "$verdir"/release.txt ]]
+then
+    relstring=$(cat "$verdir"/release.txt)
+    if [[ "$relstring" != "$base" ]]
+    then
+        echo "ERROR: release.txt exists, but contents do not match base version" 1>&2
```
mharms, line 32: Can you give examples of the role of `release.txt` and `candidate.txt`?

mharms: i.e., Maybe there should be a section at the top of the script that explains what it is doing, and how it is designed to work?

show_version
```diff
@@ -1,40 +1,132 @@
...
+then
+    echo "ERROR: both release.txt and candidate.txt exist" 1>&2
+    exit 1
+fi
+
+if [[ "${HCPPIPE_HIDE_RC:-}" != "" ]]
+then
+    if [[ "$HCPPIPE_HIDE_RC" != "$base" ]]
+    then
+        echo "ERROR: HCPPIPE_HIDE_RC is set, but contents do not match base version" 1>&2
```
mharms, line 59: `HCPPIPE_HIDE_RC` sounds to me more like the name for a boolean (0/1; N/Y) variable, not a string that would be checked against the value of `base.txt`.

show_version
```diff
@@ -1,40 +1,132 @@
...
+        exit 1
+    fi
+    if ((candidate))
```
mharms, line 62: Having a hard time (without examples) following the use-case here.

show_version
```diff
@@ -1,40 +1,132 @@
...
+then
+    modified=no
+    if ! (cd "$HCPPIPEDIR"; git diff-index --quiet HEAD)
+    then
+        modified=YES
+        verstring="$verstring"-MOD
+    fi
+    commit=$(cd "$HCPPIPEDIR"; git rev-parse HEAD)
+    shortcommit=$(cd "$HCPPIPEDIR"; git rev-parse --short HEAD)
+    if ((! release))
```
mharms, line 97: Why not just allow the commit tag to be included (if a git repo) for "releases" as well?  Would be similar to what FS and FSL is doing for reporting their "version".

show_version
```diff
@@ -1,40 +1,132 @@
...
 
-if [ -e "${branch_file}" ] ; then
-	echo -n "     BRANCH: "
-	cat ${branch_file}
+commit=''
+#plot twist: mac's which has -s for silent, but debian's doesn't
+if [[ -d "$HCPPIPEDIR"/.git ]] && which git &> /dev/null
+then
+    modified=no
+    if ! (cd "$HCPPIPEDIR"; git diff-index --quiet HEAD)
```
mharms, line 90: Is this just detecting if someones local git repo has been locally modified?

mharms (commented):
I think I just some practical examples of use cases.

show_version
coalsont: This isn't really meant for end users, it is a hack to allow RCs to pretend to be the release.  If it doesn't check for being the same version, then using the hack for one version could inappropriately carry over to the next version - if someone just decided to set it once in their .bashrc and it was a simple boolean, the version info would lie to them/us forever without any effort on their part.  The variable name could be changed, but its purpose is to behave like a boolean, but need to be "set" very intentionally.

show_version
coalsont: 1. a "rc pretending to be release" qunex build would have a different version string than the official release
2. anything that isn't a git clone doesn't include a commit in the version anyway
3. we will already know what commit a qunex build used
4. the commit hash (when available) is in the extended information anyway, which is what we would ask users for

show_version
coalsont: Yes.

show_version
coalsont: This test means if you ask it to hide RC status, but it isn't an RC, it shouldn't pretend to be a release.

coalsont:
Explanation has been added to the top of the script.  I have also changed the modification check by git to ignore edits made in `Examples/`, so that instructions that have people copy or edit the examples in-place don't trigger false-positives.

show_version
mharms: `HCPPIPE_HIDE_RC` is fine if that's what you want to go with.

show_version
mharms: You meant, "shouldn't pretend to be a candidate", right?

coalsont: No, I meant pretend to be a release, there is no variable to make a non-RC version pretend to be an RC.  Here is the effect of properly setting the "hide RC" variable:
```
(format: actual->reported)
dev->dev (because of this check)
RC->release
release->release
```
With the `else` in the original version, both of the "do nothing" cases would produce errors instead.

mharms: If `base.txt` = `candidate.txt` = `HCPPIPE_HIDE_RC`, then isn't the result that the version will appear to be an official release?

mharms: Never mind, that's exactly what we want to occur in that situation.

mharms:
Looks good to me!

coalsont:
I have edited the places I saw that used version.txt, and replaced them, such that their behavior is mostly the same, except for the git additions to non-release version strings.   An exception is the diffusion pipelines, which already called show_version, so they are the only things that will currently display the extended info that now looks a bit different.  I have not added calls to show_version to any script that didn't previously report the version.

mharms:
Sounds good to me, and that it is done?

coalsont:
I believe it is done, unless we also want to make more of the pipelines report the version (thereby sanity checking the versioning files/variables more reliably).

glasserm:
Should we merge it?

PR #215, opened 2021-07-28
[TYPO] Remove colon from path of git diff-index

tashrifbillah:
This PR addresses a minor TYPO/ERROR that does not cause the rest of the pipeline to fail.

I believe the current command has a mistaken `:` 
```bash
[user@~ HCPpipelines]$ git diff-index --quiet HEAD -- ':!/Examples/'
fatal: Unimplemented pathspec magic '!' in ':!/Examples/'
```

I have removed the `:`. In addition, I have removed `--quiet` because I could not find this flag in `git diff-index` help message. Morever, I don't see why the preceding `/` is necessary. In total, I have removed `:/` and `--quiet`. 

Now everything is fine:

```
[user@~ HCPpipelines]$ git diff-index --quiet HEAD -- '!Examples/'
[user@~ HCPpipelines]$
```

(Any change in `Examples` directory has been suppressed. You can edit any file inside `Examples/` and test it.)

### DIFF
--- a/show_version
+++ b/show_version
@@ -104,7 +104,7 @@ if [[ -d "$HCPPIPEDIR"/.git ]] && which git &> /dev/null
 then
     modified=no
     #ignore modification of things in Examples?
-    if ! (cd "$HCPPIPEDIR"; git diff-index --quiet HEAD -- ':!/Examples/')
+    if ! (cd "$HCPPIPEDIR"; git diff-index HEAD -- '!Examples/')
     then
         modified=YES
         verstring="$verstring"-MOD


glasserm:
Tim,

Please merge if this is okay.

Thanks,

Matt.

coalsont:
This edit no longer does what is intended:

```
tim@timsdev:/mnt/myelin/tim/pipelinedevel/pr_eval/HCPpipelines$ nano -w README.md 
tim@timsdev:/mnt/myelin/tim/pipelinedevel/pr_eval/HCPpipelines$ ./show_version 
  DIRECTORY: /mnt/myelin/tim/pipelinedevel/pr_eval/HCPpipelines
    PRODUCT: HCP Pipeline Scripts
    VERSION: Post-v4.3.0-c782fe0
     COMMIT: c782fe0e1e5520c467ae10df1864a4f8917a7846
   MODIFIED: no
tim@timsdev:/mnt/myelin/tim/pipelinedevel/pr_eval/HCPpipelines$ git diff
diff --git a/README.md b/README.md
index 666aa34..6a2ecd0 100644
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # HCP Pipelines 
-
+foo
 The HCP Pipelines product is a set of tools (primarily, but not exclusively,
 shell scripts) for processing MRI images for the [Human Connectome Project][HCP
 Among other things, these tools implement the Minimal Preprocessing Pipeline 
```

What version of git are you using?  I probably tested this with a newer git, but these options are present in git diff-index 1.9.1 (which does complain about the current pathspec in a different way):

```
       --exit-code
           Make the program exit with codes similar to diff(1). That is, it
           exits with 1 if there were differences and 0 means no differences.

       --quiet
           Disable all output of the program. Implies --exit-code.
```

The ':!' invokes the exclusion pathspec magic, but we also want the "top" magic (shortcut is `:/`) so that we don't match "Examples" in a subfolder, and `'*' ':!/Examples/'` seems to work in git 1.9.1 (the '*' is to resolve its "nothing to exclude from" complaint).  However, it seems to have some false positives related to some kind of caching behavior (manually edit a file back to commit state, it still says modified, then after running `git diff`, it no longer says modified).

Let me know if you find a way to detect changes while excluding a folder in your version of git.

tashrifbillah:
> What version of git are you using?

1.8.3.1

tashrifbillah:
> '*' ':!/Examples/' seems to work in git 1.9.1 (the '*' is to resolve its "nothing to exclude from" complaint)

There is no `*` in the current command.

coalsont:
> There is no `*` in the current command.

Correct, 1.9.1 is older than what I originally tested it on, newer versions work without the `'*'`.  If we don't have a solution for 1.8.3.1 to detect modified files, it would be good to at least add a `'*'` to the original version.

tashrifbillah:
> Let me know if you find a way to detect changes while excluding a folder in your version of git.

So what is the expected output of the current command? Let's say I do:

```bash
[tb571@grx06 HCPpipelines]$ cd Examples/
[tb571@grx06 Examples]$ echo hi >> Scripts/FreeSurferPipelineBatch.sh
```
Then what should `git diff-index --quiet HEAD -- ':!/Examples/'` output?

coalsont:
It should exit with status code 1, and produce no output (whereas without modifications, it should exit with code 0).

It doesn't have to be done with status codes, that was just convenient.  Something like `if [[ $(cd "$HCPPIPEDIR"; git diff HEAD) != "" ]]` might be sufficient.

coalsont:
> `[user@~ HCPpipelines]$ cd Examples/`

Wups, missed this - because your example edit is in `Examples`, we want that to still be counted as "not modified" (exit code 0).  But, modification to a file outside Examples should cause it to say modified (exit code 1).

tashrifbillah:
Observation with git 2.32.0:

Modification inside `Examples/` and `DiffusionPreprocessing/` folders

```bash
[user@~ HCPpipelines]$ git diff-index --quiet HEAD -- ':!Examples/'
[user@~ HCPpipelines]$ echo $?
0

[user@~ HCPpipelines]$ if ! (git diff-index --quiet HEAD -- ':!Examples/'); then echo hi; fi
hi
[user@~ HCPpipelines]$
```
- Parenthetical exit code is 0, hence the if evaluated to true.

---

Update inside `Examples` folder

```bash
[user@~ HCPpipelines]$ git diff-index --quiet HEAD -- ':!Examples/'
[user@~ HCPpipelines]$ echo $?
1

[user@~ HCPpipelines]$ if ! (git diff-index --quiet HEAD -- ':!Examples/'); then echo hi; fi
[user@~ HCPpipelines]$
```
- Parenthetical exit code is 1, hence the if evaluated to false.

tashrifbillah:
I am out of ideas, I appreciate the discussion though. People reaching here in the future would know that they should use an advanced version of git.

coalsont:
A workaround is to move or delete the .git folder before running things so that it doesn't use git to try to check for changes.

Issue #217, opened 2021-12-21
Can't process dMRI acquisitions with an odd number of slices if using latest 'eddy' features


mharms:
`eddy` needs to know the slice timing when using advanced features such as outlier replacement (`--repol` with `--ol_type=both`) or slice-to-volume (s2v) motion correction.  This information is most conveniently provided using the `--json` option, which then reads the slice-timing info (to define the "MB-groups") from the json produced by DICOM conversion (using `dcm2niix`).

However, these features of `eddy` came after the `DiffusionPreprocessing` pipeline was originally written.  At that time, the recommendation for `topup` was to remove one slice if there was an odd number of slices (due to the assumptions built into the config file that `topup` needs). Consequently, `DiffusionPreprocessing` was written to automatically remove one slice from the dMRI data in the case of an odd number slices.  However, this leads to a crash in `eddy` if using the advanced features mentioned above.

This issue has evaded detection until now because (1) the HCP-LS dMRI acquisitions used MB4 with an even number of slices, and (2) the HCP-YA acquisitions, while using MB3 with an odd number of slices, were processed before these advanced features were available.

Notably, the FSL guidance on how to handle the situation of an odd number of slices has also changed, with the suggestion being to handle the situation of an odd number of slices by using a different config file for `topup`, rather than discarding a slice: https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=FSL;6c4c9591.2002

The proposed solution is to modify `DiffusionPreprocessing` to no longer automatically discard a slice in the case of an odd number of slices (although that could potentially still be provided as an option), and instead provide an option for the user to select the config file that they want to use, per the FSL link above.

glasserm:
This makes sense.  Also at some point we may change the default topup configs for the HCP Pipelines anyway, as we have found issues with the current ones and there is room for improvement.

MichielCottaar:
Makes sense to me. Removing the slice removal is very straightforward (see #218). So, the main difficulty is in deciding on:
1. what to do with the topup config file? Do we want to use a different one if we have an even number of slices (as Jesper recommends) or keep the code simpler but take the performance hit when users have images with an even number of slices.
[...]

PR #218, opened 2022-01-07
No more slice removal for odd number of slices

MichielCottaar:
This is meant to address #217 . This pull request is still in draft, because the topup will crash for any images with odd number of slices. We still need to adjust the topup config file used (either for all input volumes or only for those with odd number of slices).


glasserm (commented):
Thanks Michiel.  Is this tested?  It looks good from a code skim.

MichielCottaar:
I'm running some overnight tests on HCP YA data. I'll let you know tomorrow

MichielCottaar:
This is working, I've tested it in three specific cases. In all cases the input was the HCP YA adult data, which has an odd number of slices.
1. The default pipeline without any additional flags fails in that case, because the default topup configuration file is incompatible with an odd number of slices. I've copied the error message raised by topup below.
2. When adding the `--ensure-even-slices` flag, the pipeline removed a slice from both the input to topup and eddy. Topup ran succesfully afterward. I've not tested the rest of the pipeline, but see no reason why it would not.
3. When adding the `--topup-config-file=<file>` flag with a topup configuration without subsampling, topup also ran succesfully (without removing any slices).

The error message raised by topup if the subsampling levels are incompatible with the image dimensions
```
Part of FSL (ID: 6.0.6:e380285d)
topup

Usage: 
topup --imain=<some 4D image> --datain=<text file> --config=<text file with parameters> --out=my_topup_results


Compulsory arguments (You MUST set one or more of):
	--imain		name of 4D file with images
	--datain	name of text file with PE directions/times

Optional arguments (You may optionally specify one or more of):
	--out		base-name of output files (spline coefficients (Hz) and movement parameters)
	--fout		name of image file with field (Hz)
	--iout		name of 4D image file with unwarped images
	--logout	Name of log-file
	--warpres	(approximate) resolution (in mm) of warp basis for the different sub-sampling levels, default 10
	--subsamp	sub-sampling scheme, default 1
	--fwhm		FWHM (in mm) of gaussian smoothing kernel, default 8
	--config	Name of config file specifying command line arguments
	--miter		Max # of non-linear iterations, default 5
	--lambda	Weight of regularisation, default depending on --ssqlambda and --regmod switches. See user documetation.
	--ssqlambda	If set (=1), lambda is weighted by current ssq, default 1
	--regmod	Model for regularisation of warp-field [membrane_energy bending_energy], default bending_energy
	--estmov	Estimate movements if set, default 1 (true)
	--minmet	Minimisation method 0=Levenberg-Marquardt, 1=Scaled Conjugate Gradient, default 0 (LM)
	--splineorder	Order of spline, 2->Qadratic spline, 3->Cubic spline. Default=3
	--numprec	Precision for representing Hessian, double or float. Default double
	--interp	Image interpolation model, linear or spline. Default spline
	--scale		If set (=1), the images are individually scaled to a common mean, default 0 (false)
	--regrid		If set (=1), the calculations are done in a different grid, default 1 (true)
	-h,--help	display help info
	-v,--verbose	Print diagonostic information while running
	-h,--help	display help info



Topup: msg=topup_clp::topup_clp: Subsampling levels incompatible with image data.
Mon Jan 10 14:40:58 GMT 2022:run_topup.sh: While running '/home/fs0/ndcn0236/Work/datalad/HCP_test_slice_removal/Pipelines/DiffusionPreprocessing/scripts/run_topup.sh /home/fs0/ndcn0236/Work/datalad/HCP_test_slice_removal/base/103818/Diffusion/topup /home/fs0/ndcn0236/Work/datalad/HCP_test_slice_removal/Pipelines/global/config/b02b0.cnf':
Mon Jan 10 14:40:58 GMT 2022:run_topup.sh: ERROR: '/opt/fmrib/fsl/bin/topup' command failed with return code: 1
```

glasserm (changes requested):
This is nearly there.  I do think the topup error message is actually not as userfriendly as we would like.  It would be better to catch the error early in the pipeline (before much processing has taken place) and to suggest the user to 1) Use a different config file (recommended) or 2) Use the --ensure-even-slices flag (if trying to replicate legacy behavior or if not using advanced eddy features).  Does that seem reasonable to you?

MichielCottaar:
Sounds reasonable. I thought that I might need to parse the topup config file to do that. However, I just realised that we could simply raise an error if (1) there is an odd number of slices *and* (2) there is no `--topup-config-file` or `--ensure-even-slices` flag. I'll add an error in that case.

That would mean that if the user provides a topup config file that is incompatible with the data dimensions, they would still get the topup error. Is that okay with you?

glasserm:
Is parsing the file hard?  @coalsont might be able to help with the code for this.  Basically it needs to grep the line: 
--subsamp=2,2,2,1,1 from the planned topup config and then check whether any digits are not 1.

MichielCottaar:
It would take me quite a while to code that up in bash (I'm not very good in bash, grep, etc.). If someone can give me the code, I'll be happy to put it in.

glasserm:
@coalsont is better than me, but I think it will be a one-liner.  Something like testing whether `cat <config> | grep "--subsamp=" | sed 's/--subsamp=//g' | sed 's/,//g' | sed 's/1//g'` returns a non-blank string.

coalsont:
Honestly, if the user supplies their own topup config, I think they can deal with the errors they get from topup themselves, so I don't know that I would try to parse it to give an error in advance (maybe a future topup feature will allow subsampling without requirements on dimensions, which could make our error on someone else's config wrong and annoying).  I would recommend an option to ignore this check if we want our script to check custom config files by default.

If you want to try to protect users from their own custom config "mistakes" anyway, `sed 's/,//g'` is a bit brutal, and theoretically wouldn't catch "11", though hopefully nobody would use that.  Something like `| grep -Evq '^(1,)*1$'` on the end instead of the comma, 1 pair of `sed`s would give an exit status indication you could use in a conditional:

```bash
if grep -e "--subsamp=" <config> | sed 's/^.*--subsamp=//g' | grep -Evq '^[:space:]*(1,)*1[:space:]$' && [[ <slice count is odd> && "$EnsureEvenSlices" != "true" ]]
then
    log_Err_Abort "topup config uses subsampling, use --ensure-even-slices"
fi
```

If topup configs forbid whitespace, then some of the extra stuff in the regexps can be removed.

Technically, this assumes that the only kind of subsampling that will ever be in anyone's config is 2, since it assumes that an even number makes anything fine.  The current code also assumes that only the slice direction can be an issue.  Not sure how deep into this rabbit hole we really want to go, as long as we give the user the option of "just try it and see what topup thinks".  Really, the right solution is to ask FSL to improve topup's error messages.

mharms:
Given that a 4x subsampling config file is now an option that FSL provides, and that the proposed check is only on the slice dimension, maybe it is indeed best to not get into the checking business at all (given that what we are proposing is not a complete check for the ways that the config file can be incorrect relative to the image dimensions).

But I defer to the consensus of others.

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -134,6 +134,10 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
   [--b0maxbval=<b0-max-bval>]
                           Volumes with a bvalue smaller than this value will be
                           considered as b0s. Defaults to ${DEFAULT_B0_MAX_BVAL}
+  [--topup-config-file=<filename>]
+                          file containing the FSL topup configuration.
```
mharms, line 138:
```suggestion
                          File containing the FSL topup configuration.
```

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -134,6 +134,10 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
...
+                          Defaults to b02b0.cnf in the HCP configuration
+                          directory (as defined by ${HCPPIPEDIR_Config}).
   [--printcom=<print-command>]
```
mharms, line 141: Can we move the [--printcom] usage to the end of the usage statement?

glasserm:
The topup error doesn't really tell the user how to resolve the problem as far as the HCP Pipelines are concerned.  If we wanted to catch and parse the topup error, that would be another option...

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -145,6 +149,12 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
                           using equally spaced b0's throughout the scan. The best b0
                           is identified as the least distorted (i.e., most similar to
                           the average b0 after registration).
+  [--ensure-even-slices]
+                          If set will ensure the input images to FSL's topup and eddy
+                          have an even number of slices by removing one slice if necessary.
+                          This behaviour used to be the default, but is now optional,
+                          because it is incompatible with using slice-to-volume correction
```
mharms, line 156:
```suggestion
                          because discarding a slice is incompatible with using slice-to-volume correction
```

DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -145,6 +149,12 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
...
+                          in FSL eddy.
```
mharms, line 157:
```suggestion
                          in FSL's eddy.
```

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -92,6 +92,8 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
                             data_AP1@data_AP2@...data_APn
   --echospacing=<echo-spacing>
                           Effective Echo Spacing in msecs
+  --topup-config-file=<filename>
+                          file containing the FSL topup configuration.
```
mharms, line 96:
```suggestion
                          File containing the FSL topup configuration.
```

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -103,6 +105,12 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
                           using equally spaced b0's throughout the scan. The best b0
                           is identified as the least distorted (i.e., most similar to
                           the average b0 after registration).
+  [--ensure-even-slices]
+                          If set will ensure the input images to FSL's topup and eddy
+                          have an even number of slices by removing one slice if necessary.
+                          This behaviour used to be the default, but is now optional,
+                          because it is incompatible with using slice-to-volume correction
```
mharms, line 112:
```suggestion
                          because discarding a slice is incompatible with using slice-to-volume correction
```

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -103,6 +105,12 @@ PARAMETERs are: [ ] = optional; < > = user supplied value
...
+                          in FSL eddy.
```
mharms, line 113:
```suggestion
                          in FSL's eddy.
```

coalsont:
Detecting that topup crashed with a custom config and saying "if your config contains subsampling, make sure your image dimensions are an exact multiple" wouldn't require parsing anything.  When using the pipelines' own config files, we can provide more accurate errors in advance (because we know what they require), again without parsing anything.

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -138,8 +146,12 @@ EOF
 #                         encoding direction
 #  ${echospacing}         Echo spacing in msecs
 #  ${DWIName}             Name to give DWI output directories
+#  ${TopupConfig}         Filename with topup configuration
 #  ${b0maxbval}           Volumes with a bvalue smaller than this value will
 #                         be considered as b0s
+#  ${SelectBestB0}        If "true" will select the least distorted B0 for topup
+#  ${EnsureEvenSlices}    If "true" will ensure input images to topup/eddy has
```
mharms, line 153:
```suggestion
#  ${EnsureEvenSlices}    If "true" will ensure input images to topup/eddy have
```

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -444,6 +474,18 @@ main() {
 		exit 1
 	fi
 
+	# if the number of slices are odd, check that the user has a way to deal with that
+	if [ "${EnsureEvenSlices}" == "false" ] && [ "${TopupConfig}" == "${HCPPIPEDIR_Config}/b02b0.cnf" ] ; then
+		dimz=$(${FSLDIR}/bin/fslval ${topupdir}/Pos_b0 dim3)
+		if [ $(isodd $dimz) -eq 1 ]; then
+			log_Msg "Input images have an odd number of slices. This is incompatible with the default topup configuration file."
+			log_Msg "Either supply your own topup configuration file using the --topup-config-file=<file> flag (recommended)"
```
mharms, line 482:
```suggestion
			log_Msg "Either supply a topup configuration file that doesn't use subsampling (e.g., FSL's 'b02b0_1.cnf') using the --topup-config-file=<file> flag (recommended)"
```

DiffusionPreprocessing/DiffPreprocPipeline_PreEddy.sh
```diff
@@ -444,6 +474,18 @@ main() {
...
+			log_Msg "or instruct the HCP pipelines to remove a slice using the --ensure-even-slices flag (legacy option)."
+			log_Msg "Note that the legacy option is incompatible with slice-to-volume correction in FSL eddy"
```
mharms, line 484:
```suggestion
			log_Msg "Note that the legacy option is incompatible with slice-to-volume correction in FSL's eddy"
```

mharms (commented):
Looks good.  Thanks @MichielCottaar.  I've suggested some minor tweaks to some of the messages.

DiffusionPreprocessing/DiffPreprocPipeline.sh
MichielCottaar: I've made this change for all diffusion preprocessing scripts.

glasserm:
Ok

mharms:
So, that does mean we are basically done, pending any further requests/suggestions from either @coalsont or @glasserm ?

glasserm:
Well, I'd like things to be as userfriendly as possible, but don't want to hold things up if there isn't consensus on that.

glasserm:
Shall we merge this so that we can tag a release once the tICA testing is done?

mharms:
That's fine with me.

PR #224, opened 2022-04-07
Clarify error message if eddy and eddy_openmp are identical

MichielCottaar:
This error is raised when "run_eddy.sh" is looking for a GPU enabled eddy to run. If eddy and eddy_openmp are identical, this means that the search has failed. The new error message explains what the user might do about that.


glasserm:
Thanks Michiel!

PR #225, opened 2022-04-15
Remove old num_cores related code from FreeSurferPipeline.sh

mharms:
FreeSurferPipeline.sh does not need explicit code for setting the number of cores.

The deleted code in the PR actually results in undesired behavior on the WU SLURM system. (i.e., "-openmp 8" is getting set).

If someone still wants to set the "-openmp" flag in the call to `recon-all`, they can use the `--extra-reconall-arg` flag.  Additionally, the `OMP_NUM_THREADS` environment variable is available.

### DIFF
--- a/FreeSurfer/FreeSurferPipeline.sh
+++ b/FreeSurfer/FreeSurferPipeline.sh
@@ -574,7 +574,6 @@ main()
 	local extra_reconall_args
 	local conf2hires="TRUE"
 
-	local num_cores
 	local zero_threshold_T1wImage
 	local return_code
 	local recon_all_cmd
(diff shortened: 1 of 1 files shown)


glasserm:
I believe this is in place because of bad OpenMP behavior of some FreeSurfer binaries in the setting of hyperthreading.

mharms:
`recon-all` in FS defaults to setting `OMP_NUM_THREADS` to 1, which seems appropriate.

glasserm:
Yes that is fine as a default if not otherwise specified.

mharms:
The code I removed in the PR needs to go, as it is incorrectly resulting in `-openmp 8` in the `recon-all` call.

coalsont:
If someone is using 5.3.0-HCP, and this code is removed, will that bring back the bad behavior it was designed to fix?  Maybe the better solution is to leave it, and change the default to 1.  If NSLOTS is set by SLURM and isn't what we want to tell freesurfer to use, then we could have it check a different environment variable that isn't set by any known system.

coalsont:
Ah, right, 5.3.0-HCP requires running a different script anyway.  Removing it here seems fine.

PR #238, opened 2022-10-18
Move new 'MyelinTargetFile' argument to end of positional list

mharms:


glasserm (approved):

coalsont:
Looks good, can we test it, or should I just merge it?

coalsont:
Though, this does leave the default value set in 2 different places, which isn't ideal.

glasserm:
How is that?  I changed where the default was set.

coalsont:
These two lines:
https://github.com/Washington-University/HCPpipelines/blob/515c5636db9114c3222fa43303b004b2d4a95d1d/MSMAll/MSMAllPipeline.sh#L284
https://github.com/Washington-University/HCPpipelines/blob/515c5636db9114c3222fa43303b004b2d4a95d1d/MSMAll/MSMAllPipeline.sh#L367
Also, the second one incorrectly uses `${p_MSMAllTemplates}`, which is supposed to be "private" to the named option parsing code.

Ideally, the named parameter code would pass an empty string when it isn't provided, so that `main()` could be the only place that sets that default.

glasserm:
Not sure how best to address that.

mharms:
How about I just fix L367 to use `$MSMAllTemplates` rather than `$p_MSMAllTemplates` and we leave it at that?

mharms:
For example, `MatlabRunMode` similarly has defaults set in two places (once in the named parameter parsing in `get_options`; and once in `main()` (in case anyone actually invokes via positional arguments).  And its probably the same in other scripts that use this same style (that haven't yet been adapted), so changing it only here seems a bit ad hoc.

glasserm:
Probably should leave it alone then.

coalsont:
This default is longer and more complicated than the matlab default.  I pushed a change, but if we feel strongly it should remain redundant (and easy to forget to change both), it can be rolled back.

coalsont:
Also, the matlab default uses a variable, so that it could be changed in exactly one place in the script, and affect both bits of redundant code.

mharms:
Maybe we should just create a `G_DEFAULT_MYELINTEMPLATEFILE` variable then?  It's not a big deal, but the current version no longer reports the `MyelinTemplateFile` value used in the `get_options` parsing code (which is a different "asymmetry" that we've now introduced).

coalsont:
A global variable for the default can't work, the default uses another argument to set the directory.

I put an "exit 0" just after the argument printing code in main to check the logic and it appears to work:
```
$ MSMCONFIGDIR=/home/tim/pipelinedevel/HCPpipelines/MSMConfig HCPPIPEDIR=/home/tim/pipelinedevel/HCPpipelines MSMAll/MSMAllPipeline.sh --path=/mnt/myelin/brainmappers/Connectome_Project --subject=100307 --fmri-names-list=foo --output-fmri-name=bar --high-pass=0 --fmri-proc-string=baz --msm-all-templates=/home/tim/pipelinedevel/HCPpipelines/global/templates/MSMAll/ --output-registration-name=foo --high-res-mesh=164 --low-res-mesh=32 --input-registration-name=MSMSulc
  DIRECTORY: /home/tim/pipelinedevel/HCPpipelines
    PRODUCT: HCP Pipeline Scripts
    VERSION: Post-v4.3.0-MOD-eab500f
     COMMIT: eab500f3543ffbec39ace9677e24d0121024e3fa
   MODIFIED: YES
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: HCPPIPEDIR: /home/tim/pipelinedevel/HCPpipelines
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: MSMCONFIGDIR: /home/tim/pipelinedevel/HCPpipelines/MSMConfig
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Showing HCP Pipelines version
Post-v4.3.0-MOD-eab500f
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Using named parameters
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Study Folder: /mnt/myelin/brainmappers/Connectome_Project
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Subject: 100307
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: fMRI Names: foo
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Output fMRI Name: bar
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: ICA+FIX HighPass setting: 0
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: fMRI Proc string: baz
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: MSM All Templates: /home/tim/pipelinedevel/HCPpipelines/global/templates/MSMAll/
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Output Registration Name: foo
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: High Resolution Mesh: 164
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Low Resolution Mesh: 32
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Input Registration Name: MSMSulc
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Myelin Target File: 
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: MATLAB run mode: 1 - Use interpreted MATLAB
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Starting main functionality
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: StudyFolder: /mnt/myelin/brainmappers/Connectome_Project
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: Subject: 100307
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: fMRINames: foo
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: OutputfMRIName: bar
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: HighPass: 0
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: fMRIProcSTRING: baz
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: MSMAllTemplates: /home/tim/pipelinedevel/HCPpipelines/global/templates/MSMAll/
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: OutputRegName: foo
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: HighResMesh: 164
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: LowResMesh: 32
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: InputRegName: MSMSulc
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: MyelinTargetFile: /home/tim/pipelinedevel/HCPpipelines/global/templates/MSMAll//Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: MatlabRunMode: 1
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: mrfixNames: 
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: mrfixConcatName: 
Tue 18 Oct 2022 05:30:15 PM CDT:MSMAllPipeline.sh: mrfixNamesToUse: 
```
Note that the myelin target file is empty in the first print (in the parsing code), but properly set in `main()`.  This could be suppressed.

mharms:
Why can't we have a global default for the _file name_ and leave the setting of the directory as is? I was trying to have the myelin target file included with both prints, which again is more stylistically consistent.

coalsont:
Because if someone wants to use a different file, it probably won't be in the MSMAll templates directory, and detecting "is this path relative" is more complication (and potentially more confusion, if rooted in the template dir instead of the current dir) than expecting the value to be valid as-is.

coalsont:
As for style, I don't really care if the log file looks a bit strange.  It is already inconsistent with the `mrfix` and `fmriName` values, as it prints empty strings for them in `main()`, but suppresses them in the named arguments section if it detects they are not needed.  It is easy to suppress an empty value in the named option parsing, but I don't think we should add anything more complicated than that.  I feel this is better (either suppressed or not) than leaving two places in the code that give the same literal value, which would set up a careless editing trap.

coalsont:
Finally, if we really want this script to be clean, we should to convert it to `newopts` rather than putting effort into discussing and patching things.

glasserm:
I agree.

mharms:
This is fine, although FWIW, I don't follow your objection to dealing with the "editing trap" by creating a global variable defining a default value for `MyelinTemplateFile` within the script.  Doing so would merely have been a convenience for us inside the script.  It would have had no bearing on the situation of someone wanting to supply their own file, which the current script already requires to be in the `MSMAllTemplates` directory.

coalsont:
I was not talking about the global variable idea, I was talking about the redundant default code.  What part of the script would fail if the user-specified myelin template were outside the `MSMAllTemplates` folder?

mharms:
Ok, then I'm confused as to what your objection was to using 
`G_DEFAULT_MYELINTARGETFILE=Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii`
to deal with the problem of needing to edit something in two different places (i.e., just the default file name itself, not the full location; maybe that was the source of confusion).

That would have worked.  But I'm fine with leaving it "as is".

Looking at the code again, I was wrong about a user-specified template outside of `MSMAllTemplates` folder not working (precisely because in that case, the user has to specify the full path as part of specifying `--myelin-target-file`).

coalsont:
The user does not need to specify an absolute path, a relative path should work as long as there are no unreversed `cd` or `pushd` lines to get in the way (or the script could convert to absolute beforehand if necessary).  Because the option currently accepts files that are outside the `MSMAllTemplates` folder, I don't consider that global variable proposal workable, because it would change the semantics of the option, because a user providing the option with a bare filename with no path means "the file is in the current directory", not "inside the MSMAllTemplates folder".

coalsont:
Actually, I think I missed a sort of hybrid option: have redundant code that gives a default of `"${p_MSMAllTemplates}/${G_DEFAULT_MYELINTARGETFILE}"`, such that that line wouldn't need to be edited, and only the default logic has the special "in this folder" logic.  I don't think a cosmetic logging issue deserves this, and the more effective solution to logging aesthetics would be to remove all logging from the parsing logic (if the logging is there for checking for errors in the transition to positional arguments, then it shouldn't need to match the post-default-logic values, as making those match inherently requires handling the same defaults at two different times depending on whether positional parameters are used, which invites problems that are more serious than aesthetics).

mharms:
Yes, that's precisely what I was proposing all along, and what I was trying to convey when I wrote "Why can't we have a global default for the _file name_ and leave the setting of the directory as is"

PR #239, opened 2022-11-24
Batch queue rework

coalsont:
This updates the `QUEUE` variable in all applicable example batch scripts to contain only the queue name without the `-q` flag (adding logic later on to do this for you), and an empty `QUEUE` or specifying the `--runlocal` option now both cause capturing of stdout/stderr to log files (via fsl_sub without `-q`).

It also obsoletes `PRINTCOM` in most launch scripts, as it was inconsistent and incomplete (the exception is Diffusion, where it was close to complete, and I added it to the few lines that were missing it), and adds some quoting.

This branch was tested by Jenn, who also added some documentation updates.


Examples/Scripts/MSMAllPipelineBatch.sh
```diff
@@ -88,13 +86,14 @@ fMRINames=""
...
 # the original MR FIX concatenated name (only one group)
 mrfixConcatName="fMRI_CONCAT"
-# @-separate list of runs to use for this new MSMAll run of MR FIX
+# @-separated list of runs to use for this new MSMAll run of MR FIX
 mrfixNamesToUse="rfMRI_REST1_RL@rfMRI_REST1_LR@rfMRI_REST2_LR@rfMRI_REST2_RL"
 # FIX output concat name for this new MSMAll run of MR FIX
 OutfMRIName="rfMRI_REST_CONCAT"
 
-
+#Use HighPass = 2000 for single-run FIX data, HighPass = 0 for MR FIX data
```
coalsont, line 94: We don't recommend hp=2000 for any new data, so perhaps it would be clearer if this said something like "HighPass=2000 was used for HCP YA, HighPass=0 is recommended for any new processing"

Examples/Scripts/7T/IcaFixProcessingBatch.7T.sh
```diff
@@ -163,17 +164,17 @@ main() {
 
 				bandpass=2000
```
coalsont, line 165: Do we want to edit this 7T default?

Examples/Scripts/MSMAllPipelineBatch.sh
coalsont: The ICAFIX batch says something like "single-run fix used hp=2000 in HCP processing", but also says "cannot be 0 for single-run fix", which is a restriction I was not aware of, is it still true?

Examples/Scripts/7T/IcaFixProcessingBatch.7T.sh
```diff
@@ -129,32 +129,33 @@ main() {
...
 
 	# establish list of directions on which to run ICA+FIX
-	DirectionList=""
-	DirectionList="${DirectionList} PA"
-	DirectionList="${DirectionList} AP"
-	DirectionList="${DirectionList} PA"
-	DirectionList="${DirectionList} AP"
+	DirectionList=()
+	DirectionList+=(PA)
+	DirectionList+=(AP)
```
mharms, line 148: If `CondList` contains 4 runs, doesn't `DirectionList` need to contain 4 values as well?

Examples/Scripts/7T/IcaFixProcessingBatch.7T.sh
```diff
@@ -129,32 +129,33 @@ main() {
...
 	do
 		echo ${Subject}
 
-		for Condition in ${CondList}
+		for Condition in "${CondList[@]}"
 		do
 			echo "  ${Condition}"
 
-			for Direction in ${DirectionList}
+			for Direction in "${DirectionList[@]}"
```
mharms, line 158: This `for` loop inside another `for` loop doesn't seem right.

Examples/Scripts/MSMAllPipelineBatch.sh
mharms: I don't think that restriction has changed, unless `fix` itself has changed.

mharms (commented):
I commented on a couple things that caught my eye.

glasserm (approved):
I think we should probably make a record keeping folder and move the 3T size checking and 7T examples there.  I think they could be potentially confusing as it stands now.  We should go ahead and get a release out with these changes and then work on getting legacy mode for sICA+FIX and the transmit field correction pipeline into the next release.

Examples/Scripts/7T/IcaFixProcessingBatch.7T.sh
coalsont: With the for loop below, having 4 didn't make sense.  Since this is only used in the input filenames, then it should just be included in the run names array to simplify things.

mharms: Right.  A single `for` loop, with the task names and polarities combined seems like the way to go (and consistent with how other batch scripts are set up).

coalsont:
Is "historical" a good name for the record-keeping folder?

coalsont:
Please check the `Examples/Scripts/Historical/README.md` for accuracy.

PR #241, opened 2022-12-13
MR FIX legacy data

coalsont:
Expose number of wisharts and make a melodic-like mode for icaDim, update example scripts to match.

### DIFF
--- a/Examples/Scripts/IcaFixProcessingBatch.sh
+++ b/Examples/Scripts/IcaFixProcessingBatch.sh
@@ -168,14 +168,17 @@ main() {
 	
 	#MR FIX config support for non-HCP settings
 	config=""
-	#uncomment the below for legacy-style data
+	processingmode="HCPStyleData"
+	#uncomment the below two lines for legacy-style data
 	#config="$HCPPIPEDIR"/ICAFIX/config/legacy.conf
+	#processingmode="LegacyStyleData"
 	
 	#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
 	#DO NOT include "-q " at the beginning
 	#default to no queue, implying run local
 	QUEUE=""
 	#QUEUE="hcp_priority.q"
+	
 	if [[ "$RunLocal" == "TRUE" || "$QUEUE" == "" ]]; then
 		queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
 	else
@@ -230,7 +233,7 @@ main() {
 
 				echo "  InputFile: ${InputFile}"
 
-				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --config="$config")
+				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --config="$config" --processing-mode="$processingmode")
 				echo "About to run: ${cmd[*]}"
 				"${cmd[@]}"
 			done
--- a/Examples/Scripts/IcaFixProcessingBatch.sh
+++ b/Examples/Scripts/IcaFixProcessingBatch.sh
@@ -234,7 +234,7 @@ main() {
 
 				echo "  InputFile: ${InputFile}"
 
-				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --vol-wisharts=$volWisharts --cifti-wisharts=$ciftiWisharts --icadim-mode=$icadimMode)
+				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --vol-wisharts=${volWisharts} --cifti-wisharts=${ciftiWisharts} --icadim-mode="${icadimMode}")
 				echo "About to run: ${cmd[*]}"
 				"${cmd[@]}"
 			done
--- a/Examples/Scripts/IcaFixProcessingBatch.sh
+++ b/Examples/Scripts/IcaFixProcessingBatch.sh
@@ -166,6 +166,15 @@ main() {
 	#delete highpass files (note that delete intermediates=TRUE is not recommended for MR+FIX)
 	DeleteIntermediates=FALSE
 	
+	#defaults for icaDim
+	volWisharts=2
+	ciftiWisharts=3
+	icadimMode=default
+	#for data with far fewer timepoints than HCP, uncomment the three lines below to behave more like melodic
+	#volWisharts=1
+	#ciftiWisharts=1
+	#icadimMode=melodiclike
+	
 	#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
 	#DO NOT include "-q " at the beginning
 	#default to no queue, implying run local
@@ -184,8 +193,6 @@ main() {
 		
 		if [ -z "${ConcatNames}" ]; then
 			# single-run FIX
-			FixScript=${HCPPIPEDIR}/ICAFIX/hcp_fix
-			
 			fMRINamesFlat=$(echo ${fMRINames} | sed 's/[@%]/ /g')
 			
 			for fMRIName in ${fMRINamesFlat}; do
@@ -193,7 +200,7 @@ main() {
 
 				InputFile="${ResultsFolder}/${fMRIName}/${fMRIName}"
 
-				cmd=("${queuing_command[@]}" "${FixScript}" "${InputFile}" ${bandpass} ${domot} "${SRTrainingData}" ${FixThreshold} "${DeleteIntermediates}")
+				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix" "${InputFile}" ${bandpass} ${domot} "${SRTrainingData}" ${FixThreshold} "${DeleteIntermediates}")
 				echo "About to run: ${cmd[*]}"
 				"${cmd[@]}"
 			done
@@ -213,7 +220,6 @@ main() {
 				ConcatName="${concatarray[$i]}"
 				fMRINamesGroup="${fmriarray[$i]}"
 				# multi-run FIX
-				FixScript=${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run
 				ConcatFileName="${ResultsFolder}/${ConcatName}/${ConcatName}"
 
 				IFS=' @' read -a namesgrouparray <<< "${fMRINamesGroup}"
@@ -228,7 +234,7 @@ main() {
 
 				echo "  InputFile: ${InputFile}"
 
-				cmd=("${queuing_command[@]}" "${FixScript}" "${InputFile}" ${bandpass} "${ConcatFileName}" ${domot} "${MRTrainingData}" ${FixThreshold} "${DeleteIntermediates}")
+				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --vol-wisharts=$volWisharts --cifti-wisharts=$ciftiWisharts --icadim-mode=$icadimMode)
 				echo "About to run: ${cmd[*]}"
 				"${cmd[@]}"
 			done
--- a/Examples/Scripts/IcaFixProcessingBatch.sh
+++ b/Examples/Scripts/IcaFixProcessingBatch.sh
@@ -166,14 +166,10 @@ main() {
 	#delete highpass files (note that delete intermediates=TRUE is not recommended for MR+FIX)
 	DeleteIntermediates=FALSE
 	
-	#defaults for icaDim
-	volWisharts=2
-	ciftiWisharts=3
-	icadimMode=default
-	#for data with far fewer timepoints than HCP, uncomment the three lines below to behave more like melodic
-	#volWisharts=1
-	#ciftiWisharts=1
-	#icadimMode=melodiclike
+	#MR FIX config support for non-HCP settings
+	config=""
+	#uncomment the below for legacy-style data
+	#config="$HCPPIPEDIR"/ICAFIX/config/legacy.conf
 	
 	#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
 	#DO NOT include "-q " at the beginning
@@ -234,7 +230,7 @@ main() {
 
 				echo "  InputFile: ${InputFile}"
 
-				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --vol-wisharts=${volWisharts} --cifti-wisharts=${ciftiWisharts} --icadim-mode="${icadimMode}")
+				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --config="$config")
 				echo "About to run: ${cmd[*]}"
 				"${cmd[@]}"
 			done
--- a/ICAFIX/ReApplyFixMultiRunPipeline.sh
+++ b/ICAFIX/ReApplyFixMultiRunPipeline.sh
@@ -112,10 +112,10 @@ get_options()
 	p_LowResMesh="${G_DEFAULT_LOW_RES_MESH}"            # ${7}
 	p_MatlabRunMode="${G_DEFAULT_MATLAB_RUN_MODE}"      # ${8}
 	p_MotionRegression="${G_DEFAULT_MOTION_REGRESSION}" # ${9}
-    p_DeleteIntermediates="${G_DEFAULT_DELETE_INTERMEDIATES}"
-    p_volWisharts="${G_DEFAULT_VOLWISHARTS}"
-    p_ciftiWisharts="${G_DEFAULT_CIFTIWISHARTS}"
-    p_icadimMode="${G_DEFAULT_ICADIMMODE}"
+	p_DeleteIntermediates="${G_DEFAULT_DELETE_INTERMEDIATES}"
+	p_volWisharts="${G_DEFAULT_VOLWISHARTS}"
+	p_ciftiWisharts="${G_DEFAULT_CIFTIWISHARTS}"
+	p_icadimMode="${G_DEFAULT_ICADIMMODE}"
 	
 	# parse arguments
 	local num_args=${#arguments[@]}
@@ -418,17 +418,17 @@ have_hand_reclassification()
 
 function interpret_as_bool()
 {
-    case $(echo "$1" | tr '[:upper:]' '[:lower:]') in
-    (true | yes | 1)
-        echo 1
-        ;;
-    (false | no | none | 0)
-        echo 0
-        ;;
-    (*)
-        log_Err_Abort "error: '$1' is not valid for this argument, please use TRUE or FALSE"
-        ;;
-    esac
+	case $(echo "$1" | tr '[:upper:]' '[:lower:]') in
+	(true | yes | 1)
+		echo 1
+		;;
+	(false | no | none | 0)
+		echo 0
+		;;
+	(*)
+		log_Err_Abort "error: '$1' is not valid for this argument, please use TRUE or FALSE"
+		;;
+	esac
 }
 
 # ------------------------------------------------------------------------------
@@ -563,7 +563,7 @@ main()
 	# But, then generate the absolute path so we can reuse the code from hcp_fix_multi_run
 	ConcatName="${StudyFolder}/${Subject}/MNINonLinear/Results/${ConcatNameOnly}/${ConcatNameOnly}"
 
-    # If we have a hand classification and no regname, reapply fix to the volume as well
+	# If we have a hand classification and no regname, reapply fix to the volume as well
 	if have_hand_reclassification ${StudyFolder} ${Subject} ${ConcatNameOnly} ${hp}
 	then
 		fixlist="HandNoise.txt"
@@ -600,7 +600,7 @@ main()
 
 	local regenConcatHP=0
 	if [[ ! -f "${ConcatNameNoExt}_Atlas${RegString}_hp${hp}.dtseries.nii" || \
-			( $DoVol == "1" && `$FSLDIR/bin/imtest "${ConcatNameNoExt}_hp${hp}"` != 1 ) ]]
+		( $DoVol == "1" && `$FSLDIR/bin/imtest "${ConcatNameNoExt}_hp${hp}"` != 1 ) ]]
 	then
 		regenConcatHP=1
 	else  # Generate some messages that we are going to use already existing files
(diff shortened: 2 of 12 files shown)


mharms:
I'm a bit confused.  Is this functionality already in `v4.7.0`?

coalsont:
No, this is not in 4.7.0.  There was a "bugfix" for identifying problem runs and rerunning them with 1 volume wishart, but it did not include these additional options.

Examples/Scripts/IcaFixProcessingBatch.sh
```diff
@@ -228,7 +234,7 @@ main() {
 
 				echo "  InputFile: ${InputFile}"
 
-				cmd=("${queuing_command[@]}" "${FixScript}" "${InputFile}" ${bandpass} "${ConcatFileName}" ${domot} "${MRTrainingData}" ${FixThreshold} "${DeleteIntermediates}")
+				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --vol-wisharts=$volWisharts --cifti-wisharts=$ciftiWisharts --icadim-mode=$icadimMode)
```
mharms, line 237:
```suggestion
				cmd=("${queuing_command[@]}" "${HCPPIPEDIR}/ICAFIX/hcp_fix_multi_run" --fmri-names="${InputFile}" --high-pass=${bandpass} --concat-fmri-name="${ConcatFileName}" --motion-regression=${domot} --training-file="${MRTrainingData}" --fix-threshold=${FixThreshold} --delete-intermediates="${DeleteIntermediates}" --vol-wisharts=${volWisharts} --cifti-wisharts=${ciftiWisharts} --icadim-mode="${icadimMode}")
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -73,6 +73,13 @@ PARAMETERs are [ ] = optional; < > = user supplied value
      If TRUE, deletes both the concatenated high-pass filtered and non-filtered timeseries files
      that are prerequisites to FIX cleaning.
      (The concatenated, hpXX_clean timeseries files are preserved for use in downstream scripts).
+   [--vol-wisharts=<integer>] - Number of wisharts to fit to volume data in
+     icaDim, default ${G_DEFAULT_VOLWISHARTS}
+   [--cifti-wisharts=<integer>] - Number of wisharts to fit to cifti data in
+     icaDim, default ${G_DEFAULT_CIFTIWISHARTS}
+   [--icadim-mode=<"default" or "melodiclike">] - Choose how to run icaDim:
+     "default" - start with a VN dimensionality of 1 and rerun until convergence
```
mharms, line 81: Can we elaborate slightly on what "VN dimensionality" is?

mharms: What is "variance normalization dimensionality"?

mharms: i.e., "variance normalization" is division by an estimate of the std. dev.  In what sense does that have a "dimensionality"?

coalsont: I would guess it is the number of PCA components to assume are "signal" for the first iteration, treating the rest as the noise estimate.  I didn't bother trying to figure out how icaDim works.  @glasserm ?

glasserm: VN=1 means divide by the timeseries standard deviation (VN=0 likely means don't do variance normalization at all).  VN=#TimePoints/2 means use the last half of the eigen spectrum to estimate the unstructured noise.  Usually the VN is adjusted dynamically with the icaDim fitting (variance normalizing with the eigenvalues smaller than the fit dimensionality.  This will tend to converge, so VN=1 is fine to start with; however with legacy data it converges to too small a number and hence we stick with a single iteration of fitting and VN=#TimePoints/2.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -32,94 +34,125 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
+        end
+    end
 end
-if isempty(hpstring) && hp>=0
+
+%write out the hpstring logic explicitly, rather than tracing a bunch of code paths
+%must NOT be moved below the hp == 0 check below
+if pdflag
+    %explicit polynomial detrending can say whatever
     hpstring = ['_hp' pdstring num2str(hp)];
```
mharms, line 112: With the changes made above, should this be?
```suggestion
    hpstring = ['_hp' pdstring num2str(polydeg)];
```

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -32,94 +34,125 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
-if ~pdflag
-  pdstring = '';
+% Check whether polynomial detrend is being requested for the high-pass filtering.
+% Coincidentally deals with compiled matlab arguments
+if ischar(hp)
+    hp = lower(hp);
+    if strncmp(hp,pdstring,numel(pdstring))
+        pdflag = true;
+        polydeg = str2double(hp(numel(pdstring)+1:end));  % different purposes should use a different variable name
+        if (~isscalar(hp) || hp < 0 || hp ~= fix(hp))
```
mharms, line 97: Should this be using `polydeg` rather than `hp`?

coalsont: probably

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -175,8 +208,8 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
     ciftisave(BO,[fmri '_Atlas' regstring hpstring '.dtseries.nii'],WBC);
     delete('Atlas.nii.gz');
 
-elseif hp<0  % If no hp filtering, still need to at least demean the volumetric time series
-    if dovol > 0
+elseif hp < 0  % If no hp filtering, still need to at least demean the volumetric time series, assumes concatenated input
```
mharms, line 211: One could choose to run single-run with no hp filtering, so the comment added here of "assumes concatenated input" isn't appropriate.
```suggestion
elseif hp < 0  % If no hp filtering, still need to at least demean the volumetric time series
```

coalsont: If you set the current code to do no hp filtering, it currently assumes it is a concatenated run, so basically, no you can't tell it to do that (you might want to, but the code will act differently than you want).  If this specific conditional doesn't contain such assumptions, it could be removed from this comment.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -192,29 +225,32 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
-if hp>=0
-    if dovol > 0
-        Outcts=icaDim(cts,0,1,-1,volhpwishart); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, Volume fits two distributions to deal with MNI transform 
+if singlerun
+    if dovol
+        Outcts = icaDim(cts, 0, VN, iters, volwishart); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, Volume fits 2 distributions by default to deal with MNI transform
     end
     
-    OutBO=icaDim(BO.cdata,0,1,-1,3); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, CIFTI fits three distributions to deal with volume to CIFTI mapping
+    OutBO = icaDim(BO.cdata, 0, VN, iters, ciftiwishart); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, CIFTI fits 3 distributions by default to deal with volume to CIFTI mapping
```
mharms, line 242: Very ambiguous what variable the comment is referring to here.  Is it the second argument ("0"), or the 3rd argument ("VN")?

mharms: And looking at `icaDim.m` doesn't help, because `icaDim.m` has no usage/help information at all.

coalsont: The values in the comment are the arguments in order, excluding the input data.  Some still contain the previous hardcoded values rather than explicitly saying they are the defaults of what are now variables.

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: It would be really helpful if, as part of this PR, we added some usage/guidance/help info to the top of `icaDim.m`.  It's pretty much impossible to understand what exactly MR-FIX is doing without being able to understand the purpose/functionality of `icaDim.m`.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
mharms: hp<0 does not _have_ to mean concatenated data. The previous and existing code code supports use of hp<0 on single run data as well.  Given that, I find the introduction of the `singlerun` variable as a proxy for "hp > 0" to not be particularly clarifying.  How about the following instead:
`dohp = pdflag || (hp >= 0);`
and then remove the comments you added regarding single vs. concatenated runs?

ICAFIX/scripts/functionhighpassandvariancenormalize.m
mharms: That's very confusing -- no one is going to get that. I think users would be much better served by simply adding some usage/functionality info into `icaDim.m` itself.

mharms:
Have you done any testing already?  We should test with `hp = -1, 0, 200, 'pd2'` to cover the four main supported "modes" of hp specification.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: This comment block already existed:

https://github.com/Washington-University/HCPpipelines/blob/ba9369d068c7ea510b79f42cb5247e8e8549d5be/ICAFIX/scripts/functionhighpassandvariancenormalize.m#L130

In particular, if hp < 0, it completely skips loading the cifti file, so none of the code that works on cifti can be run.  The only obvious way to make it clear is to change the arguments so that the user can separately specify single run or concatenated, and whatever hp value.  At that point, because virtually no lines in this file are executed by both hp < 0 and hp >= 0 (aka concatenated and single run) cases, it might as well be split into two different functions.  We could take the time and testing to do that, and it would probably improve the code.

coalsont:
I have done some testing on HCP-style and legacy-style data, with the recommended settings (hp=0, and hp=-1 during the concatenated section).  I didn't test polynomial detrending.

mharms:
Can we test `hp = 2000, 'pd2'` as well?

ICAFIX/ReApplyFixMultiRunPipeline.sh
coalsont: I have taken a stab at describing icaDim.

coalsont:
I have tested hp 2000 and pd2, and didn't see any errors.

global/matlab/icaDim/icaDim.m
```diff
@@ -1,9 +1,19 @@
...
+function [Out] = icaDim(Origdata, DEMDT, VN, Iterate, NDist)
+%
+% ICADIM(Origdata, DEMDT, VN, Iterate, NDist)
+%
+% Estimates the number of ica components that are not due to random noise,
+%   and makes a standard deviation map of the random noise estimate
+%
+% DEMDT: 1 to detrend, 0 to only demean, -1 to do nothing
+% VN: initial variance normalization dimensionality
+% Iterate: mode/count for iterating the estimation
```
mharms, line 10: These make sense, except `Iterate` remains ambiguous. e.g., what are the different modes?  What does -1 represent?

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -32,94 +34,125 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
+dovol = true;
 regstring = '';
 pdflag = false;  % Polynomial detrend
 pdstring = 'pd';  % Expected string at start of HP variable to request a "polynomial detrend"
-hpstring = '';
-volhpwishart = 2;
+volwishart = 2;
+ciftiwishart = 3;
+VN = 1; %first iteration uses 1 for dimensionality
+iters = -1; %iterate until new convergence criteria
```
mharms, line 45: Where do these "new convergence criteria" come from?  Bit more in the usage/help for `icaDim.m` on this parameter would help.

global/matlab/icaDim/icaDim.m
coalsont: It is complicated, ad-hoc, and shouldn't really be combined into a single input argument, so it may not deserve description.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: Ad-hoc, as far as I know.  There may not be an existing description outside of the icaDim code that implements it.

global/matlab/icaDim/icaDim.m
mharms: Even if ad-hoc, it would be helpful to have _something_, short of having to decipher the code itself.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
mharms: Where do the "new convergence criteria" come from?  
"Iterate until convergence", while vague (doesn't explain how convergence is defined, which isn't appropriate for here anyway) would be a simpler and clearer comment, IMO.

global/matlab/icaDim/icaDim.m
coalsont: Looks like 0 will crash, anything negative triggers one block of code that messes with the loop variables to cause an exit based on averaging all the dimensionality estimates after iteration 3 (also unfortunately changing what the value of stabThresh means), 1-3 do the loop as written (looks like number of agreements on dimensionality, without needing to be consecutive), and > 3 messes with the loop variables to cause it to iterate exactly that many times.  This was already summarized in a comment about 5 lines below the comments I added.  Some of these modes may not get used anyway...

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: icaDim currently implements two modes of convergence testing.

mharms: So, I think "%iterate until convergence using new algorithm" better captures what is being implied here.

global/matlab/icaDim/icaDim.m
mharms: So, that comment ~5 lines later in the code, is almost sufficient for a high-level overview of what is going on, but needs to include:
1. A brief explanation of the difference between "legacy" vs "new" convergence mode.
2. Explanation of what "1 for VNDIM=1 results" means.  (Does it just mean 1 iteration and then done?)

The inclusion of that extra bit of info, and then moving the whole explanation up to the top, would make me happy :)

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: "new algorithm" implies the computation is different, "criteria" is a correct term for simply testing whether to stop, without changing the computation.

mharms: Why don't we just duplicate the nice language you added to `icaDim.m`:
"iterate until the average of the dimensionality history doesn't change much"

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: In the info for the VN parameter in `icaDim.m`, would the following addition be accurate:
```
0: No variance normalization
1: Use standard deviation of entire timeseries
> 1: Use that many time points from the tail of the eigen spectrum to estimate the standard deviation of the unstructured noise
```
Or, is > 1 specifying the opposite, namely the number of tp's at the front of eigen spectrum to _exclude_ from the eigen spectrum for estimating the noise?

coalsont: I believe it is the opposite:

https://github.com/Washington-University/HCPpipelines/blob/29a6e7b19d598a6a7a699010b7eaacaebfb36cb0/global/matlab/icaDim/icaDim.m#L94

Also, it updates each iteration based on the dimensionality estimate (unless the argument is zero, where it skips normalization regardless).

mharms: I'm not following the code.  How is the estimate of the unstructured noise derived from the signal components of the SVD?  I thought the unstructured noise is represented in the tail of the eigen spectrum?

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: I have clarified this comment to indicate that the cifti and motion confounds will not have been loaded when this branch is taken.

coalsont:
@mharms I have made edits to the comments which should resolve the discussions.  Please check them and say whether you are happy with the PR now.

ICAFIX/ReApplyFixMultiRunPipeline.sh
coalsont: That is why the signal components are excluded: it reconstructs just the noise components into what should (in theory) be a timeseries of images containing only noise, and then takes the standard deviation of that.

coalsont: To simplify as pseudocode (with some liberties):
```
[u s v] = svd(cov(data)); %strongest components (signal) first
noiseseries = u(numsignal:end) * s(numsignal:end) * v(numsignal:end);
noisestd = stdev(noiseseries);
```
So technically it is using the ICA dimensionality estimate on a PCA decomposition.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
mharms: FWIW, I still think the comment revisions/additions on this issue conflate two different things: Saying in the pre-PR version that `functionhighpassandvariancenormalize` expects that `hcp_multi_run_fix` will pass in concatenated data with `hp < 0` is not the same as assuming that "hp < 0 is equivalent to concatenated input".

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: But that doesn't seem to be what the code is doing.
We have this:
```
[u,EigS,v]=nets_svds(data',0); %Eigenvalues of data
Out.DOF=sum(diag(EigS)>(std(diag(EigS))*0.1)); %Compute degrees of freedom
```
and then this:
`noise_unst = (u(:,Out.VNDIM(c):Out.DOF)*EigS(Out.VNDIM(c):Out.DOF,Out.VNDIM(c):Out.DOF)*v(:,Out.VNDIM(c):Out.DOF)')';`

`Out.DOF` is the equivalent of `numsignal` in your pseudocode.  But that is getting used as the end of the range of SVD modes to include, not the beginning.
i.e., what it appears to be doing is:
`noiseseries = u(1:numsignal) * s(1:numsignal) * v(1:numsignal);`

coalsont: No, DOF is not the number of signal components, it is only there because there may be some missing degrees of independence from prior processing.  `c` is the current iteration, and Out.VNDIM is the history of what each iteration estimated for the number of signal components.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: functionhighpass... doesn't currently have a separate flag to convey concatenated/single input mode, and it does in fact currently assume that if hp < 0, it is being run on concatenated data.  It is not assuming they are conceptually equivalent, it is saying that the current logic of the code assumes that each implies the other.  I did not change the logic, that would be more invasive and need more testing, I just substituted a different variable so that we didn't have "hp < 0" on conditionals that actually only care whether it is running in concatenated mode.

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: Ok, I think two different comment would be helpful:
1. Something prior to the `Out.DOF` calculation explaining what that variable represents, and why we are only including eigenvalues > 10% of the eigenvalue std as eligible "degrees of freedom" (rather than the number of total time points).  The fact that that line is selecting just the highest subset of eigenvalues made me think that it was some algorithm for selecting the "signal" components.
2. Something (maybe right in the top usage/help info) that explains more fully what "variance normalization dimensionality" represents.  If I'm understanding correctly, it is basically synonymous with "estimated or assumed number of ICA signal components"?

coalsont: The VN argument is the (signal) dimensionality used during the variance normalization of the first iteration, which is exactly how I read the current explanation.  All other iterations use the estimate produced by the previous iteration.  Other than changing the VN mode, which is now also in the explanation, it does nothing else (the weird conditional that sets `stabCount` is dead code, I have now deleted it).

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: This version is not general, which is the whole point of these comments.  The comments are specifically pointing out that it can't currently be used to do single run without hp, because the code *makes assumptions* that it probably shouldn't.  That is why the comments I added use the word "assumes".  They are documenting what assumptions it *does* make, right now.  I am not defending them, I am pointing them out so they are less likely to stab someone in the face.

If you think there is a clearer way to convey this shortcoming, then suggest the phrasing you would use.  The comments should match what the code actually is, not what it could be in the future, so your criticism seems off track.

global/matlab/icaDim/icaDim.m
```diff
@@ -78,6 +78,9 @@
 %Precompute PCA reconstruction
 %octave doesn't have "rng", but it does accept "rand('twister', 0);" for matlab compatibility
 
+% previous regressions will remove degrees of freedom (DOF), putting zeros on the end of our eigenvalues
+% but, we use the end of the eigenvalues to estimate the wishart distribution, so we need to detect and exclude them
```
mharms, line 82:
```suggestion
% but, we use the end of the eigenvalues to estimate the wishart distribution, so we need to detect and exclude those zeroes
```

global/matlab/icaDim/icaDim.m
```diff
@@ -176,10 +176,6 @@
     %Maximum of laplacian is estimate of optimal dimensionality
     [~, Out.calcDim]=max(Out.pcaDim_lambdaAdj.lap(1:DOF));
     
-    if VN~=1
-      stabCount = 2;
-    end
-
```
mharms, line 161: I saw this code, but are you sure it can be deleted?  After all, `stabCount` is a variable used elsewhere, so I assumed that there was a reason for these 3 lines...

coalsont: It is unconditionally assigned to below the display block, with no intervening flow control or uses:

https://github.com/Washington-University/HCPpipelines/blob/fcef464fb50e8ec4787f5a11538ba71a263b29d5/global/matlab/icaDim/icaDim.m#L200

If it did have an effect on the loop control, it would already cause a premature exit under strange conditions.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
mharms: I'm not convinced that the function couldn't in fact be used "as is" to run a single run with no HP.  Why do you say that it can't?  Personally, I would use a variable called `doHP` rather than `singlerun` (which is just a convenience variable to avoid having to check for `pdflag || (hp >= 0)`, and remove the comments about assumed concatenated input.

global/matlab/icaDim/icaDim.m
coalsont: As for a reason, it was probably an attempt to force loop exit after a single iteration in the "no normalize" mode (which would actually be a reasonable feature, but I don't think it did that), without considering the VN > 1 or stabThresh = 3 cases (some of which may not have been implemented at the time), and then was forgotten and never cleaned up.

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms: Generic "dimensionality" in the help is ambiguous.  Why can't we just make the connection explicit that "variance normalization dimensionality" is referring to the assumed (first iteration) or estimated (other iterations) _number of ICA signal components_?

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: If you do give hp = -1, it will not currently open the cifti file (or motion confounds, which are a different story), and if you don't open that, you can't write out a cifti VN map, which is, as I understand it, part of what single-run mode's requirements are.  Thus, in order to actually tell it to do something different in single-run no-hp mode, compared to concat-run no-hp mode, an additional mode must be conveyed to the code somehow, and the code changed to make use of it.

coalsont: The entire reason I made the "singlerun" variable is because some conditionals contain code that is *not* related to the hp setting, but in fact is relevant to single-run versus concat-run modes.  Those conditionals get the "singlerun" test so that they make more sense.  Saying "if doHP, then do this single-run relevant code, otherwise do the concat-run code" would be a regression to the confusion that I specifically made the "singlerun" variable to clarify.

mharms: Ok.  The lack of a cifti VN map with hp < 0 may be relevant.  Maybe add a comment along those lines to the code, as a reminder if/when we come back to this in the future.

mharms: Maybe with a link as well referring back to your previous comment.

ICAFIX/ReApplyFixMultiRunPipeline.sh
coalsont: This guess of the dimensionality is only used for the variance normalization (and only in the first iteration), nothing else.  That is the only change between loop iterations, as far as I can tell.  The actual estimation code (`FitWishart`) doesn't take a dimensionality guess as input.

I would like to point out, that the intial PR treated icaDim as a black box, with no edits whatsoever (and I took a brief look at the icaDim code and decided I'd prefer not to try to understand it, if possible).  You seem to want to instead make this PR into a marathon to document the entirety of icaDim as pseudocode in the icaDim help info.  Please try to show some restraint.

mharms: What I want to do is try to give people that come to `icaDim.m` from `functionhighpass` some basic sense of its purpose and what it is doing.  The additions you have made are very helpful in that regard.  I feel we are basically there if we just make the connection for people that all this talk of dimensionality is referring to the number of _signal components_.  I don't understand the reluctance to add that final piece of the puzzle to the icaDim help info given how helpful it would be.

mharms: Do what you wish to wrap this up.  Thanks for all the additions.

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: A later conditional specifically doesn't even write out a volume VN map in concat-run mode.  The only unifying principle in these conditionals is just that, single-run versus concat-run mode, so your suggestion amounts to adding a unique comment to every conditional saying what the code inside it does, which seems excessive.  I consider the new "singlerun" variable to make those conditionals closer to self-documenting, unlike their previously misleading test of the highpass setting.

coalsont: To be clear, as far as I can tell, the only conditional in the entire function that should actually care about the high pass setting is the first one that actually does the high pass filtering, and the rest are solely introducing quirks that differentiate the concat-run and single-run mode computation/outputs.  None of them should have ever depended on the `hp` setting in the first place, instead they should have used a new input argument to change the mode.

coalsont:
@glasserm any comments?

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -32,94 +34,125 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
+    polydeg = 1;
+end
+
+% NOTE: To interpret the logic of the following "Compute, Save, Apply" sections need to know that this function expects that: 
+% (1) individual run time series will be passed in with hp requested
+% (2) concatenated (multi-run) time series will be passed in with hp<0
+%% [This is a bit of hack: would be cleaner if we disentangled the operation of this function from the specific needs of 'hcp_fix_multi_run'].
+
+%TSC: so, let's make a new variable that actually conveys some understanding of the logic
+%HACK: assume hp < 0 is equivalent to concatenated input
```
mharms, line 136:
```suggestion
%HACK: assume hp < 0 is equivalent to concatenated input
Note that concatenated input (hp < 0) never generates a VN version of the CIFTI.
```

ICAFIX/scripts/functionhighpassandvariancenormalize.m
mharms: I've suggested one additional line to go with your "HACK" comment, which I think is all that is necessary regarding my request. (Don't think a unique comment is necessary for every conditional).

ICAFIX/scripts/functionhighpassandvariancenormalize.m
coalsont: Concatenated mode also doesn't save a VN file for volume (or cifti, obviously).  On the other hand, the volume VN map does get computed and used internally in concatenated mode, just not saved to a file.  So, your comment, while technically accurate, could be misinterpreted to imply that concatenated mode saves a volume VN file (which it doesn't).  Who are you trying to save from being surprised by adding this comment?

mharms: It's for me, and anyone else trying to better (and more quickly) understand the practical difference between the two modes, without having to comb the code for 30 minutes.  I've edited my suggestion to be more precise.

mharms: Incidentally, I've realized (consistent with your earlier point) that mods to `functionhighpass` would be necessary if wanted to use it as part of processing just a single run through FIX (ala `hcp_fix` style).  A bunch of the stuff that it is doing in `singlerun` mode is only necessary in the context of prep'ing individual runs for use in MR-FIX.  Specifically, I think all we would need/want is the filtering related stuff.  The VN stuff would be unnecessary.

glasserm (commented):
I think the main thing missing here is the --legacy flag to trigger the legacy behavior and the --processing-mode=(HCPStyleData|LegacyStyleData)] stuff that the other scripts with legacy options have.  I don't think we want legacy mode users to have to figure out what things like volWisharts, ciftiWisharts, or icadimMode mean.  Such users are already unlikely to be super comfortable with the HCP Pipelines, so we need to keep things simple.  Those are fine as expert flags for someone like Takuya or me, but are not a good thing to expose to the typical user.

coalsont:
We wanted to be able to change the wisharts for other purposes, which is why I made them options.  Having more than one option that changes the wisharts would create problems with precedence, and hide additional sets of defaults in the code.  This is why I suggested a config file to contain the "legacy" settings (config handling is already built into the new options library), to simplify the three arguments into one, without creating new problems.  That would require converting ReApply to newopts, but it would also allow the possibility for hcp_fix_multi_run to pass more of its arguments to ReApply... via a config file saved into the subject dir while running, rather than needing to be copied manually into the ReApply batch file.

Regardless, "--legacy" is not a good option name for triggering this functionality, since we established there could be data other than legacy-style that would benefit from it, and it isn't clear from the name what aspect of the processing it changes.  A "--fix-mode=legacy" option might be a better way to name it, allowing easy extension to an NHP mode.  But, of course, using the config functionality would allow new settings without editing the code itself.  I don't see a reason not to use a config file for this, unless we specifically want the wishart settings and other details to not be exposed in the usage (and not allow users to use other combinations without editing the code).

I am not as clear whether this processing mode has drawbacks considerable enough to disable it without the "legacy-style" processing mode.  One end of the scale is slice timing interpolation, which interferes with motion correction.  It also raises the question of what settings for the wisharts should require the processing mode check.

mharms:
@coalsont nicely articulate the unease I was feeling about automatically triggering these new features as part of the other "legacy"-style code.

coalsont:
I think even with Matt's idea, there would be separate "--processing-mode=legacy" to enable, and "--fix-mode=melodic" or similar options to actually change the mode.  Config files to contain the groups of settings, rather than writing them into the code, just seems more maintainable and appropriate to me.

coalsont:
I have added the config handling (and changed ReApply MR to newopts, and rebased in case the edits to examples created a conflict).  The basic idea is summarized in the batch file changes of this commit:

https://github.com/Washington-University/HCPpipelines/pull/241/commits/c3f7d913516ac02f49b904d2c1ac43e332bf7ae7#diff-f5b4d31760d3145df7fb2810b30eb80ca92545e28bcf86c512286c45467a444f

glasserm:
What is next for this?  Does it need testing?

coalsont:
I have done some testing of it, I can test a couple extra highpass settings.  Does it need to have this set of options disabled without `--processing-mode=LegacyStyleData`?

glasserm:
Well it really is those specific settings that constitute legacy mode.  Pretty much anything else might be reasonable for non legacy use-cases.

coalsont:
So, volwisharts=1, icadimmode=melodiclike, but ciftiwisharts=2 should be considered reasonable "HCPStyleData" settings?

glasserm:
Maybe the icadimmode would be most specific for this situation.

coalsont:
To put it another way, is icadimmode=melodiclike a bad idea on HCP-style data?

glasserm:
Yes that basically turns off the fitting.

mharms:
If we required `--processing-mode=LegacyStyleData` does that mean you need to include that to use any user-supplied `--config` file option, or would the `--config` option be available regardless of `--processing-mode`?  I'm not a fan of tying the two together.

coalsont:
No, it would depend on the settings inside the config file.  `--config` is an automatic option whenever some options are enabled for config.

mharms:
Are you saying that whatever you specify in the `--config` file would go through without regard to `--processing-mode`?  Or that there would be separate stage of checks in which the actual values specified in the `--config` file would be checked to see if they were allowed for the given `--processing-mode`, and an error raised at that stage if they are not?

coalsont:
The `--config` file is just a way to specify multiple options that are already available.  The `--processing-mode` check will be done against whatever final settings result from the combination of config and direct options.

mharms:
Ok.  Thinking about it a bit more, it makes sense that `--config` based values shouldn't get any "special preference" in terms of parameter checks relative to directly specifying a value via a direct option.

What takes precedence if there is both direct specification of an option, and it exists in the `--config`, and the two are in conflict?  Is there an warning (or error) in that situation?

coalsont:
The direct option takes precedence, there is no warning.

coalsont:
For reference, the message that the processing mode check gives in legacy mode specifically talks about Glasser et. al. 2013:

https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/processingmodecheck.shlib#L26

coalsont:
And at present, the message comes out looking like this, with line-wrapping in a non-maximized terminal window:
```
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run: LegacyStyleData mode requested.
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run: NOTICE: You are using a mode that enables processing of acquisitions that do not
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         conform to the HCP specification as described in Glasser et al. (2013)!
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         Be aware that if the HCP requirements are not met, the level of data quality
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         can not be guaranteed and the Glasser et al. (2013) paper should not be used
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         in support of this workflow. A manuscript with comprehensive evaluation for
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         the LegacyStyleData processing mode is in active preparation and should be
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         appropriately cited when published.
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run:         
Wed Dec 21 17:36:33 CST 2022:hcp_fix_multi_run: The following LegacyStyleData settings were requested: --icadim-mode=melodiclike
```
It would probably be better to stick it all into a single `log_Msg` (or `log_Warn`) so there aren't redundant timestamps making the lines longer.

glasserm:
I think there are some specific messages for different options to.  For this one, the message is that you are turning off the iterative automatic estimation of the structured and unstructured portions of the signal and simply assuming that the back half of the eigen spectrum is unstructured.

coalsont:
Yes, I put an additional message into hcp_fix_multi_run itself for the option, to similar effect.

It may also be good to make the script pause for about 10 seconds after the generic legacy message (which comes last) so the user has a chance to see these messages before the processing code can spam anything to the terminal.

glasserm:
I wouldn't do that as usually these end up redirected into logfiles.

coalsont:
I could test whether stdout is a tty.  10 seconds isn't very long regardless, and it only applies to modes used on suboptimal data.

coalsont:
Let me know if my message is good enough:

https://github.com/Washington-University/HCPpipelines/pull/241/commits/a9ee57536d7aed6ef89d11a87f38bc96d5b7df4e#diff-42bcf9398cb5468c0d739867747e72454868f68dfb7eb73cc0ee5921657fe80aR100

glasserm:
That looks fine.

mharms:
@glasserm : The new mode assumes that the last half of the eigen spectrum is unstructured.  We are calling that mode `melodiclike`, but `melodic` has its own heuristic for determining the number of components for the spatial ICA.  So, in what sense is it "melodiclike"?

mharms:
Should there be an error or warning if someone is using `melodiclike`, but also tries setting `volwisharts` or `ciftiwisharts`?  That would imply confusion on their part about the relationships between the various options, right?

coalsont:
I don't think melodiclike prevents fitting multiple wisharts, though the "half spectrum" noise estimate may not be good enough for stable results.

mharms:
Isn't the point of the fitting to determine the dimensionality?  If so, then if `melodiclike` fixes the dimensionality, then what role do `volwisharts` and `ciftiwisharts` play?

coalsont:
I think it is true that melodic doesn't fit multiple wisharts.  That may be an unimportant detail, and the actual effect is described in the help info.

glasserm:
You wouldn't use multiple Wisharts with melodiclike, so that can be an error.  Melodic uses a fixed number of eigenvalues to do variance normalization from the rest, that is why this is melodiclike.

mharms:
I was perhaps conflating the estimate of the dimensionality for the VN (which apparently `melodic` does with a hard-coded assumption that the back-half of the eigen spectrum is unstructured?), from the estimate of the dimensionality to use for the ICA decomposition itself...
In the iterative approach, are those two different dimensionalities the same, or are they still different?

glasserm:
1) Melodiclike is an option for the variance normalization approach.  Melodic uses a fixed d=30 for variance normalization (i.e., do variance normalization on the eigenvalues to the right of 30; Line 580 of https://git.fmrib.ox.ac.uk/fsl/melodic/-/blob/master/meldata.cc).  I thought this was a bad idea, so that was part of why ica_dim converges to do variance normalization with the number of eigenvalues to the right of the estimated dimensionality.  This doesn't work correctly with legacy data; however, so I just use the right-hand half of the eigenvalues, which allows for some dynamic scaling according to number of timepoints (rather than just picking a fixed number like melodic did).  
2) Melodic estimates the spatial degrees of freedom a bit differently than ica_dim.  It looks at the 3D data smoothness to estimate the spatial degrees of freedom and creates a Wishart distribution based on that, rather than fitting to the back half of the eigen spectrum as ica_dim does to figure out the Wishart distribution.  The 3D smoothness approach will never work for CIFTI data, so we still need to use our ica_dim approach for such data; however, multiple Wishart distributions will also not work well for legacy data with few timepoints, so those should be set to 1.

global/scripts/processingmodecheck.shlib
```diff
@@ -22,33 +22,48 @@ check_mode_compliance()
...
-      log_Msg "        appropriately cited when published."
-      log_Msg "        "
+        log_Msg "LegacyStyleData mode requested.
+NOTICE: You are using a mode that enables processing of acquisitions that do not
+        conform to the HCP specification as described in Glasser et al. (2013)!
+        Be aware that if the HCP requirements are not met, the level of data
+        quality can not be guaranteed and the Glasser et al. (2013) paper should
+        not be used in support of this workflow. A manuscript with comprehensive
+        evaluation for the LegacyStyleData processing mode is in active
+        preparation and should be appropriately cited when published.
```
mharms, line 32: While we are tweaking the message, should we remove the last sentence about a manuscript in "active preparation", which doesn't seem forthcoming any time soon...

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -32,94 +34,126 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
-
 %% Defaults
-dovol = 1;
+dovol = true;
 regstring = '';
 pdflag = false;  % Polynomial detrend
 pdstring = 'pd';  % Expected string at start of HP variable to request a "polynomial detrend"
-hpstring = '';
-volhpwishart = 2;
+volwishart = 2;
```
mharms, line 42: Move comment below up to here, where it now more appropriately belongs
```suggestion
volwishart = 2;  %Volume fits 2 distributions by default to deal with MNI transform
```

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -192,29 +226,32 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
+    end
+end
 
 %% Compute VN map
-if hp>=0
-    if dovol > 0
-        Outcts=icaDim(cts,0,1,-1,volhpwishart); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, Volume fits two distributions to deal with MNI transform 
+if singlerun
+    if dovol
+        Outcts = icaDim(cts, 0, VN, iters, volwishart); %Volume fits 2 distributions by default to deal with MNI transform
```
mharms, line 240: Comment no longer necessary here, since moved up front
```suggestion
        Outcts = icaDim(cts, 0, VN, iters, volwishart); 
```

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -192,29 +226,32 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
     end
     
-    OutBO=icaDim(BO.cdata,0,1,-1,3); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, CIFTI fits three distributions to deal with volume to CIFTI mapping
+    OutBO = icaDim(BO.cdata, 0, VN, iters, ciftiwishart); %CIFTI fits 3 distributions by default to deal with volume to CIFTI mapping
```
mharms, line 243:
```suggestion
    OutBO = icaDim(BO.cdata, 0, VN, iters, ciftiwishart);
```

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -192,29 +226,32 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
 else
-    if dovol > 0
-        Outcts=icaDim(cts,0,1,-1,2); %0=Don't detrend, 1=Initialize variance normalization at 1, -1=Converge with running dim average, Volume fits two distributions to deal with MNI transform  
+    %concatenated input
+    if dovol
+        Outcts = icaDim(cts, 0, VN, iters, volwishart); %Volume fits 2 distributions by default to deal with MNI transform
```
mharms, line 247:
```suggestion
        Outcts = icaDim(cts, 0, VN, iters, volwishart);
```

ICAFIX/scripts/functionhighpassandvariancenormalize.m
```diff
@@ -32,94 +34,126 @@ function functionhighpassandvariancenormalize(TR,hp,fmri,WBC,varargin)
...
 %% Defaults
-dovol = 1;
+dovol = true;
 regstring = '';
 pdflag = false;  % Polynomial detrend
 pdstring = 'pd';  % Expected string at start of HP variable to request a "polynomial detrend"
-hpstring = '';
-volhpwishart = 2;
+volwishart = 2;
+ciftiwishart = 3;
```
mharms, line 43:
```suggestion
ciftiwishart = 3;  %CIFTI fits 3 distributions by default to deal with volume to CIFTI mapping
```

mharms:
Thanks @glasserm.  That's helpful additional context in your comment above.  One other thing might be worth commenting on here (or even including as a comment in `functionhighpass...` where the default number of Wisharts is set) is why/how did the need for multiple Wisharts come into existence in the first place?  (E.g., there isn't by any chance a figure of the fit for the differing number of Wisharts in the supplement of a paper, is there?)

glasserm:
I've asked Alex to put that in his paper.  I did an OHBM abstract on it in 2020.

coalsont:
Since the VN step is orthogonal to the wishart modeling step, I don't see a reason to make it an error to combine them in a way we aren't currently expecting.  A warning makes sense for combinations that we expect to be suboptimal, but haven't had a reason to try.  We already print a bunch of stuff just for enabling the option anyway.

At present, warnings are printed to stdout, with just the addition of the word "WARNING".  Do we want to stick with that, or have all warnings go to stderr?  We could even test if stderr is a tty, and if not, log to both stderr and stdout.

coalsont:
On the other hand, maybe the point is that we want to warn users that what we call "melodiclike" is intended for low timepoints, and could check the number of timepoints against some threshold (300?) and warn the user that default mode should work better.  This could also suggest that the mode should be named something like "fewtimepoints".

glasserm:
I don't know exactly how many timepoints is too few.   The HCP-YA Emotion task was 176 timepoints and lasted 2.1 minutes and works fine.  The legacy data had 90 timepoints and lasted 4.5 minutes and didn't work.  There may be an interaction between slow TR and few timepoints as well.

glasserm:
Warnings could go to both stderr and stdout if not a tty.

mharms:
I'm fine with either `melodiclike` or `fewtimepoints`.  Regarding putting WARNINGS into stderr, is that something we do anywhere else?  Personally, I think error messages should go to both stdout and stderr (perhaps they already do via our error handling functions?)  Under that model, if warnings go to stderr, they should be duplicated in stdout as well.  But there seems to be differing opinions online about whether warnings properly belong in stderr in the first place.

coalsont:
For unixlike tools where the stdout is likely to be parsed by some other command, warnings (and any other stuff besides the parseable output) should absolutely go to stderr.  For other tools, stderr is for things you want to make extra sure the user sees (if they redirect just stdout to a file, stderr will still show up in the terminal), so stderr is still a good place for warnings (as long as warnings aren't overused for minor things).

I didn't remember it, but yes our log functions already did put errors into both stderr and stdout, if stderr wasn't a tty, but left warnings on stdout.  Checking whether stdout and stderr go to the same place might be better, but I don't know if there is an easy way to do that.  To be clear, I am suggesting changing the shlib so all warnings in all pipelines would get printed to the same place(s) as errors.

mharms:
I don't have a strong opinion.  It would probably elevate the likelihood that warnings would be seen, but would also mean that a non-empty stderr file is no longer a sign of a error per se (i.e., the stderr file may be non-empty, even though the processing completed).
This is all further complicated by the additional layer that QuNex adds.  QuNex generates just a single log of the pipeline output, so I assume that it is redirecting stdout and stderr into the same file.

coalsont:
If we output a warning from the pipelines, it is generally serious enough that it should be looked at as if it could be an error, right?  Hopefully no tools are specifically using log file sizes as a check.

Yes, that would suggest that at present, qunex logs that have a pipelines-produced error have the message printed twice in the log.  I don't see this as a significant problem, given the technical challenge of avoiding it (versus the more important feature of showing it in both .o and .e logs), and would be fine with having warnings do the same.

glasserm:
I agree.

mharms:
Will this have any impact on where the various "Warnings" from `wb_command` end up?  I assume those will still be stdout only?

coalsont:
No, it does not affect how other commands choose to report warnings.  If they choose to use stderr, it will be on stderr.

PR #242, opened 2023-02-15
replace old argument parse in MSMAllPipeline with Tim's newopt

cyang31:
### Background
It was not flexible to add/remove arguments in the original location-based argument parsing. Tim's newopt library is a lifesaver. In addition to the argument part, the original script used `MSMAllRegs` to store the arguments for the main script MSMAll.sh, which is not necessary. 

### Main changes
- replace the location-based argument parsing with newopt;
- remove the logic related to `MSMAllRegs`;
- add optional arguments to set default file locations and default parameters such as `ICAdim`, `IterationModes`, `method`, etc.;

### Evaluation
One subject with resting state fMRI is evaluated on the original version of MSMAllPipeline and this latest version. The results are identical.

### DIFF
--- a/MSMAll/MSMAllPipeline.sh
+++ b/MSMAll/MSMAllPipeline.sh
@@ -58,7 +58,7 @@ opts_AddMandatory '--multirun-fix-names-to-use' 'mrfixNamesToUse' 'rfMRI_REST1_L
 opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to give to concatenated single subject scan"
 opts_AddMandatory '--high-pass' 'HighPass' 'integer' 'the high pass value that was used when running FIX' '--melodic-high-pass'
 opts_AddMandatory '--fmri-proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing already done, e.g. '_Atlas_hp0_clean'"
-opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'string' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
+opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'path' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
 opts_AddMandatory '--input-registration-name' 'InputRegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMSulc'"
 opts_AddMandatory '--output-registration-name' 'OutputRegName' 'MSMAll' "the registration string corresponding to the output files, e.g. 'MSMAll_InitalReg'"
 opts_AddMandatory '--high-res-mesh' 'HighResMesh' 'meshnum' "high resolution mesh node count (in thousands), like '164' for 164k_fs_LR"
@@ -82,11 +82,11 @@ opts_AddOptional '--vn' 'VN' 'YES/NO' "whether to perform variance normalization
 opts_AddOptional '--rerun-if-exists' 'ReRunIfExists' 'YES/NO' "whether to re-run even if output already exists, defaults to 'YES'" 'YES'
 opts_AddOptional '--registration-configure-path' 'RegConfPath' 'string' "it can be either the relative path where the registration configuration exists in 'MSMCONFIGDIR', or an absolute path" 'MSMAllStrainFinalconf1to1_1to3'
 opts_AddOptional '--registration-configure-override-variables' 'RegConfOverrideVars' 'string' "the registration configure variables to override instead of using the configuration file. Please use quotes, and space between parameters is not recommended. e.g. 'REGNUMBER=1,REGPOWER=3', defaults to 'NONE'" 'NONE'
-opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
-opts_AddOptional '--rsn-template-file' 'RSNTemplates' 'string' "alternate rsn template file, relative path to 'MSMAllTemplates'" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/melodic_oIC.dscalar.nii'
-opts_AddOptional '--rsn-weights-file' 'RSNWeights' 'string' "alternate rsn weights file, relative path to 'MSMAllTemplates'" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/Weights.txt'
-opts_AddOptional '--topography-roi-file' 'TopographyROIs' 'string' "alternate topography roi file, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.atlas_Topographic_ROIs.32k_fs_LR.dscalar.nii'
-opts_AddOptional '--topography-target-file' 'TopographyTarget' 'string' "alternate topography target, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.atlas_Topography.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--rsn-template-file' 'RSNTemplates' 'string' "alternate rsn template file, relative to the --msm-all-templates folder" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/melodic_oIC.dscalar.nii'
+opts_AddOptional '--rsn-weights-file' 'RSNWeights' 'string' "alternate rsn weights file, relative to the --msm-all-templates folder" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/Weights.txt'
+opts_AddOptional '--topography-roi-file' 'TopographyROIs' 'string' "alternate topography roi file, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.atlas_Topographic_ROIs.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--topography-target-file' 'TopographyTarget' 'string' "alternate topography target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.atlas_Topography.32k_fs_LR.dscalar.nii'
 opts_AddOptional '--matlab-run-mode' 'MatlabRunMode' '0, 1, or 2' "defaults to $g_matlab_default_mode
 0 = compiled MATLAB
 1 = interpreted MATLAB
--- a/MSMAll/MSMAllPipeline.sh
+++ b/MSMAll/MSMAllPipeline.sh
@@ -49,7 +49,7 @@ opts_SetScriptDescription "implements MSM-All Registration Pipeline"
 
 #mandatory
 #general inputs
-opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects"
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects" '--path'
 opts_AddMandatory '--subject' 'Subject' '100206' "one subject ID"
 opts_AddMandatory '--fmri-names-list' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of single-run fmri run names separated by @s"
 opts_AddMandatory '--multirun-fix-names' 'mrfixNames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of multi-run fmri run names separated by @s"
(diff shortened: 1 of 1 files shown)


MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
+g_matlab_default_mode=1
+
+#description to use in usage - syntax of parameters is now explained automatically
+opts_SetScriptDescription "implements MSM-All Registration Pipeline"
+
+#mandatory
+#general inputs
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects"
```
cyang31, line 52: I noticed `--path` is used in the example batch script, however, I only keep `--study-folder` in the updated version. Do we want to include both options in the new version? I am not aware of the proper usage in newopt to achieve this.

coalsont: https://github.com/Washington-University/HCPpipelines/blob/b8db803befadf5f80d1ec9004fe64b3d68bf7d25/ICAFIX/ReApplyFixMultiRunPipeline.sh#L50

Legacy flags that mean the same thing are specified on the end (warning, for an optional parameter, you MUST provide a default value before you can add legacy flags, but the empty string can be used).

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+opts_AddMandatory '--output-registration-name' 'OutputRegName' 'MSMAll' "the registration string corresponding to the output files, e.g. 'MSMAll_InitalReg"
+opts_AddMandatory '--high-res-mesh' 'HighResMesh' 'meshnum' "high resolution mesh node count (in thousands), like '164' for 164k_fs_LR"
+opts_AddMandatory '--low-res-mesh' 'LowResMesh' 'meshnum' "low resolution mesh node count (in thousands), like '32' for 32k_fs_LR"
+#MSMAll inputs
+opts_AddOptional '--module-name' 'ModuleName' 'string' "name of script or code used to run registration, defaults to 'MSMAll.sh'" 'MSMAll.sh'
+opts_AddOptional '--iteration-modes' 'IterationModes' 'string' "Specifieds what modalities:
+                            C=RSN Connectivity
+                            A=Myelin Architecture
+                            T=RSN Topography
+                          and number is the number of elements delimited by _
```
coalsont, line 72: `newopts.shlib` includes word wrapping with extra indentation for lines other than the first, please do not manually wrap sentences like this.  This is also more indentation than advisable, as it will cause additional wrapping in a default size terminal (or when redirecting to a file, where it assumes a default 80 columns for wrapping).

Try 4 spaces for the modes and 0 spaces for the rest of the text and see how the help info looks.

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+opts_AddOptional '--method' 'Method' 'string' "Possible values: DR, DRZ, DRN, WR, WRZ, WRN, defaults to 'WRN'" 'WRN'
+opts_AddOptional '--use-migp' 'UseMIGP' 'YES/NO' "whether to use MIGP (MELODIC's Incremental Group Principal Component Analysis), defaults to 'NO'" 'NO'
+opts_AddOptional '--ica-dim' 'ICAdim' 'integer' "ICA (Independent Component Analysis) dimension, defaults to '40'" '40'
+opts_AddOptional '--low-sica-dims' 'LowsICADims' 'num_num_num...' "the low sICA dimensionalities to use for determining weighting for individual projection, defaults to '7_8_9_10_11_12_13_14_15_16_17_18_19_20_21'" '7_8_9_10_11_12_13_14_15_16_17_18_19_20_21'
+opts_AddOptional '--vn' 'VN' 'YES/NO' "whether to perform variance normalization, defaults to 'NO'" 'NO'
+opts_AddOptional '--rerun-if-exists' 'ReRunIfExists' 'YES/NO' "whether to re-run even if output already exists, defaults to 'YES'" 'YES'
+opts_AddOptional '--registration-configure-path' 'RegConfPath' 'string' "the relative path where the registration configuration exists in 'MSMCONFIGDIR', defaults to 'MSMAllStrainFinalconf1to1_1to3'" "MSMAllStrainFinalconf1to1_1to3"
```
coalsont, line 82: This is a bit messy, and doesn't allow for a config file outside that directory that is set in a different file (or at least, not in a friendly manner).  This also means if `MSMCONFIGDIR` is set outside the HCP pipelines dir, the default config may not be found (so, we may want the default to actually ignore `MSMCONFIGDIR`).  I think my approach would be to have the argument take a filename, and if it doesn't start with a '/', assume it is relative to the config directory.

Also, simply because the default config name is messy, I'm not sure if I would mention what the default is, or even put the default filename here (as the default value will be reported by `opts_ShowValues`), but instead check for empty string later in the script (specifically, right before it checks for starting with '/').

These are both topics for discussion (@glasserm), and I am not clear why `MSMCONFIGDIR` is exposed as user-editable in the first place, as I would consider it more obvious and convenient for the user for a custom config to be specified as an absolute path in an argument.

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+opts_AddOptional '--registration-configure-override-variables' 'RegConfOverrideVars' 'string' "the registration configure variables to override instead of using the configuration file, e.g. 'REGNUMBER=1, REGPOWER=3', defaults to 'NONE'" 'NONE'
```
coalsont, line 83: We may want to tell the user to use quotes here, if the space after the comma is necessary or recommended.

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative path to 'MSMAllTemplates', defaults to 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
```
coalsont, line 84: More defaults that have messy file names (that include the specific dimensionality of the default template), discuss.

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+                            A=Myelin Architecture
+                            T=RSN Topography
+                          and number is the number of elements delimited by _
+                          So CA_CAT means one iteration using RSN Connectivity and Myelin
+                          Architecture, followed by another iteration using RSN Connectivity,
+                          Myelin Architecture, and RSN Topography. Defaults to 'CA_CAT'" 'CA_CAT'
+opts_AddOptional '--method' 'Method' 'string' "Possible values: DR, DRZ, DRN, WR, WRZ, WRN, defaults to 'WRN'" 'WRN'
+opts_AddOptional '--use-migp' 'UseMIGP' 'YES/NO' "whether to use MIGP (MELODIC's Incremental Group Principal Component Analysis), defaults to 'NO'" 'NO'
+opts_AddOptional '--ica-dim' 'ICAdim' 'integer' "ICA (Independent Component Analysis) dimension, defaults to '40'" '40'
+opts_AddOptional '--low-sica-dims' 'LowsICADims' 'num_num_num...' "the low sICA dimensionalities to use for determining weighting for individual projection, defaults to '7_8_9_10_11_12_13_14_15_16_17_18_19_20_21'" '7_8_9_10_11_12_13_14_15_16_17_18_19_20_21'
```
coalsont, line 79: I would use `,` or `@` as the separator - `@` is common in other pipelines, while `,` is more natural-looking than `_`.

cyang31: I agree. I kept this because the current `_` is used in the `MSMAll.sh` as the separator for inputs. I haven't planned to change the main `MSMAll.sh` script but maybe this one should also be updated. @glasserm Matt, am I right that we should also update the `MSMAll.sh`?

MSMAll/MSMAllPipeline.sh
cyang31: This `MyelinTarget` was fixed as the current default file name in the old script. We can either hard code it as an intermediate parameter or use optional parsing. I think the file name is the only information that matters for later processing instead of the dimensionality of the template. @glasserm Any comments?

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,232 @@
...
+log_Msg "output_proc_string: ${output_proc_string}"
+
+expected_concatenated_output_dir="${StudyFolder}/${Subject}/MNINonLinear/Results/${OutputfMRIName}"
+expected_concatenated_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}${output_proc_string}.dtseries.nii"
+before_vn_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}_novn.dtseries.nii"
+
+if [[ "$fMRINames" == "" ]]
+then
+	log_Msg "Running MSM on Multi-run FIX timeseries"
+    runSplits=()
```
coalsont, line 141: Indentation appears to be spaces on these two lines.

MSMAll/MSMAllPipeline.sh
coalsont: I think the original reason for using _ was that the internal loop in `MSMAllPipeline.sh` was made to be able to run multiple registrations in sequence, and it did that by a nested set of delimiters (looks like `_`, `@`, and `+`).  That loop is now gone in this version (@glasserm, please take note).  If we don't need the loop functionality, then it would probably be good to simplify/standardize things like this (and perhaps other no longer needed abstractions, like `ModuleName`).

MSMAll/MSMAllPipeline.sh
glasserm: I wouldn't want to change that we set the default to the existing MSMAll configuration, but I don't at this point recall the reason for this being "messy" and am happy to change it to specifying an absolute path (or a relative path if both can be achieved and properly documented).

MSMAll/MSMAllPipeline.sh
glasserm: I don't understand this concern.  We want to be able to specify the existing files as the default, but allow others to be used by the user if needed.  As for dimensionalities, it was useful to test different ones at one time and could be again.

MSMAll/MSMAllPipeline.sh
glasserm: I would rather not change this at this time.

glasserm (changes requested):
I responded to the questions directed at me and I also noted that I thought Alex should include the rest of his changes either in this pull request or another.

MSMAll/MSMAllPipeline.sh
coalsont: The option doesn't necessarily need to mention what the default filename is.  Because the default filenames are a bit messy, not having them in the help info may increase the overall readability.

MSMAll/MSMAllPipeline.sh
coalsont: @glasserm by "not change this", do you mean the `_` separator, or the previous loop functionality (to run multiple registrations)?

It wouldn't be very hard to support more than one allowed separator character (if we still wanted the loop, separate arrays for each setting is another way to do that).

MSMAll/MSMAllPipeline.sh
cyang31: I see. I will leave the default file name empty and then, in the main script, replace the empty file name with the current default names if there's nothing specified for those arguments.

cyang31:
> I responded to the questions directed at me and I also noted that I thought Alex should include the rest of his changes either in this pull request or another.

I have resolved most of the issues. I would prefer to include the rest of the changes in another pull request.

MSMAll/MSMAllPipeline.sh
```diff
@@ -112,6 +113,37 @@ opts_ShowValues
...
+	TopographyROIs="Q1-Q6_RelatedParcellation210.atlas_Topographic_ROIs.32k_fs_LR.dscalar.nii"
+fi
+
+if [[ "$TopographyTarget" == "" ]]
+then
+	TopographyTarget="Q1-Q6_RelatedParcellation210.atlas_Topography.32k_fs_LR.dscalar.nii"
+fi
+
 # MSMAll config file
 RegConfPath="${MSMCONFIGDIR}/${RegConfPath}"
```
coalsont, line 148: Could put an `if [[ "$RegConfPath" != /* ]]` around this, to support absolute paths.

glasserm:
I don't understand why the defaults are not a part of the argument line now?

cyang31:
> 

I would say that the current version highlights the default file names in the main script while making the help info less messy.

glasserm:
Well in the interest of keeping things shorter and easier to maintain I think it might be better to put the defaults in the one liner.  What does Tim think?

cyang31:
k;. lvu

> Well in the interest of keeping things shorter and easier to maintain I think it might be better to put the defaults in the one liner. What does Tim think?

@coalsont

coalsont:
Having it a default in the `opts_AddOptional` line puts the default filename into the log during `opts_ShowValues`.  I had suggested taking it out of that line because of that.  However, having it in a log file isn't as much of an eyesore as having it in the help info, so we could put it in the `opts_AddOptional` default, and just not mention what the default is in the help string.

I wouldn't say that having messy-looking default filenames on the end of a long line necessarily makes the logic more obvious or maintainable, though.  Default values are more obvious-looking for simple options that take a number or a boolean.  Yes, it saves some lines compared to an `if` statement, but so does a construct like `test -f "$foo" && cp "$foo" "$bar"`, which in my opinion is nearly unreadable compared to `if [[ -f "$foo" ]]; then cp "$foo" "$bar"; fi`.

I don't have a strong preference in this particular case.

cyang31:
> I would prefer that this all be in a single line, but don't care if it shows up in the help info. Is that agreeable to everyone?

I don't have a strong opinion on this, so I am okay with it. I will soon update with a new commit.

cyang31:
I have updated this pull request with the changes for the default file names in opt.

glasserm (approved):
I am okay with this if Tim is.  Tim can do the merge.

MSMAll/MSMAllPipeline.sh
```diff
@@ -31,693 +32,237 @@
...
+opts_AddOptional '--vn' 'VN' 'YES/NO' "whether to perform variance normalization, defaults to 'NO'" 'NO'
+opts_AddOptional '--rerun-if-exists' 'ReRunIfExists' 'YES/NO' "whether to re-run even if output already exists, defaults to 'YES'" 'YES'
+opts_AddOptional '--registration-configure-path' 'RegConfPath' 'string' "it can be either the relative path where the registration configuration exists in 'MSMCONFIGDIR', or an absolute path" 'MSMAllStrainFinalconf1to1_1to3'
+opts_AddOptional '--registration-configure-override-variables' 'RegConfOverrideVars' 'string' "the registration configure variables to override instead of using the configuration file. Please use quotes, and space between parameters is not recommended. e.g. 'REGNUMBER=1,REGPOWER=3', defaults to 'NONE'" 'NONE'
+opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
```
coalsont, line 85: Noticed this a bit late, but MSMAllTemplates is an internal variable, and shouldn't be referenced in the help info as it the user should know what it means.  The phrasing is also a bit awkward.  These should be changed to ", relative to --msm-all-templates folder".  Additionally, `--msm-all-templates` should probably say "path" or "folder" instead of "string".

cyang31: Thanks, Tim. This is helpful! I have added a new commit to address your comments.

PR #245, opened 2023-02-28
Rikendev

takuya-hayashi:
I 'm suggesting modification of aff2rigid and ACPCAlignment for adapting to NHP or other brains. Now the world format of matrix is used as a new code, aff2rigid_world, so that ACPCAlignment can properly generate a 6-dof matrix from 12-dof affine transformation converting subject's brain to the ACPC space of ANY species of brain, as long as the species' brain template is created L-R symmetrically with AC as an origin, and RPI.

### DIFF
--- a/global/scripts/aff2rigid_world
+++ b/global/scripts/aff2rigid_world
@@ -77,9 +77,9 @@ from numpy import *
 def usage():
     print("Usage: " + argv[0] + " <input2standard mat> <output mat>")
     print(" ")
-    print("       First argument is the FLIRT transform (12 DOF) from the input image to standard")
-    print("       Second argument is the output matrix which will go from the input image to standard space (6 DOF)")
-    print("          aligning the AC, the AC-PC line and the mid-sagittal plane (in order of decreasing accuracy)")
+    print("       First argument is the NIFTI world transform (12 DOF) from the input image to standard")
+    print("       Second argument is the output world matrix which will go from the input image to standard space (6 DOF)")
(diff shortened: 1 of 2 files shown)


glasserm (commented):
I am okay with this if Tim is.

global/scripts/aff2rigid_world
```diff
@@ -0,0 +1,138 @@
...
+    print("       Second argument is the output matrix which will go from the input image to standard space (6 DOF)")
+    print("          aligning the AC, the AC-PC line and the mid-sagittal plane (in order of decreasing accuracy)")
+    sys.exit(1)
+
+if len(argv) < 2:
+    usage()
+
+# Load in the necessary info
+a=loadtxt(argv[1])
+# set specific AC and PC coordinates in FLIRT convention (x1=AC, x2=PC, x3=point above x1 in the mid-sag plane)
```
coalsont, line 90: Change this comment to say something like "AC and PC coordinates in nifti mm convention".

takuya-hayashi: I edited it by clicking "view changes" and "edit file" and commited. I hope it is properly modified.

coalsont: Updating the usage was also a good idea, which I should have suggested.  However, I think the comments immediately above the `x1`, `x2`, and `x3` coordinates should be changed so that they don't claim to be "FLIRT convention", because they aren't anymore.

takuya-hayashi: Sorry, I changed that part.

PR #247, opened 2023-02-28
update SingleSubjectConcat with frame selection; update MSMAll pipeline

cyang31:
## Background
Previously the script `SingleSubjectConcat.sh` doesn't enable the frame range selection for each fMRI runs. Hence, the MSMAll is limited to be only applied to the full timeseries. 

## Main changes
- replace the old opt with Tim's newopt library in `SingleSubjectConcat.sh`;
- remove the bias correction and computation of VN files in `SingleSubjectConcat.sh`;
- add two optional arguments `StartFrame` and `EndFrame` in `SingleSubjectConcat.sh` and `MSMAllPipeline.sh`;
- validate the input frame range and throw an error if invalid range is provided in `SingleSubjectConcat.sh`;

## Evaluation
One subject with resting state fMRI is evaluated. No error.

### DIFF
--- a/MSMAll/MSMAllPipeline.sh
+++ b/MSMAll/MSMAllPipeline.sh
@@ -142,9 +142,6 @@ expected_concatenated_output_dir="${StudyFolder}/${Subject}/MNINonLinear/Results
 expected_concatenated_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}${output_proc_string}.dtseries.nii"
 before_vn_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}_novn.dtseries.nii"
 
-OutputfMRINameOverride=""
-expected_concatenated_output_file_override=""
-
 if [[ "$fMRINames" == "" ]]
 then
 	log_Msg "Running MSM on Multi-run FIX timeseries"
@@ -211,22 +208,7 @@ else
 		--start-frame="${StartFrame}" \
 		--end-frame="${EndFrame}" \
 		--matlab-run-mode="${MatlabRunMode}"
-		
-	# TODO: the current save and load for new OutputfMRIName might cause trouble in parallel mode
-	OutputfMRINameOverride=$(cat "/tmp/OutputfMRIName.txt")
-	rm /tmp/OutputfMRIName.txt
-	# override the expected_concatenated_output_file
-	expected_concatenated_output_file_override="${StudyFolder}/${Subject}/MNINonLinear/Results/${OutputfMRINameOverride}/${OutputfMRINameOverride}${fMRIProcSTRING}${output_proc_string}.dtseries.nii"
 fi
-
-if [[ "$OutputfMRINameOverride" != "" ]]; then
-	OutputfMRIName="$OutputfMRINameOverride"
-fi
-
-if [[ "$expected_concatenated_output_file_override" != "" ]]; then
-	expected_concatenated_output_file="$expected_concatenated_output_file_override"
-fi
-
 log_File_Must_Exist "${expected_concatenated_output_file}"
 
 # fMRIProcSTRING now should reflect the name expected by registrations done below
--- a/MSMAll/MSMAllPipeline.sh
+++ b/MSMAll/MSMAllPipeline.sh
@@ -58,7 +58,7 @@ opts_AddMandatory '--multirun-fix-names-to-use' 'mrfixNamesToUse' 'rfMRI_REST1_L
 opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to give to concatenated single subject scan"
 opts_AddMandatory '--high-pass' 'HighPass' 'integer' 'the high pass value that was used when running FIX' '--melodic-high-pass'
 opts_AddMandatory '--fmri-proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing already done, e.g. '_Atlas_hp0_clean'"
-opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'string' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
+opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'path' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
 opts_AddMandatory '--input-registration-name' 'InputRegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMSulc'"
 opts_AddMandatory '--output-registration-name' 'OutputRegName' 'MSMAll' "the registration string corresponding to the output files, e.g. 'MSMAll_InitalReg'"
 opts_AddMandatory '--high-res-mesh' 'HighResMesh' 'meshnum' "high resolution mesh node count (in thousands), like '164' for 164k_fs_LR"
@@ -82,11 +82,11 @@ opts_AddOptional '--vn' 'VN' 'YES/NO' "whether to perform variance normalization
 opts_AddOptional '--rerun-if-exists' 'ReRunIfExists' 'YES/NO' "whether to re-run even if output already exists, defaults to 'YES'" 'YES'
 opts_AddOptional '--registration-configure-path' 'RegConfPath' 'string' "it can be either the relative path where the registration configuration exists in 'MSMCONFIGDIR', or an absolute path" 'MSMAllStrainFinalconf1to1_1to3'
 opts_AddOptional '--registration-configure-override-variables' 'RegConfOverrideVars' 'string' "the registration configure variables to override instead of using the configuration file. Please use quotes, and space between parameters is not recommended. e.g. 'REGNUMBER=1,REGPOWER=3', defaults to 'NONE'" 'NONE'
-opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
-opts_AddOptional '--rsn-template-file' 'RSNTemplates' 'string' "alternate rsn template file, relative path to 'MSMAllTemplates'" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/melodic_oIC.dscalar.nii'
-opts_AddOptional '--rsn-weights-file' 'RSNWeights' 'string' "alternate rsn weights file, relative path to 'MSMAllTemplates'" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/Weights.txt'
-opts_AddOptional '--topography-roi-file' 'TopographyROIs' 'string' "alternate topography roi file, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.atlas_Topographic_ROIs.32k_fs_LR.dscalar.nii'
-opts_AddOptional '--topography-target-file' 'TopographyTarget' 'string' "alternate topography target, relative path to 'MSMAllTemplates'" 'Q1-Q6_RelatedParcellation210.atlas_Topography.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--rsn-template-file' 'RSNTemplates' 'string' "alternate rsn template file, relative to the --msm-all-templates folder" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/melodic_oIC.dscalar.nii'
+opts_AddOptional '--rsn-weights-file' 'RSNWeights' 'string' "alternate rsn weights file, relative to the --msm-all-templates folder" 'rfMRI_REST_Atlas_MSMAll_2_d41_WRN_DeDrift_hp2000_clean_PCA.ica_dREPLACEDIM_ROW_vn/Weights.txt'
+opts_AddOptional '--topography-roi-file' 'TopographyROIs' 'string' "alternate topography roi file, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.atlas_Topographic_ROIs.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--topography-target-file' 'TopographyTarget' 'string' "alternate topography target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.atlas_Topography.32k_fs_LR.dscalar.nii'
 opts_AddOptional '--matlab-run-mode' 'MatlabRunMode' '0, 1, or 2' "defaults to $g_matlab_default_mode
 0 = compiled MATLAB
 1 = interpreted MATLAB
--- a/MSMAll/MSMAllPipeline.sh
+++ b/MSMAll/MSMAllPipeline.sh
@@ -49,7 +49,7 @@ opts_SetScriptDescription "implements MSM-All Registration Pipeline"
 
 #mandatory
 #general inputs
-opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects"
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects" '--path'
 opts_AddMandatory '--subject' 'Subject' '100206' "one subject ID"
 opts_AddMandatory '--fmri-names-list' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of single-run fmri run names separated by @s"
 opts_AddMandatory '--multirun-fix-names' 'mrfixNames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of multi-run fmri run names separated by @s"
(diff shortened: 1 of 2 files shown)


MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -31,504 +31,182 @@
...
+	
+	# remove the mean files & dense timeseries with selected frame range to save space
+	rm ${ResultsFolder}/${fMRIName}${fMRIProcSTRING}${FrameString}_mean.dscalar.nii
+	rm ${ResultsFolder}/${fMRIName}${fMRIProcSTRING}${FrameString}.dtseries.nii
+	
+	# construct the merge string
+	MergeSTRING=`echo "${MergeSTRING} -cifti ${ResultsFolder}/${fMRIName}${fMRIProcSTRING}${FrameString}${OutputProcSTRING}.dtseries.nii"`
+	
+	# construct the remove string
+	RemoveSTRING=`echo "${RemoveSTRING} ${ResultsFolder}/${fMRIName}${fMRIProcSTRING}${FrameString}${OutputProcSTRING}.dtseries.nii"`
```
coalsont, line 189: If these always get deleted, consider `tempfiles.shlib`.

The backtick echo construct is not needed, `RemoveSTRING="$RemoveSTRING <whatever>"` works (and += can be used), but for whitespace safety (and not needing to remember the space character) arrays are a better approach: `RemoveARRAY+=("<whatever>")`, with a `RemoveARRAY=()` at the top and `rm -f "${RemoveARRAY[@]}"` at the bottom.  `tempfiles.shlib` does this all for you with just a `tempfiles_add "<whatever>"` command and no setup or `rm` at the end, and you can also instead use `tempfiles_create somename_XXXXXX.dtseries.nii mytempname` (the variable `mytempname` contains the actual filename to use) to create a file in the temporary directory, which can often do faster IO than writing to the subject dir.

MSMAll/MSMAllPipeline.sh
```diff
@@ -224,11 +208,23 @@ else
...
-		--variance-normalization="${VarianceNormalization}" \
-		--compute-variance-normalization="${ComputeVarianceNormalization}" \
-		--revert-bias-field="${RevertBiasField}" \
+		--start-frame="${StartFrame}" \
+		--end-frame="${EndFrame}" \
 		--matlab-run-mode="${MatlabRunMode}"
+		
+	# TODO: the current save and load for new OutputfMRIName might cause trouble in parallel mode
+	OutputfMRINameOverride=$(cat "/tmp/OutputfMRIName.txt")
+	rm /tmp/OutputfMRIName.txt
```
cyang31, line 217: This can be improved if I separate the part of frame checking and duration computation from `SingleSubjectConcat.sh` in the main script `MSMAllPipeline.sh`. Otherwise, @coalsont do you have any thought on this potential issue?

MSMAll/scripts/SingleSubjectConcat.sh
cyang31: I added the `tempfiles_add`. This is a really cool function!

MSMAll/MSMAllPipeline.sh
coalsont: I strongly suggest that the modified OutputfMRIName be generated in advance and passed into `SingleSubjectConcat.sh`, rather than coding a special naming convention into the internal script and trying to pass it back out.  I would generate the modified name in the launch script, and pass it to MSMAllPipeline.sh (assuming it doesn't need to actually use files from a previous concatenation), and not have any of this naming logic internal to the pipelines.

glasserm: My suggestion was not to programmatically modify OutputfMRIName, but to physically change the name when you run the code, along with specifying the optional arguments controlling the frame range e.g.,:

1) OutputfMRIName=rfMRI_REST; OutputRegName=MSMAll (all resting state fMRI data).
2) OutputfMRIName=rfMRI_REST_5min_1To100; OutputRegName=MSMAll_5min_1To100 (first ~1.25mins of each resting state fMRI run).
3) OutputfMRIName=fMRI_REST_TASK_5min_1To100; OutputRegName=MSMAll_REST_TASK_5min_1To100 (first ~1.25mins of each resting state and task fMRI run).

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -31,504 +31,177 @@
...
+# loop for demean+vn+concat
+for ((index = 0; index < ${#fMRINamesArray[@]}; ++index)) ; do
+    fMRIName="${fMRINamesArray[$index]}"
+	log_Msg "fMRIName: ${fMRIName}"
+	ResultsFolder="${AtlasFolder}/Results/${fMRIName}"
+	log_Msg "ResultsFolder: ${ResultsFolder}"
+
+	# pick frames
+	if [ $((StartFrame)) -eq 1 ] && [ $((EndFrame)) -eq "${runFrames[index]}" ]; then
+		cp ${ResultsFolder}/${fMRIName}${fMRIProcSTRING}.dtseries.nii ${ResultsFolder}/${fMRIName}${fMRIProcSTRING}${FrameString}.dtseries.nii
```
glasserm, line 159: What is this line needed for?

MSMAll/MSMAllPipeline.sh
cyang31: I found one advantage to programmatically modify the name is that it can automatically calculate the duration with format such as "5mins36secs", "40secs". It won't be a problem if we intentionally omit the seconds part in the file names, or users are fully aware of how to calculate the duration.

glasserm: That looks nice actually.  If you want to do it that way, make sure that it appends to both OutputfMRIName and OutputRegName.

MSMAll/scripts/SingleSubjectConcat.sh
cyang31: This condition means that the current frame range selects the full frame range of the data, which doesn't have to be put into -cifti-merge. This duplicate file is for consistency with the file to be processed afterwards, such as the file to set as temporary file, the file to use for demean and variance normalization. Otherwise, we may have duplicate operations for -cifti-merge, -cifti-reduce.

coalsont: Technically, doing a -cifti-merge in advance of the other code and concatenation will alter the provenance information.  However, it might simplify the logic to instead check whether the start and end frame raw options are something other than the default of "use everything", rather than trying to detect when the frame numbers specify the full run.

coalsont: However, making a copy of a file is a less efficient solution than using a variable to contain the file to use in later steps, and setting it appropriately according to whether a subset of the frames was requested.

coalsont:
I will point out that this PR is currently marked as a draft, which suggests it isn't ready for full scrutiny yet.

MSMAll/MSMAllPipeline.sh
coalsont: I would vote for predictable filenames rather than adding nontrivial-to-calculate things to the filename.  Outputting the duration to the logfile might be nice, but having it in the filename itself sounds like a pain to run later scripts on.

cyang31:
Thanks for your comments. I have addressed all of them.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -31,504 +31,176 @@
...
+duration=0
+for ((index = 0; index < ${#fMRINamesArray[@]}; ++index))
+do
+	TR="${runTR[index]}"
+	Frames=$((EndFrame-StartFrame+1))
+	duration=$(bc -l <<< $duration+$TR*$Frames)
+done
+
+# convert to mins secs
+m=$(bc <<< "(${duration}%3600)/60")
```
coalsont, line 152: If you don't output hours, don't do the `% 3600` part, or 63 minutes will show as "3 minutes".

MSMAll/MSMAllPipeline.sh
glasserm: Tim this is showing up in pre-specified variable names that we have complete freedom over.  This won't mess up downstream scripts.

glasserm:
Please produce some test outputs after you address Tim's concern above.

MSMAll/MSMAllPipeline.sh
glasserm: Well, that was my original suggestion.

cyang31:
> 

Are you pointing to the duration calculation part?

MSMAll/MSMAllPipeline.sh
cyang31: The latest commit doesn't use programmed filenames anymore.

cyang31:
For the whole pipeline test, here are the arguments.
`Tue Feb 28 19:38:09 CST 2023:MSMAllPipeline.sh: arguments: --study-folder=/media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final --subject=103818 --fmri-names-list=rfMRI_REST1_LR@rfMRI_REST1_RL@rfMRI_REST2_LR@rfMRI_REST2_RL --multirun-fix-names= --multirun-fix-concat-name= --multirun-fix-names-to-use= --output-fmri-name=rfMRI_REST_38mins26secs_400To1200 --high-pass=2000 --fmri-proc-string=_Atlas_hp2000_clean --msm-all-templates=/media/myelin/alex/pipelines/HCPpipelines/global/templates/MSMAll --output-registration-name=MSMAll_AlexTest_rest_38mins26secs_400To1200 --high-res-mesh=164 --low-res-mesh=32 --input-registration-name=MSMSulc --start-frame=400 --end-frame=1200 --matlab-run-mode=1`

Now `RegName` and `OutfMRIName` are specified with frame range and the duration. The purpose of including these information in the file name is to avoid overwriting and meanwhile prepare for the following experiments on the MSMAll optimization with varying amount of fMRII data. 

The examples for output files:
- `/media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final/103818/MNINonLinear/fsaverage_LR32k/103818.ArealDistortion_MSMAll_AlexTest_rest_38mins26secs_400To1200_1_d40_WRN.32k_fs_LR.dscalar.nii`
- `/media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final/103818/MNINonLinear/103818.ArealDistortion_MSMAll_AlexTest_rest_38mins26secs_400To1200_2_d40_WRN.164k_fs_LR.dscalar.nii`
- `/media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final/103818/MNINonLinear/Native/103818.ArealDistortion_MSMAll_AlexTest_rest_38mins26secs_400To1200_1_d40_WRN.native.dscalar.nii`
- `/media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final/103818/MNINonLinear/Results/rfMRI_REST_38mins26secs_400To1200/rfMRI_REST_38mins26secs_400To1200_Atlas_hp2000_clean_vn.dtseries.nii`

glasserm:
Is this a temporary file: /media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final/103818/MNINonLinear/Results/rfMRI_REST_38mins26secs_400To1200/frames_duration.txt?

cyang31:
> Is this a temporary file: /media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final/103818/MNINonLinear/Results/rfMRI_REST_38mins26secs_400To1200/frames_duration.txt?

No, this is an output file that aims to let user know explicitly how long this concatenation is and which frame range is selected.

glasserm:
But that is in the file name now, right?

cyang31:
> But that is in the file name now, right?

The file name is set by user. This file is calculated via the script. I think we can provide a **correct** source on this.

glasserm:
I see, we have gone back and forth on this.  I know Tim had concerns about this being programmatically set.  Keeping a file around and having the user set the values also seems not ideal.  Perhaps we should not keep the file and programmatically set things?

cyang31:
> 



> I see, we have gone back and forth on this. I know Tim had concerns about this being programmatically set. Keeping a file around and having the user set the values also seems not ideal. Perhaps we should not keep the file and programmatically set things?

This `frame_duration.txt` file is not used to set the file names, it is just a reminder to help user confirm that they have set the right name. If adding another file to the directory is not ideal, maybe I should just drop the logic to save the file and only log the calculated frame and duration.

glasserm:
Logging would make a lot of sense, though I see the desirability of programmatically setting the file too.

cyang31:
The latest commit deletes the matlab run mode from `SingleSubjectConcat.sh` and comments out the part to save the frame and duration file.

glasserm:
Let's try to come to consensus with @coalsont on the programmatic vs user defined naming before we merge this.

coalsont:
As I understand it, he specified `--output-fmri-name=rfMRI_REST_38mins26secs_400To1200` in that run.  It looks auto-generated because it was copied and pasted from the earlier runs.  It is likely that future runs would not have that particular format of duration in the name.

I think it is fine to generate a "frames_duration.txt" file when using non-default start and end frames (or always, why not), I don't see a drawback to having that information in that filename, since it appears to be in a folder specific to that concatenation.

As long as the output filenames are built with `--output-fmri-name` and constants, with no extra internally computed stuff added, I am okay with it.

coalsont:
If you aren't already, please test the edit I made from `MergeSTRING` to `MergeArray`.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -31,504 +31,171 @@
...
+	EndFrame="$minFrames"
+	log_Msg "EndFrame: ${EndFrame}"
 fi
-
-# Verify that HCPPIPEDIR environment variable is set
-if [ -z "${HCPPIPEDIR}" ]; then
-	echo "${g_script_name}: ABORTING: HCPPIPEDIR environment variable must be set"
-	exit 1
+# check valid frame range
+if [ $((StartFrame)) -gt "${minFrames}" ] || [ $((EndFrame)) -gt "${minFrames}" ] || [ $((StartFrame)) -lt 1 ] || [ $((EndFrame)) -lt 1 ] || [ $((StartFrame)) -gt $((EndFrame)) ]; then
```
coalsont, line 132: This would be cleaner like `if ((StartFrame > minFrames || EndFrame > minFrames`, etc.

However, this doesn't seem to allow the situation of concatenating full runs when the inputs aren't exactly the same length, and `--start-frame` and `--end-frame` aren't specified.  I think we should support this situation, which means preserving the original `--end-frame` default empty string, instead of overwriting that specific input variable with the computed minimum number of frames, and modifying the logic in the loop to distinguish these cases (and not run the `-cifti-merge` in the default case).

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -31,504 +31,171 @@
...
+	VarDemean="-var Mean ${FrameMean} -select 1 1 -repeat"
+	
+	# vn file
+	OutputVN="${ResultsFolder}/${fMRIName}${fMRIProcSTRING}_vn.dscalar.nii"
+	log_File_Must_Exist "$OutputVN"
+	MATHVN=" / max(VN,0.001)"
+	VarVN="-var VN ${OutputVN} -select 1 1 -repeat"
+	
+  	# math expression
+	MATH="(TCS${MATHDemean})${MATHVN}"
```
coalsont, line 180: If we aren't going to have conditionals controlling whether we do the demean and variance normalize, please just have the final `"TCS - Mean / max(VN, 0.001)"` expression as a literal in the -cifti-math line itself, rather than splitting it across multiple variables.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -31,504 +31,171 @@
 #
 #~ND~END~
```
coalsont, line 33: I prefer to have `set -eu` on new-format scripts, which can catch when you make a typo on a variable name, or use a variable before you set it.  The only false positives I know of are if you are used to the lazy "default empty string" behavior (I think it is better style to explicitly do `var=""`, as it makes it obvious what behavior you wanted), or intentionally wanting an empty array to expand to zero words (fairly rare, has been fixed in new bash versions, and there is a workaround for old bash).

We have workarounds to deal with sourcing external things like fix's `settings.sh`, which by default isn't clean for exit codes and undefined variables, but since you don't source anything external here, you shouldn't need to use them.

glasserm:
Sounds like Tim has said what he wants, so if this is fully tested I think we can merge it.

cyang31:
I have updated a commit to address Tim's comments. Regarding the default behavior of EndFrame, here are two passed tests.

Test1: StartFrame=200; EndFrame="";
/media/myelin/alex/MSMAll_optimization/SingleSubjectConcat.sh.o2192569
**Passed**

Test2: StartFrame=1; EndFrame="";
/media/myelin/alex/MSMAll_optimization/SingleSubjectConcat.sh.o2181189
**Passed**

glasserm:
Looks like this needs some cleanup to be able to merge...

cyang31:
> Looks like this needs some cleanup to be able to merge...

Which part exactly?

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -163,31 +141,41 @@ for ((index = 0; index < ${#fMRINamesArray[@]}; ++index)) ; do
 	FrameOutput=${ResultsFolder}/${fMRIName}${fMRIProcSTRING}_${FrameString}${OutputProcSTRING}.dtseries.nii
 
 	# pick frames
-	${Caret7_Command} -cifti-merge ${FrameDenseTCS} -cifti ${FullDenseTCS} -column ${StartFrame} -up-to ${EndFrame}
+	if [[ "$EndFrame" == "" ]]; then
+		if ((StartFrame==1)); then
+			# override the FrameDenseTCS with the full dense timeseries
+			FrameDenseTCS=${FullDenseTCS}
+			# mark temp files for mean, and timeseries after the above process
+			tempfiles_add ${FrameMean} ${FrameOutput}
```
coalsont, line 149: It is fine to `tempfiles_add` something that won't exist, this can be moved out of the conditionals for simplicity.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -163,31 +141,41 @@ for ((index = 0; index < ${#fMRINamesArray[@]}; ++index)) ; do
...
+	else
+		${Caret7_Command} -cifti-merge ${FrameDenseTCS} -cifti ${FullDenseTCS} -column ${StartFrame} -up-to ${EndFrame}
+		# mark temp files for mean, selected range and timeseries after the above process
+		tempfiles_add ${FrameMean} ${FrameDenseTCS} ${FrameOutput}
+	fi
 
 	# mean file
 	${Caret7_Command} -cifti-reduce ${FrameDenseTCS} MEAN ${FrameMean}
-	MATHDemean=" - Mean"
 	VarDemean="-var Mean ${FrameMean} -select 1 1 -repeat"
```
coalsont, line 163: As long as these are going to remain unconditional, I wouldn't put the `-var` argument sequences in variables, either.

MSMAll/scripts/SingleSubjectConcat.sh
cyang31: I don't add `tempfiles_add {FrameDenseTCS}` because in this particular case, I don't want the full dense timeseries to be deleted...

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -163,31 +141,41 @@ for ((index = 0; index < ${#fMRINamesArray[@]}; ++index)) ; do
...
 	
 	# vn file
 	OutputVN="${ResultsFolder}/${fMRIName}${fMRIProcSTRING}_vn.dscalar.nii"
 	log_File_Must_Exist "$OutputVN"
-	MATHVN=" / max(VN,0.001)"
 	VarVN="-var VN ${OutputVN} -select 1 1 -repeat"
 	
   	# math expression
-	MATH="(TCS${MATHDemean})${MATHVN}"
+	MATH="(TCS- Mean) / max(VN,0.001)"
```
coalsont, line 171: I meant put that string into the -cifti-math line, rather than having this `$MATH` variable.

MSMAll/scripts/SingleSubjectConcat.sh
coalsont: Right, but if you do it before the conditional on the temporary file that might not get created, it is fine.  In fact, you should generally `tempfiles_add` files before you actually create them, so that they don't get left behind if the script errors or is interrupted.

cyang31: I see. Thanks for the explanation.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -1,534 +1,180 @@
...
+	log_Msg "FullDenseTCS: ${FullDenseTCS}"
+	NumFrames=`${Caret7_Command} -file-information "${FullDenseTCS}" -only-number-of-maps`
+	runFrames[index]="$NumFrames"
+	TR=`${Caret7_Command} -file-information "${FullDenseTCS}" -only-step-interval`
+	runTR[index]="$TR"
+	if [ $((index)) -eq 0 ]; then
+		minFrames=${runFrames[index]}
 	else
-		log_Msg "Demean: ${p_Demean}"
+		(( ${runFrames[index]}<minFrames )) && minFrames=${runFrames[index]}
```
coalsont, line 93: Please convert this to `if ((${runFrames[index]} < minFrames)); then` format.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -1,534 +1,180 @@
...
+	fMRIName="${fMRINamesArray[index]}"
+	ResultsFolder="${AtlasFolder}/Results/${fMRIName}"
+	log_Msg "ResultsFolder: ${ResultsFolder}"
+	FullDenseTCS=${ResultsFolder}/${fMRIName}${fMRIProcSTRING}.dtseries.nii
+	log_Msg "FullDenseTCS: ${FullDenseTCS}"
+	NumFrames=`${Caret7_Command} -file-information "${FullDenseTCS}" -only-number-of-maps`
+	runFrames[index]="$NumFrames"
+	TR=`${Caret7_Command} -file-information "${FullDenseTCS}" -only-step-interval`
+	runTR[index]="$TR"
+	if [ $((index)) -eq 0 ]; then
```
coalsont, line 90: This could be `if ((index == 0))`.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -1,534 +1,180 @@
...
+# convert to mins secs
+m=$(bc <<< "${duration}/60")
+s=$(bc <<< "${duration}%60")
+DurationString="${m}mins${s}secs"
+
+MergeArray=()
+cnt=0
+# loop for demean+vn+concat
+for ((index = 0; index < ${#fMRINamesArray[@]}; ++index)) ; do
+    fMRIName="${fMRINamesArray[$index]}"
```
coalsont, line 132: This line uses space for indentation, the rest use tabs (except for the stuff from the `newopts.shlib` template).  Please try not to mix indentation types in consecutive lines, at least.  Personally, I prefer 4 spaces for indentation, but that isn't standard across the pipelines.

glasserm: I usually use two spaces in shell and a tab in Matlab.

cyang31:
I think the current commit should have addressed all the comments above.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -87,10 +87,12 @@ do
 	runFrames[index]="$NumFrames"
 	TR=`${Caret7_Command} -file-information "${FullDenseTCS}" -only-step-interval`
 	runTR[index]="$TR"
-	if [ $((index)) -eq 0 ]; then
+	if ((index -eq 0 )); then
```
coalsont, line 90: `-eq` is not valid in `(( ))` (it gets interpreted as "minus variable `eq`"), use `==`.  The main reason for switching to `(( ))` is that `==` (and needing fewer `$`) is more readable in my opinion.

cyang31: fixed now

glasserm:
Still has conflicts to merge.

coalsont:
The conflicts were caused by merging master into the remote branch.  Rebasing would have kept it able to rebase cleanly.  Squash and merge says it will work cleanly, and might be a better choice than adding the 15 commits separately.

cyang31:
I opened a new PR where there's only one commit including the latest changes. I wonder if this is an appropriate solution.

coalsont:
"Squash and merge" is an option on github, you didn't need to do anything.  I'd rather merge this one, because we know we reviewed the code in it, than to check that the new one has exactly the same changes.

cyang31:
Okay, I see.

PR #249, opened 2023-03-01
add myelin map BC module

cyang31:
## Background
Previously the scripts `MSMAll.sh`, `DeDriftAndResamplePipeline.sh` has repeated code blocks for generating the MyelinMap_BC files. The script `CreateMyelinMaps.sh` generated the same files based on GIFTI files. An abstract module is needed.

## `MyelinMap_BC.sh`
- generate the MyelinMap_BC file given an input mesh and an output mesh
- the input mesh and output mesh can be `native`, `164k`, `32k`
- An option `--use-ind-mean` is provided to match the reference map's mean with the individual map's mean value
- An option `RegName` is provided if the files to use is after registration such as `MSMAll`

## Evaluation
Not done yet.

### DIFF
--- a/global/scripts/MyelinMap_BC.sh
+++ b/global/scripts/MyelinMap_BC.sh
@@ -11,21 +11,23 @@ fi
 
 source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
 source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
+source "$HCPPIPEDIR/global/scripts/tempfiles.shlib"
 
 defaultSigma=$(echo "sqrt(200)" | bc -l)
 
 #description to use in usage - syntax of parameters is now explained automatically
-opts_SetScriptDescription "implements CIFTI-based Myelin bias fields from native mesh to low resolution then back to native mesh"
+opts_SetScriptDescription "implements CIFTI-based Myelin bias fields from native mesh to 32k resolution then back to native mesh; by default changing the group reference's mean with individual map's mean"
 
 #mandatory
 #general inputs
 opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects"
 opts_AddMandatory '--subject' 'Subject' '100206' "one subject ID"
-opts_AddMandatory '--registration-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMAll' or 'MSMSulc' or ''"
-opts_AddMandatory '--low-res-mesh' 'LowResMesh' 'meshnum or native' "the low resolution mesh node count (in thousands), like '164' for 164k_fs_LR, '32' for 32k_fs_LR"
+opts_AddMandatory '--registration-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMAll' or 'MSMSulc'"
+opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'path' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
 #optional inputs
-opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'NO'" 'NO'
+opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'YES'" 'YES'
 opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, default '$defaultSigma'" "$defaultSigma"
+opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
 opts_ParseArguments "$@"
 
 if ((pipedirguessed))
@@ -61,34 +63,22 @@ NativeFolder=${AtlasSpaceFolder}/Native
 log_Msg "NativeFolder: $NativeFolder"
 NativeT1wFolder=${T1wFolder}/Native
 log_Msg "NativeT1wFolder: $NativeT1wFolder"
+# MSMAll templates Myelin Target which is the group average MyelinMap_BC
+MyelinTarget="${MSMAllTemplates}/${MyelinTarget}"
+# 32k setting
+LowResMesh=32
+LowResMeshString=${LowResMesh}k_fs_LR
+LowResFolder=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k
+LowResT1wFolder=${T1wFolder}/fsaverage_LR${LowResMesh}k
+LowResBCMapName="atlas_MyelinMap"
 
-case "$LowResMesh" in
-	(164)
-		LowResMeshString=164k_fs_LR
-		LowResFolder=${AtlasSpaceFolder}
-		LowResT1wFolder=${AtlasSpaceFolder}
-		LowResBCMapName="MyelinMap"
-		;;
-	(*)
-		if [[ $LowResMesh =~ ^[0-9]+$ ]]; then
-			LowResMeshString=${LowResMesh}k_fs_LR
-			LowResFolder=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k
-			LowResT1wFolder=${T1wFolder}/fsaverage_LR${LowResMesh}k
-			LowResBCMapName="atlas_MyelinMap"
-		else
-        	log_Err_Abort "unrecognized --low-res-mesh value '$LowResMesh', valid options are numbers like '32', '164', etc."
-		fi
-        ;;
-esac
+# atlas group average reference map
+tempfiles_create atlas_MyelinMap_BC_XXXXXX.dscalar.nii ReferenceMap
+cp "${MyelinTarget}" "$ReferenceMap"
 
 case "$RegName" in
-	("")
-		RegNameInOutputName=""
-		RegNameInT1wName=""
-		RegNameStructString=""
-		;;
 	(MSMSulc)
-		RegNameInOutputName=_${RegName}
+		RegNameInOutputName=""
 		RegNameInT1wName=""
 		RegNameStructString=.${RegName}
 		;;
@@ -118,39 +108,26 @@ ${Caret7_Command} -cifti-resample ${NativeMyelinMap} \
 
 log_Msg "Resampled MyelinMap in the low res mesh space using registration: ${IndividualLowResMap}"
 
-ReferenceMap=${LowResFolder}/${Subject}.${LowResBCMapName}_BC.${LowResMeshString}.dscalar.nii
-log_File_Must_Exist "${ReferenceMap}"
-ReferenceMapOld=${LowResFolder}/${Subject}.${LowResBCMapName}_oldBC.${LowResMeshString}.dscalar.nii
-ReferenceMapToUse=${ReferenceMap}
-
-# TODO: if _oldBC exists, then _BC is from the new method; if _oldBC doesn't exists, then move forward to the if block.
 # match the reference map's mean with the individual map's mean
 if ((UseIndMeanBool)); then
-	log_Msg "match the reference map's mean with the individual map's mean"
+	log_Msg "match the group reference map's mean with the individual map's mean"
 	log_Msg "UseIndMeanBool: $UseIndMeanBool"
-	# the new reference map
-	# TODO: Do we want to distinguish the mean-matched reference map by inserting a string in the file name?
-	ReferenceMapMeanMatch=${LowResFolder}/${Subject}.${LowResBCMapName}_BC.${LowResMeshString}.dscalar.nii
 	# calcualte means of reference and individual maps
 	# not average in vertex areas, but it is fine
 	MeanRef=$(${Caret7_Command} -cifti-stats -reduce MEAN ${ReferenceMap})
 	MeanInd=$(${Caret7_Command} -cifti-stats -reduce MEAN ${IndividualLowResMap})
 	log_Msg "MeanRef: $MeanRef"
 	log_Msg "MeanInd: $MeanInd"
-	# to save the old maps as _oldBC
-	if [ -f "$ReferenceMap" ]; then
-		cp ${ReferenceMap} ${ReferenceMapOld}
-	fi
+	
 	# match the reference map's mean as individual map's mean
-	${Caret7_Command} -cifti-math "Reference - $MeanRef + $MeanInd" ${ReferenceMapMeanMatch} -var Reference ${ReferenceMap}
-	ReferenceMapToUse=${ReferenceMapMeanMatch}
+	${Caret7_Command} -cifti-math "Reference / $MeanRef * $MeanInd" ${ReferenceMap} -var Reference ${ReferenceMap}
 fi
 
 # generate BiasField in the input mesh space
 LowResBiasField=${LowResFolder}/${Subject}.BiasField${RegNameInOutputName}.${LowResMeshString}.dscalar.nii
 ${Caret7_Command} -cifti-math "Individual - Reference" ${LowResBiasField} \
 	-var Individual ${IndividualLowResMap} \
-	-var Reference ${ReferenceMapToUse}
+	-var Reference ${ReferenceMap}
 	
 log_Msg "BiasField in the low res mesh space: ${LowResBiasField}"
 # smooth the bias field
@@ -178,6 +155,13 @@ log_Msg "Resampled BiasField in the native mesh space: ${NativeBiasField}"
 
 # generate bias corrected map in the output mesh space
 NativeBCMap=${NativeFolder}/${Subject}.MyelinMap_BC${RegNameInOutputName}.native.dscalar.nii
+NativeOldBCMap=${NativeFolder}/${Subject}.MyelinMap_oldBC${RegNameInOutputName}.native.dscalar.nii
+# TODO: maybe we should move the file operation outside of this module
+if [[ ! -f "$NativeOldBCMap" ]]; then
+	mv "$NativeBCMap" "$NativeOldBCMap"
+	log_Msg "the old _BC Myelin map in the native mesh space is now saved as : ${NativeOldBCMap}"
+fi
+
 ${Caret7_Command} -cifti-math "Var - Bias" ${NativeBCMap} \
 	-var Var ${NativeMyelinMap} \
 	-var Bias ${NativeBiasField}
--- a/global/scripts/MyelinMap_BC.sh
+++ b/global/scripts/MyelinMap_BC.sh
@@ -27,7 +27,7 @@ opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'path' "path to direct
 #optional inputs
 opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'YES'" 'YES'
 opts_AddOptional '--low-res-mesh' 'LowResMesh' 'meshnum' "low resolution mesh node count (in thousands), defaults to '32' for 32k_fs_LR" '32'
-opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, default '$defaultSigma'" "$defaultSigma"
+opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, this option is probably only intended for non-human-adult data, defaults to '$defaultSigma'" "$defaultSigma"
 opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
 opts_ParseArguments "$@"
 
--- a/global/scripts/MyelinMap_BC.sh
+++ b/global/scripts/MyelinMap_BC.sh
@@ -11,7 +11,7 @@ fi
 
 source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
 source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
-g_matlab_default_mode=1
+
 defaultSigma=$(echo "sqrt(200)" | bc -l)
 
 #description to use in usage - syntax of parameters is now explained automatically
@@ -27,10 +27,6 @@ opts_AddMandatory '--output-mesh' 'OutputMesh' 'meshnum or native' "low resoluti
 #optional inputs
 opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'NO'" 'NO'
 opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, default '$defaultSigma'" "$defaultSigma"
-opts_AddOptional '--matlab-run-mode' 'MatlabRunMode' '0, 1, or 2' "defaults to $g_matlab_default_mode
-0 = compiled MATLAB
-1 = interpreted MATLAB
-2 = Octave" "$g_matlab_default_mode"
 opts_ParseArguments "$@"
 
 if ((pipedirguessed))
@@ -55,25 +51,24 @@ opts_ShowValues
 log_Msg "Starting main functionality"
 Caret7_Command=${CARET7DIR}/wb_command
 RegNameToUse=""
+RegNameStructString=""
 if [[ "$RegName" != "" ]]; then
 	RegNameToUse=_${RegName}
+	RegNameStructString=.${RegName}
 fi
+log_Msg "RegNameToUse: $RegNameToUse"
 
+UseIndMeanBool=$(opts_StringToBool "$UseIndMean")
 # default folders
 SubjFolder=${StudyFolder}/${Subject}
+log_Msg "SubjFolder: $SubjFolder"
 AtlasSpaceFolder=${SubjFolder}/MNINonLinear
+log_Msg "AtlasSpaceFolder: $AtlasSpaceFolder"
 T1wFolder=${SubjFolder}/T1w
-#NativeT1wFolder=${T1wFolder}/Native
-#NativeFolder=${AtlasSpaceFolder}/Native
-#DownSampleFolder=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k
-#DownSampleT1wFolder=${T1wFolder}/fsaverage_LR${LowResMesh}k
-
-# convert upper case to lower case
-InputMesh=`echo "$InputMesh" | tr '[:upper:]' '[:lower:]'`
-OutputMesh=`echo "$OutputMesh" | tr '[:upper:]' '[:lower:]'`
+log_Msg "T1wFolder: $T1wFolder"
 
 if [[ "$InputMesh" == "$OutputMesh" ]]; then
-	log_Err_Abort "the input mesh ${InputMesh} and the output mesh ${OutputMesh} shouldn't be the same!"
+	log_Err_Abort "the input mesh ${InputMesh} and the output mesh ${OutputMesh} shouldn't be the same resolution!"
 fi
 
 case "$InputMesh" in
@@ -89,14 +84,15 @@ case "$InputMesh" in
 		InputT1wFolder=${AtlasSpaceFolder}
 		InputBCMapName="MyelinMap"
 		;;
-	(32)
-		InputMeshString=32k_fs_LR
-		InputFolder=${AtlasSpaceFolder}/fsaverage_LR32k
-		InputT1wFolder=${T1wFolder}/fsaverage_LR32k
-		InputBCMapName="atlas_MyelinMap"
-		;;
 	(*)
-        log_Err_Abort "unrecognized --input-mesh value '$InputMesh', valid options are native, 32, 164"
+		if [[ $InputMesh =~ ^[0-9]+$ ]]; then
+			InputMeshString=${InputMesh}k_fs_LR
+			InputFolder=${AtlasSpaceFolder}/fsaverage_LR${InputMesh}k
+			InputT1wFolder=${T1wFolder}/fsaverage_LR${InputMesh}k
+			InputBCMapName="atlas_MyelinMap"
+		else
+        	log_Err_Abort "unrecognized --input-mesh value '$InputMesh', valid options are 'native' or numbers like '32', '164', etc."
+		fi
         ;;
 esac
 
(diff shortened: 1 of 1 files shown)


cyang31:
This is a **draft** PR. @glasserm Please check if the current implementation is appropriate regarding the input and output.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,180 @@
...
+		InputT1wFolder=${T1wFolder}/Native
+		InputBCMapName="MyelinMap"
+		;;
+	(164)
+		InputMeshString=164k_fs_LR
+		InputFolder=${AtlasSpaceFolder}
+		InputT1wFolder=${AtlasSpaceFolder}
+		InputBCMapName="MyelinMap"
+		;;
+	(32)
```
coalsont, line 92: This doesn't support other meshes like 58k, or any new meshes we might make in the future.  The 32k case should probably use `(*)` as a catch-all, and use the argument assuming it is a number for `fsaverage_*k` and `*k_fs_LR`, etc.  You could use a regexp test to sanity check that it consists entirely of digits.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,180 @@
...
+AtlasSpaceFolder=${SubjFolder}/MNINonLinear
+T1wFolder=${SubjFolder}/T1w
+#NativeT1wFolder=${T1wFolder}/Native
+#NativeFolder=${AtlasSpaceFolder}/Native
+#DownSampleFolder=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k
+#DownSampleT1wFolder=${T1wFolder}/fsaverage_LR${LowResMesh}k
+
+# convert upper case to lower case
+InputMesh=`echo "$InputMesh" | tr '[:upper:]' '[:lower:]'`
+OutputMesh=`echo "$OutputMesh" | tr '[:upper:]' '[:lower:]'`
```
coalsont, line 73: I don't think we need to support people giving "NaTiVe" in either of these arguments.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,180 @@
...
+
+IndividualMap=${InputFolder}/${Subject}.MyelinMap${RegNameToUse}.${InputMeshString}.dscalar.nii
+log_File_Must_Exist "${IndividualMap}"
+ReferenceMap=${InputFolder}/${Subject}.${InputBCMapName}_BC.${InputMeshString}.dscalar.nii
+log_File_Must_Exist "${ReferenceMap}"
+ReferenceMapOld=${InputFolder}/${Subject}.${InputBCMapName}_oldBC.${InputMeshString}.dscalar.nii
+ReferenceMapToUse=${ReferenceMap}
+
+# match the reference map's mean with the individual map's mean
+if ((UseIndMeanBool)); then
```
coalsont, line 135: I think you are missing the `UseIndMeanBool=$(opts_StringToBool "$UseIndMean")` step for this.

cyang31: Oops. My bad.

global/scripts/MyelinMap_BC.sh
glasserm: Agree this should be agnostic to mesh resolution.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,180 @@
...
+#general inputs
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects"
+opts_AddMandatory '--subject' 'Subject' '100206' "one subject ID"
+opts_AddMandatory '--registration-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMAll' or ''"
+opts_AddMandatory '--input-mesh' 'InputMesh' 'meshnum or native' "input resolution mesh node count (in thousands) or 'native', like '164' for 164k_fs_LR"
+opts_AddMandatory '--output-mesh' 'OutputMesh' 'meshnum or native' "low resolution mesh node count (in thousands) or 'native', like '32' for 32k_fs_LR"
+#optional inputs
+opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'NO'" 'NO'
+opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, default '$defaultSigma'" "$defaultSigma"
+opts_AddOptional '--matlab-run-mode' 'MatlabRunMode' '0, 1, or 2' "defaults to $g_matlab_default_mode
```
glasserm, line 30: Why are we supporting matlab here?  We don't depend on matlab anywhere else in the HCP structural pipelines.

global/scripts/MyelinMap_BC.sh
glasserm: Native is always in the native folder.  164k is always in the MNINonlinear folder.  Other meshes are in fsaverage_LR${Mesh} folder.

glasserm (commented):
This looks like a good start aside from Tim's comments.  The next step will be to test that it works properly.

global/scripts/MyelinMap_BC.sh
cyang31: I will delete it in next iteration.

cyang31:
One question is which file name we should specify for the new reference map with mean matched with individual map's mean. The current setting is `ReferenceMapMeanMatch=${InputFolder}/${Subject}.${InputBCMapName}_BC_MatchIndMean.${InputMeshString}.dscalar.nii` I wonder if we want to name it `ReferenceMapMeanMatch=${InputFolder}/${Subject}.${InputBCMapName}_BC.${InputMeshString}.dscalar.nii` after we set the old one as `${InputFolder}/${Subject}.${InputBCMapName}_oldBC.${InputMeshString}.dscalar.nii`.

glasserm:
The wording of the question is confusing to me.  Are you asking me what we want to name the _BC corrected myelin map?  We should keep it the same name and separately (before we run this on data where there is already an _BC corrected myelin map) rename the old one to _oldBC.

cyang31:
> The wording of the question is confusing to me. Are you asking me what we want to name the _BC corrected myelin map? We should keep it the same name and separately (before we run this on data where there is already an _BC corrected myelin map) rename the old one to _oldBC.

I see. So any file with _BC later is by default the mean matched version.

glasserm:
We could add a legacy-BC behavior flag in to preserve the prior behavior, but going forward, the new _BC should be the default.

cyang31:
Here're three tests that are smoothly finished after this new commit. All are using the option `--use-ind-mean=YES`.
- 164k to 32k: /media/myelin/alex/MSMAll_optimization/MyelinMap_BC.sh.o1903215 
- 164k to native: /media/myelin/alex/MSMAll_optimization/MyelinMap_BC.sh.o1905154
- 32k MSMAll to native:  /media/myelin/alex/MSMAll_optimization/MyelinMap_BC.sh.o1912405

glasserm:
I am having difficulty following these tests.

cyang31:
**Test1: 164k to 32k**:
- 164k individual map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.MyelinMap.164k_fs_LR.dscalar.nii
- 164k Reference map new: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.MyelinMap_BC.164k_fs_LR.dscalar (164kTo32k).nii
- 164k Reference map old: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.MyelinMap_oldBC.164k_fs_LR.dscalar (164kTo32k).nii
- 164k BiasField: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.BiasField.164k_fs_LR.dscalar (164kTo32k).nii
- 32k template: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.MyelinMap.32k_fs_LR.dscalar.nii
- 32k BiasField: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.BiasField.32k_fs_LR.dscalar (164kTo32k).nii
- 32k BC map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.atlas_MyelinMap_BC.32k_fs_LR.dscalar (164kTo32k).nii

**Test2: 164k to Native**:
- 164k individual map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.MyelinMap.164k_fs_LR.dscalar.nii
- 164k Reference map new: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.MyelinMap_BC.164k_fs_LR.dscalar (164kToNative).nii
- 164k Reference map old: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.MyelinMap_oldBC.164k_fs_LR.dscalar (164kToNative).nii
- 164k BiasField: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/100307.BiasField.164k_fs_LR.dscalar (164kToNative).nii
- native template: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap.native.dscalar.nii
- native BiasField: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.BiasField.native.dscalar (164kToNative).nii
- native BC map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC.native.dscalar (164kToNative).nii

**Test3: 32k to Native [MSMAll]**
- 32k MSMAll individual map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.MyelinMap_MSMAll.32k_fs_LR.dscalar.nii
- 32k MSMAll Reference map new: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.atlas_MyelinMap_BC.32k_fs_LR.dscalar (32kToNativeMSMAll).nii
- 32k MSMAll Reference map old: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.atlas_MyelinMap_oldBC.32k_fs_LR.dscalar (32kToNativeMSMAll).nii
- native template: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap.native.dscalar.nii
- native MSMAll BiasField: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.BiasField_MSMAll.native.dscalar (32kToNativeMSMAll).nii
- native MSMAll BC map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC_MSMAll.native.dscalar (32kToNativeMSMAll).nii

**Test4: 32k to Native [MSMAll_InitalReg_2_d40_WRN]**
- 32k MSMAll_InitalReg_2_d40_WRN individual map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.MyelinMap_MSMAll_InitalReg_2_d40_WRN.32k_fs_LR.dscalar.nii
- 32k MSMAll_InitalReg_2_d40_WRN Reference map new: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.atlas_MyelinMap_BC.32k_fs_LR.dscalar (32kToNativeMSMAll_InitialReg_2_d40_WRN).nii
- 32k MSMAll_InitalReg_2_d40_WRN Reference map old: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/fsaverage_LR32k/100307.atlas_MyelinMap_oldBC.32k_fs_LR.dscalar (32kToNativeMSMAll_InitialReg_2_d40_WRN).nii
- native template: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap.native.dscalar.nii
- native MSMAll_InitalReg_2_d40_WRN BiasField: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.BiasField_MSMAll_InitalReg_2_d40_WRN.native.dscalar (32kToNativeMSMAll_InitialReg_2_d40_WRN).nii
- native MSMAll_InitalReg_2_d40_WRN BC map: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC_MSMAll_InitalReg_2_d40_WRN.native.dscalar (32kToNativeMSMAll_InitialReg_2_d40_WRN).nii

cyang31:
I was trying to generate result for test5: 32k to native MSMSulc, but found an error for this command:
`# resample the bias field to output mesh space
${Caret7_Command} -cifti-resample ${InputFolder}/${Subject}.BiasField${RegNameToUse}.${InputMeshString}.dscalar.nii \
	COLUMN ${OutputFolder}/${Subject}.MyelinMap.${OutputMeshString}.dscalar.nii \
	COLUMN ADAP_BARY_AREA ENCLOSING_VOXEL \
	${OutputFolder}/${Subject}.BiasField${RegNameToUse}.${OutputMeshString}.dscalar.nii \
	-surface-postdilate 40 \
	-left-spheres ${InputFolder}/${Subject}.L.sphere.${InputMeshString}.surf.gii ${OutputFolder}/${Subject}.L.sphere${RegNameStructString}.${OutputMeshString}.surf.gii \
	-left-area-surfs ${InputT1wFolder}/${Subject}.L.midthickness${RegNameToUse}.${InputMeshString}.surf.gii ${OutputT1wFolder}/${Subject}.L.midthickness.${OutputMeshString}.surf.gii \
	-right-spheres ${InputFolder}/${Subject}.R.sphere.${InputMeshString}.surf.gii ${OutputFolder}/${Subject}.R.sphere${RegNameStructString}.${OutputMeshString}.surf.gii \
	-right-area-surfs ${InputT1wFolder}/${Subject}.R.midthickness${RegNameToUse}.${InputMeshString}.surf.gii ${OutputT1wFolder}/${Subject}.R.midthickness.${OutputMeshString}.surf.gii 
`
The reason why it failed is because the file `100307.L.midthickness_MSMSulc.32k_fs_LR.surf.gii` is not available in /media/myelin/shared/HCPpipelines_test_rc/100307/T1w/fsaverage_LR32k. Shall I use `100307.L.midthickness.32k_fs_LR.surf.gii` when `RegName=MSMSulc`?

coalsont:
MSMSulc files do not contain the string `MSMSulc` in their filename (except for the native registered sphere, it seems).  This is generally specified by giving `--reg-name` the empty string (or leaving it unspecified, or using the `NONE` placeholder that was used before the code could reliably accept an empty string as an argument).

coalsont:
So, hold on, why do we want this module to resample to a (single) different mesh than the input?  We wouldn't want to upsample a 32k myelin map to produce a 164k output either, as it would be missing data, and all of our existing myelin maps are resampled from native mesh directly to final resolution, I believe.  It might make sense for this module to always produce the same-mesh output as the input, and then an additional downsampled version, but that doesn't line up with the current arguments all that well.

Do we actually resample the bias correction template to native mesh in the current code?  Do we ever want to?  wb_command's metric smoothing should be robust enough to handle the haphazard native mesh properly, but do we expect any benefit to doing it that way?

cyang31:
Examples when resampling bias field from downsample to native mesh:
- https://github.com/Washington-University/HCPpipelines/blob/master/DeDriftAndResample/DeDriftAndResamplePipeline.sh#L877-L883
- https://github.com/Washington-University/HCPpipelines/blob/master/MSMAll/scripts/MSMAll.sh#L1149-L1153

glasserm:
I am still confused about these tests.  Also I see a bunch of parentheses and spaces in file names (and in between the parts of the file extension).  I am hoping to see _BC and oldBC for all of the _BC operations we do in a subject (i.e., both MSMSulc and MSMAll) and also in all of the spaces (Native, 164k, and 32k).  I agree with Tim that ideally we would compute _BC on the native mesh according to MSMSulc or MSMAll and then resample the results to 164k or 32k.  If there is somewhere where that isn't being done, it is likely because I was lazy and focused on getting something implemented, not because it is what we want to do going forward.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,187 @@
...
+then
+    pipedirguessed=1
+    #fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
+
+defaultSigma=$(echo "sqrt(200)" | bc -l)
```
coalsont, line 15: This is a strange default sigma, as it is not in the form of a fwhm-conversion, `FWHM / (2 * sqrt(2 l(2)))`.  What did we previously use?

coalsont:
Those examples both appear to do the bias field smoothing on the low-res mesh, presumably for efficiency, then upsample to the native mesh before subtracting the bias from the original native data.  They do not upsample the myelin data from 32k to native or 164k.  They appear to make use of the fact that downsampled versions of those exact native-mesh files already exist, rather than downsampling them immediately prior to the smoothing step.

Similarly, the PR seems to rely on `${OutputFolder}/${Subject}.MyelinMap.${OutputMeshString}.dscalar.nii` already existing, and applies the bias correction to it, using the input resolution only for the bias computation and smoothing steps.  This may be a reasonable approach, but the help info probably needs some adjustment to make it clear that the low res mesh is not the sole input of myelin map data.  In addition, the output mesh should generally be native or 164k, and the "smoothing mesh" would be 32k, which is effectively backwards from the examples in the help info.

global/scripts/MyelinMap_BC.sh
glasserm: That is real, but at this point I forget why it is that.  Probably something I did over a decade ago.

cyang31:
Since the current module is to resample from one resolution to another, if I understand it correctly, Matt, what you suggest is that there should be a few tests (in INPUT_MESH->OUTPUT_MESH format) from low resolution to high resolution:
- Native MSMSulc -> 164k MSMSulc
- Native MSMAll -> 164k MSMAll
- 32k MSMSulc -> Native MSMSulc
- 32k MSMAll -> Native MSMAll
- 32k MSMSulc -> 164k MSMSulc
- 32k MSMAll -> 164k MSMAll

Please correct me if I am wrong.

global/scripts/MyelinMap_BC.sh
coalsont: Technically it has a fairly obtuse connection to 200 mm^2 on a plane (a square aligned with one quarter of the kernel, reaching 1 sigma at the non-diagonal points, which could be cut on the other diagonal and rearranged to very approximately cover a bit less than half the 1-sigma circle)...

cyang31: Here're several cases in the existing scripts:
- https://github.com/Washington-University/HCPpipelines/blob/master/PostFreeSurfer/PostFreeSurferPipeline.sh#L50
- https://github.com/Washington-University/HCPpipelines/blob/master/MSMAll/scripts/MSMAll.sh#L529

coalsont:
If I understand correctly, the 32k downsampling's purpose is only computation efficiency (to deal with the rather large smoothing kernel), and/or avoiding doing smoothing on the haphazard native mesh, and is not a primary feature of correcting the myelin maps.  It probably only needs to generate the native _BC versions, and then optionally downsample those to both 164k and 32k (and possibly 58k or others in the future).

It would probably be most internally consistent for it to resample the native myelin map to the myelin template mesh, do the subtraction and smoothing there, then resample back to the native mesh and apply it to the native myelin map.  This would mean it always uses the same individual myelin data for both the smoothing and correction steps (rather than relying on consistency between two input files).

glasserm:
Sorry, still confused.  I think the point of this module is to resample the native mesh uncorrected myelin map according to the desired registration to the low resolution space, compare with the reference map in the low resolution space to compute a bias field, and then resample that bias field back to the native mesh to compute an _BC myelin map on the native mesh.  That _BC myelin map can then subsequently be split into GIFTIs, resampled to standard meshes according to the registration, etc.  Thus, I am not really clear on why we are trying all of these permutations.  When it really comes down to it, we just need to compare _BC with MSMSulc  and MSMAll to  _oldBC on the native mesh and then verify that we can produce the desired standard mesh myelin maps and GIFTIs.

glasserm:
I see Tim and I just wrote something similar.

coalsont:
As an additional detail for the "individual input is only native mesh" plan, the downsampling to 32k should use the ADAP_BARY_AREA method, which should use the `-*-area-surfs` options (because this is individual data, we don't need to use average vertex area metrics).

cyang31:
A latest commit is pushed to address the comments. The two tests are available to check. There are no clear visual difference between the _BC and _oldBC maps on native mesh.

Test1: native mesh to 32k MSMAll, then back to native mesh
- MyelinMap_BC on native mesh /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC_MSMAll_from_32k_MSMAll.native.dscalar.nii
- MyelinMap_oldBC on native mesh /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_oldBC_MSMAll_from_32k_MSMAll.native.dscalar.nii

Test2: native mesh to 32k MSMSulc, then back to native mesh
- MyelinMap_BC on native mesh /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC_MSMSulc_from_32k_MSMSulc.native.dscalar.nii
- MyelinMap_oldBC on native mesh /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_oldBC_MSMSulc_from_32k_MSMSulc.native.dscalar.nii

glasserm:
Is the oldBC really the old method, or just prior runs of the new method?

cyang31:
> Is the oldBC really the old method, or just prior runs of the new method?

They are generated from the old method where the reference map in low resolution mesh is directly used without mean matching operation.

glasserm:
My goal was to actually compare the original pipeline outputs as _oldBC not newly generated outputs.

cyang31:
I see. I am not sure if data in `/media/myelin/shared/HCPpipelines_test_rc/100307/` is generated from the original pipeline. If it does, here's the raw Myeline_BC_MSMAll on native mesh: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC_MSMAll.native.dscalar.nii

Regarding the MSMSulc version, i wonder if `/media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC.native.dscalar.nii` is MSMSulc based. If yes, then this is the raw file from the original pipeline.

glasserm:
The MSMSulc file seemed to be the original but the MSMAll file had been overwritten, so I pulled it from the snapshots.  For MSMAll, everything looks in good alignment, but for MSMSulc, there is quite a difference between your approach and the prior one.  Perhaps further investigation is needed?

cyang31:
Matt, which file do you look at for MSMSulc to say that's the original one?

global/scripts/MyelinMap_BC.sh
```diff
@@ -66,79 +57,84 @@ AtlasSpaceFolder=${SubjFolder}/MNINonLinear
...
 esac
 
-case "$OutputMesh" in
-	(native)
-		OutputMeshString="native"
-		OutputFolder=${AtlasSpaceFolder}/Native
-		OutputT1wFolder=${T1wFolder}/Native
-		OutputBCMapName="MyelinMap"
+case "$RegName" in
+	("")
```
coalsont, line 85: empty string should probably be taken to mean the same as MSMSulc, these settings look like they would fail (it doesn't make sense to resample the group template to individual native mesh without using a registration).

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,188 @@
...
+	ReferenceMapMeanMatch=${LowResFolder}/${Subject}.${LowResBCMapName}_BC.${LowResMeshString}.dscalar.nii
+	# calcualte means of reference and individual maps
+	# not average in vertex areas, but it is fine
+	MeanRef=$(${Caret7_Command} -cifti-stats -reduce MEAN ${ReferenceMap})
+	MeanInd=$(${Caret7_Command} -cifti-stats -reduce MEAN ${IndividualLowResMap})
+	log_Msg "MeanRef: $MeanRef"
+	log_Msg "MeanInd: $MeanInd"
+	# to save the old maps as _oldBC
+	if [ -f "$ReferenceMap" ]; then
+		cp ${ReferenceMap} ${ReferenceMapOld}
```
coalsont, line 142: If we are going to overwrite this anyway, mv is much faster than cp.  I thought we were wanting to preserve the old BC myelin map, not the old resampled group reference.

I'm also not sure the script itself should have anything to do with saving the old version, as running it twice will blow it away.  That should probably be a separate job, and this script shouldn't mess with the oldBC filenames at all.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,188 @@
...
+log_Msg "NativeFolder: $NativeFolder"
+NativeT1wFolder=${T1wFolder}/Native
+log_Msg "NativeT1wFolder: $NativeT1wFolder"
+
+case "$LowResMesh" in
+	(164)
+		LowResMeshString=164k_fs_LR
+		LowResFolder=${AtlasSpaceFolder}
+		LowResT1wFolder=${AtlasSpaceFolder}
+		LowResBCMapName="MyelinMap"
```
coalsont, line 70: This seems wrong for the name of the resampled group myelin map, it looks like the name of the individual myelin map.  Did we ever generate a file with this name previously, and was it really the group average map?

cyang31:
Matt, do you mind if you can check the tests result for the new version of this script? I checked the files and they look good alignment to me.

Test1 MSMSulc: 
- New _BC: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC.native.dscalar.nii
- Old _BC: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_oldBC.native.dscalar.nii

Test2 MSMAll: 
- New _BC: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_BC_MSMAll.native.dscalar.nii
- Old _BC: /media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/Native/100307.MyelinMap_oldBC_MSMAll.native.dscalar.nii

glasserm:
Thanks Alex.  Which reference myelin map was used for the _BC correction?

cyang31:
The reference map was the group average template map `${HCPPIPEDIR}/global/templates/MSMAll/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii` and is processed to have the same mean as the corresponding individual map.

glasserm:
Sounds good.  We should see what @coalsont has to say about the revised implementation, but you could work on getting the other scripts ready for this module.

global/scripts/MyelinMap_BC.sh
cyang31: I see. I have changed it in the latest commit.

global/scripts/MyelinMap_BC.sh
cyang31: This version has some substantial logic mistakes. I have updated it in the latest commit.

cyang31:
The latest commit treats the reference map (processed or not) as a tempfile. If there's an agreement on which file name it should be saved with (both for mean matched or unprocessed versions), I can change it back to "save" mode anytime.

global/scripts/MyelinMap_BC.sh
```diff
@@ -11,21 +11,23 @@ fi
 
 source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
 source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
+source "$HCPPIPEDIR/global/scripts/tempfiles.shlib"
 
 defaultSigma=$(echo "sqrt(200)" | bc -l)
 
 #description to use in usage - syntax of parameters is now explained automatically
-opts_SetScriptDescription "implements CIFTI-based Myelin bias fields from native mesh to low resolution then back to native mesh"
+opts_SetScriptDescription "implements CIFTI-based Myelin bias fields from native mesh to 32k resolution then back to native mesh; by default changing the group reference's mean with individual map's mean"
```
coalsont, line 19: I wouldn't mention 32k or low resolution here, that is an implementation detail, and the help info already lists the defaults.  I would have it just say "corrects an individual native mesh myelin map for bias fields based on a group average myelin map".

global/scripts/MyelinMap_BC.sh
```diff
@@ -178,6 +155,13 @@ log_Msg "Resampled BiasField in the native mesh space: ${NativeBiasField}"
 
 # generate bias corrected map in the output mesh space
 NativeBCMap=${NativeFolder}/${Subject}.MyelinMap_BC${RegNameInOutputName}.native.dscalar.nii
+NativeOldBCMap=${NativeFolder}/${Subject}.MyelinMap_oldBC${RegNameInOutputName}.native.dscalar.nii
+# TODO: maybe we should move the file operation outside of this module
+if [[ ! -f "$NativeOldBCMap" ]]; then
```
coalsont, line 160: I still think this renaming of the old version shouldn't be in the pipelines code itself.  If this code is run on fresh data, and then run a second time, it will make an _oldBC file that was the first run of the new code.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,172 @@
...
+LowResFolder=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k
+LowResT1wFolder=${T1wFolder}/fsaverage_LR${LowResMesh}k
+LowResBCMapName="atlas_MyelinMap"
+
+# atlas group average reference map
+tempfiles_create atlas_MyelinMap_BC_XXXXXX.dscalar.nii ReferenceMap
+cp "${MyelinTarget}" "$ReferenceMap"
+
+case "$RegName" in
+	(MSMSulc)
```
coalsont, line 80: I would advise `(MSMSulc|"")`, since we treat `NONE` or empty string for RegName as referring to MSMSulc in some pipelines ("MSMSulc" is almost never how the pipelines options select that registration).

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,172 @@
...
+		RegNameInOutputName=""
+		RegNameInT1wName=""
+		RegNameStructString=.${RegName}
```
coalsont, line 83: Relatedly, this should use a literal `".MSMSulc"` when MSMSulc is selected, since there are multiple settings of RegName that mean MSMSulc, and we want them all to act identically.

global/scripts/MyelinMap_BC.sh
cyang31: Now I comment out these lines. I agree that this shouldn't be in the official pipeline as it will cause troubles for users and is not explicitly mentioned to them. The comment now is just a reminder to us if we want to generate new _BC and save the old ones.

coalsont: Looks like the lines were deleted, with no remaining comment, but deleting the lines from this script is what I would suggest.  As long as we remember (some other way) to do this rename before running the modified pipeline with the new template on the full dataset.

global/scripts/MyelinMap_BC.sh
```diff
@@ -0,0 +1,168 @@
...
+T1wFolder=${SubjFolder}/T1w
+log_Msg "T1wFolder: $T1wFolder"
+NativeFolder=${AtlasSpaceFolder}/Native
+log_Msg "NativeFolder: $NativeFolder"
+NativeT1wFolder=${T1wFolder}/Native
+log_Msg "NativeT1wFolder: $NativeT1wFolder"
+# MSMAll templates Myelin Target which is the group average MyelinMap_BC
+MyelinTarget="${MSMAllTemplates}/${MyelinTarget}"
+# 32k setting
+LowResMesh=32
```
coalsont, line 69: I'm not sure if we want to hardcode this, or have it the default with an option to change it.  It looks like the downsampled version of the uncorrected map is produced in the subject folder, and not deleted, so this would show up if we ever changed the default fMRI surface resolution.

glasserm: Agree this should be a parameter.

coalsont:
This looks pretty close to done.

cyang31:
I will mark this PR as "Ready for review" which can be merged now. There will be a following PR to
- replace MSMAll.sh part of myelin map _BC with the new module
- replace MSMAll.sh part of arguments with the newopt library.

Does this sound good to you? @glasserm

glasserm:
What about the changes needed in the various scripts that call this module?

cyang31:
The changes can be put into the next PR as well, as long as I have a starting point which is confirmed by you and Tim.

coalsont:
They don't call it yet, sounds like he wants to do those in a separate PR.  It is not an error to have an unused script in the repo, and as I understand it he has tested the script, so we could do it that way.

cyang31:
The starting point will be the replacement in MSMAll.sh.

glasserm:
Sounds good.  If we are good with this it can be merged.

PR #250, opened 2023-03-07
update SingleSubjectConcat with frame selection; update MSMAll pipeline

cyang31:
This is the squash and merge version of a previous [PR](https://github.com/Washington-University/HCPpipelines/pull/247)

## Background
Previously the script `SingleSubjectConcat.sh` doesn't enable the frame range selection for each fMRI runs. Hence, the MSMAll is limited to be only applied to the full timeseries. 

## Main changes
- replace the old opt with Tim's newopt library in `SingleSubjectConcat.sh`;
- remove the bias correction and computation of VN files in `SingleSubjectConcat.sh`;
- add two optional arguments `StartFrame` and `EndFrame` in `SingleSubjectConcat.sh` and `MSMAllPipeline.sh`;
- validate the input frame range and throw an error if invalid range is provided in `SingleSubjectConcat.sh`;

## Evaluation
One subject with resting state fMRI is evaluated. No error.


glasserm (commented):
I forget, is this fully tested?  If so we might just merge this one.

cyang31:
> I forget, is this fully tested? If so we might just merge this one.
Yes, it has been fully tested.

coalsont:
Let's use the original one, so we don't need to compare whether this one has completely identical changes.

PR #251, opened 2023-03-07
replace _BC with MyelinMap_BC module in MSMAll.sh

cyang31:
## Background
Previously the script `MSMAll.sh` doesn't use newopt and has code block that can be replaced with the new module `global/scripts/MyelinMap_BC.sh`.

## Main changes
- replace the old opt with the newopt library in `MSMAll.sh`;
- replace _BC generation in `MSMAll.sh` with the module `MyelinMap_BC.sh`;
- adjust the arguments in `MSMAll/MSMAllPipeline.sh`

## Evaluation
Not done yet.


MSMAll/scripts/MSMAll.sh
```diff
@@ -1,1195 +1,254 @@
...
-		${Caret7_Command} -cifti-math "Var - Bias" ${NativeFolder}/${Subject}.MyelinMap_BC_${RegName}.native.dscalar.nii -var Var ${NativeFolder}/${Subject}.MyelinMap.native.dscalar.nii -var Bias ${NativeFolder}/${Subject}.BiasField_${RegName}.native.dscalar.nii
```
cyang31, line 1153: Here are the replacements using `MyelinMap_BC.sh`.

MSMAll/scripts/MSMAll.sh
```diff
@@ -1,1195 +1,254 @@
...
+AtlasFolder="${StudyFolder}/${Subject}/MNINonLinear"
+DownSampleFolder="${AtlasFolder}/fsaverage_LR${LowResMesh}k"
+NativeFolder="${AtlasFolder}/Native"
+ResultsFolder="${AtlasFolder}/Results/${OutputfMRIName}"
+T1wFolder="${StudyFolder}/${Subject}/T1w"
+DownSampleT1wFolder="${T1wFolder}/fsaverage_LR${LowResMesh}k"
+NativeT1wFolder="${T1wFolder}/Native"
+
+if [[ $(echo -n "${Method}" | grep "WR") ]] ; then
+	LowICAdims=$(echo "${RegressionParams}" | sed 's/@/ /g')
```
coalsont, line 90: I think Matt wanted to still support `_` here.  You could do this as-is with sed using `'s/[_@]/ /g'`.  There is also a way to do it with arrays, but since these will all be numbers, we don't need to worry about handling whitespace.

MSMAll/scripts/MSMAll.sh
```diff
@@ -1,1195 +1,254 @@
...
+log_Msg "CorrectionSigma: ${CorrectionSigma}"
+log_Msg "BC: ${BC}"
+log_Msg "nTPsForSpectra: ${nTPsForSpectra}"
+log_Msg "VolParams: ${VolParams}"
+
+IndArealDistortionFile=${NativeFolder}/${Subject}.ArealDistortion_${RegNameStem}_${NumIterations}_d${ICAdim}_${Method}.native.dscalar.nii
+if [[ -e ${IndArealDistortionFile} &&  ${ReRun} = "NO" ]]; then
+	log_Msg "IndArealDistortionFile exists: ${IndArealDistortionFile} and ReRun is disabled"
+	log_Msg "the main functionality of MSMAll.sh script is finished"
+	exit 0
```
coalsont, line 121: This is a much better solution for this behavior than the previous `if` statement over the entire function body.  However, the log message should perhaps say "--rerun is set to 'no'" and "skipping MSMAll.sh" rather than saying the functionality is finished.

Also, please use `opts_StringToBool` on `ReRun`, to also accept true/false, 0/1, and all capitalizations.  The previous implementation was "literal `YES`, capitalized, everything else means no, no sanity checking", which is a little unfortunate, but the current PR instead treats anything ambiguous as a yes, which could bite unsuspecting users.

coalsont:
For anyone not aware, github has a feature to show the diffs while ignoring whitespace changes, which is particularly useful for this PR.  Right below the files changed tab, there is a gear icon, which sets the diff type, which has a checkbox at the bottom of the popup labeled "hide whitespace".

coalsont:
Has this been tested?  Looks pretty good to me.

cyang31:
I fixed the bugs for this script and tested on it. The full output is here, /media/myelin/alex/MSMAll_optimization/MSMAllPipeline.sh.o1964568
The test passed without any error on a test subject.

glasserm:
When you do this test, is it on a clean run before MSMAll?  That is safer for testing purposes.

cyang31:
No, the testing data in that folder `/media/myelin/shared/HCPpipelines_test_rc/100307/MNINonLinear/` doesn't have _vn file for process string 'Atlas_hp0_clean'. I can see if I can copy a fully processed subject to this directory `/media/myelin/shared/HCPpipelines_test_rc/`.

cyang31:
I tested the script on a new testing subject, the log output is here `/media/myelin/alex/MSMAll_optimization/MSMAllPipeline.sh.o2367006` without an error. The fMRI process string is `_Atlas_hp2000_clean_rclean`.

glasserm:
Thanks Alex.  This can be committed.

PR #252, opened 2023-03-11
replace _BC with MyelinMap_BC module in DeDriftAndResamplePipeline.sh

cyang31:
## Background
Previously the script `DeDriftAndResamplePipeline.sh` doesn't use newopt and has code block that can be replaced with the new module `global/scripts/MyelinMap_BC.sh`.

## Main changes
- replace the old opt with the newopt library in `DeDriftAndResamplePipeline.sh`;
- replace _BC generation in `DeDriftAndResamplePipeline.sh` with the module `MyelinMap_BC.sh`;

## Evaluation
Test is passed on a HCD subject. 
launch script: `/media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipelineBatch.sh`
log output without error:
`/media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipeline.sh.o770799`

### DIFF
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+set -eu
 #
 # # DeDriftAndResamplePipeline.sh
 #
@@ -228,8 +228,8 @@ log_Msg "ResultsFolder: ${ResultsFolder}"
 DownSampleFolderNames=""
 DownSampleT1wFolderNames=""
 for LowResMesh in ${LowResMeshes} ; do
-	DownSampleFolderNames=`echo "${DownSampleFolderNames}${AtlasFolder}/fsaverage_LR${LowResMesh}k "`
-	DownSampleT1wFolderNames=`echo "${DownSampleT1wFolderNames}${T1wFolder}/fsaverage_LR${LowResMesh}k "`
+	DownSampleFolderNames+="${DownSampleFolderNames}${AtlasFolder}/fsaverage_LR${LowResMesh}k "
+	DownSampleT1wFolderNames+="${DownSampleT1wFolderNames}${T1wFolder}/fsaverage_LR${LowResMesh}k "
 done
 log_Msg "DownSampleFolderNames: ${DownSampleFolderNames}"
 log_Msg "DownSampleT1wFolderNames: ${DownSampleT1wFolderNames}"
@@ -263,7 +263,7 @@ for Hemisphere in L R ; do
 	${Caret7_Command} -surface-distortion ${NativeFolder}/${Subject}.${Hemisphere}.sphere.native.surf.gii ${NativeFolder}/${Subject}.${Hemisphere}.sphere.${OutputRegName}.native.surf.gii ${NativeFolder}/${Subject}.${Hemisphere}.EdgeDistortion_${OutputRegName}.native.shape.gii -edge-method
 	
 	${Caret7_Command} -surface-distortion "${NativeFolder}"/"${Subject}"."${Hemisphere}".sphere.native.surf.gii "${NativeFolder}"/"${Subject}"."${Hemisphere}".sphere.${OutputRegName}.native.surf.gii "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -local-affine-method
-    ${Caret7_Command} -metric-merge "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii -metric "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -column 1
+	${Caret7_Command} -metric-merge "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii -metric "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -column 1
     ${Caret7_Command} -metric-merge "${NativeFolder}"/"$Subject"."$Hemisphere".StrainR_${OutputRegName}.native.shape.gii -metric "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -column 2
     ${Caret7_Command} -metric-math "ln(var) / ln (2)" "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii -var var "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii
     ${Caret7_Command} -metric-math "ln(var) / ln (2)" "${NativeFolder}"/"$Subject"."$Hemisphere".StrainR_${OutputRegName}.native.shape.gii -var var "${NativeFolder}"/"$Subject"."$Hemisphere".StrainR_${OutputRegName}.native.shape.gii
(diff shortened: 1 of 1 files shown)


DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,614 @@
...
+				log_Debug_Msg "Point 1.2"
+				# Myelin Map BC using 32k
+				"$HCPPIPEDIR"/global/scripts/MyelinMap_BC.sh \
+					--study-folder="$StudyFolder" \
+					--subject="$Subject" \
+					--registration-name="$RegName" \
+					--msm-all-templates="$MSMAllTemplates" \
+					--use-ind-mean="$UseIndMean" \
+					--low-res-mesh="$LowResMesh" \
+					--myelin-target-file="$MyelinTargetFile"
```
cyang31, line 468: replace with MyelinMap_BC.sh

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,614 @@
...
-					${Caret7_Command} -cifti-resample ${DownSampleFolder}/${Subject}.BiasField_${OutputRegName}.${LowResMesh}k_fs_LR.dscalar.nii COLUMN ${NativeFolder}/${Subject}.MyelinMap.native.dscalar.nii COLUMN ADAP_BARY_AREA ENCLOSING_VOXEL ${NativeFolder}/${Subject}.BiasField_${OutputRegName}.native.dscalar.nii -surface-postdilate 40 -left-spheres ${DownSampleFolder}/${Subject}.L.sphere.${LowResMesh}k_fs_LR.surf.gii ${NativeFolder}/${Subject}.L.sphere.${OutputRegName}.native.surf.gii -left-area-surfs ${DownSampleT1wFolder}/${Subject}.L.midthickness_${OutputRegName}.${LowResMesh}k_fs_LR.surf.gii ${NativeT1wFolder}/${Subject}.L.midthickness.native.surf.gii -right-spheres ${DownSampleFolder}/${Subject}.R.sphere.${LowResMesh}k_fs_LR.surf.gii ${NativeFolder}/${Subject}.R.sphere.${OutputRegName}.native.surf.gii -right-area-surfs ${DownSampleT1wFolder}/${Subject}.R.midthickness_${OutputRegName}.${LowResMesh}k_fs_LR.surf.gii ${NativeT1wFolder}/${Subject}.R.midthickness.native.surf.gii
```
cyang31, line 879: the old _BC is here

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,614 @@
...
+		log_Debug_Msg "Point 2.0"
 
-					# ----- End moved statements -----
+		if [[ ${Map} = "ArealDistortion" || ${Map} = "EdgeDistortion" || ${Map} = "StrainJ" || ${Map} = "StrainR" || ${Map} = "MyelinMap_BC" || ${Map} = "SmoothedMyelinMap_BC" ]] ; then
+			NativeMap="${Map}_${OutputRegName}"
+		else
+			NativeMap="${Map}"
+		fi
 
-					${Caret7_Command} -cifti-math "Var - Bias" ${NativeFolder}/${Subject}.${Map}_BC_${OutputRegName}.native.dscalar.nii -var Var ${NativeFolder}/${Subject}.${Map}.native.dscalar.nii -var Bias ${NativeFolder}/${Subject}.BiasField_${OutputRegName}.native.dscalar.nii
```
cyang31, line 883: The old _BC is here.

cyang31:
I also tried to test this new script but it always has errors when running the `ReApplyFixPipeline.sh` part. 
`/media/myelin/alex/pipelines/HCPpipelines/ICAFIX/ReApplyFixPipeline.sh \
	--path=/media/myelin/shared/HCPpipelines_test_rc \
	--subject=103818 --fmri-name=rfMRI_REST1_LR \
	--high-pass=2000 --reg-name=MSMAll \
	--matlab-run-mode=1 --motion-regression=TRUE`

- The error in the compiled version said that `ERROR: To use MATLAB run mode: 0, the FSL_FIX_MCR environment variable must be set` which I am not sure of how to set it in the $FSL_FIXDIR/settings.sh. 
- The error in the interactive version seems to be a bug that the TR is calculated as an empty vector `[]` in line60 of `fix_3_clean.m`.

glasserm:
I am not having any issues with up to date versions of this code.  I expect there are earlier errors in the interpreted version.  Which FIX directory are you using?

cyang31:
I thought mine is the latest version as well. `export FSL_FIXDIR=/media/myelin/brainmappers/fix1.06.15`

glasserm:
It doesn't look like you have a complete dataset in your testing location.  Also, we aren't using --motion-regression=TRUE any more.  Finally, I think you are running these registrations on MSMSulc, recleaned, tICA cleaned inputs.  ReApplyFixPipeline is not set up to replicate all that and at this point is a deprecated pipeline (given our recommendation to use Multi-Run Fix instead).  

It might be easier to write some custom code to clean the native mesh GIFTI timeseries (/media/myelin/shared/HCPpipelines_test_rc/103818/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR.L.native.func.gii
/media/myelin/shared/HCPpipelines_test_rc/103818/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR.R.native.func.gii) using hp2000, sICA+FIX with the reclean reclassification, and tICA cleanup.  

An alternative, potentially simpler approach might be to run fMRISurface on the already fully cleaned (hp2000, reclean, tclean) NIFTI files to generate fully cleaned native mesh timeseries.  Then you could "don't FIX" those fMRI timeseries.  

Yet another alternative could be to run fMRISurface on the fully cleaned data and use the MSMAll registration spheres output from DeDriftAndResample (and not resample fMRI timeseries at all with that).

cyang31:
Are you suggesting to run the `GenericfMRISurfaceProcessingPipeline.sh` on the `/media/myelin/shared/HCPpipelines_test_rc/103818/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR_hp2000_clean_rclean_tclean.nii.gz`? I don't understand how these two can be related...

coalsont:
So to clarify, the problem with testing these myelin map changes to DeDriftAndResample is that the unedited fMRI section of the code errors with specific settings that don't match how it was previously run?  This sounds like you should test (not using our main processing directory for the outputs, hopefully) using whatever FIX settings were previously used just to get through those steps, ignore the fMRI outputs (or check them without expecting them to have the latest cleaning), and only check the myelin maps from the outputs.

Matt seems to be suggesting a hacky workaround to force the tICA cleaned fMRI data through a pipeline it has never been through before, which seems overcomplicated for testing a feature unrelated to fMRI.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,614 @@
...
-	if [ -z "${p_StudyFolder}" ]; then
-		log_Err "Study Folder (--path= or --study-folder=) required"
-		error_count=$(( error_count + 1 ))
-	else
-		log_Msg "Study Folder: ${p_StudyFolder}"
+if [[ "${OutputRegName}" == "" ]]; then
+	OutputRegName="${RegName}"
+else
+	if [[ "${DeDriftRegFiles}" == "" ]]; then
+		log_Err "--concat-reg-name must not be used unless dedrifting (just remove --concat-reg-name=<whatever> from your command)"
```
coalsont, line 101: This should probably be `log_Err_Abort`.  `log_Err` continues without exiting, which is generally not what should happen after an error.

coalsont: The reason the previous parsing code used `log_Err` was to report all the issues with the arguments before exiting (hence the error count variable).  The new parsing does that internally for the mandatory arguments, but I think it would be unnecessary complication to implement it for the additional "these can't be used together" or other detailed sanity checking logic.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,614 @@
...
-	fi
-	log_Msg "After delimiter substitution, fixNames: ${fixNames[@]}"
+	# Make MSM Registration Areal Distortion Maps
+	log_Msg "Make MSM Registration Areal Distortion Maps"
+	${Caret7_Command} -surface-vertex-areas ${NativeFolder}/${Subject}.${Hemisphere}.sphere.native.surf.gii ${NativeFolder}/${Subject}.${Hemisphere}.sphere.native.shape.gii
+	${Caret7_Command} -surface-vertex-areas ${NativeFolder}/${Subject}.${Hemisphere}.sphere.${OutputRegName}.native.surf.gii ${NativeFolder}/${Subject}.${Hemisphere}.sphere.${OutputRegName}.native.shape.gii
+	${Caret7_Command} -metric-math "ln(spherereg / sphere) / ln(2)" ${NativeFolder}/${Subject}.${Hemisphere}.ArealDistortion_${OutputRegName}.native.shape.gii -var sphere ${NativeFolder}/${Subject}.${Hemisphere}.sphere.native.shape.gii -var spherereg ${NativeFolder}/${Subject}.${Hemisphere}.sphere.${OutputRegName}.native.shape.gii
+	rm ${NativeFolder}/${Subject}.${Hemisphere}.sphere.native.shape.gii ${NativeFolder}/${Subject}.${Hemisphere}.sphere.${OutputRegName}.native.shape.gii
 
-	if [ "${dontFixNames}" = "NONE" ] ; then
```
coalsont, line 545: There may be legacy launch scripts using NONE for some of these.  I would be happy to remove the NONE convention from the pipelines entirely, but our users may not be quite as happy if their old launcher scripts error for what seem like silly reasons.  @glasserm ?

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,614 @@
...
-       2 = Use interpreted Octave
-
-EOF
-}
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
+g_matlab_default_mode=1
+
+#description to use in usage - syntax of parameters is now explained automatically
+opts_SetScriptDescription "implements DeDrift and Resample Pipeline"
```
coalsont, line 43: This is the dedrift and resample pipeline, it is a little odd to say that it implements itself.  You could just put "De-Drift and Resample" to match the previous help info.  Or, we could make it more informative, and say "resamples many types of data after a surface registration".  The "dedrift" part is basically an unfortunate red herring, made more unfortunate by legacy considerations requiring the use of the dedrifting option for MSMAll.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Also, if we don't support NONE, these if statements should be unnecessary, as `read -a` should generate an empty array when there is no input, as desired.

cyang31:
> So to clarify, the problem with testing these myelin map changes to DeDriftAndResample is that the unedited fMRI section of the code errors with specific settings that don't match how it was previously run? This sounds like you should test (not using our main processing directory for the outputs, hopefully) using whatever FIX settings were previously used just to get through those steps, ignore the fMRI outputs (or check them without expecting them to have the latest cleaning), and only check the myelin maps from the outputs.
> 
> Matt seems to be suggesting a hacky workaround to force the tICA cleaned fMRI data through a pipeline it has never been through before, which seems overcomplicated for testing a feature unrelated to fMRI.

Yes, you are right, Tim. Matt and I had a meeting this afternoon and the current plan is to just test out this script with the simple replacement of `MyelinMap_BC.sh`, the part involving tICA cleanup is related to my MSMAll optimization project which will be addressed soon, but not in this PR.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
glasserm: It probably comes from positional arguments.  I would probably prefer to remove it from the launcher but not the back end code for now.

cyang31: Okay, so to make the users happy with either empty string or "NONE" for the arguments like this, we will enable the if check `if [ "${ARGUMENT}" = "NONE" ] ; then`, but in the launcher script and the default value in the newopt, we should specify empty string instead of "NONE". Am I correct about this?

coalsont: That is correct.

cyang31:
The test is passed without an error and the log is saved here `/media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipeline.sh.o757932`

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,616 @@
 # <!-- References -->
 # [HCP]: http://www.humanconnectome.org
 #
+pipedirguessed=0
```
coalsont, line 30: `set -eu` appears to be missing (it is usually just after `#!/bin/bash`), please add it and test again.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -27,1094 +27,616 @@
...
+log_Msg "NativeT1wFolder: ${NativeT1wFolder}"
+
+ResultsFolder="${AtlasFolder}/Results"
+log_Msg "ResultsFolder: ${ResultsFolder}"
+
+#Naming Conventions
+DownSampleFolderNames=""
+DownSampleT1wFolderNames=""
+for LowResMesh in ${LowResMeshes} ; do
+	DownSampleFolderNames=`echo "${DownSampleFolderNames}${AtlasFolder}/fsaverage_LR${LowResMesh}k "`
```
coalsont, line 231: Could consider simplifying this to `DownSampleFolderNames+="${AtlasFolder}/fsaverage_LR${LowResMesh}k "`, or switching to arrays.

coalsont:
I see some indentation mismatch at line 266, though it seems to have already existed before this PR.  The reapply section at line 563 switches to using spaces, but it is internally consistent, so that is tolerable.

It looks like both cases are because I forgot to switch my editor to tabs...

cyang31:
I tested the latest commit and it finished without error.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
coalsont: Do not do `variable+="${variable}otherstuff "`, that will grow exponentially.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -263,7 +263,7 @@ for Hemisphere in L R ; do
...
-    ${Caret7_Command} -metric-merge "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii -metric "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -column 1
+	${Caret7_Command} -metric-merge "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii -metric "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -column 1
     ${Caret7_Command} -metric-merge "${NativeFolder}"/"$Subject"."$Hemisphere".StrainR_${OutputRegName}.native.shape.gii -metric "${NativeFolder}"/"$Subject"."$Hemisphere".Strain_${OutputRegName}.native.shape.gii -column 2
     ${Caret7_Command} -metric-math "ln(var) / ln (2)" "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii -var var "${NativeFolder}"/"$Subject"."$Hemisphere".StrainJ_${OutputRegName}.native.shape.gii
     ${Caret7_Command} -metric-math "ln(var) / ln (2)" "${NativeFolder}"/"$Subject"."$Hemisphere".StrainR_${OutputRegName}.native.shape.gii -var var "${NativeFolder}"/"$Subject"."$Hemisphere".StrainR_${OutputRegName}.native.shape.gii
```
coalsont, line 269: Please fix the indentation on these nearby lines, too.

cyang31: done!

coalsont:
This looks reasonable to me.

cyang31:
@glasserm Matt, any other comment on this PR? Otherwise I think this one can be merged.

glasserm (approved):

PR #253, opened 2023-03-15
fix np.int deprecation issue

cyang31:
## Background
Since numpy 1.20.0, ["np.int" is no longer supported](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations). Instead, the python "int" can be used to replace "np.int" without changing the behavior. A user reported this [issue](https://groups.google.com/a/humanconnectome.org/g/hcp-users/c/7ewa-Wxp-kM) in the HCP users group.

## Main changes
- replace "np.int" with "int" in `global/scripts/average_bvecs.py`

## Evaluation
The user has tested this new setting and fixed the bug.

### DIFF
--- a/global/scripts/average_bvecs.py
+++ b/global/scripts/average_bvecs.py
@@ -177,8 +177,8 @@ def extract_overlaps(bvals1, bvecs1, bvals2, bvecs2, overlap1, overlap2, only_co
         log('Data set 1 session {0} offset: {1}'.format(session + 1, offset1))
         log('Data set 2 session {0} offset: {1}'.format(session + 1, offset2))
 
-        bv1_indices = np.arange(offset1, offset1 + overlaps, dtype=np.int)
-        bv2_indices = np.arange(offset2, offset2 + overlaps, dtype=np.int)
+        bv1_indices = np.arange(offset1, offset1 + overlaps, dtype=int)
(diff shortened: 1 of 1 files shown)


cyang31:
I try to run a simple test

cyang31:
@glasserm Matt, I didn't find the diffusion directory in myelin that has example data including the arguments in the call [line#117 of eddy_postproc.sh](https://github.com/Washington-University/HCPpipelines/blob/master/DiffusionPreprocessing/scripts/eddy_postproc.sh#L117)

`${globalscriptsdir}/average_bvecs.py ${eddydir}/Pos.bval ${eddydir}/Pos_rotated.bvec ${eddydir}/Neg.bval ${eddydir}/Neg_rotated.bvec ${datadir}/avg_data ${CombineDataFlag} ${eddydir}/Pos_SeriesVolNum.txt ${eddydir}/Neg_SeriesVolNum.txt`

Any idea how I can do a simple test for this PR?

glasserm:
I don't think we are set up to easily test the diffusion pipeline.  If you did a simple find and replace and verified that the replacements were all correct, I think this could just be merged since it fixed the user's issue.

glasserm (approved):

coalsont:
I think numpy's habit of turning deprecations into errors is misguided...

PR #256, opened 2023-03-23
fix the high pass argument

cyang31:
## Background
The `Examples/Scripts/DeDriftAndResamplePipelineBatch.sh` doesn't set the correct argument of high pass value, and as a result the `DeDriftAndResample/DeDriftAndResamplePipeline.sh` was wrongly updated about this argument.

## Main change
replace the `--highpass` with `--high-pass` to stay consistency of the same argument in the other scripts, e.g. ICAFix, MSMAll.

### DIFF
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -51,7 +51,7 @@ opts_AddMandatory '--low-res-meshes' 'LowResMeshes' 'meshnum@meshnum@...' "low r
 opts_AddMandatory '--registration-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMAll_InitalReg'"
 opts_AddMandatory '--maps' 'Maps' 'non@myelin@maps' "@-delimited map name strings corresponding to maps that are not myelin maps, e.g. 'sulc@curvature@corrThickness@thickness'"
 opts_AddMandatory '--smoothing-fwhm' 'SmoothingFWHM' 'number' "Smoothing FWHM that matches what was used in the fMRISurface pipeline"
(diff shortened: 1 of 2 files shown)


DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -51,7 +51,7 @@ opts_AddMandatory '--low-res-meshes' 'LowResMeshes' 'meshnum@meshnum@...' "low r
 opts_AddMandatory '--registration-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files, e.g. 'MSMAll_InitalReg'"
 opts_AddMandatory '--maps' 'Maps' 'non@myelin@maps' "@-delimited map name strings corresponding to maps that are not myelin maps, e.g. 'sulc@curvature@corrThickness@thickness'"
 opts_AddMandatory '--smoothing-fwhm' 'SmoothingFWHM' 'number' "Smoothing FWHM that matches what was used in the fMRISurface pipeline"
-opts_AddMandatory '--highpass' 'HighPass' 'integer' 'the high pass value that was used when running FIX' '--melodic-high-pass'
+opts_AddMandatory '--high-pass' 'HighPass' 'integer' 'the high pass value that was used when running FIX' '--melodic-high-pass'
```
coalsont, line 54: The old parsing code allowed both `--high-pass` and `--highpass`, please add that to the list of alternate option names (like `--melodic-high-pass` already is).

cyang31: Do you mean like `opts_AddMandatory '--high-pass' 'HighPass' 'integer' 'the high pass value that was used when running FIX' '--melodic-high-pass' '--highpass'`?

coalsont: Yes.

PR #257, opened 2023-03-24
fix DeDriftAndResample and MyelinMap_BC module

cyang31:
## Background
- The example script was not updated with the new mandatory argument "--msm-all-templates".
- MyelinMap_BC is only applied to MyelinMap and is not applied to a given map name, e.g. 'SmoothedMyelinMap'.

## Main change
- Add the template argument and the variable to the example script `Examples/Scripts/DeDriftAndResamplePipelineBatch.sh`.
- Fix MyelinMap_BC and the reference in `DeDriftAndResamplePipeline.sh`

## Evaluation
/media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipeline.sh.o3003663
/media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipeline.sh.e3003663
without error

### DIFF
--- a/DeDriftAndResample/DeDriftAndResamplePipeline.sh
+++ b/DeDriftAndResample/DeDriftAndResamplePipeline.sh
@@ -123,8 +123,10 @@ log_Msg "After delimiter substitution, Maps: ${Maps}"
 #these elses result in empty when given the empty string, make NONE do the same
 if [[ "${MyelinMaps}" == "NONE" ]] ; then
 	MyelinMaps=""
+	MyelinMapsArray=()
 else
 	MyelinMaps=`echo "$MyelinMaps" | sed s/"@"/" "/g`
+	IFS=' ' read -a MyelinMapsArray <<< "${MyelinMaps}"
 fi
 log_Msg "After delimiter substitution, MyelinMaps: ${MyelinMaps}"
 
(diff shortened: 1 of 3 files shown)


global/scripts/MyelinMap_BC.sh
```diff
@@ -156,13 +157,19 @@ ${Caret7_Command} -cifti-resample ${LowResBiasField} \
 log_Msg "Resampled BiasField in the native mesh space: ${NativeBiasField}"
 
 # generate bias corrected map in the output mesh space
-NativeBCMap=${NativeFolder}/${Subject}.MyelinMap_BC${RegNameInOutputName}.native.dscalar.nii
+NativeBCMapToUse=${NativeFolder}/${Subject}.MyelinMap_BC${RegNameInOutputName}.native.dscalar.nii
+NativeMyelinMapToUse=${NativeMyelinMap}
+if [[ "$MapName" != "MyelinMap" ]]; then
```
coalsont, line 162: It looks like when MapName is MyelinMap, the lines inside this `if` would result in the same filenames, so this conditional should be unnecessary.  Please try to simplify the logic and remove any variables that aren't needed.

cyang31: fixed

coalsont:
It appears that it uses a hardcoded "MyelinMap" and the template to create the bias field, and then applies that bias field to whatever `--map` was set to.  For the smoothed myelin map, this will give a slightly different answer than just smoothing the corrected myelin map, so it isn't clear to me whether that is the desired logic.

Is `--map` intended for anything other than the smoothed myelin map?

coalsont:
This also suggests that the intended usage is to call the myelin map BC script twice when both normal and smoothed versions are desired, computing the bias field twice.  I would think the more obvious solution is to have a second optional smoothed output file, or to call a smoothing command in the outer script after this script completes.

cyang31:
Here's the original version without the replacement using `MyelinMap_BC.sh`, https://github.com/Washington-University/HCPpipelines/blob/9da3a70a661fa4ee500154897603802c7e2e2183/DeDriftAndResample/DeDriftAndResamplePipeline.sh#L867-L883

It appears that in the original version, the smoothed myelin map can be given by the input argument `--myelin-maps`, and the smoothing is not applied directly to the MyelinMap_BC_${RegName}.native.dscalar.nii. So the current PR replicates the same behavior of the original code block, even if the MyelinMap_BC module is called twice. 

I agree that we can optimize the current nested loop by removing the part `for MapMap in ${MyelinMaps} ; do`, checking if `SmoothedMyelinMap` is specified, then run the smoothing command `					${Caret7_Command} -cifti-math "Var - Bias" ${NativeFolder}/${Subject}.${Map}_BC_${OutputRegName}.native.dscalar.nii -var Var ${NativeFolder}/${Subject}.${Map}.native.dscalar.nii -var Bias ${NativeFolder}/${Subject}.BiasField_${OutputRegName}.native.dscalar.nii
`.

cyang31:
However, I think, even if the bias field is computed twice in the current PR and the original version, it can handle the situation nicely when 
- `--myelin-maps="MyelinMap SmoothedMyelinMap"` both types of myelin maps are specified
- `--myelin-maps="SmoothedMyelinMap"` or `--myelin-maps="MyelinMap"` only one type of myelin maps is specified

Otherwise, we might want to design a logic to check which types of myelin maps are specified and implement a if-elseif-elseif-else code block to handle all the situations, while checking if the bias field exists.

coalsont:
I think we want it to recompute the bias field when it is called, regardless, to avoid using a result from a previous group template.  It seems like the logic should actually be "always generate MyelinMap_BC, optionally generate SmoothedMyelinMap_BC", and while it isn't clear that there will ever be another myelin map name, if it should specifically apply the bias as-is to another file (rather than smoothing MyelinMap_BC), my instinct is to handle this with an `--additional-myelin-maps="SmoothedMyelinMap"` argument.

cyang31:
I see. This is actually a good idea, but it also requires to add a logic to determine the following parts that `MyelinMap_BC.sh` need.
- what is the main myelin map, we can default it to be `MyelinMap` or the first element in the `--myelin-maps` (in case a user only specify `--myelin-maps="SmoothedMyelinMap"`)
- what is/are the additional myelin maps that should be fed into `--additional-myelin-maps`.

Actually, I have pushed a new commit that doesn't add a new argument to the `MyelinMap_BC` and handle the several situations in the pipeline script, meanwhile, limit the computation of bias field into one time when several myelin maps are given.

I think both strategies are reasonable.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -411,67 +413,72 @@ for Mesh in ${LowResMeshes} ${HighResMesh} ; do
...
-
-		for MapMap in ${MyelinMaps} ; do
+		
+		for ((index = 0; index < ${#MyelinMapsArray[@]}; ++index))
+		do
+			MapMap="${MyelinMapsArray[index]}"
 			log_Msg "MapMap: ${MapMap}"
-
+			
 			if [ ${MapMap} = ${Map} ] ; then
```
coalsont, line 422: For the record, this confusing structure for "only run this part on the myelin maps" should really be a loop over only myelin maps, above the `for Map in ...` loop (and the "run this part on only the first myelin map" shouldn't be inside the myelin map loop either), populating a new list of `_BC` map names to be used in the `for Map in ...` loop (see the modification of `$Map` at the end of the myelin map section).  The loops with `MapMap` at the end of this same `for Map in ...` loop should also be moved after the outer loop.

However, that should probably be a separate PR after we have the errors fixed.

cyang31: I agree that this legacy part is confusing. The `for MapMap in ${MyelinMaps} ; do` loop should be changed to `if [[ " ${MyelinMapsArray[*]} " =~ " ${Map} " ]]; then`, to only apply the bias field correct on myelin map when `${Map}` is an element in the array `${MyelinMapsArray}`.

coalsont: Also, I would not use `=~` to resolve this, trusting variables to not contain regexp-special characters is not a great idea.

cyang31: Okay, I see. So you are suggesting to divide the loop into two sequential loops, one for the myelin maps list (`${MyelinMaps}`), the following one for the rest of maps. This should make the code more readable.

coalsont: Correction: the myelin map BC script shouldn't be called if the list of myelin maps is empty, so that is a conditional it should be inside.  Other than that, it should only be inside the loop over resolutions, not inside any additional loop.

Given that the myelin map list is already expected to contain the value 'MyelinMap' if it contains it at all, the myelin map BC script should probably take that entire list as-is if it handles them internally (which would be my suggestion, to avoid needing a myelin-specific loop in DeDrift...).

coalsont: Yes, specifically, a loop to do the myelin-only steps first, and then a loop over the other maps AND the _BC myelin maps, to do the resampling that is common to all of them.

cyang31: The latest commit should address the comments. I would rather not add the `addtional-myelin-map` part into the `MyelinMap_BC.sh`.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -412,74 +412,77 @@ for Mesh in ${LowResMeshes} ${HighResMesh} ; do
...
+	MyelinMapsToUse=""
+	# myelin map only loop
+	for MyelinMap in ${MyelinMaps} ; do
+		if [ "$BiasFieldComputed" = false ]; then
+			# ----- Begin moved statements -----
+
+			# Recompute Myelin Map Bias Field Based on Better Registration
+			log_Msg "Recompute Myelin Map Bias Field Based on Better Registration"
+
+			cifti_in=${NativeFolder}/${Subject}.MyelinMap.native.dscalar.nii
```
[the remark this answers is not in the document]
cyang31: fixed

global/scripts/MyelinMap_BC.sh
```diff
@@ -94,19 +94,38 @@ log_Msg "RegNameInOutputName: $RegNameInOutputName"
...
+LowResCiftiTemplate=${LowResFolder}/${Subject}.MyelinMap.${LowResMeshString}.dscalar.nii
+log_File_Must_Exist "${LowResCiftiTemplate}" # 2 
+LeftSphereCurrentSphere=${NativeFolder}/${Subject}.L.sphere${RegNameStructString}.native.surf.gii
+log_File_Must_Exist "${LeftSphereCurrentSphere}" # 3
+LeftSphereNewSphere=${LowResFolder}/${Subject}.L.sphere.${LowResMeshString}.surf.gii
+log_File_Must_Exist "${LeftSphereNewSphere}" # 4
+LeftAreaSurfCurrentArea="${NativeT1wFolder}/${Subject}.L.midthickness.native.surf.gii"
+log_File_Must_Exist "${LeftAreaSurfCurrentArea}" # 5
+LeftAreaSurfNewArea=${LowResFolder}/${Subject}.L.midthickness${RegNameInOutputName}.${LowResMeshString}.surf.gii
+log_File_Must_Exist "${LeftAreaSurfNewArea}" # 6 - This is the one that doesn't exist
```
coalsont, line 109: The "doesn't exist" part of the comment is ancient cruft that should be deleted.

DeDriftAndResample/DeDriftAndResamplePipeline.sh
```diff
@@ -408,75 +410,46 @@ for Mesh in ${LowResMeshes} ${HighResMesh} ; do
...
+		else
+			# bias field in native space is already generated
+			# BC the other types of given myelin maps
+			${Caret7_Command} -cifti-math "Var - Bias" ${NativeFolder}/${Subject}.${MyelinMap}_BC_${OutputRegName}.native.dscalar.nii -var Var ${NativeFolder}/${Subject}.${MyelinMap}.native.dscalar.nii -var Bias ${NativeFolder}/${Subject}.BiasField_${OutputRegName}.native.dscalar.nii
+		fi
+		MyelinMapsToUse+="${MyelinMap}_BC "
+	done
+	
+	log_Debug_Msg "Point 2.0"
+	if [ "$BiasFieldComputed" = false ]; then
```
coalsont, line 446: This can only happen if `MyelinMaps` is empty, which already matches the initialization of `MyelinMapsToUse`, so this isn't needed.

global/scripts/MyelinMap_BC.sh
coalsont: Also, I don't think the numbers serve any purpose, so I would just delete the comments from all of these.

coalsont:
Looks good, but I need to add an edit to deal with QuNex using an older bash that doesn't like empty arrays.  You'll need to pull my edit to make further edits on this branch.  Is it okay to do that now?

cyang31:
I can do it now.

cyang31:
pulled

coalsont:
If you have tested it with my edit, we should probably merge it so Jure can use it more easily.

cyang31:
testing logs: 
- /media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipeline.sh.o3003663
- /media/myelin/alex/MSMAll_optimization/DeDriftAndResamplePipeline.sh.e3003663
without error

PR #258, opened 2023-03-26
fix typo in MyelinMap_BC module

cyang31:
The area files should be loaded from low resolution T1w folder instead of the low resolution folder in atlas space folder.

Now they are fixed.


coalsont:
Looks reasonable.

coalsont:
Rebase manages to ignore the duplicate edit, so I am merging this to give Jure this fix sooner.

PR #259, opened 2023-03-26
replace the old _BC block with MyelinMap_BC

cyang31:
## Background
There remain several legacy _BC code block that is not replaced by the module `MyelinMap_BC.sh`. 

## Main changes
- replace the old _BC code block in `MSMAll.sh` with `MyelinMap_BC.sh`
- `CreateMyelinMaps.sh` used gifti based _BC which is now replaced by the cifti-based `MyelinMap_BC.sh`, while the gifti files are kept
- add arguments to `PostFreeSurferPipeline.sh` to adapt the module `MyelinMap_BC.sh`
- update the example launch script `PostFreeSurferPipelineBatch.sh`
- change the argument `--myelin-target-file` as a single argument with absolute file path that doesn't depend on `--msm-all-templates`

## Evaluation
- MSMAllPipeline logs with no error: 
- `/media/myelin/alex/MSMAll_optimization/MSMAllPipeline.sh.e49165`
- `/media/myelin/alex/MSMAll_optimization/MSMAllPipeline.sh.o49165`

- PostFreeSurferPipeline logs with no error: 
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e162416`
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o162416`

### DIFF
--- a/Examples/Scripts/PostFreeSurferPipelineBatch.sh
+++ b/Examples/Scripts/PostFreeSurferPipelineBatch.sh
@@ -87,7 +87,8 @@ for Subject in $Subjlist ; do
     FreeSurferLabels="${HCPPIPEDIR_Config}/FreeSurferAllLut.txt"
     ReferenceMyelinMaps="${HCPPIPEDIR_Templates}/standard_mesh_atlases/Conte69.MyelinMap_BC.164k_fs_LR.dscalar.nii"
     RegName="MSMSulc" #MSMSulc is recommended, if binary is not available use FS (FreeSurfer)
-
+	MSMAllTemplates="${HCPPIPEDIR}/global/templates/MSMAll"
+	UseIndMean="YES"
     if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
         echo "About to locally run ${HCPPIPEDIR}/PostFreeSurfer/PostFreeSurferPipeline.sh"
         queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
@@ -107,7 +108,9 @@ for Subject in $Subjlist ; do
         --subcortgraylabels="$SubcorticalGrayLabels" \
         --freesurferlabels="$FreeSurferLabels" \
         --refmyelinmaps="$ReferenceMyelinMaps" \
-        --regname="$RegName"
+        --regname="$RegName" \
+        --msm-all-templates="$MSMAllTemplates" \
+        --use-ind-mean="$UseIndMean"
 
     # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
 
@@ -121,7 +124,9 @@ for Subject in $Subjlist ; do
         --subcortgraylabels=$SubcorticalGrayLabels \
         --freesurferlabels=$FreeSurferLabels \
         --refmyelinmaps=$ReferenceMyelinMaps \
-        --regname=$RegName"
+        --regname=$RegName \
+        --msm-all-templates="$MSMAllTemplates" \
+        --use-ind-mean="$UseIndMean""
 
     echo ". ${EnvironmentScript}"
 done
--- a/Examples/Scripts/PostFreeSurferPipelineBatch.sh
+++ b/Examples/Scripts/PostFreeSurferPipelineBatch.sh
@@ -87,9 +87,9 @@ for Subject in $Subjlist ; do
     FreeSurferLabels="${HCPPIPEDIR_Config}/FreeSurferAllLut.txt"
     ReferenceMyelinMaps="${HCPPIPEDIR_Templates}/standard_mesh_atlases/Conte69.MyelinMap_BC.164k_fs_LR.dscalar.nii"
     RegName="MSMSulc" #MSMSulc is recommended, if binary is not available use FS (FreeSurfer)
-	MSMAllTemplates="${HCPPIPEDIR}/global/templates/MSMAll"
-	MyelinTargetFile="${MSMAllTemplates}/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii"
-	UseIndMean="YES"
+    MSMAllTemplates="${HCPPIPEDIR}/global/templates/MSMAll"
+    MyelinTargetFile="${MSMAllTemplates}/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii"
+    UseIndMean="YES"
     if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
         echo "About to locally run ${HCPPIPEDIR}/PostFreeSurfer/PostFreeSurferPipeline.sh"
         queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
@@ -110,7 +110,7 @@ for Subject in $Subjlist ; do
         --freesurferlabels="$FreeSurferLabels" \
         --refmyelinmaps="$ReferenceMyelinMaps" \
         --regname="$RegName" \
-		--myelin-target-file="$MyelinTargetFile" \
+        --myelin-target-file="$MyelinTargetFile" \
         --use-ind-mean="$UseIndMean"
 
     # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
--- a/Examples/Scripts/PostFreeSurferPipelineBatch.sh
+++ b/Examples/Scripts/PostFreeSurferPipelineBatch.sh
@@ -88,6 +88,7 @@ for Subject in $Subjlist ; do
     ReferenceMyelinMaps="${HCPPIPEDIR_Templates}/standard_mesh_atlases/Conte69.MyelinMap_BC.164k_fs_LR.dscalar.nii"
     RegName="MSMSulc" #MSMSulc is recommended, if binary is not available use FS (FreeSurfer)
 	MSMAllTemplates="${HCPPIPEDIR}/global/templates/MSMAll"
+	MyelinTargetFile="${MSMAllTemplates}/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii"
 	UseIndMean="YES"
     if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
         echo "About to locally run ${HCPPIPEDIR}/PostFreeSurfer/PostFreeSurferPipeline.sh"
@@ -109,7 +110,7 @@ for Subject in $Subjlist ; do
         --freesurferlabels="$FreeSurferLabels" \
         --refmyelinmaps="$ReferenceMyelinMaps" \
         --regname="$RegName" \
-        --msm-all-templates="$MSMAllTemplates" \
+		--myelin-target-file="$MyelinTargetFile" \
         --use-ind-mean="$UseIndMean"
 
     # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
@@ -125,7 +126,7 @@ for Subject in $Subjlist ; do
         --freesurferlabels=$FreeSurferLabels \
         --refmyelinmaps=$ReferenceMyelinMaps \
         --regname=$RegName \
-        --msm-all-templates="$MSMAllTemplates" \
+        --myelin-target-file="$MyelinTargetFile" \
         --use-ind-mean="$UseIndMean""
 
     echo ". ${EnvironmentScript}"
--- a/PostFreeSurfer/PostFreeSurferPipeline.sh
+++ b/PostFreeSurfer/PostFreeSurferPipeline.sh
@@ -226,7 +226,7 @@ if ((doProcessing)); then
     argList+=("$SubcorticalGrayLabels")     # ${21}
     argList+=("$RegName")                   # ${22}
     argList+=("$InflateExtraScale")         # ${23}
-    #"$PipelineScripts"/FreeSurfer2CaretConvertAndRegisterNonlinear.sh "${argList[@]}"
+    "$PipelineScripts"/FreeSurfer2CaretConvertAndRegisterNonlinear.sh "${argList[@]}"
 
     log_Msg "Create FreeSurfer ribbon file at full resolution"
 
@@ -238,7 +238,7 @@ if ((doProcessing)); then
     argList+=("$AtlasSpaceT1wImage")        # ${6}
     argList+=("$T1wRestoreImage")           # ${7}  Called T1wImage in CreateRibbon.sh
     argList+=("$FreeSurferLabels")          # ${8}
-    #"$PipelineScripts"/CreateRibbon.sh "${argList[@]}"
+    "$PipelineScripts"/CreateRibbon.sh "${argList[@]}"
 
     log_Msg "Myelin Mapping"
     log_Msg "RegName: ${RegName}"
--- a/PostFreeSurfer/PostFreeSurferPipeline.sh
+++ b/PostFreeSurfer/PostFreeSurferPipeline.sh
@@ -63,12 +63,15 @@ opts_AddMandatory '--lowresmesh' 'LowResMeshes' 'number' "usually '32', the stan
 opts_AddMandatory '--subcortgraylabels' 'SubcorticalGrayLabels' 'file' "location of FreeSurferSubcorticalLabelTableLut.txt"
 opts_AddMandatory '--freesurferlabels' 'FreeSurferLabels' 'file' "location of FreeSurferAllLut.txt"
 opts_AddMandatory '--refmyelinmaps' 'ReferenceMyelinMaps' 'file' "high-resolution group myelin map to use for bias correction"
+opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'path' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
 
 opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, default '$defaultSigma'" "$defaultSigma"
 opts_AddOptional '--regname' 'RegName' 'name' "surface registration to use, default 'MSMSulc'" 'MSMSulc'
 opts_AddOptional '--inflatescale' 'InflateExtraScale' 'number' "surface inflation scaling factor to deal with different resolutions, default '1'" '1'
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData|LegacyStyleData' "disable some HCP preprocessing requirements to allow processing of data that doesn't meet HCP acquisition guidelines - don't use this if you don't need to" 'HCPStyleData'
 opts_AddOptional '--structural-qc' 'QCMode' 'yes|no|only' "whether to run structural QC, default 'yes'" 'yes'
+opts_AddOptional '--myelin-target-file' 'MyelinTarget' 'string' "alternate myelin map target, relative to the --msm-all-templates folder" 'Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'
+opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'YES'" 'YES'
 
 opts_ParseArguments "$@"
 
@@ -223,7 +226,7 @@ if ((doProcessing)); then
     argList+=("$SubcorticalGrayLabels")     # ${21}
     argList+=("$RegName")                   # ${22}
     argList+=("$InflateExtraScale")         # ${23}
-    "$PipelineScripts"/FreeSurfer2CaretConvertAndRegisterNonlinear.sh "${argList[@]}"
+    #"$PipelineScripts"/FreeSurfer2CaretConvertAndRegisterNonlinear.sh "${argList[@]}"
 
     log_Msg "Create FreeSurfer ribbon file at full resolution"
 
@@ -235,7 +238,7 @@ if ((doProcessing)); then
     argList+=("$AtlasSpaceT1wImage")        # ${6}
     argList+=("$T1wRestoreImage")           # ${7}  Called T1wImage in CreateRibbon.sh
     argList+=("$FreeSurferLabels")          # ${8}
-    "$PipelineScripts"/CreateRibbon.sh "${argList[@]}"
+    #"$PipelineScripts"/CreateRibbon.sh "${argList[@]}"
 
     log_Msg "Myelin Mapping"
     log_Msg "RegName: ${RegName}"
@@ -279,6 +282,9 @@ if ((doProcessing)); then
     argList+=("$ReferenceMyelinMaps")
     argList+=("$CorrectionSigma")
     argList+=("$RegName")                                  # ${39}
+    argList+=("$MSMAllTemplates")
+    argList+=("$MyelinTarget")
+    argList+=("$UseIndMean")
     "$PipelineScripts"/CreateMyelinMaps.sh "${argList[@]}"
 fi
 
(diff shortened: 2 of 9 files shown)


global/scripts/MyelinMap_BC.sh
```diff
@@ -105,15 +105,15 @@ LeftSphereNewSphere=${LowResFolder}/${Subject}.L.sphere.${LowResMeshString}.surf
 log_File_Must_Exist "${LeftSphereNewSphere}"
 LeftAreaSurfCurrentArea="${NativeT1wFolder}/${Subject}.L.midthickness.native.surf.gii"
 log_File_Must_Exist "${LeftAreaSurfCurrentArea}"
-LeftAreaSurfNewArea=${LowResFolder}/${Subject}.L.midthickness${RegNameInOutputName}.${LowResMeshString}.surf.gii
+LeftAreaSurfNewArea=${LowResT1wFolder}/${Subject}.L.midthickness${RegNameInOutputName}.${LowResMeshString}.surf.gii
```
coalsont, line 108: Don't make the same edits on two different branches, it causes issues with merging.  I think we should probably close the other PR that does only this, assuming this PR looks reasonable.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -214,12 +217,6 @@ for Hemisphere in L R ; do
       ${CARET7DIR}/wb_command -metric-mask "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere"."$Map"_s"$CorrectionSigma".native.func.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere"."$Map"_s"$CorrectionSigma".native.func.gii
```
coalsont, line 217: We probably don't need to continue to generate these `_s"$CorrectionSigma"` files, since they are no longer involved in generating the _BC files.  However, maybe the _BC script should do a -cifti-separate at the end to generate the expected gifti versions of the _BC files.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -214,12 +217,6 @@ for Hemisphere in L R ; do
...
-    ${CARET7DIR}/wb_command -metric-math "(Individual - Bias) * Mask" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap_BC.native.func.gii -var Individual "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap.native.func.gii -var Bias "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".BiasField.native.func.gii -var Mask "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii
-    ${CARET7DIR}/wb_command -metric-math "(Individual - Bias) * Mask" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".SmoothedMyelinMap_BC.native.func.gii -var Individual "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".SmoothedMyelinMap.native.func.gii -var Bias "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".BiasField.native.func.gii -var Mask "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii
     rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap_s"$CorrectionSigma".native.func.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap_s"$CorrectionSigma".native.func.gii
```
coalsont, line 220: In fact, if you just trace the `rm`s like this backwards through the code, you may find that many of these commands will now do nothing other than waste time generating files that are then deleted, please try to clean this up and see if there is anything left of this "T2 present" conditional.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
cyang31: Since the gifti output is not necessary every time this module is called, I would choose to add another optional argument `--gifti-output=True` to control this behavior.

coalsont: We may want this to be a deliberate design decision, where we either always generate the gifti versions, or never, modifying the behavior of the existing code.  @glasserm ?

glasserm: Well, I have thought about it, but ended up thinking we should maintain existing behavior for now.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -205,19 +205,6 @@ for Hemisphere in L R ; do
...
     rm "$T1wFolder"/temp_ribbon.nii.gz
     ${CARET7DIR}/wb_command -metric-smoothing "$T1wFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".midthickness.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap.native.func.gii "$SurfaceSmoothingSigma" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".SmoothedMyelinMap.native.func.gii -roi "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii  
-    ${CARET7DIR}/wb_command -metric-resample "$AtlasSpaceFolder"/"$Subject"."$Hemisphere".RefMyelinMap."$HighResMesh"k_fs_LR.func.gii "$AtlasSpaceFolder"/"$Subject"."$Hemisphere".sphere."$HighResMesh"k_fs_LR.surf.gii ${RegSphere} ADAP_BARY_AREA "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap.native.func.gii -area-surfs "$AtlasSpaceFolder"/"$Subject"."$Hemisphere".midthickness."$HighResMesh"k_fs_LR.surf.gii "$T1wFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".midthickness.native.surf.gii -current-roi "$AtlasSpaceFolder"/"$Subject"."$Hemisphere".atlasroi."$HighResMesh"k_fs_LR.shape.gii
```
cyang31, line 205: The commands related with `RefMyelinMap` are removed.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
cyang31: So we should NOT add optional argument `--gifti-output=True` and the gifti generation is a separated step outside the module `MyelinMap_BC.sh`?

glasserm: my thought had been that this module would only do the stuff we want done across all usecases and that the surrounding code would be individualized per each use case.

global/scripts/MyelinMap_BC.sh
```diff
@@ -185,6 +187,9 @@ ${Caret7_Command} -cifti-math "Var - Bias" ${NativeBCMapToUse} \
 	-var Bias ${NativeBiasField}
 
 log_Msg "_BC Myelin map in the native mesh space: ${NativeBCMapToUse}"
-# TODO: add gifti generation according to one argument
-# -cifti-separate-all
+# gifti generation in native mesh space (BiasField and _BC files)
+if ((GiftiOutputBool)); then
+	${Caret7_Command} -cifti-separate-all ${NativeBiasField} -left ${NativeFolder}/${Subject}.L.BiasField${RegNameInOutputName}.native.dscalar.nii -right ${NativeFolder}/${Subject}.R.BiasField${RegNameInOutputName}.native.dscalar.nii
```
cyang31, line 192: the file extensions are not correct.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -223,7 +226,7 @@ if ((doProcessing)); then
     argList+=("$SubcorticalGrayLabels")     # ${21}
     argList+=("$RegName")                   # ${22}
     argList+=("$InflateExtraScale")         # ${23}
-    "$PipelineScripts"/FreeSurfer2CaretConvertAndRegisterNonlinear.sh "${argList[@]}"
+    #"$PipelineScripts"/FreeSurfer2CaretConvertAndRegisterNonlinear.sh "${argList[@]}"
```
coalsont, line 229: Presumably these comments were to speed up testing, and shouldn't have been committed.

cyang31: oh no...Thanks for pointing it out! Otherwise it will definitely become a disaster...

global/scripts/MyelinMap_BC.sh
coalsont: This (and its option) should probably be removed.

cyang31: apparently

coalsont:
Should we also edit the _1res version while we are at it?

cyang31:
I didn't find lines in `CreateMyelinMaps_1res.sh` that are performing the same operations in _BC module. It seems that the gifti files to construct cifti ${MyelinMap}_BC are already generated elsewhere. Maybe @glasserm can confirm it?

coalsont:
I said that before I looked at it, yes it seems _1res it just resamples the existing native mesh maps, which makes sense.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -202,25 +205,6 @@ for Hemisphere in L R ; do
...
-    ${CARET7DIR}/wb_command -metric-math "(Individual - Bias) * Mask" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap_BC.native.func.gii -var Individual "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap.native.func.gii -var Bias "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".BiasField.native.func.gii -var Mask "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii
-    ${CARET7DIR}/wb_command -metric-math "(Individual - Bias) * Mask" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".SmoothedMyelinMap_BC.native.func.gii -var Individual "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".SmoothedMyelinMap.native.func.gii -var Bias "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".BiasField.native.func.gii -var Mask "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".roi.native.shape.gii
-    rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".MyelinMap_s"$CorrectionSigma".native.func.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".RefMyelinMap_s"$CorrectionSigma".native.func.gii
   fi
 
   for STRING in $MapListFunc ; do
```
coalsont, line 210: If there are multiple low res meshes, this will probably fail when it gets to `Myelin` on a mesh that isn't the first one.  You may need to split this loop (and add a special myelin-only native cifti generation step) so that the _BC code is done in the same order compared to other steps as the old version.

cyang31: Tim, I am confusing about this. Are you indicating that this `for STRING in $MapListFunc ;` loop will fail when `"${T2wPresent}" != "YES"` and there are multiple low res meshes?

coalsont: I thought the loop at 217 was doing something other than it is, never mind that part.  However, it looks like _BC will only get calculated for the first low-res mesh because there is no loop over `LowResMeshes` after _BC gets run except the one adding things to spec files, so with `LowResMeshes=32@59`, the 59k spec file will then contain a _BC dscalar filename for 59k that doesn't exist.

cyang31: I see. So currently the MyelinMap_BC and SmoothedMyelinMap_BC cifti files in the other low res mesh space are not generated. There are two ways to keep this part, 1) follow the previous setting but in a separate loop, namely, generate gifti files with command `-metric-resample` and `-metric-mask` and then create cifti files, 2) resample the MyelinMap_BC and SmoothedMyelinMap_BC cifti files in native mesh space to the other low res mesh space. The main difference between these two choices is whether we want to keep the gifti files for those low res mesh spaces.

coalsont: I think Matt wants all the output files that were previously generated.  You could resample in cifti and then -cifti-separate, though.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -63,12 +63,15 @@ opts_AddMandatory '--lowresmesh' 'LowResMeshes' 'number' "usually '32', the stan
 opts_AddMandatory '--subcortgraylabels' 'SubcorticalGrayLabels' 'file' "location of FreeSurferSubcorticalLabelTableLut.txt"
 opts_AddMandatory '--freesurferlabels' 'FreeSurferLabels' 'file' "location of FreeSurferAllLut.txt"
 opts_AddMandatory '--refmyelinmaps' 'ReferenceMyelinMaps' 'file' "high-resolution group myelin map to use for bias correction"
+opts_AddMandatory '--msm-all-templates' 'MSMAllTemplates' 'path' "path to directory containing MSM All template files, e.g. 'YourFolder/global/templates/MSMAll'"
```
glasserm, line 66: This is an odd variable name for PostFreeSurfer that precedes MSMAll by quite a few lines.  I am not a huge fan of tying the myelin target file to a specific folder either.  I would have preferred a full path and a single argument.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
glasserm: Yes we don't want to change the outputs of the pipelines at this time.

glasserm:
I didn't write the 1res code.  That is a patch for if someone does not want to rerun PostFreeSurfer but wants to add a new lowres mesh resolution.

PostFreeSurfer/PostFreeSurferPipeline.sh
cyang31: I changed the myelin target as a single argument with an absolute path wherever the MyelinMap_BC is involved.

Examples/Scripts/PostFreeSurferPipelineBatch.sh
```diff
@@ -109,7 +110,7 @@ for Subject in $Subjlist ; do
         --freesurferlabels="$FreeSurferLabels" \
         --refmyelinmaps="$ReferenceMyelinMaps" \
         --regname="$RegName" \
-        --msm-all-templates="$MSMAllTemplates" \
+		--myelin-target-file="$MyelinTargetFile" \
```
coalsont, line 113: This looks like it changed indentation type.

Examples/Scripts/PostFreeSurferPipelineBatch.sh
```diff
@@ -87,7 +87,9 @@ for Subject in $Subjlist ; do
     FreeSurferLabels="${HCPPIPEDIR_Config}/FreeSurferAllLut.txt"
     ReferenceMyelinMaps="${HCPPIPEDIR_Templates}/standard_mesh_atlases/Conte69.MyelinMap_BC.164k_fs_LR.dscalar.nii"
     RegName="MSMSulc" #MSMSulc is recommended, if binary is not available use FS (FreeSurfer)
-
+	MSMAllTemplates="${HCPPIPEDIR}/global/templates/MSMAll"
```
coalsont, line 90: Check indentation type here.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -63,12 +63,14 @@ opts_AddMandatory '--lowresmesh' 'LowResMeshes' 'number' "usually '32', the stan
...
+opts_AddMandatory '--myelin-target-file' 'MyelinTarget' 'string' "myelin map target file, absolute folder, e.g. 'YourFolder/global/templates/MSMAll/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii'"
 
 opts_AddOptional '--mcsigma' 'CorrectionSigma' 'number' "myelin map bias correction sigma, default '$defaultSigma'" "$defaultSigma"
 opts_AddOptional '--regname' 'RegName' 'name' "surface registration to use, default 'MSMSulc'" 'MSMSulc'
 opts_AddOptional '--inflatescale' 'InflateExtraScale' 'number' "surface inflation scaling factor to deal with different resolutions, default '1'" '1'
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData|LegacyStyleData' "disable some HCP preprocessing requirements to allow processing of data that doesn't meet HCP acquisition guidelines - don't use this if you don't need to" 'HCPStyleData'
 opts_AddOptional '--structural-qc' 'QCMode' 'yes|no|only' "whether to run structural QC, default 'yes'" 'yes'
+opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean , defaults to 'YES'" 'YES'
```
coalsont, line 73: Unnecessary space before comma.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -257,6 +241,57 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
   done
 done
 
+BiasFieldComputed=false
+#Reduce memory usage by smoothing on downsampled mesh (match the gifti version by using the first lowresmesh)
+LowResMesh=`echo ${LowResMeshes} | cut -d " " -f 1`
```
coalsont, line 246: `LowResMesh="${LowResMeshArray[0]}"`

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -257,6 +241,57 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
+	else
+		# bias field in native space is already generated
+		# BC the other types of given myelin maps
+		${CARET7DIR}/wb_command -cifti-math "Var - Bias" ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}_BC.native.dscalar.nii -var Var ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}.native.dscalar.nii -var Bias ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.BiasField.native.dscalar.nii
+	fi
+	${CARET7DIR}/wb_command -cifti-separate-all ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}_BC.native.dscalar.nii -left ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.${MyelinMap}_BC.native.func.gii -right ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.${MyelinMap}_BC.native.func.gii
+done
+
+# create cifti and gift MyelinMap in the low res mesh spaces
+for ((index = 0; index < ${#LowResMeshesArray[@]}; ++index)) ; do
```
coalsont, line 276: This could be `for LowResMesh in "${LowResMeshArray[@]}"`.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -114,6 +116,7 @@ SurfaceSmoothingFWHM="4"
 MyelinMappingSigma=`echo "$MyelinMappingFWHM / ( 2 * ( sqrt ( 2 * l ( 2 ) ) ) )" | bc -l`
 SurfaceSmoothingSigma=`echo "$SurfaceSmoothingFWHM / ( 2 * ( sqrt ( 2 * l ( 2 ) ) ) )" | bc -l`
 
+IFS='@' read -a LowResMeshesArray <<< "${LowResMeshes}"
```
coalsont, line 119: I would use `IFS=' @'` to accept spaces also (which the string-based version technically would accept), and try to remove direct use of `LowResMeshes` below, including the `sed` line, as using two different variables for the same list looks sloppy.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -257,6 +241,57 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
+			${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.${MyelinMap}_BC.${LowResMesh}k_fs_LR.dscalar.nii \
+			-surface-postdilate 40 \
+			-left-spheres ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.sphere.MSMSulc.native.surf.gii ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.L.sphere.${LowResMesh}k_fs_LR.surf.gii \
+			-left-area-surfs ${StudyFolder}/${Subject}/T1w/${NativeFolder}/${Subject}.L.midthickness.native.surf.gii ${StudyFolder}/${Subject}/T1w/fsaverage_LR${LowResMesh}k/${Subject}.L.midthickness.${LowResMesh}k_fs_LR.surf.gii \
+			-right-spheres ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.sphere.MSMSulc.native.surf.gii ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.R.sphere.${LowResMesh}k_fs_LR.surf.gii \
+			-right-area-surfs ${StudyFolder}/${Subject}/T1w/${NativeFolder}/${Subject}.R.midthickness.native.surf.gii ${StudyFolder}/${Subject}/T1w/fsaverage_LR${LowResMesh}k/${Subject}.R.midthickness.${LowResMesh}k_fs_LR.surf.gii
+		# gifti files
+		${CARET7DIR}/wb_command -cifti-separate-all ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.${MyelinMap}_BC.${LowResMesh}k_fs_LR.dscalar.nii \
```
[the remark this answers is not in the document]
cyang31: I checked that `-cifti-separate` doesn't seem to have simple argument `-left` or `-right` that can be easily used for the same purpose... Probably the same functionality can be achieved by `${CARET7DIR}/wb_command -cifti-separate ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.${MyelinMap}_BC.${LowResMesh}k_fs_LR.dscalar.nii COLUMN -metric CORTEX_LEFT ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.L.${MyelinMap}_BC.${LowResMesh}k_fs_LR.func.gii -metric CORTEX_RIGHT ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.R.${MyelinMap}_BC.${LowResMesh}k_fs_LR.func.gii`?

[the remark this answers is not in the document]
coalsont: Correct.

cyang31:
tested without error: `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e2761221` and `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o2761221`

PostFreeSurfer/PostFreeSurferPipeline.sh
coalsont: You fixed this in MyelinMap_BC.sh's help, but not the help of this pipeline.

cyang31: fixed

coalsont:
Looks reasonable to me, and I think @glasserm's comment was addressed.

PR #261, opened 2023-04-11
fix #260

### LINKED ISSUE #260: Bugs in PR 259
Looks like PR #259 got merged before I had a chance to look at the outputs.  I see several issues:

1) It seems like we are specifying two kinds of reference myelin maps to the input arguments.  
2) I am hesitant to change the input arguments as dramatically as was done including having "MSMAll" appear in variable names long before the MSMAll Pipeline.  It would be better to stick with the original variable name now accepting a 32k mesh (perhaps it should still also be capable of accepting a 164k mesh as it did before?).  
3) It looks like, again, this was not tested on a fresh directory.
4) There were various scene errors that may be unrelated, but we need to at least try to address.  
5) It would be good to see a diff of the files produced by the old and new code to see what has changed and if we need to make any adjustments.

cyang31:
## Background
Address the issues in #260 

## Main changes
- remove `--myelin-target-file` and combine it with `--refmyelinmaps`
- automatically detect the number of vertices in the specified reference map, if it is a 164k or 59k reference map, it will be resampled into 32k and then be fed into `MyelinMap_BC.sh` via `--myelin-target-file="$MyelinTargetFile"`
- The scene errors are **NOT** addressed.
- tested on fresh directories
- diff of output structures is checked and provided in this PR

## Diff of output structures
I compared an old version without the recent updates of `MyelinMap_BC.sh`, whose commit is 5fc6bae7d95fd407c53ce708f0660562686b3656, with the current PR. PreFreeSurfer+FreeSurfer+PostFreeSurfer are launched on two fresh directories.
 
Output Files from the commit 5fc6bae7d95fd407c53ce708f0660562686b3656 **BUT NOT** from the current PR:
MNINonLinear/Native/102311.L.RefMyelinMap.native.func.gii
MNINonLinear/Native/102311.R.RefMyelinMap.native.func.gii

 Output Files from the current PR **BUT NOT** from the commit 5fc6bae7d95fd407c53ce708f0660562686b3656:
MNINonLinear/fsaverage_LR32k/102311.BiasField.32k_fs_LR.dscalar.nii
MNINonLinear/fsaverage_LR32k/102311.RefMyelinMap.32k_fs_LR.dscalar.nii
MNINonLinear/Native/102311.BiasField.native.dscalar.nii
T1w/102311.L.midthickness.164k_fs_LR.surf.gii
T1w/102311.R.midthickness.164k_fs_LR.surf.gii

## Evaluation
The tests are finished without pipeline errors, but there remain scene errors indicating missing files.

When `--refmyelinmaps="${HCPPIPEDIR_Templates}/standard_mesh_atlases/Conte69.MyelinMap_BC.164k_fs_LR.dscalar.nii"`
/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e2650115
/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o2650115

When `refmyelinmaps=${HCPPIPEDIR}/global/templates/MSMAll/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii`
/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e2650964
/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o2650964

### DIFF
--- a/PostFreeSurfer/scripts/CreateMyelinMaps.sh
+++ b/PostFreeSurfer/scripts/CreateMyelinMaps.sh
@@ -254,11 +254,11 @@ if [ "${T2wPresent}" = "YES" ] ; then
 	# append the HighResMesh into the full ResMesh array
 	AllAvailableMeshesArray="${LowResMeshesArray[@]}"
 	AllAvailableMeshesArray+=(${HighResMesh})
-	NumRefSurfVertice=$(${CARET7DIR}/wb_command -file-information "$ReferenceMyelinMaps" -only-cifti-xml | grep -m 1 -oP 'SurfaceNumberOf(Vertices|Nodes)="\K\d+')
+	NumRefSurfVertices=$(${CARET7DIR}/wb_command -file-information "$ReferenceMyelinMaps" -only-cifti-xml | grep -m 1 -oP 'SurfaceNumberOf(Vertices|Nodes)="\K\d+')
 	# compare vertex numbers between mesh files in the template directory and the input reference myelin map
 	for ResMesh in "${AllAvailableMeshesArray[@]}" ; do
-		NumSurfVertice=$(grep -m 1 -oP 'Dim0="\K\d+' ${HCPPIPEDIR}/global/templates/standard_mesh_atlases/L.atlasroi.${ResMesh}k_fs_LR.shape.gii)
-		if [ "$NumRefSurfVertice" = "$NumSurfVertice" ]; then
+		NumSurfVertices=$(grep -m 1 -oP 'Dim0="\K\d+' ${HCPPIPEDIR}/global/templates/standard_mesh_atlases/L.atlasroi.${ResMesh}k_fs_LR.shape.gii)
+		if [ "$NumRefSurfVertices" = "$NumSurfVertices" ]; then
 			RefResMesh=${ResMesh}
 			IsRefValid=true
 			log_Msg "Find the template file with the same resolution mesh as the reference myelin map! The ResMesh is ${RefResMesh}"
@@ -271,9 +271,8 @@ if [ "${T2wPresent}" = "YES" ] ; then
 		log_Err_Abort "The mesh resolution of the input reference map ${ReferenceMyelinMaps} doesn't match with any template files in ${HCPPIPEDIR}/global/templates/standard_mesh_atlases!"
 	fi
 	
-	# resample the reference map into 32k
 	case "$RefResMesh" in
-		("164")
+		(${HighResMesh})
 			SphereFolder=${AtlasSpaceFolder}
 			T1wSurfFolder=${StudyFolder}/${Subject}/T1w
 			;;
@@ -282,9 +281,13 @@ if [ "${T2wPresent}" = "YES" ] ; then
 			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR${RefResMesh}k
 			;;
 	esac
+
+	#Reduce memory usage by smoothing on downsampled mesh (match the gifti version by using the first lowresmesh)
+	LowResMesh="${LowResMeshesArray[0]}"
 	MyelinTargetFile=${ReferenceMyelinMaps}
-	if [ "$RefResMesh" != "32" ]; then
-		log_Msg "resample the reference map with ${NumRefSurfVertice} ~ ${RefResMesh}k vertices into 32k"
+	# only resample the reference map into low res mesh when it is a HighResMesh
+	if [ "$RefResMesh" == "${HighResMesh}" ]; then
+		log_Msg "resample the reference map with ${NumRefSurfVertices} ~ ${RefResMesh}k vertices into low res mesh"
 		MyelinTargetFile=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.RefMyelinMap.${LowResMesh}k_fs_LR.dscalar.nii
 		${CARET7DIR}/wb_command -cifti-resample ${ReferenceMyelinMaps} \
 				COLUMN ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.MyelinMap.${LowResMesh}k_fs_LR.dscalar.nii \
@@ -301,8 +304,6 @@ if [ "${T2wPresent}" = "YES" ] ; then
 		-metric CORTEX_LEFT "$SphereFolder"/"$Subject".L.RefMyelinMap."$RefResMesh"k_fs_LR.func.gii \
 		-metric CORTEX_RIGHT "$SphereFolder"/"$Subject".R.RefMyelinMap."$RefResMesh"k_fs_LR.func.gii
 	
-	#Reduce memory usage by smoothing on downsampled mesh (match the gifti version by using the first lowresmesh)
-	LowResMesh="${LowResMeshesArray[0]}"
 	# ----- Begin moved statements -----
 	# Recompute Myelin Map Bias Field Based on Better Registration
 	log_Msg "Recompute Myelin Map Bias Field Based on Better Registration"
--- a/PostFreeSurfer/scripts/CreateMyelinMaps.sh
+++ b/PostFreeSurfer/scripts/CreateMyelinMaps.sh
@@ -286,7 +286,7 @@ if [ "${T2wPresent}" = "YES" ] ; then
 	LowResMesh="${LowResMeshesArray[0]}"
 	MyelinTargetFile=${ReferenceMyelinMaps}
 	# only resample the reference map into low res mesh when it is a HighResMesh
-	if [ "$RefResMesh" == "${HighResMesh}" ]; then
+	if [ "$RefResMesh" != "${LowResMesh}" ]; then
 		log_Msg "resample the reference map with ${NumRefSurfVertices} ~ ${RefResMesh}k vertices into low res mesh"
 		MyelinTargetFile=${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.RefMyelinMap.${LowResMesh}k_fs_LR.dscalar.nii
 		${CARET7DIR}/wb_command -cifti-resample ${ReferenceMyelinMaps} \
@@ -294,10 +294,10 @@ if [ "${T2wPresent}" = "YES" ] ; then
 				COLUMN ADAP_BARY_AREA ENCLOSING_VOXEL \
 				${MyelinTargetFile} \
 				-surface-postdilate 40 \
-				-left-spheres ${SphereFolder}/${Subject}.L.sphere.${RefResMesh}k_fs_LR.surf.gii ${AtlasSpaceFolder}/fsaverage_LR32k/${Subject}.L.sphere.32k_fs_LR.surf.gii \
-				-left-area-surfs ${T1wSurfFolder}/${Subject}.L.midthickness.${RefResMesh}k_fs_LR.surf.gii ${StudyFolder}/${Subject}/T1w/fsaverage_LR32k/${Subject}.L.midthickness.32k_fs_LR.surf.gii \
-				-right-spheres ${SphereFolder}/${Subject}.R.sphere.${RefResMesh}k_fs_LR.surf.gii ${AtlasSpaceFolder}/fsaverage_LR32k/${Subject}.R.sphere.32k_fs_LR.surf.gii \
-				-right-area-surfs ${T1wSurfFolder}/${Subject}.R.midthickness.${RefResMesh}k_fs_LR.surf.gii ${StudyFolder}/${Subject}/T1w/fsaverage_LR32k/${Subject}.R.midthickness.32k_fs_LR.surf.gii
+				-left-spheres ${SphereFolder}/${Subject}.L.sphere.${RefResMesh}k_fs_LR.surf.gii ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.L.sphere.${LowResMesh}k_fs_LR.surf.gii \
+				-left-area-surfs ${T1wSurfFolder}/${Subject}.L.midthickness.${RefResMesh}k_fs_LR.surf.gii ${StudyFolder}/${Subject}/T1w/fsaverage_LR${LowResMesh}k/${Subject}.L.midthickness.${LowResMesh}k_fs_LR.surf.gii \
+				-right-spheres ${SphereFolder}/${Subject}.R.sphere.${RefResMesh}k_fs_LR.surf.gii ${AtlasSpaceFolder}/fsaverage_LR${LowResMesh}k/${Subject}.R.sphere.${LowResMesh}k_fs_LR.surf.gii \
+				-right-area-surfs ${T1wSurfFolder}/${Subject}.R.midthickness.${RefResMesh}k_fs_LR.surf.gii ${StudyFolder}/${Subject}/T1w/fsaverage_LR${LowResMesh}k/${Subject}.R.midthickness.${LowResMesh}k_fs_LR.surf.gii
 	fi
 	# the gifti files from reference maps are generated in previous versions
 	${CARET7DIR}/wb_command -cifti-separate "$ReferenceMyelinMaps" COLUMN \
--- a/PostFreeSurfer/scripts/CreateMyelinMaps.sh
+++ b/PostFreeSurfer/scripts/CreateMyelinMaps.sh
@@ -170,9 +170,9 @@ if [ "${T2wPresent}" = "YES" ] ; then
   ${CARET7DIR}/wb_command -volume-palette "$T1wFolder"/T1wDividedByT2w_ribbon.nii.gz MODE_AUTO_SCALE_PERCENTAGE -pos-percent 4 96 -interpolate true -palette-name videen_style -disp-pos true -disp-neg false -disp-zero false
   ${CARET7DIR}/wb_command -add-to-spec-file "$T1wFolder"/"$NativeFolder"/"$Subject".native.wb.spec INVALID "$T1wFolder"/T1wDividedByT2w_ribbon.nii.gz
   
-	${CARET7DIR}/wb_command -cifti-separate "$ReferenceMyelinMaps" COLUMN \
-  	-metric CORTEX_LEFT "$AtlasSpaceFolder"/"$Subject".L.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii \
-  	-metric CORTEX_RIGHT "$AtlasSpaceFolder"/"$Subject".R.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii
+  ${CARET7DIR}/wb_command -cifti-separate "$ReferenceMyelinMaps" COLUMN \
+  -metric CORTEX_LEFT "$AtlasSpaceFolder"/"$Subject".L.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii \
+  -metric CORTEX_RIGHT "$AtlasSpaceFolder"/"$Subject".R.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii
 fi
 
 
--- a/PostFreeSurfer/scripts/CreateMyelinMaps.sh
+++ b/PostFreeSurfer/scripts/CreateMyelinMaps.sh
@@ -250,27 +250,36 @@ ${CARET7DIR}/wb_command -surface-resample ${StudyFolder}/${Subject}/T1w/${Native
 # BC processing
 if [ "${T2wPresent}" = "YES" ] ; then	
 	# determine the resolution of the reference myelin map
-	# TODO: is there a better way to directly get the information of 164k, 59k or 32k? 1) hack into the xml, 2) hack info file names, 3)???
+	IsRefValid=false
+	# append the HighResMesh into the full ResMesh array
+	AllAvailableMeshesArray="${LowResMeshesArray[@]}"
+	AllAvailableMeshesArray+=(${HighResMesh})
 	NumRefSurfVertice=$(${CARET7DIR}/wb_command -file-information "$ReferenceMyelinMaps" -only-cifti-xml | grep -m 1 -oP 'SurfaceNumberOf(Vertices|Nodes)="\K\d+')
+	# compare vertex numbers between mesh files in the template directory and the input reference myelin map
+	for ResMesh in "${AllAvailableMeshesArray[@]}" ; do
+		NumSurfVertice=$(grep -m 1 -oP 'Dim0="\K\d+' ${HCPPIPEDIR}/global/templates/standard_mesh_atlases/L.atlasroi.${ResMesh}k_fs_LR.shape.gii)
+		if [ "$NumRefSurfVertice" = "$NumSurfVertice" ]; then
+			RefResMesh=${ResMesh}
+			IsRefValid=true
+			log_Msg "Find the template file with the same resolution mesh as the reference myelin map! The ResMesh is ${RefResMesh}"
+			break
+		fi
+	done
+	
+	# error when the number of vertex doesn't have a match
+	if [ "$IsRefValid" = false ]; then
+		log_Err_Abort "The mesh resolution of the input reference map ${ReferenceMyelinMaps} doesn't match with any template files in ${HCPPIPEDIR}/global/templates/standard_mesh_atlases!"
+	fi
+	
 	# resample the reference map into 32k
-	case "$NumRefSurfVertice" in
-		("163842")
-			RefResMesh="164"
+	case "$RefResMesh" in
+		("164")
 			SphereFolder=${AtlasSpaceFolder}
 			T1wSurfFolder=${StudyFolder}/${Subject}/T1w
 			;;
-		("59292")
-			RefResMesh="59"
-			SphereFolder=${AtlasSpaceFolder}/fsaverage_LR59k
-			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR59k
-			;;
-		("32492")
-			RefResMesh="32"
-			SphereFolder=${AtlasSpaceFolder}/fsaverage_LR32k
-			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR32k
-			;;
 		(*)
-			log_Err_Abort "unrecognized NumRefSurfVertice value '$NumRefSurfVertice', valid options are 163842, 59292, 32492"
+			SphereFolder=${AtlasSpaceFolder}/fsaverage_LR${RefResMesh}k
+			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR${RefResMesh}k
 			;;
 	esac
 	MyelinTargetFile=${ReferenceMyelinMaps}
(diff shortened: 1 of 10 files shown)


PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -170,10 +169,10 @@ if [ "${T2wPresent}" = "YES" ] ; then
...
   ${CARET7DIR}/wb_command -volume-palette "$T1wFolder"/T1wDividedByT2w_ribbon.nii.gz MODE_AUTO_SCALE_PERCENTAGE -pos-percent 4 96 -interpolate true -palette-name videen_style -disp-pos true -disp-neg false -disp-zero false
   ${CARET7DIR}/wb_command -add-to-spec-file "$T1wFolder"/"$NativeFolder"/"$Subject".native.wb.spec INVALID "$T1wFolder"/T1wDividedByT2w_ribbon.nii.gz
-
+  
   ${CARET7DIR}/wb_command -cifti-separate "$ReferenceMyelinMaps" COLUMN \
-  	-metric CORTEX_LEFT "$AtlasSpaceFolder"/"$Subject".L.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii \
-  	-metric CORTEX_RIGHT "$AtlasSpaceFolder"/"$Subject".R.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii
+  -metric CORTEX_LEFT "$AtlasSpaceFolder"/"$Subject".L.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii \
+  -metric CORTEX_RIGHT "$AtlasSpaceFolder"/"$Subject".R.RefMyelinMap."$HighResMesh"k_fs_LR.func.gii
```
coalsont, line 175: I would indent the two `-metric` lines one more level, since they are a continuation of the -cifti-separate.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -242,60 +241,128 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
-		BiasFieldComputed=true
-		${CARET7DIR}/wb_command -cifti-separate ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.BiasField.native.dscalar.nii COLUMN \
-			-metric CORTEX_LEFT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.BiasField.native.func.gii \
-			-metric CORTEX_RIGHT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.BiasField.native.func.gii
-	else
-		# bias field in native space is already generated
-		# BC the other types of given myelin maps
-		${CARET7DIR}/wb_command -cifti-math "Var - Bias" ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}_BC.native.dscalar.nii -var Var ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}.native.dscalar.nii -var Bias ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.BiasField.native.dscalar.nii
+	# determine the resolution of the reference myelin map
+	# TODO: is there a better way to directly get the information of 164k, 59k or 32k? 1) hack into the xml, 2) hack info file names, 3)???
```
coalsont, line 257: Probably just expect this to match HighResMesh?

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -242,60 +241,128 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
-		# ----- Begin moved statements -----
-		# Recompute Myelin Map Bias Field Based on Better Registration
-		log_Msg "Recompute Myelin Map Bias Field Based on Better Registration"
+# BC processing
+if [ "${T2wPresent}" = "YES" ] ; then	
+	# Create surface on HighResMesh in subject's T1w space
+	${CARET7DIR}/wb_command -surface-resample ${StudyFolder}/${Subject}/T1w/${NativeFolder}/${Subject}.L.midthickness.native.surf.gii \
+		${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.sphere.MSMSulc.native.surf.gii \
+		${AtlasSpaceFolder}/${Subject}.L.sphere.${HighResMesh}k_fs_LR.surf.gii \
+		BARYCENTRIC ${StudyFolder}/${Subject}/T1w/${Subject}.L.midthickness.${HighResMesh}k_fs_LR.surf.gii
```
coalsont, line 250: If we aren't going to delete these high res T1w surfaces, we should probably generate them unconditionally, rather than only when T2w images are present.  Don't create differences between datasets when you don't have to.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -242,60 +241,128 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
+			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR59k
+			;;
+		("32492")
+			RefResMesh="32"
+			SphereFolder=${AtlasSpaceFolder}/fsaverage_LR32k
+			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR32k
+			;;
+		(*)
+			log_Err_Abort "unrecognized NumRefSurfVertice value '$NumRefSurfVertice', valid options are 163842, 59292, 32492"
+      ;;
```
coalsont, line 278: The indentation on this line is not correct.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -242,60 +241,128 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
+				-metric CORTEX_LEFT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.BiasField.native.func.gii \
+				-metric CORTEX_RIGHT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.BiasField.native.func.gii
+		else
+			# bias field in native space is already generated
+			# BC the other types of given myelin maps
+			${CARET7DIR}/wb_command -cifti-math "Var - Bias" ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}_BC.native.dscalar.nii -var Var ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}.native.dscalar.nii -var Bias ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.BiasField.native.dscalar.nii
+		fi
+		${CARET7DIR}/wb_command -cifti-separate ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}_BC.native.dscalar.nii COLUMN \
+			-metric CORTEX_LEFT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.${MyelinMap}_BC.native.func.gii \
+			-metric CORTEX_RIGHT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.${MyelinMap}_BC.native.func.gii
```
coalsont, line 326: This final command of this loop can be moved to the start of the next loop, and combined with the above comments, that means this loop doesn't need to exist anymore.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
cyang31: Actually I don't think the `-cifti-separate` here is necessary. I will just remove it, since the resulting files aren't used in the following processings.

cyang31: @glasserm Matt, can you confirm if we still need to generate gifti reference myelin map files? I don't think they are necessary because they are basically useless in this processing stage.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -293,7 +289,7 @@ if [ "${T2wPresent}" = "YES" ] ; then
 	fi
 	BiasFieldComputed=false
 	#Reduce memory usage by smoothing on downsampled mesh (match the gifti version by using the first lowresmesh)
-	LowResMesh="${LowResMeshesArray[0]}"
+	LowResMesh="${LowResMeshesArray[0]}"		
 	# myelin map only loop
 	for MyelinMap in MyelinMap SmoothedMyelinMap ; do
 		if [ "$BiasFieldComputed" = false ]; then
```
coalsont, line 295: I don't see a comment that I thought I made here, so making it again...this loop is hardcoded to always execute twice, and the first time it will do this `= false` section, and the second time it will do the `else`.  Therefore, all of this `if` statement can instead be translated into appropriate unconditional commands before this loop.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -293,7 +289,7 @@ if [ "${T2wPresent}" = "YES" ] ; then
 	fi
 	BiasFieldComputed=false
 	#Reduce memory usage by smoothing on downsampled mesh (match the gifti version by using the first lowresmesh)
-	LowResMesh="${LowResMeshesArray[0]}"
+	LowResMesh="${LowResMeshesArray[0]}"
```
coalsont, line 292: Please don't add whitespace onto the ends of lines.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
cyang31: I see. Since the BC operation for `MyelinMap` is already part of the module `MyelinMap_BC.sh`, there will still be a `if` condition to allow single BC operation (`Var-bias`) on the other types of myelin maps, such as `SmoothedMyelinMap`.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
cyang31: The `-cifti-separate` above are not necessary IMO. Then I am not sure if the reference myelin map is always matched with `$HighResMesh`. Otherwise, it would be nice if the script can automatically detect the resolution of the reference map and then decide whether to downsample to 32k or not.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -287,33 +287,30 @@ if [ "${T2wPresent}" = "YES" ] ; then
...
-				--myelin-target-file="$MyelinTargetFile" \
-				--map="$MyelinMap"
-			# ----- End moved statements -----
-			# bias field is computed in the module MyelinMap_BC.sh
-			BiasFieldComputed=true
-			${CARET7DIR}/wb_command -cifti-separate ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.BiasField.native.dscalar.nii COLUMN \
-				-metric CORTEX_LEFT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.BiasField.native.func.gii \
-				-metric CORTEX_RIGHT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.BiasField.native.func.gii
-		else
+		if [ "$MyelinMap"!="MyelinMap" ]; then
```
coalsont, line 311: The only item that does this is SmoothedMyelinMap.  Unless we plan to add more things to this loop, please just explicitly do this on SmoothedMyelinMap before the loop.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
coalsont: This is a sub-script of `PostFreeSurfer.sh`, it is very narrow-purpose (it doesn't get called by any other pipeline, or on its own, otherwise we would convert it to newopts.shlib).  If `PostFreeSurfer.sh` doesn't need that flexibility, we shouldn't add complication to try to support it.

cyang31: This is related to a mandatory argument `--refmyelinmaps` in the `PostFreeSurferPipeline.sh`, which is then passed to this sub-script. I recalled in last meeting with Matt @glasserm he wanted the script to support automatic detection of the resolution of the input reference myelin map. I would like to see if he changed his mind.

glasserm: For backwards compatibility I think we should keep the ability to use 164k files without specifying anything additional and also keep all the files that the prior command generated.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -242,60 +237,125 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
+			SphereFolder=${AtlasSpaceFolder}/fsaverage_LR59k
+			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR59k
+			;;
+		("32492")
+			RefResMesh="32"
+			SphereFolder=${AtlasSpaceFolder}/fsaverage_LR32k
+			T1wSurfFolder=${StudyFolder}/${Subject}/T1w/fsaverage_LR32k
+			;;
+		(*)
+			log_Err_Abort "unrecognized NumRefSurfVertice value '$NumRefSurfVertice', valid options are 163842, 59292, 32492"
```
glasserm, line 273: We have additional options in NHPs.  We probably should test if equal to 163842 or else assume it matches one of the LowResMeshes.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -242,60 +237,125 @@ for STRING in "$AtlasSpaceFolder"/"$NativeFolder"@native@roi "$AtlasSpaceFolder"
...
+		-var Var ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.SmoothedMyelinMap.native.dscalar.nii \
+		-var Bias ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.BiasField.native.dscalar.nii
+
+	# myelin map only loop
 	for MyelinMap in MyelinMap SmoothedMyelinMap ; do
+		${CARET7DIR}/wb_command -cifti-separate ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${MyelinMap}_BC.native.dscalar.nii COLUMN \
+			-metric CORTEX_LEFT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.L.${MyelinMap}_BC.native.func.gii \
+			-metric CORTEX_RIGHT ${AtlasSpaceFolder}/${NativeFolder}/${Subject}.R.${MyelinMap}_BC.native.func.gii
+			
+		# create cifti and gift MyelinMap in the high res mesh space
```
glasserm, line 323: gift --> gifti

PostFreeSurfer/scripts/CreateMyelinMaps.sh
glasserm: I think we should keep all the old files for now.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
cyang31: I don't get it. How to "assume it matched one of the $LowResMeshes"? I would suggest to list all the valid options in this case-esac statement to check if the reference myelin map falls into valid options or not.

glasserm: There could be a variety of numbers.  The possibilities are either the 164k mesh or one of the user specified LowResMeshes to the PostFreeSurfer pipeline.

cyang31: I see. I just feel it is technically difficult to implement without knowing the ground truth relation between the number of vertices and the corresponding LowResMesh.

glasserm: Maybe @coalsont  has a suggestion.

cyang31: One thing I would like to confirm is, except for the case `ResMesh="164" when NumRefSurfVertice=163842`, do all the rest cases have the relation `echo "$NumRefSurfVertice" | cut -c 1-2 == $ResMesh`?

glasserm: @takuya-hayashi might know

coalsont: Don't count on taking the first digits.  If anything, it would probably be rounding the `/ 1000` to the nearest integer, but they are not automatically computed (someone looked at the number, rounded it in their head, and used that for the filename of the spheres).

I don't see a reason why CreateMyelinMaps (and PostFreeSurfer itself) can't just expect the template myelin map to be HighResMesh.  This is something that we will rarely regenerate, and we can resample it to the appropriate mesh for the species at that time, so I see no substantial benefit to supporting other meshes in the per-subject code.  I also don't really like having the script modify what commands it runs based on the contents of a file, rather than based on options passed in by the user (which would suggest an option like `--myelin-template-mesh=32`, and logic that uses HighResMesh otherwise).

If we decide we really, truly don't want to have to resample the myelin template to the HighRes mesh in advance...does this line contain what you need to proceed?

https://github.com/Washington-University/HCPpipelines/blob/a8cea8aeee6bfc53dfd5e7936aad9bf6505e17c4/PostFreeSurfer/scripts/CreateMyelinMaps.sh#L214

glasserm: The issue is that we haven't been terribly consistent about these myelin templates (e.g., the original MSMSulc template was 164k.  The MSMAll template was 32k.  The transmit field corrected MSMAll template was 32k.  Macaques could be yet different resolutions).  Thus, we need some flexibility here if we want to maintain backwards compatibility, which I think we should.

coalsont:
If I understand correctly, this PR is trying to fix a bug that wasn't caught by the testing that was done (because the directories had previous runs done in them).  What is the status here?  Should we move the "support other resolutions of group myelin map" feature to a different PR/issue so we can get the bug fixed?

cyang31:
I thought we were waiting on detailed information about all the available resolutions, which can be ignored if we decide to use the `HighResMesh` or `--myelin-template-mesh` to inform the pipeline with the resolution of the reference myelin map. (I lean towards the `--myelin-template-mesh` option)

I now know Tim's point of controlling the behavior of the script only with what is fed into the arguments. If @glasserm also agrees with this, I will update the code with an optional `--myelin-template-mesh` with default value `164k`, test the script again, if passed commit it for another round of review. Does this sound okay?

glasserm:
We are already specifying these resolutions to PostFreeSurfer in the --hiresmesh and --lowresmesh parameters.  The myelin template needs to be one of these and the script should detect which one.  PostFreeSurfer will already have errors if there is non-matching between a mesh resolution specified in these parameters and the templates provided with the pipelines.  I don't see the need for additional non-backwards compatible parameters or setting the default simply for backwards compatibility when we will likely use something else by default in the future.

coalsont:
I am opposed to having scripts examine the contents of data files (or parse filenames) to change which commands the script runs, as that creates complications in the code, and surprising behavior by simply changing data inputs.  Adding an option needed for future data, so that the usage with previous data can use the default setting, doesn't seem that bad to me.  cyang31's plan is what I had in mind (except per convention, the default should be `164`, and not include the `k`).

coalsont:
As for doing it Matt's way, a loop through the `HighResMesh` and the set of `LowResMesh`es, and `grep`ing the output of `wb_command -file-information -no-map-info` for the number of vertices, comparing a resampling sphere to the myelin reference map, would be a relatively clean way to find the correct mesh identifier.  I would prefer actually telling the script (with a new option) what to expect for new data, rather than making it guess.

glasserm:
In general, I am okay with Tim's suggestion; however in this particular case we need to both maintain backwards compatibility and change the default to something new, so I don't see a way to do that other than what I suggested.

coalsont:
I don't think we need to change the pipeline default to something new (the option didn't even exist yet, so it isn't really a "change" anyway).  The default on a new option is for backwards compatibility with previous data, doing something different going forward can have an additional parameter in the launcher script.

glasserm:
Perhaps we need to discuss tomorrow.

cyang31:
It works for me. Let's discuss tomorrow.

glasserm:
Tim and I discussed.  While Tim's concerns make sense from an overall design philosophy, in this particular instance he is okay with us doing what I suggest (basically compare the number of vertices in the myelin template to the meshes specified in HighResMesh and LowResMesh to figure out what it is.  These meshes must existing in the pipelines template folder also, so there is no ambiguity.  If a myelin template is specified that is not one of these meshes, it is an error.  This particular situation arises due to the need for backwards compatibility while at the same time making a change to a default going forward.

cyang31:
I see. I think this is a clear instruction about the next steps but I would like to confirm with @glasserm and @coalsont about some implementation details in advance.
- Again, we are not aware of all the possible mappings from number of vertices to the `ResMesh`, and the `echo "$NumRefSurfVertice" | cut -c 1-2 == $ResMesh` doesn't seem to be the solution to determine all the mappings. What are the other solutions to validate if a number of vertices is matching a `ResMesh`
- To check if the meshes exist in the pipeline template, is it sufficient to check the existence of files with "ResMesh" in the file names from the `global/templates/standard_mesh_atlases` folder?

glasserm:
Just look for the mesh files in the templates directory using the mesh string and compare vertex numbers.  There is no ambiguity that way.  If you don't find the files in the templates directory, it is an error.

cyang31:
Test passed:
using 164k reference myelin map: 
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e2793935`
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o2793935`

using 32k reference myelin map: 
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e2815676`
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o2815676`

glasserm:
Some of this code is reliant on a 32k mesh, but in macaques it is a 10k mesh, marmosets a 2k mesh, and chimps a 20k mesh.  Perhaps it should instead revolve around not being the ${HighResMesh}.  Also, was it tested on empty folders (no prior PostFreeSurfer run)?

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -250,27 +250,36 @@ ${CARET7DIR}/wb_command -surface-resample ${StudyFolder}/${Subject}/T1w/${Native
...
 	# determine the resolution of the reference myelin map
-	# TODO: is there a better way to directly get the information of 164k, 59k or 32k? 1) hack into the xml, 2) hack info file names, 3)???
+	IsRefValid=false
+	# append the HighResMesh into the full ResMesh array
+	AllAvailableMeshesArray="${LowResMeshesArray[@]}"
+	AllAvailableMeshesArray+=(${HighResMesh})
 	NumRefSurfVertice=$(${CARET7DIR}/wb_command -file-information "$ReferenceMyelinMaps" -only-cifti-xml | grep -m 1 -oP 'SurfaceNumberOf(Vertices|Nodes)="\K\d+')
+	# compare vertex numbers between mesh files in the template directory and the input reference myelin map
+	for ResMesh in "${AllAvailableMeshesArray[@]}" ; do
+		NumSurfVertice=$(grep -m 1 -oP 'Dim0="\K\d+' ${HCPPIPEDIR}/global/templates/standard_mesh_atlases/L.atlasroi.${ResMesh}k_fs_LR.shape.gii)
```
coalsont, line 260: Typo, "vertices" is the correct word for the variable name.

cyang31:
> PostFreeSurfer

cyang31:
oops, I accidentally close it. Now it is reopened.

cyang31:
> Some of this code is reliant on a 32k mesh, but in macaques it is a 10k mesh, marmosets a 2k mesh, and chimps a 20k mesh. Perhaps it should instead revolve around not being the ${HighResMesh}. Also, was it tested on empty folders (no prior PostFreeSurfer run)?

1. How about using the first element in the LowResMeshArray?
2. They are both tested on empty folders without prior PostFreeSurfer run. the 164k was first failed with a bug in CreateMyelinMap.sh so then I commented out the first two steps in PostFreeSurfer and rerun it only on the CreateMyelinMap.sh, which I don't think it is a big deal.

glasserm:
You should only need to resample if it is HighResMesh though.  Else, it presumably should match a LowResMesh and not need to be resampled, right?

coalsont:
Please try not to make merge commits.  If you need to use something that has been added to master, rebasing is cleaner.

coalsont:
> You should only need to resample if it is HighResMesh though. Else, it presumably should match a LowResMesh and not need to be resampled, right?

So, you want to additionally change the mesh the BC smoothing is performed on based on the resolution of the input?  I vote against that, as it makes it more complicated to understand what the script actually does.  The sooner in the process you can deal with input inconsistencies, the easier it is to understand the remaining code.

cyang31:
I think It is hard to fix the mesh the BC smoothing is performed on given the variety of options...So, it has to be controlled by certain arguments, externally through command line (use the existing arguments or add a new one) or internally by some complicated if-else statements (not preferred from an understandable point).

> Some of this code is reliant on a 32k mesh, but in macaques it is a 10k mesh, marmosets a 2k mesh, and chimps a 20k mesh

glasserm:
There is a bunch of stuff currently hardcoded to a 32k mesh, which will not be correct for other species.  If it were instead hard coded in reference to a 164k mesh, that would be okay for now.

cyang31:
Is this the current plan?
- resample the reference myelin map to hardcoded 32k whenever the reference myelin map is verified not to be 32k
- apply MyelinMap_BC using hardcoded 32k reference myelin map

If we want to support other species then we may consider to include a new argument such as `--myelin-target-file-res=10` to control the above two steps explicitly.

glasserm:
Just use the LowResMesh for what you resample to and it should work.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -286,18 +286,18 @@ if [ "${T2wPresent}" = "YES" ] ; then
 	LowResMesh="${LowResMeshesArray[0]}"
 	MyelinTargetFile=${ReferenceMyelinMaps}
 	# only resample the reference map into low res mesh when it is a HighResMesh
```
coalsont, line 288: This comment should be edited to match the logic.

glasserm:
Please run another fresh test before we commit.

cyang31:
New rounds of tests finished without error:
164k reference myelin map:
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e25259`
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o25259`

32k reference myelin map: 
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.e25398`
- `/media/myelin/alex/MSMAll_optimization/PostFreeSurferPipeline.sh.o25398`

glasserm:
Thanks Alex.  This looks good to me as far as the myelin map results go.  Are we producing the same files in both cases and the same as before?

glasserm:
Okay we can merge this.

Issue #269, opened 2023-07-02
wildcard problem


MoatazAssem:
Hi,

I ran into this issue while trying to use the legacy processing option of using a reference fMRI run for alignment. 

https://github.com/Washington-University/HCPpipelines/blob/5ae05ea05c63988991b8996800234269386e0e83/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh#L850

${FSLDIR}/bin/imcp ${fMRIReferencePath}/Scout* ${fMRIFolder}

My understanding of the problem: Here using the wildcard with Scout* also recognizes the folder Scout_GradientDistortionUnwarp which then leads imcp to crash as it can only copy files.

Not sure if I am missing something here, but this probably requires a fix, either by specifying the exact Scout files, or adding an extra code/line that picks up the names of files, not folders, starting with Scout*

Moataz

glasserm:
Hopefully the QuNex team can take a look and recommend a solution, as this is part of the code they wrote.

demsarjure:
@glasserm we will take a look. It seems like we just need to make the wildcard more restrictrive.

axkralj990:
@glasserm, @demsarjure, it looks like line 850 is the only place where such a wildcard is used for copying Scout images in generic fMRIVolume and fMRISurface scripts. This can be fixed by replacing:

https://github.com/Washington-University/HCPpipelines/blob/5ae05ea05c63988991b8996800234269386e0e83/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh#L850

with:

`find ${fMRIReferencePath} -maxdepth 1 -name "Scout*" -type f -exec ${FSLDIR}/bin/imcp {} ${fMRIFolder} \; `

demsarjure:
@axkralj990 Thanks! Let us code the solution on a separate branch and prepare a pull request for the HCP team.

demsarjure:
PR that solves this is opened at https://github.com/Washington-University/HCPpipelines/pull/270.

PR #270, opened 2023-07-04
Wildcard search in Volume pipeline no longer finds directories.

demsarjure:
PR for issue https://github.com/Washington-University/HCPpipelines/issues/269

Previously the wildcard search could find directories along with files, the subsequent imcp for copying images/files would then crashed when a directory was provided as an input.

This fix now filters the wildcard search to files only.


glasserm:
I am fine with merging if you guys have tested.

coalsont:
Looks reasonable to me.

PR #194, opened 2020-09-01
Unified fieldmap

axkralj990:
Hi,

This pull request improves the performance of the distortion correction for Philips B0 map fieldmap scans, by taking the echo difference (deltaTE) into account when computing the fieldmaps. We have also included phase unwrapping and image de-meaning. We followed exactly the same steps as are implemented in the fsl_prepare_fieldmap function for Siemens scanners. The only difference is the additional brain mask erosion step for Philips scans, which results in a cleaner preprocessed fieldmap. A comparison between the current Philips and Siemens steps for fieldmap preparation is shown on the attached figure.
The code has been tested successfully on both structural and functional images from Siemens and Philips scanners. It does not change any other parts of the processing pipelines.

With kind regards,

Grega & Aleksij

Fieldmaps

PR #271, opened 2023-07-19
Correction to Philips fieldmap preprocessing

axkralj990:
## Background

In this pull request, I corrected the issues from my previous pull request, #194, regarding Philips fieldmap image preprocessing. Upon further investigation, I discovered that I had made an incorrect assumption about the units of the input fieldmap images. 

I had incorrectly assumed that the fieldmap image produced by the Philips Achieva 3.0T TX was a phase difference image in degrees. This assumption was based on the voxel values in the image, which ranged from about -180 to 180. However, upon closer inspection, the values were between about -200 and 200 (our actual dTE obtained from PAR/REC was 2.45 ms). To clarify the units, I contacted a Philips physicist and received confirmation that the fieldmap image was indeed in Hz.

The preprocessing steps are shown in the figure below, along with the old steps and current steps, in addition to the established Siemens processing, for reference. 

## Main Changes

In order to properly process the fieldmap, I have now corrected the previous steps. Since the input image is not a phase difference image, as previously assumed, it now needs to be first converted to a phase difference map, then unwrapped using FSL prelude, and finally converted back to a fieldmap.

I have also included a conditional check to determine if the echo difference parameter (dTE) has been specified. If it is not, the phase unwrapping is skipped. However, based on my research, I believe including the option for phase unwrapping is still useful. Some researchers skip phase unwrapping, especially when using the default echo difference of 3 ms during acquisition. In such cases, wrapping tends to be minimal and limited to the inferior-frontal region where BOLD signal loss usually occurs. When a larger dTE is set, the unwrapping option is necessary, as the phase is significantly wrapped (tested with dTE up to 10 ms).

Another factor contributing to the misconception that the earlier code was correctly preprocessing the fieldmaps was that the combination of all operations applied to the input image produced a similar scaling factor. Specifically, the input data was previously multiplied by `pi/180 * 1000/(3 ms) = 5.81`, whereas in the corrected code, the input data is multiplied by `2*pi = 6.28`. The old cold produces similar values, but definitely lower than they should be, leading to under-correction.

The figure also shows an unwrapped fieldmap and three sagittal slices of the distortion-corrected BOLD image with the white matter surface outline. The effective echo spacing parameter for the correction was calculated as follows ([Neurostars forum discussion](https://neurostars.org/t/converting-philips-fieldmaps-to-bids-and-finding-echo-time-1-and-echo-time-2/17592)):

```
effective_echo_spacing = (1000 * water_fat_shift)/(434.215 * (epi_factor+1))
```

I have made the necessary changes in the pull request to fix these issues and ensure proper distortion correction. Please let me know if there are any concerns or additional requests. Thank you.

Fieldmap2023 (2)


global/scripts/FieldMapPreprocessingAll.sh
```diff
@@ -101,8 +101,8 @@ case $DistortionCorrection in
         # --------------------------------------
 
         MagnitudeInputName=`getopt1 "--fmapmag" $@`
-		    PhaseInputName=`getopt1 "--fmapphase" $@`
-		    DeltaTE=`getopt1 "--echodiff" $@`
+	    PhaseInputName=`getopt1 "--fmapphase" $@`
+		DeltaTE=`getopt1 "--echodiff" $@`
```
coalsont, line 105: In terms of style, these whitespace changes show that the file has a mix of tabs and spaces (even within a single line), and the current changes also leave edited lines inconsistent with their neighbors (these two changed lines use a tab and 4 spaces, and then two tabs for the next line).  I suggest trying to match the indentation style of at least consecutive lines within lines you were going to edit anyway.

axkralj990: I changed this in my most recent commit. There were several lines in the FieldmapPreprocessingAll.sh script with such inconsistencies. I changed all the indentations in the file to 4 spaces, as this seemed as the most common pattern. There were also a few lines with 2-space indentation that I increased to 4 spaces.

glasserm (commented):
If this is tested and working well in your hands, I am happy to merge.  Thanks for your work on this!

axkralj990:
I'm glad I could help. The correction works well on my end, and I also tested it yesterday on data collected with the Philips Achieva 3.0T after a dStream upgrade. I used the effective echo spacing estimate returned by the [dcm2niix](https://github.com/rordenlab/dcm2niix) toolbox (`EstimatedEffectiveEchoSpacing`), and the correction results are practically the same as the correction with the topup method.

glasserm:
I later saw you scene and I agree it looks good.

glasserm (approved):

mharms:
Couple things:
1. Is this particular approach for correction dependent on getting "echo spacing" correct?  Because the issue of the ES for Philips is very complicated.  See https://github.com/rordenlab/dcm2niix/tree/master/Philips and https://github.com/rordenlab/dcm2niix/issues/377
2. "real fieldmap" is probably not the best terminology to use for the input in the Siemens branch of your figure.  It is actually a "phase diff(erence)" map, in Siemens arbitrary -4096:4096 convention
3.  If the input for the Philips branch is a "real fieldmap (Hz)", then the range of that is not going to be [-500:500]/dTE, but rather whatever the actual field discrepancy (from B0) is in Hz, right?

mharms:
Re (1):  Did the EES (EffectiveEchoSpacing) value reported by `dcm2niix` change by the SENSE (in-plane) factor as you changed the SENSE factor?  Can you document here which software version of Philips was used for the scanning and which `dcm2niix` version you used?

Re (2-3):  In that case, might also be better to call the first two stages of the Philips branch of the figure a "(possibly) wrapped fieldmap (Hz)" rather than a "real fieldmap (Hz)" since a "real fieldmap" should not have any wrapping (at least for how I think about that phrasing).

axkralj990:
@mharms, I have updated the figure so that it alerts users that we are not 100% sure of the specific image formats.

The software release is 5.7.1.2, and the scanner is a 2011 model. I have also included this information in the figure.

I am currently out of the office and will get back to you on Monday (July 31, 2023) on how SENSE factor changes affect dicm2niix effective echo spacing estimates, as I now only have access to images that vary in other parameters as well. I will take a few short EPI scans on a phantom with different SENSE values.

axkralj990:
@mharms. Today, we did two pairs of BOLD scans on a phantom to see how the SENSE acceleration affects the estimated effective echo spacing.

If I use the equation `(1000 * water_fat_shift)/(434.215 * (etl+1))`, it appears that I need to divide the number by the in-plane acceleration factor (SENSE) to get values comparable to those calculated by dicm2niix. So in this case SENSE has an effect on the effective echo spacing.

The distortion correction also proved to be appropriate when evaluated on real data, one of which was acquired with `SENSE = 1.9` and the other with `SENSE = 1`, both using the estimate from dicm2niix.

If you have any further questions, please let me know.

Here are the summary results of today's data.

### 1. MultiBand 3, SENSE 1.9
"EchoTime": 0.030001,
"RepetitionTime": 1.3,
"EchoTrainLength": 51,
"WaterFatShift": 13.7557,
"EstimatedEffectiveEchoSpacing": 0.000326937,
"EstimatedTotalReadoutTime": 0.031059

```
ees = (1000 * water_fat_shift)/(434.215 * (etl+1))/SENSE
    = (1000 * 13.7557)/(434.215 * (51+1))/1.9
    = 0.32064 s
    = 0.00032064 ms
```

### 2. MultiBand 3, SENSE 1
"EchoTime": 0.031024,
"RepetitionTime": 1.54233,
"EchoTrainLength": 95,
"WaterFatShift": 25.4869,
"EstimatedEffectiveEchoSpacing": 0.0006112,
"EstimatedTotalReadoutTime": 0.058064

```
ees = (1000 * water_fat_shift)/(434.215 * (etl+1))/SENSE
    = (1000 * 25.4869)/(434.215 * (95+1))/1
    = 0.6114 s
    = 0.0006114 ms
```

### 3. MultiBand 4, SENSE 1.9
"EchoTime": 0.03,
"RepetitionTime": 1,
"EchoTrainLength": 51,
"WaterFatShift": 13.7557,
"EstimatedEffectiveEchoSpacing": 0.000326937,
"EstimatedTotalReadoutTime": 0.031059

```
ees = (1000 * water_fat_shift)/(434.215 * (etl+1))/SENSE
    = (1000 * 13.7557)/(434.215 * (51+1))/1.9
    = 0.32064 s
    = 0.00032064 ms
```

### 4. MultiBand 4, SENSE 1
"EchoTime": 0.03111,
"RepetitionTime": 1.18339,
"EchoTrainLength": 95,
"WaterFatShift": 25.4869,
"EstimatedEffectiveEchoSpacing": 0.0006112,
"EstimatedTotalReadoutTime": 0.058064

```
ees = (1000 * water_fat_shift)/(434.215 * (etl+1))/SENSE
    = (1000 * 25.4869)/(434.215 * (95+1))/1
    = 0.6114 s
    = 0.0006114 ms
```

mharms:
In the interest of being very specific, I think that the relevant equations used currently by `dcm2niix` are at the following link. Note that they use "EPI_Factor" (2001,1013) rather than "ETL", and the two are not always the same, which is an important consideration.
https://github.com/rordenlab/dcm2niix/issues/377#issuecomment-1328530596

If you were to use the formulas at the above link, you should be able to duplicate the values returned by `dcm2niix` exactly (rather than just very close).

mharms:
Good to see that you can indeed (exactly) match the current output of `dcm2niix`.

PR #278, opened 2023-11-04
update rsn regression with the tICA regression

cyang31:
Introduce a `tICA_weighted` mode which performs the following steps in one mode:

-  weighted spatial regression given a group sICA spatial map;
- tICA unmix given a group tICA mixing matrix;
- single regression to generate individual tICA spatial maps.

### DIFF
--- a/global/scripts/RSNregression.m
+++ b/global/scripts/RSNregression.m
@@ -181,10 +181,6 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
             
             if strcmp(Method, "tICA_weighted") % based on tICA/scripts/ComputeGroupTICA.m but for one subject as a group
                 NODEts=NODEts';
-                disp(['NODEts'])
-                size(NODEts)
-                SpectraArray = textscan(optional.SpectraParams, '%s', 'Delimiter', {'@'});
-                nTPsForSpectra =str2double(SpectraArray{1}{1});
                 thisStart = 1;
                 TCSRunVarSub = [];
                 for i = 1:size(inputArray, 1)
--- a/global/scripts/RSNregression.m
+++ b/global/scripts/RSNregression.m
@@ -1,5 +1,5 @@
 function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, varargin)
-    optional = myargparse(varargin, {'GroupMaps' 'VAWeightsName' 'VolInputFile' 'VolInputVNFile' 'VolCiftiTemplate' 'OldBias' 'OldVolBias' 'GoodBCFile' 'VolGoodBCFile' 'SpectraParams' 'OutputZ' 'OutputVolBeta' 'OutputVolZ' 'SurfString' 'ScaleFactor' 'WRSmoothingSigma'});
+    optional = myargparse(varargin, {'GroupMaps' 'tICAMM' 'VAWeightsName' 'VolInputFile' 'VolInputVNFile' 'VolCiftiTemplate' 'OldBias' 'OldVolBias' 'GoodBCFile' 'VolGoodBCFile' 'SpectraParams' 'OutputZ' 'OutputVolBeta' 'OutputVolZ' 'SurfString' 'ScaleFactor' 'WRSmoothingSigma'});
     
     %InputFile - text file containing filenames of timeseries to concatenate
     %InputVNFile - text file containing filenames of the variance maps of each input
@@ -139,13 +139,13 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
         clear vnvolsum;
     end
     
-    if strcmp(Method,'weighted') || strcmp(Method,'dual')
+    if strcmp(Method,'weighted') || strcmp(Method,'dual') || strcmp(Method,'tICA_weighted') 
         GroupMapcii = ciftiopen(optional.GroupMaps, wbcommand);
         weightscii = ciftiopen(optional.VAWeightsName, wbcommand); %normalized vertex areas, voxels are all 1s
         AreaWeights = weightscii.cdata;
     end
     switch Method
-        case 'weighted'
+        case {'weighted', 'tICA_weighted'}
             if strcmp(optional.WRSmoothingSigma, '')
                 error '"weighted" method requires WRSmoothingSigma to be specified'
             end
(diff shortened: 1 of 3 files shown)


glasserm (approved):
I think this looks good to me.

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,56 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+                disp(['NODEts'])
+                size(NODEts)
+                SpectraArray = textscan(optional.SpectraParams, '%s', 'Delimiter', {'@'});
+                nTPsForSpectra =str2double(SpectraArray{1}{1});
+                thisStart = 1;
+                TCSRunVarSub = [];
+                for i = 1:size(inputArray, 1)
+                    dtseriesName=[inputArray{i}];
+                    if exist(dtseriesName, 'file')
+                        [~, runLengthStr] = system(['wb_command -file-information -only-number-of-maps ' dtseriesName]);
```
coalsont, line 193: Consider updating `my_system` below to return the stdout, and using it, which avoids passing matlab's library env variables to other executables.

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,56 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+                end
+                
+                % unmix the sICA timeseries
+                W = pinv(A);
+                normicasig = W * NODEts;
+                icasig = normicasig .* repmat(std(A ./ repmat(sICAtcsvars, 1, size(A, 2)))', 1, size(NODEts, 2)); %Unormalize the icasig assuming sICAtcs with std = 1 (approximately undo the original variance normalization)
+                
+                tICAtcs = icasig';
+                TSTDs = std(tICAtcs, [], 1); %unnormalized tICA temporal standard deviations
+                TIs = [1:1:length(TSTDs)];
```
coalsont, line 220: `1:1:num` is more verbose than needed, `1:num` is equivalent.

coalsont: Also, `TSTDs` values don't seem to get used?  So is it not normalizing after the weird `std(A ./ repmat(...` command?

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,56 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+
+                tICAtcs(:, [1:1:length(TIs)]) = single(tICAtcs(:, TIs));
```
coalsont, line 222: What in the world is this trying to do?  It looks like it is trying to do `tICAtcs = single(tICAtcs);` in the strangest way (and it isn't clear whether it would convert to single when using indexed assignment).  Was `TIs` originally testing whether some timepoints are 0?

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,56 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+                tICAtcs=tICAtcs';
+                tICAtcsAll = reshape(tICAtcs, size(A,1), nTPsForSpectra, 1);
```
coalsont, line 224: It isn't clear to me that `nTPsForSpectra` would always equal the input file dimensions, so this might fail in subjects that don't have full data.

coalsont: Also, I would remove the `, 1)` giving it an unwanted third dimension, so that the next line also doesn't have to `squeeze`.

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,56 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+                tICATCS = squeeze(tICAtcsAll(:, std(tICAtcsAll(:, :, 1), [], 1) > 0, 1));
```
coalsont, line 225: `range(...) > 0` is much more stable (and a bit faster) than `std(...) > 0`.

coalsont: Also, if this is testing for timepoints that are all zeros, I'm not sure that is even necessary in this code (we don't have the pad-subjects-to-expected-timepoints stuff in play here, right?).  With the several transposes in the code, I'm not sure exactly what dimension is what, though.

cyang31:
The code redundancy is mainly because I am too lazy to dig into the dimension and stuff so I keep the same workflow as in `tICA/scripts/ComputeGroupTICA.sh`. Now I remove the redundancy: 1) There's no need to check the zeros since this is an inference mode without padding. 2) No need to check the subject timepoints.

global/scripts/RSNregression.m
cyang31: The TSTDs in `tICA/scripts/ComputeGroupTICA.sh` is used for calculating the explained variances, but in current case, it is not necessary to have it.

cyang31:
The latest commit has been tested without error. The log file: /media/myelin/alex/pipelines/pipeline_areal_classifier/areal_classifier/twregression_log/RSNregression.sh.o1994865

global/scripts/RSNregression.m
```diff
@@ -355,7 +348,7 @@ function cleanupFunc(tempbase)
 end
 
 %like call_fsl, but without sourcing fslconf
-function my_system(command)
+function [exitStatus, stdout]=my_system(command)
```
coalsont, line 351: Since this function already errors if the exit status is nonzero, it doesn't need to return the exit status.

cyang31: I think it might be a good idea to have the same output structure as `system`. Don't you think so?

coalsont: No.  `my_system` is intended to do the error detection that most people forget to do when they use `system`.  If `my_system` returns, the exit status was 0, so there is no reason to return it, so the only time you should use its return value is when you want the stdout.  It is not a drop-in replacement or wrapper, it is a simplification.

I might also argue that it should only use `'-echo'` when the nargout is 0.

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,49 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+                A = load(optional.tICAMM);
+
+                if size(A,1) ~= size(NODEts,1)
+                    error('Mixing matrix to be used does not match dimensions of the sICA components');
+                end
+                
+                % unmix the sICA timeseries
+                W = pinv(A);
+                normicasig = W * NODEts;
+                icasig = normicasig .* repmat(std(A ./ repmat(sICAtcsvars, 1, size(A, 2)))', 1, size(NODEts, 2)); %Unormalize the icasig assuming sICAtcs with std = 1 (approximately undo the original variance normalization)
```
coalsont, line 216: It would be nice if we could get closure on whether this is doing something reasonable before merging this.

cyang31: @glasserm might have his explanation on this.

glasserm: I think I put this in an e-mail somewhere. The point is to get the timeseries to have the amplitudes rather than std=1.

coalsont: It appears to be taking the standard deviation of the mixing matrix divided by sICA standard deviations.  I don't follow how that implements what you said.

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,49 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
             
             betaICAone = weightedDualRegression(GroupMapcii.cdata, inputConcat, AreaWeights .* ScaledAlignmentQuality);
             [betaICA, NODEts] = weightedDualRegression(normalise(betaICAone), inputConcat, AreaWeights);
+            
+            if strcmp(Method, "tICA_weighted") % based on tICA/scripts/ComputeGroupTICA.m but for one subject as a group
+                NODEts=NODEts';
+                disp(['NODEts'])
```
coalsont, line 184: This looks like it might be debugging code.

global/scripts/RSNregression.m
glasserm: If I recall correctly, the goal was to have the tICA component timeseries variances be not one but have the result reflect what it would have been if the sICA component timeseries variances were one.

coalsont: If the timeseries in question are orthogonal, then you can combine their stdevs by root mean square without reconstructing the actual timeseries.  This does not appear to be doing that (stdev calculates and subtracts the mean first, though it is otherwise root mean square, so perhaps this is why its answer may look reasonable).

global/scripts/RSNregression.m
```diff
@@ -178,6 +178,49 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
...
+                size(NODEts)
+                SpectraArray = textscan(optional.SpectraParams, '%s', 'Delimiter', {'@'});
+                nTPsForSpectra =str2double(SpectraArray{1}{1});
```
coalsont, line 187: This code no longer uses nTPsForSpectra, so these lines can be removed.

global/scripts/RSNregression.m
coalsont: Also, `A` is the matrix that turns tICA components back into sICA components, right?  That doesn't seem like the correct matrix to use in combination with sICA stdevs.

coalsont: I may be starting to puzzle out the intent, ICA outputs `A` and `normicasig`, such that `A * normicasig` equals the input, but `normicasig` has rows with...probably stdev equal to 1.  If we think of it in terms of `[u s v] = svd(...);`, we want `s * v'`, but ICA gives us something like `u * s` and `v'` instead.  In the case of SVD, because `s` is diagonal and `u` is orthonormal, we can recover the diagonal of `s` from `u * s` via `sqrt(sum((u * s) .^ 2))`.  `A` has the amplitudes corresponding to the tICA of `NODEts` (after per-run normalization and mostly-restoration of original stdev (missing runs stay 0, throwing off the per-run normalization's outcome a bit)), and we...want to pretend `NODEts` had been normalized before ICA?  To stdev=1?  We would have:

input = A * normicasig
pretendinput = diag(1 / stdev(input)) * A * normicasig
...assume normicasig doesn't change...
pretendA = diag(1 / stdev(input)) * A = A ./ repmat(stdev(input), ...)

which starts to look familiar.  If ica spits out timeseries with stdev=1, it would almost work out if the mixing matrix is orthonormal...it probably isn't, and the demeaning part of the stdev is likely undesired on A, but it may be close enough to not be worth the trouble of improving it.  Some of this should get written down (here and in tICA) so we don't have to guess next time.

coalsont: Someone please double check that fastica and icasso spit out stdev = 1 timeseries, and see if you agree with the explanation I added.

glasserm: They are.  Explanation is fine.  The purpose of this is to help reorder to components from strongest to weakest according to the tICA decomposition alone.

PR #281, opened 2024-01-19
Add compiled reclean pipeline

cyang31:
The reclean pipeline with compiled feature generation step is added in this PR. 

The test is still running. I will update with the test result soon.


glasserm (commented):
If this is tested, we should merge so that Jure can test.

coalsont:
Merge comes after testing, not before.  Jure should be able to test a branch on a different repo.

cyang31:
> If this is tested, we should merge so that Jure can test.

I will update in this thread if it fully passes the test. Currently, it has already passed the step of this new change (compiled feature generation script), but I haven't tested the rest of the compiled steps (FIX, etc.). It should work though, because the rest is simply FIX pipeline that has been tested.

ICAFIX/scripts/computeRecleanFeatures.m
```diff
@@ -154,12 +154,25 @@ function computeRecleanFeatures(StudyFolder, ...
         % structure idx
         CIFTI_LENGTH=length(CIFTI.cdata);
 
-        CORTEX_Left_idx_list = cifti_diminfo_dense_get_surface_info(CIFTI.diminfo{1}, 'CORTEX_LEFT').ciftilist;
```
coalsont, line 157: This is what the matlab compiler was complaining about?  And it can't handle `size(..., [list])` either?  Such weird restrictions...

cyang31: Yes, the size function in 2017 version cannot handle the list and can only use positive integer...


```matlab
>> volume=read_avw("/media/myelin/shared/HCPpipelines_test_rc/HCD0001305_V1_MR/MNINonLinear/Results/fMRI_CONCAT_ALL/fMRI_CONCAT_ALL_hp0.ica/filtered_func_data.ica/melodic_IC.nii.gz");
>> size(volume,[1,2,3])
Error using size
Dimension argument must be a positive integer scalar within indexing range.
 
>> size(volume,1)

ans =

    91

>> size(volume,2)

ans =

   109

>> size(volume,3)

ans =

    91
```

cyang31: The matlab compiler also thinks that `cifti_diminfo_dense_get_surface_info(CIFTI.diminfo{1}, 'CORTEX_LEFT')` is a function which cannot be indexed by dot...

cyang31:
The current PR passed the test on my end.

ICAFIX/scripts/computeRecleanFeatures.m
coalsont: Ah, I didn't realize interpreted 2017b didn't support dimension lists in `size` either.

mharms: The following should work in pretty much every matlab version:
```
sz = size(volume)
sz(1:3)
```

PR #284, opened 2024-02-08
Extract and concatenate runs after tICA cleanup

coalsont:
This basically adds the REST data extract/concatenate call (and `_vn` copying) we did manually into the tICA pipeline options.  Please sanity check the naming conventions of the outputs.  It doesn't regenerate the .csv, because the file name wasn't specific to which kind of cleanup, and it should already exist due to `DeDrift...` previously calling `Extract...`


glasserm (approved):
Thanks Tim.  This looks good to me.

PR #290, opened 2024-04-10
Stand Alone MSMSulc.sh script

nagedem:


coalsont:
You didn't need to close the other PR, it will automatically update whenever you push.  You have also added the merge commit and `msmsulc.sh` back into this PR.

glasserm:
The if conditional for computing the rotated sphere needs to be there or the entire block pulled out of MSMSulc and left in PostFreeSurfer.  We should not be overwriting files that we make available in package downloads necessarily.

coalsont:
I thought the idea was to replace the MSMSulc section of PostFS with a call to this script.  Only a patch script should need a conditional for whether to generate a file, and patch scripts don't need to be in the repo.

glasserm:
The purpose of this script is either to replace MSMSulc in the pipelines or to allow different MSMSulc to be run after the fact.  The same script can do both.

coalsont:
Putting the sphere.rot generation into PostFS outside of the MSMSulc script is also a way to avoid overwriting, without using a conditional, but at the same time, the main reason we need the rotation is for MSM.  While it appears to be used in MSMAll (side note, our "rotation" doesn't currently decompose the affine to a pure rotation, so theoretically it would be better and not exactly equivalent to provide the unrotated native sphere for cumulative distortion measurement in MSMAll, and we could even do this for MSMSulc too), since MSMAll is initialized by MSMSulc anyway, sphere.rot can be said to be purely an MSMSulc requirement, with MSMAll depending on MSMSulc, which would make the MSMSulc script/section the obvious place to generate it.  I think I somewhat prefer this option (generate sphere.rot in PostFS immediately before calling MSMSulc.sh), not having conditional behavior on the filesystem state at all.

Alternatively, having the condition instead test "if the file doesn't exist, or the regname is MSMSulc, then generate it" may better emulate the expected/current behavior of rerunning PostFS/"MSMSulc" after something like faulty memory problems.  Presumably the "after the fact" additional usage would generally be set to "MSMSulcStrain", and so "MSMSulc" can overwrite every time without issue (removing the surprise behavior change for at least the main pipeline).  Presumably it doesn't matter whether a "longitudinal combined subject" folder overwrites this internally prior to release of longitudinal results (I don't know the longitudinal expected flow well enough to say whether it isn't even an issue).

If we don't expect people to want to manually run it "after the fact" with the MSMSulc name (which would overwrite downloaded things), then maybe the default should be MSMSulcStrain, and PostFS should provide the "magic" MSMSulc name when it calls it.

glasserm:
We are using the rotated sphere for other things too.  Let's just always run it in PostFreeSurfer so that we have it even if MSMSulc is not the registration we use.

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,124 @@
...
+	cp "$NativeFolder"/"$RegName"/"$Hemisphere".sphere_rot.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".sphere.rot.native.surf.gii
+	rm "$NativeFolder"/"$RegName"/"$Hemisphere".sphere_rot.surf.gii
+  
+	(
+		cd "$NativeFolder"/"$RegName"
+
+		RefMeshFile=$(basename -- "$RefMesh")
+		ReferenceMesh="$(dirname -- "$RefMesh")/${RefMeshFile/HEMISPHERE/$Hemisphere}"
+		RefDataFile=$(basename -- "$RefData")
+		ReferenceData="$(dirname -- "$RefData")/${RefDataFile/HEMISPHERE/$Hemisphere}"
```
coalsont, line 83:
```suggestion
	#convert relative to absolute for correctness after cd, restrict HEMISPHERE replacement to filename only
	RefMeshFile=$(basename -- "$RefMesh")
	ReferenceMesh="$(cd "$(dirname -- "$RefMesh")"; pwd)/${RefMeshFile/HEMISPHERE/$Hemisphere}"
	RefDataFile=$(basename -- "$RefData")
	ReferenceData="$(cd "$(dirname -- "$RefData")"; pwd)/${RefDataFile/HEMISPHERE/$Hemisphere}"
	(
		cd "$NativeFolder"/"$RegName"

```

nagedem:
Just committed and pushed the relevant changes for your review.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -414,44 +414,15 @@ for Hemisphere in L R ; do
 
     #If desired, run MSMSulc folding-based registration to FS_LR initialized with FS affine
     if [ ${RegName} = "MSMSulc" ] ; then
-        #Calculate Affine Transform and Apply
-        if [ ! -e "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc ] ; then
-            mkdir "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc
```
coalsont, line 419: The lines below put outputs into "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc, so this mkdir should stay.  However, you can (should) use `mkdir -p` instead of the if statement, for simplicity.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -414,44 +414,15 @@ for Hemisphere in L R ; do
...
-        ${CARET7DIR}/wb_command -metric-merge "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii -metric "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_MSMSulc.native.shape.gii -column 2
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii
-        rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_MSMSulc.native.shape.gii
-
-        RegSphere="${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${Hemisphere}.sphere.MSMSulc.native.surf.gii"
+
+        ./global/scripts/MSMSulc.sh
```
coalsont, line 423: This should probably explicitly specify the regname argument to be MSMSulc.  We could have the MSMSulc script default to MSMSulcStrain so that manual running is less likely to overwrite things.

It is also missing the mandatory parameters.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -414,44 +414,15 @@ for Hemisphere in L R ; do
...
-        #Calculate Affine Transform and Apply
-        if [ ! -e "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc ] ; then
-            mkdir "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc
-        fi
+        
         ${CARET7DIR}/wb_command -surface-affine-regression "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.reg.reg_LR.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.mat
         ${CARET7DIR}/wb_command -surface-apply-affine "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.mat "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii
         ${CARET7DIR}/wb_command -surface-modify-sphere "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii 100 "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii
         cp "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.rot.native.surf.gii
```
coalsont, line 421: We may want to move this code to an unconditional location (currently, it is inside an "if MSMSulc").  We could even consider having the MSMSulc script expect to find it in the `"$AtlasSpaceFolder"/"$NativeFolder"/` sphere.rot location, instead of already existing in an MSMSulc folder.

As it stands, if someone initially runs the pipelines with "FS" instead of "MSMSulc" (not recommended, results in a warning, but looks like it will proceed), and then runs MSMSulc.sh by itself, the rotated sphere file won't exist.

glasserm: I think it would be reasonable to run the sphere rotation on any registration option.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
coalsont: My idea is that if the MSMSulc.sh default is set to "MSMSulc", then by default it would overwrite downloaded things, which we may want to discourage.  Alternatively, `--regname` could be mandatory (no default), though users running it manually may put in "MSMSulc" without realizing why it isn't the default.

glasserm: Making it mandatory would be fine.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -413,14 +413,14 @@ for Hemisphere in L R ; do
...
 
     #If desired, run MSMSulc folding-based registration to FS_LR initialized with FS affine
-    if [ ${RegName} = "MSMSulc" ] ; then
-        
-        ${CARET7DIR}/wb_command -surface-affine-regression "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.reg.reg_LR.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.mat
-        ${CARET7DIR}/wb_command -surface-apply-affine "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.mat "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii
-        ${CARET7DIR}/wb_command -surface-modify-sphere "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii 100 "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii
-        cp "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.rot.native.surf.gii
+    if [ ${RegName} = "$RegName" ] ; then
```
coalsont, line 416: This is not what I meant, and will call MSMSulc.sh when the arguments say not to.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -412,46 +412,22 @@ for Hemisphere in L R ; do
...
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii
-        rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_MSMSulc.native.shape.gii
-
-        RegSphere="${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${Hemisphere}.sphere.MSMSulc.native.surf.gii"
+        mkdir -p "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc
+
+        cp "$AtlasSpaceFolder"/"$NativeFolder"/${Hemisphere}.sphere.rot.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii
+
+        ./global/scripts/MSMSulc.sh --subject-dir="$StudyFolder" --subject="$Subject" --regname="$RegName"
```
coalsont, line 428: After a closer look, this location is inside a loop over hemispheres, but MSMSulc.sh currently runs both hemispheres in a single call.  We should decide whether to move the MSMSulc.sh call within FreeSurfer2Caret..., or make MSMSulc.sh do a single hemisphere at a time.  The resampling of MSMSulc distortion that is currently in this same hemisphere loop means that moving the MSMSulc.sh call will require a bit of restructuring.

coalsont: The restructuring required to make the both-hemispheres approach work may actually be a good thing, as it could solidify most of the "if MSMSulc" conditionals into one, improving the conceptual organization.

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,117 @@
...
+if [[ "$RefData" == "" ]]
+then
+	RefData="$SurfaceTemplateFolder"/HEMISPHERE.refsulc.164k_fs_LR.shape.gii
+fi
+
+#Make MSMSulc Directory
+mkdir -p "$NativeFolder"/"$RegName"
+
+#Loop through left and right hemispheres
+for Hemisphere in "$Hemi" ; do
```
coalsont, line 61:
```suggestion
for Hemisphere in $Hemi ; do
```
Quoting would prevent this from interpreting the space as a list separator.

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,117 @@
...
+	wb_command -surface-distortion "$NativeFolder"/"$Subject"."$Hemisphere".sphere.native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".sphere."$RegName".native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".EdgeDistortion_"$RegName".native.shape.gii -edge-method
+	wb_command -surface-distortion "$NativeFolder"/"$Subject"."$Hemisphere".sphere.native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".sphere."$RegName".native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii -local-affine-method -log2
+
+	wb_command -metric-merge "$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_"$RegName".native.shape.gii -metric "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii -column 1
+	wb_command -metric-merge "$NativeFolder"/"$Subject"."$Hemisphere".StrainR_"$RegName".native.shape.gii -metric "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii -column 2
+
+	rm "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii
+done
+
+if [[ "$Hemi" == "L R" ]] ; then
```
coalsont, line 99:
```suggestion
if [[ "$Hemi" == *L* && "$Hemi" == *R* ]] ; then
```
Keep behavior consistent even if someone reverses them, or adds extra whitespace, or...

nagedem:
I have added the hemi argument. But I am not sure if you can see the commit.

coalsont:
Looks like you didn't update your local clone after accepting my suggested edits on github, and then you made the same edits again manually.  The merge seems to have figured out what you meant, so the current version is okay, but in the future, if you ever have to commit the same edits twice, something has gone wrong.

nagedem:
Hmm, not really. I did not do the same edits manually again. I was able to
pull the changes to local computer. But perhaps I needed to do that more
than once (for each accepted change) before committing and pushing the
changes in the other script.

coalsont:
This commit appears to have several of the suggestions I made, which also have commits that look github-generated from accepting suggestions:

https://github.com/Washington-University/HCPpipelines/commit/240e67729920409867a48b7de5778be8f665e70b

After accepting suggestions on github, you should "git pull --ff-only" in your local clone before making more commits.  Any edits made using github end up as new commits that are only on github, and you have to tell your local clone to go look for updates.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -412,46 +412,22 @@ for Hemisphere in L R ; do
...
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii
-        rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_MSMSulc.native.shape.gii
-
-        RegSphere="${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${Hemisphere}.sphere.MSMSulc.native.surf.gii"
+        mkdir -p "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc
+
+        cp "$AtlasSpaceFolder"/"$NativeFolder"/${Hemisphere}.sphere.rot.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc/${Hemisphere}.sphere_rot.surf.gii
+
+        ./global/scripts/MSMSulc.sh --subject-dir="$StudyFolder" --subject="$Subject" --regname="$RegName" --hemi "$Hemisphere"
```
glasserm, line 428: This needs to be an HCP Pipelines path.

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,119 @@
...
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+#description of script/command
+opts_SetScriptDescription "Run MSMSulc registration and save distortion outputs"
+
+opts_AddMandatory '--subject-dir' 'SubjectDir' 'path' "folder containing all subjects"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject-id"
+opts_AddMandatory '--regname' 'RegName' 'my reg' "set a new registration name"
+opts_AddOptional '--msm-conf' 'ConfFile' 'conf file' "provide the name of the configuration file, default MSMSulcStrainFinalconf" "$MSMCONFIGDIR"/MSMSulcStrainFinalconf
+opts_AddOptional '--hemi' 'Hemi' 'hemisphere' "provide hemisphere for registration, default L R" "L R"
```
glasserm, line 22: L=Left, R=Right, B=Both

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,119 @@
...
+	wb_command -surface-distortion "$NativeFolder"/"$Subject"."$Hemisphere".sphere.native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".sphere."$RegName".native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".EdgeDistortion_"$RegName".native.shape.gii -edge-method
+	wb_command -surface-distortion "$NativeFolder"/"$Subject"."$Hemisphere".sphere.native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".sphere."$RegName".native.surf.gii "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii -local-affine-method -log2
+
+	wb_command -metric-merge "$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_"$RegName".native.shape.gii -metric "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii -column 1
+	wb_command -metric-merge "$NativeFolder"/"$Subject"."$Hemisphere".StrainR_"$RegName".native.shape.gii -metric "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii -column 2
+
+	rm "$NativeFolder"/"$Subject"."$Hemisphere".Strain_"$RegName".native.shape.gii
+done
+
+if [[ "$Hemi" == *L* && "$Hemi" == *R* ]] ; then
```
glasserm, line 101: if "B"

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -412,46 +412,22 @@ for Hemisphere in L R ; do
...
-        ${CARET7DIR}/wb_command -metric-merge "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii -metric "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_MSMSulc.native.shape.gii -column 2
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainJ_MSMSulc.native.shape.gii
-        ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_MSMSulc.native.shape.gii
-        rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_MSMSulc.native.shape.gii
-
-        RegSphere="${AtlasSpaceFolder}/${NativeFolder}/${Subject}.${Hemisphere}.sphere.MSMSulc.native.surf.gii"
+        mkdir -p "$AtlasSpaceFolder"/"$NativeFolder"/MSMSulc
+
```
glasserm, line 425: This is duplicate with the line in the module.

glasserm (changes requested):
There are several issues that I identified.  Please fix.

global/scripts/MSMSulc.sh
coalsont: Not unless another variable gets involved.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
coalsont: There is a cp that is currently before the MSMSulc script which uses the directory, so removing this line and doing nothing else will cause an error.  This line will only ever create MSMSulc, while the one in the script may create whatever RegName when run separately.

Having an extra `mkdir -p` is not a problem, and means we don't have to rely on something in a subscript to see that the `cp` will succeed.  I think this simply shouldn't have been removed.

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,124 @@
...
+fi
+
+for Hemisphere in $Hemi ; do
+
+	if [[ "$Hemisphere" == "L" ]] ; then
+		Structure="CORTEX_LEFT"
+	elif [[ "$Hemisphere" == "R" ]] ; then
+		Structure="CORTEX_RIGHT"
+	else
+		log_Err_Abort "unrecognized hemisphere identifier '$Hemisphere', use L and/or R"
```
coalsont, line 73: Should probably update this error text.
```suggestion
		log_Err_Abort "unrecognized hemisphere identifier '$Hemisphere', use L, R, or B for both"
```

global/scripts/MSMSulc.sh
```diff
@@ -0,0 +1,124 @@
...
+if [[ "$RefData" == "" ]]
+then
+	RefData="$SurfaceTemplateFolder"/HEMISPHERE.refsulc.164k_fs_LR.shape.gii
+fi
+
+#Make MSMSulc Directory
+mkdir -p "$NativeFolder"/"$RegName"
+
+#Loop through left and right hemispheres
+if [[ "$Hemi" == "B" ]]
```
coalsont, line 61: More bulletproof, prevents "L B" from causing the error message below by replacing any string containing B with just "L R".
```suggestion
if [[ "$Hemi" == *B* ]]
```

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -412,46 +412,22 @@ for Hemisphere in L R ; do
     ${CARET7DIR}/wb_command -metric-math "ln(var) / ln (2)" "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_FS.native.shape.gii -var var "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".StrainR_FS.native.shape.gii
     rm "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".Strain_FS.native.shape.gii
 
+    #Calculate Affine Transform and Apply
+    ${CARET7DIR}/wb_command -surface-affine-regression "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.reg.reg_LR.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/rotate.${Hemisphere}.mat
+    ${CARET7DIR}/wb_command -surface-apply-affine "$AtlasSpaceFolder"/"$NativeFolder"/${Subject}.${Hemisphere}.sphere.native.surf.gii "$AtlasSpaceFolder"/"$NativeFolder"/rotate.${Hemisphere}.mat "$AtlasSpaceFolder"/"$NativeFolder"/${Hemisphere}.sphere_rot.surf.gii
+    ${CARET7DIR}/wb_command -surface-modify-sphere "$AtlasSpaceFolder"/"$NativeFolder"/${Hemisphere}.sphere_rot.surf.gii 100 "$AtlasSpaceFolder"/"$NativeFolder"/${Hemisphere}.sphere.rot.native.surf.gii
```
coalsont, line 418: Bug, MSMSulc.sh looks for a different filename.
```suggestion
    ${CARET7DIR}/wb_command -surface-modify-sphere "$AtlasSpaceFolder"/"$NativeFolder"/${Hemisphere}.sphere_rot.surf.gii 100 "$AtlasSpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".sphere.rot.native.surf.gii
```

coalsont: Also apply this name fix to the source of the cp below.

PR #291, opened 2024-04-29
Diffusion echo spacing units

coalsont:
Allows using harmonized time units across pipelines.

### DIFF
--- a/DiffusionPreprocessing/DiffPreprocPipeline.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline.sh
@@ -260,7 +260,7 @@ then
     log_Err_Abort "You must not specify both --echospacing-seconds and --echospacing"
 fi
 
-#internally, other scripts expect milliseconds
+#internally, PreEddy script expects milliseconds
 if [[ "$echospacingsec" != "" ]]
 then
     echospacingmilli=$(echo "$echospacingsec * 1000" | bc -l)
--- a/DiffusionPreprocessing/DiffPreprocPipeline.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline.sh
@@ -184,7 +184,8 @@ opts_AddMandatory '--posData' 'PosInputImages' 'data_RL1@data_RL2@...data_RLn' "
 
 opts_AddMandatory '--negData' 'NegInputImages' 'data_LR1@data_LR2@...data_LRn' "An @ symbol separated list of data with 'negative' phase encoding direction; e.g., data_LR1@data_LR2@...data_LRn, or data_AP1@data_AP2@...data_APn"
 
-opts_AddMandatory '--echospacing' 'echospacing' 'Number in msec' "Echo spacing in msecs"
+opts_AddOptional '--echospacing-seconds' 'echospacingsec' 'Number in sec' "Echo spacing in seconds, REQUIRED (or deprecated millisec option)"
+opts_AddOptional '--echospacing' 'echospacing' 'Number in millisec' "DEPRECATED: please use --echospacing-seconds"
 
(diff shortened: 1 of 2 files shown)


DiffusionPreprocessing/DiffPreprocPipeline.sh
```diff
@@ -248,6 +249,36 @@ then
...
+then
+    log_Err_Abort "You must specify --echospacing-seconds or --echospacing"
+fi
+
+if [[ "$echospacing" != "" && "$echospacingsec" != "" ]]
+then
+    log_Err_Abort "You must not specify both --echospacing-seconds and --echospacing"
+fi
+
+#internally, other scripts expect milliseconds
```
mharms, line 263:
```suggestion
#internally, PreEddy script expects milliseconds
```

mharms (commented):
Looks good.  My only comment would be rather we want to label the `--echospacing` option as "DEPRECATED" in the usage text.

coalsont:
Do you have an opinion?  Are you suggesting it shouldn't be marked as deprecated?

I suppose we should also edit the Batch script to use seconds also.

mharms:
Yeah, I'm inclined to NOT label it as "deprecated".

demsarjure (approved):
Looks good.

coalsont:
@mharms So, the reason for implementing the seconds option was that milliseconds was inconsistent with the other pipelines.  This implies that we want users going forward to use seconds, to reduce confusion.  To me, deprecation just means "there is a better way to do this, please use it instead", with a possibility of removing the deprecated thing at some point (my thinking is whether it would benefit users for it to be removed, which may resolve to "probably never" in this case, and many other cases).

I am aware of other people having different definitions of deprecation ("deprecated things must be removed in X major versions" being one I strongly disagree with and have argued against), so I can understand it may sound scary to people with those kinds of expectations, but I don't have much sense of how much of an issue that is.

What do you see as the cost/benefit of saying "DEPRECATED" versus not?

mharms:
To me, the addition of the `--echospacing-seconds` is most useful in just calling additional attention to the units issue for those using the pipeline for the first time.   I think the odds that we actually remove the millisecond option are low (why break existing scripts for this?), and in that sense I think calling it "DEPRECATED" is unnecessary, as it may cause others that happen to notice it to wonder if they need to update their scripts for this.  Do we have any other options in other pipelines described as DEPRECATED currently?

glasserm:
I prefer to deprecate as well.  Let's have folks be consistent across pipelines in the future.

mharms:
If we are trying to consistent, should we go further and set up `PreEddy` to take both millisec and second-based options as well?

coalsont:
I don't expect anyone to manually call just PreEddy, I would say converting it is unnecessary work.  Similarly, if we wanted to convert it anyway, it would not need both options, either, as I don't think it is user-facing (in our usual conventions, it would be in a `scripts` subfolder).  Personally, I don't see much benefit to having diffusion split into three sub-scripts.

coalsont:
As for other pipelines using "DEPRECATED", I think this is the first time we are adding a parameter specifically to use different semantics than the previous equivalent parameter, so this situation may not have come up before.  Previous changes were mostly to rename options without changing their semantics (the way newopts handles simple renames, the old name isn't even listed in the help info, but still works).

mharms:
If you and @glasserm  want to include "DEPRECATED" in the usage text for `--echospacing` that's fine with me.

coalsont:
@demsarjure can you test this branch on diffusion data?

demsarjure:
Sure, I will fire up some diffusion tests on this branch.

demsarjure:
@coalsont I ran a couple of tests (some also used the new `echospacing-seconds` parameter) and there were no issues, so I think you can safely merge this.

PR #294, opened 2024-06-06
Icasso

bqrosen:
I added icasso as a non-default optional method for estimating ICs in the IcaFix pipeline.  After icasso is run, melodic is used for mixture modeling only and then FIX proceeds unchanged. It is currently only implemented for interpreted matlab mode. If hcp_fix_multi_run is called with --ica-method=ICASSO and {FSL_FIX_MATLAB_MODE} != 1, it will throw an error before starting the analysis.


ICAFIX/hcp_fix_multi_run
```diff
@@ -182,6 +189,7 @@ fi
 debug_disable_trap
 set +u
 source "${FSL_FIXDIR}/settings.sh"
+# source "${HCPPIPEDIR}/ICAFIX/settings.sh.WUSTL_CHPC2"
```
coalsont, line 192: I'm not sure we want this in here even as a comment.  I'm also not sure how relevant that CHPC2 config file is anymore.

ICAFIX/hcp_fix_multi_run
```diff
@@ -352,6 +360,10 @@ then
 	log_Err_Abort "--fallback-threshold is not allowed when --vol-wisharts=1 is used"
 fi
 
+if [[ ${FSL_FIX_MATLAB_MODE} != 1 ]] && [[ ${ICAmethod} == ICASSO ]];then
```
coalsont, line 363: We should probably remove this check and recompile.

ICAFIX/hcp_fix_multi_run
```diff
@@ -126,6 +130,9 @@ opts_AddOptional '--processing-mode' 'ProcessingMode' '"HCPStyleData" (default)
 
 opts_AddOptional '--concatenate-only' 'ConcatOnlyStr' 'TRUE or FALSE' "When set to TRUE the script stops after the concatination step. e.g., for use in experimental alternative multi-run denoising" 'FALSE'
 
+opts_AddOptional '--ica-method' 'ICAmethod' 'MELODIC or ICASSO' "Use single-pass MELODIC (default) or hundred-pass ICASSO consensus method for ICA" 'MELODIC'
```
coalsont, line 133: I don't see any sanity checking of this value, so if someone typed `--ica-method=ICASOO`, it looks like the script will neither error nor use either of these options.  My usual syntax for things like this is case:
```
case "$ICAmethod" in
    (foo|bar)
        #fine, don't error
        ;;
    (*)
        log_Err_Abort "not fine"
        ;;
esac
```
In general, I prefer case to elif when it would be repeatedly testing the same variable.

ICAFIX/hcp_fix_multi_run
coalsont: Also, this would have prevented interpreted octave from trying it, so not a good test condition, even if we wanted to not recompile for some reason.

bqrosen:
I think I addressed most of these in a new push which also includes multilevel icasso. The matlab mode issue is thornier. 

The matlab script optionally includes savefig, so if that option is invoked (as it is in the hcp_fix_multi_run) then octave will fail. I’m not sure if compiled matlab can save .fig's but I think it can. 

-Burke

ICAFIX/hcp_fix_multi_run
```diff
@@ -352,6 +359,17 @@ then
...
+    (MELODIC|ICASSO)
+        ;;
+    (*)
+        log_Err_Abort "--ica-method=${ICAmethod} must be MELODIC or ICASSO"
+        ;;
+esac
+if [[ ${FSL_FIX_MATLAB_MODE} == 0 ]] && [[ ${ICAmethod} == ICASSO ]];then
+	log_Err_Abort "--ica-method=ICASSO is not yet implemented for compiled matlab."
+fi
+
```
coalsont, line 372: Spaces vs tabs mismatch (the rest of the file seems to be tabs).

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+brainMaskFile = sprintf('%s/%s_brain_mask.nii.gz',ConcatFolder,concatfmri);
+if ~exist(vntsFile,'file');error('%s doesn''t exist!',vntsFile);end
+if ~exist(brainMaskFile,'file');error('%s doesn''t exist!',brainMaskFile);end
+
+% outputs
+outDir = sprintf('%s/%s.ica/filtered_func_data.ica',ConcatFolder,concatfmrihp);
+[~,~] = unix(['mkdir -p ' outDir]);
+
+%% load data, reshape to 2d, and apply brain mask
+vnts = double(niftiread(vntsFile));
```
glasserm, line 86: niftiread is not a function I am familiar with.  I typically use FSL's read_avw and save_avw to avoid adding dependencies.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+brainMask = logical(niftiread(brainMaskFile));
```
glasserm, line 87: Here too.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+
+%% save icasso outputs to match melodic
+% reshape betas (S_final) and z-scores back to 4-d
+mtxDim = [mtxDim(1) Dim];
+volDim = [volDim(1:3) Dim];
+S_final4d = reshape2d4d(S_final,volDim,brainMask,mtxDim);
+Z4d = reshape2d4d(Z,volDim,brainMask,mtxDim);
+
+% save betas and z-score as 4-d nifti volumes
+hdr = niftiinfo(vntsFile);
```
glasserm, line 168: Same here.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+hdr.ImageSize(end) = Dim;
+niftiwrite(single(S_final4d),sprintf('%s/melodic_oIC',outDir),hdr,'Compressed',true)
```
glasserm, line 170: Same here.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+niftiwrite(single(Z4d),sprintf('%s/melodic_IC',outDir),hdr,'Compressed',true)
```
glasserm, line 171: Same here.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+
+%% calculate FTmix spectra
+ts = struct();
+ts.Nsubjects = 1;
+ts.ts = A_final;
+[ts.NtimepointsPerSubject,ts.Nnodes] = size(ts.ts);
+ts_spectra = nets_spectra_sp(ts);
+
+% save A_final and spectra as tab-delimited text
+dlmwrite([outDir '/melodic_Tmodes'], ts.ts, 'delimiter', '\t');
```
glasserm, line 194: Is melodic_Tmodes really the same as melodic_mix?

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+[ts.NtimepointsPerSubject,ts.Nnodes] = size(ts.ts);
+ts_spectra = nets_spectra_sp(ts);
+
+% save A_final and spectra as tab-delimited text
+dlmwrite([outDir '/melodic_Tmodes'], ts.ts, 'delimiter', '\t'); 
+dlmwrite([outDir '/melodic_FTmix'], ts_spectra, 'delimiter', '\t');
+
+fid = fopen([outDir '/components.txt'],'w');
+fprintf(fid,'%i: Signal\n',1:size(S_final,2));fclose(fid);
+[~,~] = unix(['wb_command -cifti-create-scalar-series ' ...
```
glasserm, line 199: Need to do the same for the timeseries, right?  PostFix may also already take care of both of these.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,247 @@
...
+  sprintf('%s/melodic_FTmix %s/melodic_FTmix.sdseries.nii -transpose -name-file %s/components.txt -series HERTZ 0 %f',...
+  outDir,outDir,outDir,tr)]);% note: saving this file is the only thing tr is used for
+
+%% Mixture modeling
+% follows recipe from https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;6e85d498.1607
+fprintf('performing melodic mixture modeling ...\n')
+[~,~] = unix(sprintf('melodic -i %s/melodic_IC --ICs=%s/melodic_IC --mix=%s/melodic_mix -o %s --Oall --report -v --mmthresh=0.5',...
```
glasserm, line 206: I am assuming the contents of --mix doesn't matter, but this was my implementation to do mixture modelling on CIFTI data (perhaps that now works natively though?):

/media/myelin/brainmappers/HardDrives/2TBB/Connectome_Project/Pipelines/MixtureModelingRSN.sh

I also used a different --mmthresh=0.

glasserm (changes requested):
Please see my notes above.  Also, I am not a bit fan of aggressive line continuations.  They make code harder for me to read.  Screens are not tiny any more...

bqrosen:
While niftiread, niftiwrite, and niftiinfo are standard matlab toolbox functions, The script calls melodic so it’s always going to be FSL dependent. It looks like there was an attempt to add nifitread, niftiwrite, and niftiinfo to octave forge in 2022, https://savannah.gnu.org/patch/?9853, but it wan’t finished for some reason. I can swap to read_avw for better octave compatibility. 

That said, nifitread is at least twice as fast as read_avw for nii.gz because read_avw has to resave and reload the image to a temporary file to unzip it. nifitread can also read cifti cdata.  However, niftiinfo cannot read the cifti header (it returns an intent code error). niftiwrite can write nifti1 or nifti2, compressed or uncompressed, combined or with separate .hdr. 

-Burke

ICAFIX/scripts/run_icasso.m
glasserm: I didn't know that.  Perhaps could check if the function is present and fall back to read_avw/save_avw if not?

bqrosen:
Sounds good. I’ll update the matlab code in the multiecho fitting to do that as well. 

-Burke

bqrosen:
I double checked and they are. I’ll make that more explicit in the code. 

-Burke

bqrosen:
I checked (with diff) and changing --mix only alters the report .html’s and the log.txt so you’re right, it doesn’t matter.

On --mmthresh, I think you’re right for generic mixture modeling that reporting the unthresholding z’s is probably best, according to the melodic documentation <https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/MELODIC>, 0.5 is the value to choose to emulate normal, non-mixture-modeling-only melodic usage. That is also consistent with the mailing list post <https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;6e85d498.1607> I commented.  
-Burke

glasserm:
Does that threshold the data?

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,253 @@
...
+
+%% check IO dependencies
+% use imaging processing toolbox utilities, or if those are not available use FSL utilities 
+function out = out2(fun);[~,out] = fun();end
+function out = out3(fun);[~,~,out] = fun();end
+if isempty(which('niftiread'))
+  infoNIFTI = @(fName) struct('ImageSize',out2(@() read_avw(fName))','PixelDimensions',out3(@() read_avw(fName))');
+  readNIFTI = @(fName) read_avw(fName);
+  writeNIFTI = @(img,fName,hdr) save_avw(img,fName,'f',hdr.PixelDimensions);
+  if isempty(which('read_avw'))
```
coalsont, line 83: Seems like it would be better to test for this before defining things that use the _avw functions.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,253 @@
...
+%% parse paths
+% inputs
+vntsFile = sprintf('%s/%s_vnts.nii.gz',ConcatFolder,concatfmrihp);
+brainMaskFile = sprintf('%s/%s_brain_mask.nii.gz',ConcatFolder,concatfmri);
+if ~exist(vntsFile,'file');error('%s doesn''t exist!',vntsFile);end
+if ~exist(brainMaskFile,'file');error('%s doesn''t exist!',brainMaskFile);end
+
+% outputs
+outDir = sprintf('%s/%s.ica/filtered_func_data.ica',ConcatFolder,concatfmrihp);
+[~,~] = unix(['mkdir -p ' outDir]);
```
coalsont, line 101: https://www.mathworks.com/help/matlab/ref/mkdir.html

Please don't use shell calls, and particularly not the overly-specific `unix()`, to do filesystem stuff that is built into the language.  You should be able to use the status-outputting function form, `if ~mkdir(outDir); error('unable to make output folder'); end` to avoid the "already exists" warning and provide an informative message.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,253 @@
...
+
+%% load data, reshape to 2d, and apply brain mask
+vnts = double(readNIFTI(vntsFile));
```
coalsont, line 104: It would be better if the comment explained why it is being converted to double, since it is likely it was written to disk as single.  The masking and reshaping are somewhat evident from the code (though perhaps the function name could be improved to something like `maskAndSpatiallyFlatten`), and therefore are not as deserving of an explanatory comment.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,253 @@
...
+
+  imgDim = size(img);
+  mtx = reshape(img,prod(imgDim(1:3)),imgDim(4));
+  mtxDim = size(mtx);%
+  if nargin == 2
+    mtx = mtx(logical(msk(:)),:);
+  end
+end
+
+function img = reshape2d4d(mtx,imgDim,msk,mtxDim)
```
coalsont, line 238: The "2d4d" versus "4d2d" naming difference is not very obvious.  Similar to my other naming suggestion, this could be called "unflattenData" instead.

fMRIVolume/scripts/multiEchoCombine.m
```diff
@@ -94,11 +95,28 @@ function multiEchoCombine(tscPath,TE,refPath,sctPath,T2starMethod,weightMethod,f
...
-hdr = niftiinfo(tscPath);
+%% set nifti IO functions
+% use imaging processing toolbox utilities, or if those are not available, use FSL utilities 
+function out = out2(fun);[~,out] = fun();end
+function out = out3(fun);[~,~,out] = fun();end
+if isempty(which('niftiread'))
+  infoNIFTI = @(fName) struct('ImageSize',out2(@() read_avw(fName))','PixelDimensions',out3(@() read_avw(fName))');
+  readNIFTI = @(fName) read_avw(fName);
+  writeNIFTI = @(img,fName,hdr) save_avw(img,fName,'f',hdr.PixelDimensions);
+  if isempty(which('read_avw'))
```
coalsont, line 106: Again, it is more obvious to test that something exists before the lines that reference it.

global/matlab/mixtureModel.m
```diff
@@ -0,0 +1,92 @@
...
+% Dependencies:workbench, FSL
+% Written with workbench 2.0 and FSL 6.0.7.1
+%
+% ToDo: 
+% Feed arbitrary melodic arguments. 
+% Return workbench errors
+% Save melodic log?
+
+%% handle inputs 
+if nargin < 2 || isempty(outFile); outFile = inFile; end
```
coalsont, line 30: I don't like defaults that destroy their own input data files.  I think outFile should be required.

global/matlab/mixtureModel.m
```diff
@@ -0,0 +1,92 @@
+function mixtureModel(inFile,outFile,wbcmd,melocmd)
```
coalsont, line 1: This looks like it should be a bash script, I don't see any matrix math being done internally.  Perhaps the only reason it is a matlab script is because the place you currently want to call it from is matlab code?  Let's not use proprietary languages when they aren't the most obvious choice for the task.

ICAFIX/hcp_fix_multi_run
```diff
@@ -352,6 +359,17 @@ then
...
 fi
 
+case "$ICAmethod" in
+	(MELODIC|ICASSO)
+		;;
+	(*)
+		log_Err_Abort "--ica-method=${ICAmethod} must be MELODIC or ICASSO"
+		;;
+esac
+if [[ ${FSL_FIX_MATLAB_MODE} == 0 ]] && [[ ${ICAmethod} == ICASSO ]];then
```
coalsont, line 369: Again, I don't think we want this type of check in the committed version.  We can compile it as a final step before merging.

ICAFIX/scripts/run_icasso.m
```diff
@@ -0,0 +1,253 @@
+function run_icasso(Dim,concatfmri,concatfmrihp,ConcatFolder,vis,nICA,nClust,maxIter)
+run_icasso(Dim,concatfmri,concatfmrihp,ConcatFolder,vis,nICA,maxIter)
```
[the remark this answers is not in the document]
bqrosen: Yep. Fixed.

ICAFIX/scripts/run_icasso.m
bqrosen: Fixed.

ICAFIX/scripts/run_icasso.m
bqrosen: fastica needs double. But I typically cast data as double as soon as it is in memory to reduce any downstream rounding errors, and then covert back to single before storing again, unless I am RAM limited. It this script the placement of the recasting doesn't matter so long as it is before fastica. I added a comment indicating it is for fastica.

ICAFIX/scripts/run_icasso.m
bqrosen: Renamed.

ICAFIX/scripts/run_icasso.m
bqrosen: function renamed.

ICAFIX/scripts/run_icasso.m
bqrosen: I'll swapped unix for system moving forward. That's a neat mkdir trick, revised.

fMRIVolume/scripts/multiEchoCombine.m
bqrosen: Fixed.

global/matlab/mixtureModel.m
bqrosen: You are right, though I would argue that it also makes it more easy to debug 🐛. It is designed to be called from matlab scripts because in most use cases (It is supposed to be a more general utility, not just a helper function for this run_icasso) its inputs will be created by a matlab script. Also, in the future an option could be added to feed it a matrix directly rather than a file path, which bash couldn't do.

global/matlab/mixtureModel.m
bqrosen: Made required. Now returns help if nargin <2

global/matlab/mixtureModel.m
coalsont: If by "feed it a matrix" you mean "write that matrix to a file, run melodic, etc on it, and maybe read it back in later", I would still argue that the command sequence should be in a bash script, for more legibility, and potentially to be able to run the command sequence from a different script without having matlab/octave in the way.  If you wanted it to look more matlabby, you could write a secondary matlab wrapper that writes a matrix to a file and runs the bash script.  Unless a process actually uses matlab to do the math, I don't want its implementation to depend on matlab.  The less matlab we use in the pipelines, the easier they are to maintain and use (compiled matlab, MCR version, etc).

ICAFIX/scripts/run_icasso.m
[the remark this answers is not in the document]
coalsont: You should push these fixes to the PR branch so we can see them.

ICAFIX/scripts/run_icasso.m
```diff
@@ -130,19 +132,6 @@ function run_icasso(Dim,concatfmri,concatfmrihp,ConcatFolder,vis,nICA,nClust,max
 
 %% save icasso figures
 % Note: only the figures from the final level of icasso are saved
```
coalsont, line 134: Orphaned comments

global/matlab/mixtureModel.m
```diff
@@ -27,11 +27,11 @@ function mixtureModel(inFile,outFile,wbcmd,melocmd)
 % Save melodic log?
 
 %% handle inputs 
-if nargin < 2 || isempty(outFile); outFile = inFile; end 
+if nargin < 2 || isempty(inFile) || isempty(outFile); help mixtureModel; return; end 
 if nargin < 3 || isempty(wbcmd); wbcmd = 'wb_command'; end 
 if nargin < 4 || isempty(melocmd); melocmd = 'melodic'; end 
-[wbStat,~] = unix(wbcmd);
-[meloStat,~] = unix(['which ' melocmd]);
+[wbStat,~] = system(wbcmd);
```
coalsont, line 33: Using `which` is probably a better idea than trusting that wb_command will never treat "no arguments, so print the help info" as a non-error condition.

global/matlab/mixtureModel.m
```diff
@@ -0,0 +1,92 @@
...
+% Dependencies:workbench, FSL
+% Written with workbench 2.0 and FSL 6.0.7.1
+%
+% ToDo: 
+% Feed arbitrary melodic arguments. 
+% Return workbench errors
+% Save melodic log?
+
+%% handle inputs 
+if nargin < 2 || isempty(inFile) || isempty(outFile); help mixtureModel; return; end
```
coalsont, line 30: Showing the help info isn't the usual matlab behavior, nor is doing a standard return.  You could use something like `error('not enough arguments')`, or you could just do nothing and wait for the `outFile0 = outFile;` line to throw an undefined error.

global/matlab/mixtureModel.m
coalsont: I suppose `-version` would be fairly safe, and would also work on windows.  But, as long as the melodic check uses `which`, might as well use the same approach for both.

PR #297, opened 2024-08-20
--fix-only option for hcp_fix_multi_run

demsarjure:
This feature implements a new parameter (`--fix-only`) for `hcp_fix_multi_run`. This parameter can be used when to speed up ICAFix optimizations by skipping melodic and proceed with FIX on a previously computed ICA solution.


ICAFIX/hcp_fix_multi_run
```diff
@@ -379,116 +385,352 @@ log_Msg "PWD: $DIR"
...
-MeanCIFTIARRAY=()
-VNCIFTIARRAY=()
-MovementNIFTIMergeARRAY=()
-MovementNIFTIhpMergeARRAY=()
-MovementTXTMergeARRAY=()
+## ===========================================================================
+## Skip all steps until FIX if RunFIXOnly is set to TRUE
+## ===========================================================================
+
+if [[ "${RunFIXOnly}" == "FALSE" ]]; then
```
coalsont, line 393: Please use `opts_StringToBool` for sanity checking, case insensitivity, and user friendliness, see here:

https://github.com/Washington-University/HCPpipelines/blob/9cfa37a378dff26fd5aa4fd2aad24d0136830f0f/global/scripts/RSNregression.sh#L72
https://github.com/Washington-University/HCPpipelines/blob/9cfa37a378dff26fd5aa4fd2aad24d0136830f0f/global/scripts/RSNregression.sh#L211

ICAFIX/hcp_fix_multi_run
```diff
@@ -498,317 +740,97 @@ for fmri in "${fmris[@]}" ; do
...
-log_Msg "melodic_cmd: ${melodic_cmd[*]}"
-"${melodic_cmd[@]}"
-return_code=$?
-log_Msg "melodic has been run: return_code = ${return_code}"
-log_Debug_Msg "melodic has been run: Contents of ${concatfmrihp}.ica follow"
-if [ ! -z "${log_debugOn}" ] ; then
-	ls -lRa ${concatfmrihp}.ica
-fi
+	if [ -d "${concatfmrihp}.ica/fix" ]; then
+		mv "${concatfmrihp}.ica/fix" "${concatfmrihp}.ica/fix-bkp-$(date +"%Y-%m-%d.%H%M%S")"
```
coalsont, line 829: Do we want the pipeline to internally make a backup?  Granted, this option is a special case, but we generally don't like making backups the user doesn't know about.

ICAFIX/hcp_fix_multi_run
```diff
@@ -124,6 +128,8 @@ opts_AddConfigOptional '--icadim-mode' 'icadimmode' 'icadimmode' '"default" or "
 
 opts_AddOptional '--processing-mode' 'ProcessingMode' '"HCPStyleData" (default) or "LegacyStyleData"' "controls whether --icadim-mode=fewtimepoints is allowed" 'HCPStyleData'
 
+opts_AddOptional '--fix-only' 'RunFIXOnly' 'TRUE or FALSE' 'If TRUE, it skips running ICA using melodic and proceeds directly with running FIX. A previous ICA solution has to exist!' 'FALSE'
```
coalsont, line 131: Consider whether `--skip-ica` or `--reuse-existing-ica` or something might be a better name.  "Only" implies that it can never be combined with other possible process-altering flags (and that if we add another step after fix, `--fix-only` might be expected to skip it, while `--skip-ica` would be expected to run it).

ICAFIX/hcp_fix_multi_run
sivcek: Backing up previous solution is important in the process of iteratively building and evaluating FIX models, which is the use case that this option is made for. We could add another optional parameter to give the user more control, but I suspect that whoever would use `--fix-only` would also want to have the previous fix solutions backed up. To make the backup explicit, I have now added a message in the log (it will be in the next PR):

```
        if [ -d "${concatfmrihp}.ica/fix" ]; then
		fix_backup_suffix="-bkp-$(date +"%Y-%m-%d.%H%M%S")"
		mv "${concatfmrihp}.ica/fix" "${concatfmrihp}.ica/fix${fix_backup_suffix}"
		log_Msg "previous fix solution found and renamed to fix${fix_backup_suffix}"
	fi
```

glasserm: Perhaps that behavior should be a separate option?

ICAFIX/hcp_fix_multi_run
sivcek: I agree that `--skip-ica` is clearer. I changed the code for the new version of the PR.

ICAFIX/hcp_fix_multi_run
sivcek: Ok. `opts_StringToBool` is used in the upcoming new PR.

ICAFIX/hcp_fix_multi_run
sivcek: @glasserm, we can make it a separate option, but I suspect it will be always used in the use cases for `--skip-ica`.

sivcek: Also, the option only makes sense when using `--skip-ica` as with a new ica the old fix does not refer to the same components any more.

ICAFIX/hcp_fix_multi_run
coalsont: Personally, I like the descriptiveness of `--reuse-existing-ica`, so the user doesn't think they can just skip the ica on the first run.  "reuse" has some parallels to the tICA modes, too.

To be clear, please push the commits that make these changes to the same branch on github, and they will show up here.  Please don't open an additional PR for changes requested on this PR.

glasserm:
For some reason this diff is really hard to interpret.  Any idea why it is a sea of red and green even though I assume the changes were pretty minor?

glasserm:
Also, it looks like there are conflicts that need to be resolved.

coalsont:
Indentation.  In the gear button on the "files changed" page, check "hide whitespace".

Github says no conflicts for squash, but there are some changes from `-volume-math`/`-volume-merge` to `fslmaths`/`fslmerge`, which I am unsure of the purpose of.  If they weren't intentional, something may have been done something with the branch history that could cause a regression (for instance, copying an old version of a file, making edits, and then committing it to the current branch state after a reset).

ICAFIX/hcp_fix_multi_run
```diff
@@ -379,116 +385,352 @@ log_Msg "PWD: $DIR"
...
 	fi
 
-	#Demean movement regressors
-	demeanMovementRegressors Movement_Regressors.txt Movement_Regressors_demean.txt
-	MovementTXTMergeARRAY+=("$(pwd)/Movement_Regressors_demean.txt")
+	ConcatNameNoExt=$($FSLDIR/bin/remove_ext $ConcatName)  # No extension, but still includes the directory path
+
+	# Merge volumes from the individual runs
+	fslmerge -tr "${ConcatNameNoExt}_demean" "${NIFTIvolMergeARRAY[@]}" $tr
+	fslmerge -tr "${ConcatNameNoExt}_hp${hp}_vnts" "${NIFTIvolhpVNMergeARRAY[@]}" $tr
```
coalsont, line 636: With "hide whitespace", you can see these used to be `-volume-merge`s.  Was this intentional?  Perhaps to reduce memory footprint?

glasserm: Perhaps they were working from an old version of the code?  The commits are from a while ago...

ICAFIX/hcp_fix_multi_run
```diff
@@ -498,317 +740,97 @@ for fmri in "${fmris[@]}" ; do
...
-#$FSLDIR/bin/immv ${ConcatNameNoExt}_hp${hp}_vnts ${ConcatNameNoExt}_hp${hp}_vnts0
-
-# Nor do we need the concatenated demeaned time series (either volume or CIFTI)
-log_Msg "Removing the concatenated demeaned time series"
-"$FSLDIR/bin/imrm" "${ConcatNameNoExt}_demean"
-/bin/rm -f "${ConcatNameNoExt}_Atlas_demean.dtseries.nii"
-
-if ((ConcatOnly)); then
-	exit
-fi
```
glasserm, line 689: This should not be reverted.

glasserm (changes requested):
This pull request seems to be making a lot more changes to the code than are actually needed for the requested functionality.  It could be that it is reverting changes that were made subsequent to whenever these commits were first made (it appears they are from some months back).  We need a pull request based on the latest master.

sivcek:
I have committed the version with the changes requested:
* adding the functionality to the latest master to avoid reverting existing code
* changing `--fix-only` to `--skip-ica`
* using `opts_StringToBool` to process the input string
* adding message reporting that an old fix solution was found and backed up

glasserm (approved):
Thanks this is much cleaner.

ICAFIX/hcp_fix_multi_run
```diff
@@ -379,116 +387,365 @@ log_Msg "PWD: $DIR"
...
-MeanVolARRAY=()
-VNVolARRAY=()
-CIFTIMergeARRAY=()
-CIFTIhpVNMergeARRAY=()
-MeanCIFTIARRAY=()
-VNCIFTIARRAY=()
-MovementNIFTIMergeARRAY=()
-MovementNIFTIhpMergeARRAY=()
-MovementTXTMergeARRAY=()
+if ((!SkipICA))
```
coalsont, line 394: At least some versions of bash don't like this, because `!` is also a history special character.  Add a space after the `!`.
```suggestion
if ((! SkipICA))
```

sivcek: Done.

ICAFIX/hcp_fix_multi_run
coalsont: @glasserm Are we okay leaving the automatic backup stuff mandatory in this mode?  Log messages are not all that likely to get the user's attention unless the script fails.

If we are still fine with this (perhaps because of the limited use case of this feature), then I think it is ready for merge.

glasserm: I would prefer a separate option if that is okay.  I don't think one would always want to back up results.

coalsont: @demsarjure @sivcek Can you add a `--backup-fix-if-ica-skipped` boolean option that defaults to true?  Or, we could make a more general `--backup-existing-fix` option that defaults to false, and also works when not using `--skip-ica`.

mharms: What happened to @coalsont's previous suggestion for an alternative name to `--skip-ica`?  I don't recall exactly what it was, but I do recall that he made a good argument for his alternative name.

coalsont: Or, to make it more scriptable, it could be something like `--fix-backup-suffix=<string>` or `--fix-backup-folder=<fix-bkp-something>`, and if it isn't null, names the backup folder accordingly, rather than generating it from `date`.  The second form could even take an absolute path, or be used to put all the backups into a `backups/` subfolder instead of littering the `${concatfmrihp}.ica` folder directly.

mharms: Ok, found it.  It was `--reuse-existing-ica`:
https://github.com/Washington-University/HCPpipelines/pull/297#discussion_r1723691813

coalsont: @mharms I had resolved it as somewhat minor and we seemed close to merging, but I have unresolved it again since you seem to have an opinion on it.

coalsont:
The unresolved items as I see it are:

- @glasserm suggested the backup saving should be a separate option, and I suggested having the name of the backup be passed into the script, which is more predictable than being generated from the date/time at which it reached that line in the script.
- @mharms seconded (I think?) my `--reuse-existing-ica` suggestion for the flag name

sivcek:
This update:

* merges the current state of master into the PR
* renames `--skip-ica` option to `--reuse-existing-ica`
* adds `--fix-backup` option that specifies the folder to which to backup a previous fix solution if one exists

The update is currently being tested.

ICAFIX/hcp_fix_multi_run
```diff
@@ -419,427 +430,458 @@ log_Msg "PWD: $DIR"
...
+
+	# Switch to concat folder
+	cd ${ConcatFolder}
+
+    if [ -d "${concatfmrihp}.ica/fix" ] && [ -n "$FIXBackup" ]; then
+        pushd "${concatfmrihp}.ica"
+        mv fix $FIXBackup
+        popd
+		log_Msg "Previous fix solution found and moved to ${$FIXBackup}."
 	fi
```
coalsont, line 883: This has quite a mix of tabs and spaces.

sivcek: I have changed everything to spaces in the latest commit.

sivcek:
The latest commit resolves a bug and improves backup. If backup is initiated, it now looks for presence of scene files, creates their zipped version and adds them to the backup folder. 

This version is ready to be merged if there are no additional requests.

ICAFIX/hcp_fix_multi_run
```diff
@@ -419,428 +430,482 @@ log_Msg "PWD: $DIR"
...
+        fi
+
+        # Back up previous fix solution
+        if [ -d "${concatfmrihp}.ica/fix" ]; then
+            mv "${concatfmrihp}.ica/fix"/* "${FIXBackupFolder}"
+            log_Msg "Previous fix solution found and moved to ${FIXBackupFolder}."
+        fi
+
+        # Zip and back up previous scene files
+        for SceneFile in $(ls *.scene); do
```
coalsont, line 901:
```suggestion
        for SceneFile in *.scene; do
```
Globbing is whitespace-safe, use it without feeding it through `ls`.

coalsont: Also, if there is a substantial chance that there won't be any scene files, you should use `shopt -s nullglob` near the top of the file (like right under `set -eu`), to avoid "file not found: *.scene", but then you DEFINITELY need to avoid `ls`, as it does something completely different when passed zero arguments.

glasserm:
This PR is back to being impossible to follow for me.

coalsont:
> This PR is back to being impossible to follow for me.

You may have lost the "hide whitespace" setting (or turned it of for a different diff/PR).  Click the gear on the "files changed" tab.

glasserm:
Seems fine after that fix.

ICAFIX/hcp_fix_multi_run
sivcek: @coalsont thank you for spotting this! It is now changed as suggested.

sivcek:
The latest commit includes a merge from the current master and a change proposed by @coalsont regarding globbing for presence of scene files to be backed up.

PR #298, opened 2024-09-14
Transmit bias pipeline

coalsont:
I will set up compiled matlab before this gets merged.  Start with the `Examples` script to see the overall flow/settings.


glasserm (commented):
Why did we add a second copy of the 164k meshes to this?  We already had these meshes in the pipelines.

coalsont:
> Why did we add a second copy of the 164k meshes to this? We already had these meshes in the pipelines.

Because the existing names were inconsistent and confusing (and not as easily scriptable).  It just happened to be in this PR because that's what I was editing when I decided it was time to do finally something about it.

demsarjure:
When implementing the interface I noticed a minor inconsistency, `scanner-grad-coeffs` from this pipeline is called `gdcoeffs` in PreFS.

coalsont:
`--grayordinates-res` is more readable and is obviously related to `--grayordinatesres`.  I don't know that we expect anyone to copy and paste option flags between any other pipeline and TransmitBias (the batch launcher handles the argument names).  `--gdcoeffs` being harder to decipher is the only reason it wouldn't be obvious to a user that this name means the same thing.  If there were inconsistencies between MR FIX and reapply, it would matter because they are closely related, but I don't think it matters here.

mharms:
It's not just PreFS that uses `--gdcoeffs`.  fMRIVolume, and Diffusion use that as well.
Personally, I think consistency in the flag names is preferred.

glasserm (approved):
Thanks Tim.

coalsont:
I added support in case someone copy-pastes `--gdcoeffs` or `--grayordinatesres` from another pipeline.  I'd prefer to keep the more readable names in the help info.

PR #299, opened 2024-09-17
Update FSL contribution to PATH

MichielCottaar:
As of FSL 6.0.6 (release date: 22 nov 2022), the binaries are stored in $FSLDIR/share/fsl/bin, not $FSLDIR/bin

See https://fsl.fmrib.ox.ac.uk/fsl/docs/#/development/history/index?id=_606-22nd-november-2022 for more details

Including $FSLDIR/bin in the $PATH will include all the FSL-specific binaries, but also a lot of other non-FSL related ones. This can lead to confusing behaviour, where FSL versions of generic tools will be used after running `source SetupHCPPipeline.sh` rather than built-in versions.

### DIFF
--- a/Examples/Scripts/SetUpHCPPipeline.sh
+++ b/Examples/Scripts/SetUpHCPPipeline.sh
@@ -93,7 +93,14 @@ then
 fi
 
 # Add the specified versions of some things to the front of $PATH, so we can stop using absolute paths everywhere
-export PATH="$CARET7DIR:$FSLDIR/share/fsl/bin:$PATH"
+if [ -d "$FSLDIR/share/fsl/bin" ] ; then
+    # For FSL 6.0.6 (release date: 22 Nov 2022) or later
+    export PATH="$FSLDIR/share/fsl/bin:$PATH"
+else
+    # For older versions of FSL
+    export PATH="$FSLDIR/bin:$PATH"
+fi
+export PATH="$CARET7DIR:$PATH"
 
 # Source extra stuff that pipelines authors may need to edit, but users shouldn't ever need to
 # by separating them this way, a user can continue to use their previous setup file even if we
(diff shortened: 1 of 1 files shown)


coalsont:
This looks like it would make the pipelines incompatible with FSL older than 2 years.  Perhaps this should check whether `share/bin` exists and decide based on that what to add to `$PATH`.

MichielCottaar:
That is a good point. I guess it would be slightly better to check the version number of FSL, but I'm not aware of an easy way to compare version numbers in raw bash. So, checking `$FSLDIR/share/fsl/bin` sounds like a good idea. I'll adjust this PR to do that.

Examples/Scripts/SetUpHCPPipeline.sh
```diff
@@ -93,7 +93,15 @@ then
...
-export PATH="$CARET7DIR:$FSLDIR/bin:$PATH"
+if [[ -d "$FSLDIR/share/fsl/bin" ]] ; then
+    # For FSL 6.0.6 (release date: 22 Nov 2022) or later
+    export PATH="$FSLDIR/share/fsl/bin:$PATH"
+else
+    # For older versions of FSL
+    export PATH="$FSLDIR/bin:$PATH"
+fi
+export PATH="$CARET7DIR:$PATH"
+echo $PATH
```
coalsont, line 104: I don't know that we want it to echo the new `$PATH`, particularly without saying what it is that is being reported.

MichielCottaar: Oops, I just included that for some testing locally. I never meant for it to be included. I've removed it in the latest commit.

glasserm:
I am okay with accepting this.  Thanks Michiel for contributing.

PR #300, opened 2024-09-17
Updated cortical thickness and curvature regression

nagedem:
Bunch of new files and some changes in PostFreeSurfer.


PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -70,6 +70,7 @@ opts_AddOptional '--inflatescale' 'InflateExtraScale' 'number' "surface inflatio
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData|LegacyStyleData' "disable some HCP preprocessing requirements to allow processing of data that doesn't meet HCP acquisition guidelines - don't use this if you don't need to" 'HCPStyleData'
 opts_AddOptional '--structural-qc' 'QCMode' 'yes|no|only' "whether to run structural QC, default 'yes'" 'yes'
 opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the subject's myelin map as reference map's myelin map mean, defaults to 'YES'" 'YES'
+opts_AddOptional '--metric-regression' 'MetricReg' 'OLD, NEW or BOTH' "whether to use the updated curvature-thickness regression, defaults to 'B=BOTH'" 'B'
```
coalsont, line 73: Why do we want an option to not generate one of these?  That would mean any later pipelines that use thickness and don't want curvature effects would have to be adapted to what the user chose for this setting.

I would think we want both to always be generated, without giving the user a choice to make their dataset more complicated for later handling.

glasserm: The old method is deprecated because it is substantially inferior.  We are not using corrthickness much in pipelines (I think only the IDP pipeline) other than resampling it around.

coalsont: The old version is easy to compute and not a large file.  I don't see an obvious downside to keeping backward compatibility.  I assume we don't want to use the old filename for the new output.  What exactly are you suggesting by "deprecated"?

glasserm: It's purpose is entirely superseded by the new version.

coalsont: Okay, and because of that...what do you want the pipelines to *do* about it?

glasserm: In a new study, I suspect one would only generate the new version.

glasserm, line 73: You need to actually convert BOTH to B if you want to test on "BOTH."  Honestly, it would be simpler to just use BOTH instead of B throughout.

coalsont: Not generating the old version is adding a difference of a "missing" file to the dataset based on when it was processed.  What is the advantage of that situation?  What if external users have been using that file, and want to replicate the methods of their old study on a different dataset?

I see more downsides than advantages to not generating the old version.  I guess the main downside I see is possible user confusion about what file to use.

coalsont: Regardless, I'm not sure there should be a way for the user to not run the new method (is the worry just that they don't have python set up for it?).  If we decide we really don't want people to run the old method by default, an option to turn it back on might be reasonable (so they don't have to resort to an older pipelines release), but that can be just a yes/no option.

mharms: @coalsont makes a good point IMO about continuing to generate the "old" version.  Why not just automatically generate both?

glasserm: That is the current default.  We may change it in the future, as once this is available and published we will discourage use of the old method.  I think the options as defined currently are fine.

coalsont: @mharms the default at present is to generate both (though Matt seems to want to change that).  I am going further than "automatically" generating both, to "always" - why do we even want to give the user a friendly option to have "missing" files in their outputs?

glasserm: corrThickness isn't hardcoded anywhere as far as I know.  If it is, that will need to change because we would want to use this instead.  I think this discussion has run its course.

coalsont: We don't have a conclusion, so "run its course" doesn't work.  I think removing the option entirely and always generating both files results in the fewest edge cases for us to support.

glasserm: We are not doing that.  We will have a transitional period with both as a default.

coalsont: Based on what logic?  "Just do it my way" isn't very constructive.

In particular, why do we want the user to have the option to NOT generate the new version?

glasserm: That is the current behavior, so we are providing a way to support that.  We need to be able to make changes to already written pipelines.  We can have transitional periods for deprecated features.  corrThickness is obsolete with this new version.  I don't really understand why this is so controversial.  The pipelines need to be able to evolve with time.

coalsont: Huh?  I'm not saying the pipelines shouldn't change, where did you get that?  It is controversial because you aren't providing a reason for your strong position.

Why does **the user** need an option to **not run** the new method?  It doesn't overwrite a previously-released filename with different contents, right?  So the new files can't interfere with any old code (unless the user happened to already generate files with those new names), right?  If so, then why do we want an "old behavior" mode, when the new behavior doesn't change any files that the previous version generated?  Why would a user want that?

coalsont: Is the concern that someone won't like the added runtime from the new version?

glasserm: It is quite a bit longer.

coalsont: People will generally only run PostFS after they have run FreeSurfer.  How does it compare to the runtime of FreeSurfer?

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,133 @@
...
+		  python CorrThick.py "$SubjectDir" "$Subject" "$Structure" "$Hemisphere" "$Surface" "$PatchSize" "$SurfSmooth" "$MetricSmooth"
+		)
+done	
+
+#Set the Color Palette(s) and Resample to HighResMesh and LowResMesh
+for Hemisphere in $Hemi ; do
+
+	for Map in $MapListFunc ; do
+	  if [[ "$Map" == MRcorrThickness || "$Map" == MRcorrThickness_intercept ]] ; then
+		wb_command -metric-palette "$NativeFolder"/"$Subject"."$Hemisphere"."$Map".native.shape.gii MODE_AUTO_SCALE_PERCENTAGE -pos-percent 4 96 -interpolate true -palette-name videen_style -disp-pos true -disp-neg false -disp-zero false -normalization NORMALIZATION_SELECTED_MAP_DATA
```
coalsont, line 75: You are mixing tabs and spaces here, please use a consistent indendation method.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,133 @@
...
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+#description of script/command
+opts_SetScriptDescription "Run curvature corrected thickness python script and save curvatures, regression coefficients, and curvature-corrected thickness"
+
+opts_AddMandatory '--subject-dir' 'SubjectDir' 'path' "folder containing all subjects"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject-id"
+opts_AddMandatory '--regname' 'RegName' 'my reg' "set the desired registration name(s) separated by @, 'string' 'RegName@RegName@RegName@...etc.' use MSMSulc as default"
```
coalsont, line 20: `RegName` is what the loop uses.  Please choose a different variable name here, hopefully plural (perhaps `RegNamesStr`).

Mandatory arguments should generally not use the word "default", because the user is required to specify them.  We might consider making it optional with a default of MSMSulc...

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,133 @@
...
+		fi
+	done 
+	
+	for RegName in $RegNames ; do
+		RegSphere=""$NativeFolder"/"$Subject"."$Hemisphere".sphere."$RegName".native.surf.gii"
+
+		if [[ "$RegName" == "MSMSulc" ]] ; then
+			RegString=""
+		else
+			RegString="_"$RegName""
```
coalsont, line 89: Fix this below too, please.
```suggestion
			RegString="_$RegName"
```

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,133 @@
...
+for Hemisphere in $Hemi ; do
+	if [[ "$Hemisphere" == "L" ]] ; then
+		Structure="CORTEX_LEFT"
+	elif [[ "$Hemisphere" == "R" ]] ; then
+		Structure="CORTEX_RIGHT"
+	fi
+		(
+			cd "$HCPPIPEDIR"/global/scripts/
+			python CorrThick.py "$SubjectDir" "$Subject" "$Structure" "$Hemisphere" "$Surface" "$PatchSize" "$SurfSmooth" "$MetricSmooth"
+		)
```
coalsont, line 67: `global/scripts` has a few things that are used directly, so we may want to hide all the python files in a subdirectory instead.

Is the below a working substitute for the `cd`?  Or does python search specifically the working directory, and not the directory containing the input file for additional imports?
```suggestion
			python "$HCPPIPEDIR"/global/scripts/CorrThick.py "$SubjectDir" "$Subject" "$Structure" "$Hemisphere" "$Surface" "$PatchSize" "$SurfSmooth" "$MetricSmooth"
```

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,133 @@
...
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+#description of script/command
+opts_SetScriptDescription "Run curvature corrected thickness python script and save curvatures, regression coefficients, and curvature-corrected thickness"
+
+opts_AddMandatory '--subject-dir' 'SubjectDir' 'path' "folder containing all subjects"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject-id"
+opts_AddMandatory '--regname' 'RegName' 'my reg' "set the desired registration name(s) separated by @, 'string' 'RegName@RegName@RegName@...etc.' use MSMSulc as default"
+opts_AddOptional '--hemi' 'Hemi' 'hemisphere' "provide hemisphere for regression calculation, L=Left, R=Right, default 'L R' or B=Both" "B"
```
coalsont, line 21: Why do we want the user to have the option not to run it on both hemispheres?  Other scripts loop over a hardcoded `L R`.

Even if we keep it, the current description is disordered:
```suggestion
opts_AddOptional '--hemi' 'Hemi' 'hemisphere' "provide hemisphere for regression calculation, L=Left, R=Right, or B=Both, default 'B'" "B"
```

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,133 @@
...
+opts_AddOptional '--surf' 'Surface' 'surface' "provide surface for regression calculation, white or midthickness, default midthickness" "midthickness"
+opts_AddOptional '--patch-size' 'PatchSize' 'patch size' "provide patch size for regression, default 6" "6"
+opts_AddOptional '--surf-smooth' 'SurfSmooth' 'surf smooth' "provide surface smoothing fwhm, default 2.14" "2.14"
+opts_AddOptional '--metric-smooth' 'MetricSmooth' 'metric smooth' "provide metric smoothing fwhm, default 2.52" "2.52"
```
coalsont, line 25: Repeating the words from the flag name as the "type" part of the help info isn't very helpful.  The regression patch is also a gaussian weighted kernel, right?
```suggestion
opts_AddOptional '--patch-size' 'PatchSize' 'distance' "provide patch kernel size in millimeters FWHM for regression, default 6" "6"
opts_AddOptional '--surf-smooth' 'SurfSmooth' 'distance' "provide surface smoothing in millimeters FWHM, default 2.14" "2.14"
opts_AddOptional '--metric-smooth' 'MetricSmooth' 'distance' "provide metric smoothing in millimeters FWHM, default 2.52" "2.52"
```

global/scripts/CorrThick.sh
nagedem: This change won't work.  The working directory needs to be changed for importing other scripts properly. I also thought about having a subdirectory for all python scripts. Then I can cd into that. Should we do that?

coalsont: I think a `global/scripts/CorrThick` folder would be a good place for all of your python files.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,136 @@
...
+fi
+
+RegNames=`echo "$RegNamesStr" | sed s/"@"/" "/g`
+
+LowResMesh="32"
+HighResMesh="164"
+MapListFunc="MRcorrThickness MRcorrThickness_intercept MRcorrThickness_normcoeffs MRcorrThickness_curvs MRcorrThickness_coeffs"
+
+#Generate MRcorrThickness in Native Space
+if [[ "$ResampleOnly" == "NO" ]]; then
```
coalsont, line 59: Please use `opts_StringToBool` to sanity check yes/no type arguments (converts them to 0/1, which can also be used in math parenthesis for a neater-looking conditional):

https://github.com/Washington-University/HCPpipelines/blob/b4f33c5df48b0f0402f9685b7f35fd7c58ce4720/global/scripts/RSNregression.sh#L72
https://github.com/Washington-University/HCPpipelines/blob/b4f33c5df48b0f0402f9685b7f35fd7c58ce4720/global/scripts/RSNregression.sh#L211

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,136 @@
...
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject-id"
+opts_AddOptional '--regnamestr' 'RegNamesStr' 'my reg str' "set the desired registration name(s) separated by @, 'string' 'RegName@RegName@RegName@...etc.' default MSMSulc" "MSMSulc"
+opts_AddOptional '--hemi' 'Hemi' 'hemisphere' "provide hemisphere for regression calculation, L=Left, R=Right, or B=Both, default 'B'" "B"
+opts_AddOptional '--surf' 'Surface' 'surface' "provide surface for regression calculation, white or midthickness, default midthickness" "midthickness"
+opts_AddOptional '--patch-size' 'PatchSize' 'distance' "provide patch kernel size in millimeters FWHM for regression, default 6" "6"
+opts_AddOptional '--surf-smooth' 'SurfSmooth' 'distance' "provide surface smoothing in millimeters FWHM, default 2.14" "2.14"
+opts_AddOptional '--metric-smooth' 'MetricSmooth' 'distance' "provide metric smoothing in millimeters FWHM, default 2.52" "2.52"
+opts_AddOptional '--resample-only' 'ResampleOnly' 'YES or NO' "whether or not to generate the curvature-corrected cortical thickness data, if it is already available, but just to resample it to 164k and 32k, defaults to 'NO'" "NO"
```
coalsont, line 26: I try to avoid making parameters that say "only", because it causes interaction problems if you want to add a second or third "something-only" option.  Calling it "--skip-computation" or "--reuse-native" or something similar is probably better.

global/scripts/CorrThick.sh
coalsont: To be clear though, in order to do "not <flag>" with math parentheses, you need to do this:

```
if ((! DoZ))
```
The space between the `!` and the variable name is required, otherwise bash thinks it is a history expansion.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,136 @@
...
+	for Hemisphere in $Hemi ; do
+		if [[ "$Hemisphere" == "L" ]] ; then
+			Structure="CORTEX_LEFT"
+		elif [[ "$Hemisphere" == "R" ]] ; then
+			Structure="CORTEX_RIGHT"
+		fi
+			(
+				cd "$HCPPIPEDIR"/global/scripts/
+				python CorrThick.py "$SubjectDir" "$Subject" "$Structure" "$Hemisphere" "$Surface" "$PatchSize" "$SurfSmooth" "$MetricSmooth"
+			)
```
coalsont, line 69: I would leave the `(` and `)` on the same level as the preceeding `fi` (so that the interior is only one level more indended than the `fi`).

global/scripts/CorrThick.sh
glasserm: This is a good point.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,139 @@
...
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+#description of script/command
+opts_SetScriptDescription "Run curvature-corrected cortical thickness python script and save curvatures, regression coefficients, and curvature-corrected thickness"
+
+opts_AddMandatory '--subject-dir' 'SubjectDir' 'path' "folder containing all subjects"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject-id"
+opts_AddOptional '--regnamestr' 'RegNamesStr' 'my reg str' "set the desired registration name(s) separated by @, 'string' 'RegName@RegName@RegName@...etc.' default MSMSulc" "MSMSulc"
```
coalsont, line 20: The user-visible option name doesn't need to match the internal variable name.  Making it plural to make it obvious to users that it supports a list might be useful, but "str" probably isn't.  `--regnames` is probably better.

glasserm: I agree.

glasserm (approved):

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,139 @@
...
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject-id"
+opts_AddOptional '--regnames' 'RegNamesStr' 'my reg str' "set the desired registration name(s) separated by @, 'string' 'RegName@RegName@RegName@...etc.' default MSMSulc" "MSMSulc"
+opts_AddOptional '--hemi' 'Hemi' 'hemisphere' "provide hemisphere for regression calculation, L=Left, R=Right, or B=Both, default 'B'" "B"
+opts_AddOptional '--surf' 'Surface' 'surface' "provide surface for regression calculation, white or midthickness, default midthickness" "midthickness"
+opts_AddOptional '--patch-size' 'PatchSize' 'distance' "provide patch kernel size in millimeters FWHM for regression, default 6" "6"
+opts_AddOptional '--surf-smooth' 'SurfSmooth' 'distance' "provide surface smoothing in millimeters FWHM, default 2.14" "2.14"
+opts_AddOptional '--metric-smooth' 'MetricSmooth' 'distance' "provide metric smoothing in millimeters FWHM, default 2.52" "2.52"
+opts_AddOptional '--no-computation' 'NoCompute' 'YES or NO' "whether or not to compute the curvature-corrected cortical thickness, if it is already available, but just to resample it to 164k and 32k, defaults to 'NO'" "NO"
```
coalsont, line 26: `--no-compute=no` is a confusing construction.  I try to avoid negations in option flags, for this reason.  `--skip-compute` or `--reuse-native` would be better.

global/scripts/CorrThick/curvature.py
```diff
@@ -0,0 +1,252 @@
...
+            n1 = q1_x_q2/np.sqrt(np.dot(q1_x_q2,q1_x_q2))
+            n2 = q2_x_q3/np.sqrt(np.dot(q2_x_q3,q2_x_q3))
+            n2 = -n2
+            u1 = n2
+            u3 = q2/(np.sqrt(np.dot(q2,q2)))
+            u2 = np.cross(u3,u1)
+            cost = np.dot(n1, u1)
+            sint = np.dot(n1, u2)
+            theta = -math.atan2(sint,cost)
+            theta = -theta
```
coalsont: double negation
```suggestion
            theta = math.atan2(sint,cost)
```

nagedem: I kinda did that intentionally to show that we are flipping the sign convention.

coalsont: A human-readable comment would be more obvious than operations that cancel each other.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -90,6 +90,7 @@ opts_AddOptional '--inflatescale' 'InflateExtraScale' 'number' "surface inflatio
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData|LegacyStyleData' "disable some HCP preprocessing requirements to allow processing of data that doesn't meet HCP acquisition guidelines - don't use this if you don't need to" 'HCPStyleData'
 opts_AddOptional '--structural-qc' 'QCMode' 'yes|no|only' "whether to run structural QC, default 'yes'" 'yes'
 opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the session's myelin map as reference map's myelin map mean, defaults to 'YES'" 'YES'
+opts_AddOptional '--metric-regression' 'MetricReg' 'OLD, NEW or BOTH' "whether to use the updated curvature-thickness regression, defaults to 'BOTH'" 'BOTH'
```
coalsont:
```suggestion
opts_AddOptional '--metric-regression' 'MetricReg' 'OLD, NEW or BOTH' "whether to use the updated curvature-thickness regression, defaults to 'BOTH'
NOTE: if you run with 'OLD' and later decide you want the results from the new method, it is NOT RECOMMENDED to rerun PostFreeSurferPipeline.sh again, because it runs a registration that shouldn't be changed after other preprocessing pipelines are run.
Instead, run global/scripts/CorrThick.sh with appropriate arguments, which will only generate the new folding-compensated thickness outputs.
'NEW' or 'BOTH' require python 3 with numpy, nibabel, and scipy installed." 'BOTH'
```

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,139 @@
...
+if ((! SkC)); then
+	for Hemisphere in $Hemi ; do
+		if [[ "$Hemisphere" == "L" ]] ; then
+			Structure="CORTEX_LEFT"
+		elif [[ "$Hemisphere" == "R" ]] ; then
+			Structure="CORTEX_RIGHT"
+		fi
+		(
+			cd "$HCPPIPEDIR"/global/scripts/CorrThick
+			python CorrThick.py "$SubjectDir" "$Subject" "$Structure" "$Hemisphere" "$Surface" "$PatchSize" "$SurfSmooth" "$MetricSmooth"
```
coalsont: Should this call `python3` explicitly, in case someone has an environment configured to default to 2?

PostFreeSurfer/PostFreeSurferPipeline.sh
coalsont: We know that users who change this option when using the pipelines directly are very likely to have read the help info for this option, so please accept this suggested edit or similar wording to put the explanation where they can be expected to read it.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,171 @@
...
+#display the parsed/default values
+opts_ShowValues
+
+#sanity check boolean strings and convert to 1 and 0
+SkC=$(opts_StringToBool "$SkipCompute")
+
+#set paths
+NonlinearFolder="$SubjectDir"/"$Subject"/MNINonLinear
+NativeFolder="$NonlinearFolder"/Native
+T1wFolder="$SubjectDir"/"$Subject"/T1w/Native
```
coalsont: `$T1wFolder` containing "/Native" might be confusing (and possibly cause bugs if code is copy-pasted between this and a different pipeline):

https://github.com/Washington-University/HCPpipelines/blob/45d553215e0288fec620e8cb7b802bca6bb190ee/PreFreeSurfer/PreFreeSurferPipeline.sh#L357

I suggest renaming this variable to something like `$T1wNativeFolder`.

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -90,6 +90,10 @@ opts_AddOptional '--inflatescale' 'InflateExtraScale' 'number' "surface inflatio
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData|LegacyStyleData' "disable some HCP preprocessing requirements to allow processing of data that doesn't meet HCP acquisition guidelines - don't use this if you don't need to" 'HCPStyleData'
 opts_AddOptional '--structural-qc' 'QCMode' 'yes|no|only' "whether to run structural QC, default 'yes'" 'yes'
 opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the session's myelin map as reference map's myelin map mean, defaults to 'YES'" 'YES'
+opts_AddOptional '--metric-regression' 'MetricReg' 'OLD, NEW or BOTH' "whether to use the updated curvature-thickness regression, defaults to 'BOTH'
```
coalsont: "metric" is a very generic term, consider naming this `--thickness-regression` or `--thickness-compensation`.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -94,6 +94,7 @@ CorrectionSigma="${38}"
 RegName="${39}"
 UseIndMean="${40}"
 IsLongitudinal="${41}"
+MetricReg="${42}"
```
coalsont: Currently this argument doesn't get sanity checked anywhere.  If a user passes `--metric-regression=old` or `=OLDE`, it will silently not generate either file.  We may want to sanity check it in PostFS also, so the user gets immediate feedback rather than having some things run first.
```suggestion
MetricReg="${42}"

case "$MetricReg" in
    (NEW|OLD|BOTH)
        ;;
    (*)
        log_Err_Abort "unrecognized thickness compensation method: '$MetricReg', use NEW, OLD, or BOTH"
        ;;
esac
```

PostFreeSurfer/PostFreeSurferPipeline.sh
glasserm: agree.

global/scripts/CorrThick.sh
```diff
@@ -23,7 +23,39 @@ opts_AddOptional '--surf' 'Surface' 'surface' "provide surface for regression ca
...
+
+#SubjectDir="/path/to/subjects" 
+#SubjList="100206 100307 100408 ...." # space-separated subject IDs 
+
+#RegName="MSMSulc@MSMAll" # one or more registrations 
+
+#for Subject in $SubjList ; do 
+	#bash "$HCPPIPEDIR/global/scripts/CorrThick.sh" --subject-dir="$SubjectDir" --subject="$Subject" --regnames="$RegNames" --skip-computation=NO 
+#done 
+#######################################
```
glasserm: Are you wanting this info to print when someone runs the pipeline?

global/scripts/CorrThick.sh
glasserm: Agree

global/scripts/CorrThick.sh
nagedem: why not, also a help text?

global/scripts/CorrThick.sh
```diff
@@ -25,37 +25,78 @@ opts_AddOptional '--surf-smooth' 'SurfSmooth' 'distance' "provide surface smooth
...
+References:
+  - https://www.biorxiv.org/content/10.1101/2025.05.03.651968v1
+  - https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.25776
+Please cite these works when using this measure.
+
+Requirements:
+  Python dependencies: numpy, nibabel, scipy, psutil
+  Standard libraries: os, math, multiprocessing, concurrent.futures
+
+EOF
```
coalsont: This will print every time the script is run, not just when the help info is asked for.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,212 @@
...
+}
+
+# If --help is present, print help and exit
+for arg in "$@"; do
+    if [[ "$arg" == "--help" ]]; then
+        print_help
+        exit 0
+    fi
+done
```
coalsont: Other scripts that use newopts print the normal arguments info when --help is specified, I don't think we should split this behavior.  You can add any text you want (manual linebreaks are recommended, like you currently have, as only the argument descriptions get the word wrapping code applied) in the `opts_SetScriptDescription` line:

```
opts_SetScriptDescription "A script to compute things

Please be sure you want to compute things before
running this script.  The HCP and its successors take
no responsibility for the computing of things when the
user did not desire to compute things and ran this
script anyway."
```

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,212 @@
...
+    --subject       Subject ID
+
+  Optional:
+    --regnames       Registration name(s), separated by '@', default: MSMSulc
+    --hemi           Hemisphere: L=Left, R=Right, B=Both, default: B
+    --surf           Surface: white or midthickness, default: midthickness
+    --patch-size     Patch kernel size in mm FWHM, default: 6
+    --surf-smooth    Surface smoothing in mm FWHM, default: 2.14
+    --metric-smooth  Metric smoothing in mm FWHM, default: 2.52
+    --skip-computation  YES or NO, skip computation if already available, default: NO
```
coalsont: Don't do this.  `newopts.shlib` generates these descriptions from the current information used for parsing arguments, so it is never out of date, and doesn't need to be updated in two places.  Hardcoding the current information a second time defeats the point.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,212 @@
...
+SubjList="100206 100307 100408 ...."  # space-separated subject IDs
+RegNames="MSMSulc@MSMAll"
+
+for Subject in \$SubjList ; do
+    bash "\$HCPPIPEDIR/global/scripts/CorrThick.sh" \\
+        --subject-dir="\$SubjectDir" \\
+        --subject="\$Subject" \\
+        --regnames="\$RegNames" \\
+        --skip-computation=NO
+done
```
coalsont: Do we want an actual example batch in `Examples/Scripts`?  If so, it should probably support fsl_sub.  I don't know that this is complicated enough to deserve it, though (its normal usage is only 3 parameters).

You generally don't need to call "bash" explicitly to run a script (as long as it has the executable bit set).

global/scripts/CorrThick.sh
nagedem: that was the intention, but it could also go in the help only?

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -94,6 +94,15 @@ CorrectionSigma="${38}"
...
+
+case "$MetricReg" in
+    (NEW|OLD|BOTH)
+        ;;
+    (*)
+        log_Err_Abort "unrecognized thickness compensation method: '$MetricReg', use NEW, OLD, or BOTH"
+        ;;
+esac
```
coalsont: Given that `FreeSurfer2CaretConvertAndRegisterNonlinear.sh` and `CreateRibbon.sh` run in ~PreFS~ PostFS before this script, please add the same check to ~PreFS~ PostFS near the existing argument sanity checking.

global/scripts/CorrThick.sh
coalsont: It is unlikely to do much good being put into the log files when someone is actually using it for processing.  It also means the help-ish info is broken into more than one piece in the script, and thus less convenient to keep track of.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -202,7 +211,9 @@ for Hemisphere in L R ; do
     RegSphere="${AtlasSpaceFolder}/${NativeFolder}/${Session}.${Hemisphere}.sphere.reg.reg_LR.native.surf.gii"
   fi
 
-  ${CARET7DIR}/wb_command -metric-regression "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".thickness.native.shape.gii "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".corrThickness.native.shape.gii -roi "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".roi.native.shape.gii -remove "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".curvature.native.shape.gii
+  if [ "${MetricReg}" = "OLD" ] || [ "${MetricReg}" = "BOTH" ]; then
```
coalsont: This doesn't match the variable change to `$ThicknessReg`.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
```diff
@@ -228,6 +239,10 @@ for Hemisphere in L R ; do
   done
 done
 
+if [ "${MetricReg}" = "NEW" ] || [ "${MetricReg}" = "BOTH" ]; then
```
coalsont: Same.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,181 @@
...
+    - Location of the HCP Pipeline environment setup script (SetUpHCPPipeline.sh)
+
+Example batch script:
+---------------------------------------
+EnvironmentScript="/path/to/SetUpHCPPipeline.sh" # HCP Pipeline environment script
+source "\$EnvironmentScript"
+
+SubjectDir="/path/to/subjects"
+SubjList="100206 100307 100408 ...."  # space-separated subject IDs
+RegNames="MSMSulc@MSMAll"
```
coalsont: If you want quotes to show up inside quoted text, they need to be escaped.  Or, you can use single quotes for the whole help info string (`'`), which additionally avoid doing anything with `$`.

Given that the "example batch" here is literally just "run it on all subjects", and the arguments are simple, I don't know how useful it is.  I'm not sure how many people will realize the intention is to copy and paste it into a new file and edit it, but wouldn't have been able to write this themselves.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,181 @@
...
+
+for Subject in \$SubjList; do
+    \"\$HCPPIPEDIR/global/scripts/CorrThick.sh\" \\
+        --subject-dir=\"\$SubjectDir\" \\
+        --subject=\"\$Subject\" \\
+        --regnames=\"\$RegNames\" \\
+        --skip-computation=NO
```
coalsont: Unclear why this already-default setting is being provided here (also, if it is helpful, why it isn't in the fsl_sub part).

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,181 @@
...
+T1wNativeFolder="$SubjectDir"/"$Subject"/T1w/Native
+
+#Make intermediate directory to save intermediate files
+mkdir -p "$NativeFolder"/CorrThick
+
+#Loop through left and right hemispheres
+if [[ "$Hemi" == *B* ]]
+then
+	Hemi="L R"
+fi
```
coalsont: Consider a more robust check that tells the user when they did something dumb:

```suggestion
case "$Hemi" in
    (L|R|"L R"|"R L")
        ;;
    (B)
        Hemi="L R"
        ;;
    (*)
        log_Err_Abort "unrecognized hemisphere specification '$Hemi', use L, R, or B"
        ;;
esac
```

This would also forbid "L L", "L B", and other silliness (the help info doesn't actually say "L R" is valid, either, so technically it could also be forbidden).  Most scripts don't have a way to disable hemispheres, so removing this option for simplicity is another possibility.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,181 @@
...
+
+Requirements:
+  Python dependencies: numpy, nibabel, scipy, psutil
+  Standard libraries: os, math, multiprocessing, concurrent.futures
+
+Running stand-alone:
+  To run outside the full HCP pipeline, prepare a batch script that defines (at least):
+    - Subject directory
+    - Subject list
+    - Location of the HCP Pipeline environment setup script (SetUpHCPPipeline.sh)
```
coalsont: Subject directory and subject are arguments, defining them in the batch is merely for convenience in order to pass them to the script.  The script will already complain if they didn't source a `SetUp...` script, as `HCPPIPEDIR` won't be set, and `pipedirguessed` will be set, and an error will be generated if it gets through `opts_parseArguments` without other issues.  We can probably assume the user already ran at least PostFreeSurfer (potentially with custom arguments) if they are looking at the help info of this script, and knows the general idea of setting up to run our scripts.

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,181 @@
...
+opts_SetScriptDescription "Run curvature-corrected (folding-compensated) cortical thickness python script and save curvature-corrected (folding-compensated) thickness, curvatures, regression coefficients, and resampled outputs.
+
+References:
+  - https://www.biorxiv.org/content/10.1101/2025.05.03.651968v1
+  - https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.25776
+Please cite these works when using this measure.
+
+Requirements:
+  Python dependencies: numpy, nibabel, scipy, psutil
+  Standard libraries: os, math, multiprocessing, concurrent.futures
```
coalsont: The requirements/dependencies are to tell the user what they need to install beyond a working python3.  A python install that is missing any standard libraries is not a usable python install in the first place.  We don't need to mention these.

global/scripts/CorrThick.sh
glasserm: I think an example Batch script would be preferred to putting something in the help.

global/scripts/CorrThick.sh
nagedem: example usage of different arguments...

global/scripts/CorrThick.sh
nagedem: So, you prefer not to have a help section altogether?

PostFreeSurfer/scripts/CreateMyelinMaps.sh
nagedem: I didn't get this?

global/scripts/CorrThick.sh
coalsont: The dependencies that need to be installed by the user are fine to list.  The standard libraries used would already be installed along with python itself, and don't tell the user anything they would need to know.  I'd like the help info to be short and useful.

PostFreeSurfer/scripts/CreateMyelinMaps.sh
coalsont: Sorry, I shouldn't have typed "PreFS".

If the user passes `--thickness-regression=WTF` to PostFS, we want to give an error immediately.  We don't want to run FreeSurfer2Caret... and CreateRibbon.sh, and then error, given such an obvious mistake.  Copy and paste this block into PostFS, adding on to its existing sanity checking section (so that both scripts have the same check).

global/scripts/CorrThick.sh
```diff
@@ -0,0 +1,186 @@
...
+SubjectDir="/path/to/subjects"
+SubjList="100206 100307 100408 ..."  # space-separated subject IDs
+RegNames="MSMSulc@MSMAll"
+
+for Subject in $SubjList; do
+    "$HCPPIPEDIR/global/scripts/CorrThick.sh" \
+        --subject-dir="$SubjectDir" \
+        --subject="$Subject" \
+        --regnames="$RegNames" \
+        --patch-size=6
```
coalsont: Missing line continuation.
```suggestion
        --patch-size=6 \
```
I don't see much value in demonstrating changes to the advanced parameters we optimized, as the user isn't going to understand how they should decide on the value to use (because generally, they shouldn't be changing them).  NHP or infants would be the main cases I would expect to need different parameters, but we would do a new optimization, presumably.

global/scripts/CorrThick.sh
nagedem: you never know. I would be happy if this pipeline reaches as many people as possible with different skills and backgrounds. This very brief script could be useful to at least some people.

glasserm: I would do this as an Example Batch script like all the other pipelines have.

nagedem: just drafted the example script

Examples/Scripts/CorrThickPipelineBatch.sh
```diff
@@ -0,0 +1,91 @@
+get_batch_options() {
```
coalsont, line 3: To make it clear this isn't a tcsh script:
```suggestion
#!/bin/bash

get_batch_options() {
```

Examples/Scripts/CorrThickPipelineBatch.sh
```diff
@@ -0,0 +1,91 @@
...
+            --runlocal)
+                command_line_specified_run_local="TRUE"
+                index=$(( index + 1 ))
+                ;;
+	    *)
+		echo ""
+		echo "ERROR: Unrecognized Option: ${argument}"
+		echo ""
+		exit 1
+		;;
```
coalsont: This has both tabs and spaces.
```suggestion
        *)
            echo ""
            echo "ERROR: Unrecognized Option: ${argument}"
            echo ""
            exit 1
            ;;
```

Examples/Scripts/CorrThickPipelineBatch.sh
```diff
@@ -0,0 +1,91 @@
...
+        echo "About to locally run ${HCPPIPEDIR}/global/scripts/CorrThick.sh"
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/global/scripts/CorrThick.sh"
+        queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
+    fi
+
+    "${queuing_command[@]}" "$HCPPIPEDIR"/global/scripts/CorrThick.sh \
+        --subject-dir="$StudyFolder" \
+        --subject="$Subject" \
```
coalsont: Stray line continuation.
```suggestion
        --subject="$Subject"
```

global/scripts/CorrThick.sh
```diff
@@ -20,41 +20,8 @@ References:
...
-OR example usage with fsl_sub:
-
-for Subject in $SubjList; do
-    fsl_sub -q matlabparallel.q "$HCPPIPEDIR/global/scripts/CorrThick.sh" \
-        --subject-dir="$SubjectDir" \
-        --subject="$Subject" \
-done
----------------------------------------'
+# Requirements for this script:
+#  Python3 with dependencies numpy, nibabel, scipy, and psutil'
```
coalsont: If it displays, it shouldn't look like a comment.
```suggestion
This script requires python3 with numpy, nibabel, scipy, and psutil'
```

PostFreeSurfer/PostFreeSurferPipeline.sh
```diff
@@ -90,6 +90,10 @@ opts_AddOptional '--inflatescale' 'InflateExtraScale' 'number' "surface inflatio
...
 opts_AddOptional '--structural-qc' 'QCMode' 'yes|no|only' "whether to run structural QC, default 'yes'" 'yes'
 opts_AddOptional '--use-ind-mean' 'UseIndMean' 'YES or NO' "whether to use the mean of the session's myelin map as reference map's myelin map mean, defaults to 'YES'" 'YES'
+opts_AddOptional '--thickness-regression' 'ThicknessReg' 'OLD, NEW or BOTH' "whether to use the updated curvature-thickness regression, defaults to 'BOTH'
+NOTE: if you run with 'OLD' and later decide you want the results from the new method, it is NOT RECOMMENDED to rerun PostFreeSurferPipeline.sh again, because it runs a registration that shouldn't be changed after other preprocessing pipelines are run.
+Instead, run global/scripts/CorrThick.sh with appropriate arguments, which will only generate the new folding-compensated (curvature-corrected) thickness outputs.
+'NEW' or 'BOTH' require python 3 with numpy, nibabel, and scipy installed." 'BOTH'
```
coalsont: Synchronize requirements info.
```suggestion
'NEW' or 'BOTH' require python 3 with numpy, nibabel, scipy, and psutil installed." 'BOTH'
```

PR #319, opened 2024-12-29
Modifications to generic fMRIVolume and fMRISurface pipelines to run on longitudinal sessions. 

mmilch01:
This pull request comprises modifications to generic fMRIVolume and fMRISurface to run on longitudinal timepoints. Prerequisites for GenericfMRIVolumeProcessingPipeline are:

1. All cross-sectional pipelines must have been run up to post-Freesurfer for all timepoints.
2. All longitudinal pipelines for the corresponding timepoints must have been run up to post-Freesurfer.
3.  GenericfMRIVolumeProcessingPipeline must have been run on all cross-sectional timepoints.
4. All timepoints must be under the same processing directory.

Prerequisites for fMRISurface are the same plus GenericfMRIVolumeProcessingPipeline must have been run on all longitudinal timepoints.

Note that neither multi-echo fMRI nor fMRIReference is currently supported in longitudinal mode.


demsarjure:
@mmilch01 

1. I assume that in 
```
Prerequisites for fMRISurface are the same plus GenericfMRIVolumeProcessingPipeline must have been run on all longitudinal timepoints.
```
`GenericfMRIVolumeProcessingPipeline ` should be `GenericfMRISurfaceProcessingPipeline`?


2. It would be probably great, if I ran some tests for this on our end. Misha, can you please paste 2 example calls that you used for tests on your end, one for long volume and one for long surface. Just so I have something to quickly compare my command calls with. Thanks!

demsarjure:
It looks like this MR now includes changes from Volume + Surface and also changes from the long FS which were already merged into master? So the review tab shows a lot of code that is already tested and reviewed. Maybe it would be best to have a new branch for the new stuff?

Also, the branch is not in sync with the current master, so it would be best to merge the current master into this and resolve the conflicts. Or the cleanest way, which is to create new branch of the master and apply only long volume and long surface changes to it.

fMRIVolume/GenericfMRIVolumeProcessingPipeline-0.sh
demsarjure, line 1: What is the `-0` suffix in `GenericfMRIVolumeProcessingPipeline-0.sh`?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -626,146 +649,172 @@ fi
...
+    # --- Do slice time correction if indicated
+    # Note that in the case of STC, $fMRIFolder/$OrigTCSName will NOT be the "original" time-series
+    # but rather the slice-time corrected version thereof.
 
-    IFS='@' read -a SliceTimerCorrectionParametersArray <<< "$SliceTimerCorrectionParameters"
-    ${FSLDIR}/bin/immv "$fMRIFolder"/"$OrigTCSName" "$fMRIFolder"/"$OrigTCSName"_prestc
-    ${FSLDIR}/bin/slicetimer -i "$fMRIFolder"/"$OrigTCSName"_prestc -o "$fMRIFolder"/"$OrigTCSName" -r ${TR} -v "${SliceTimerCorrectionParametersArray[@]}"
-    ${FSLDIR}/bin/imrm "$fMRIFolder"/"$OrigTCSName"_prestc
-fi
+    if [ $DoSliceTimeCorrection = "TRUE" ] ; then
```
demsarjure, line 703: Weird offset at the beginning of this line.

demsarjure (commented):
I did a somewhat quick review and things look good. As usual I might report some bugs later on, when I actually start running tests.

coalsont:
> It looks like this MR now includes changes from Volume + Surface and also changes from the long FS which were already merged into master? So the review tab shows a lot of code that is already tested and reviewed. Maybe it would be best to have a new branch for the new stuff?
> 
> Also, the branch is not in sync with the current master, so it would be best to merge the current master into this and resolve the conflicts. Or the cleanest way, which is to create new branch of the master and apply only long volume and long surface changes to it.

Rebase is a cleaner way to resolve normal conflicts (this situation isn't normal).  Merge commits are weird and awkward, and hard to look at diffs of.  "In sync" isn't really applicable here, the whole purpose of a PR is to propose changes, conflicts mean that someone else has also made changes to the same files (or, it can be a symptom that your branch's starting point is wrong or contains extra stuff).

I can try to salvage the intended commits.

mmilch01:
> 1. I assume that in
> 
> ```
> Prerequisites for fMRISurface are the same plus GenericfMRIVolumeProcessingPipeline must have been run on all longitudinal timepoints.
> ```
> 
> `GenericfMRIVolumeProcessingPipeline ` should be `GenericfMRISurfaceProcessingPipeline`?
> 
Longitudinal fMRISurface needs longitudinal fMRIVolume to have been run on the same timepoint. Cross-sectional fMRISurface is not required to have been run. 

> 2. It would be probably great, if I ran some tests for this on our end. Misha, can you please paste 2 example calls that you used for tests on your end, one for long volume and one for long surface. Just so I have something to quickly compare my command calls with. Thanks!

I'll post my example batch scripts on Slack shortly.

mmilch01:
> It looks like this MR now includes changes from Volume + Surface and also changes from the long FS which were already merged into master? So the review tab shows a lot of code that is already tested and reviewed. Maybe it would be best to have a new branch for the new stuff?
> 
> Also, the branch is not in sync with the current master, so it would be best to merge the current master into this and resolve the conflicts. Or the cleanest way, which is to create new branch of the master and apply only long volume and long surface changes to it.

This should have been fixed now.

fMRIVolume/GenericfMRIVolumeProcessingPipeline-0.sh
mmilch01: > What is the `-0` suffix in `GenericfMRIVolumeProcessingPipeline-0.sh`?
That's a backup file that somehow made it into one of commits. should be fixed now.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -149,6 +149,13 @@ opts_AddOptional '--fmriref' 'fMRIReference' 'folder' "Specifies whether to use
 
 opts_AddOptional '--fmrirefreg' 'fMRIReferenceReg' 'linear or nonlinear' "Specifies whether to compute and apply a nonlinear transform to align the inputfMRI to the reference fMRI, if one is specified using --fmriref. The nonlinear transform is computed using 'fnirt' following the motion correction using the mean motion corrected fMRI image." "linear"
 
+opts_AddOptional '--output-native' 'OutputNative' 'binary' "Flag to output in native T1w space [no by default]" "0"
```
coalsont, line 152: This doesn't seem to be implemented.

"binary" and "flag" don't seem like explanations that would be obvious to users.  I typically use `'TRUE or FALSE'` or `'YES or NO'` for these, and try to be consistent within a script.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -626,145 +649,168 @@ fi
...
+    #copy directory structure and create symbolic link per each file under source. 
+    cp -rf "$ResultsFolder" "$ResultsFolderLong"
+    if (( $? )); then
+        log_Err_Abort "Copying cross-sectional output $ResultsFolder to longitudinal session folder $ResultsFolderLong failed."
+    fi
+    fMRIFolderLong="$Path"/"$SessionLong"/"$NameOffMRI"
+    cp -rf "$fMRIFolder" "$fMRIFolderLong"
+	if (( $? )); then
+		log_Err_Abort "Copying cross-sectional output $fMRIFolder to longitudinal session folder $fMRIFolderLong failed."
+	fi
```
coalsont, line 681:
```suggestion
    cp -rf "$ResultsFolder" "$ResultsFolderLong"
    fMRIFolderLong="$Path"/"$SessionLong"/"$NameOffMRI"
    cp -rf "$fMRIFolder" "$fMRIFolderLong"
```
Under `set -eu` and `debug.shlib`, all nonzero exit statuses will be caught before they reach your conditionals.  If you really feel these custom messages are necessary, you would need to do it this way:

```
    if ! cp -rf "$ResultsFolder" "$ResultsFolderLong"
    then
        log_Err_Abort "Copying cross-sectional output $ResultsFolder to longitudinal session folder $ResultsFolderLong failed."
    fi
```

You need whitespace on both sides of the `!`, due to bash history syntax also using that symbol.

The second conditional also somehow ended up with tabs instead of spaces.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -626,145 +649,168 @@ fi
...
-    zorient=$($FSLDIR/bin/fslval "$fMRIFolder"/"$OrigTCSName" qform_zorient | tr -d ' ')
+    if [ "$fMRIReference" != "NONE" ]; then
+        # --- copy over existing scout images
+        log_Msg "Copying Scout from Reference fMRI"
+        ${FSLDIR}/bin/imcp ${fMRIReferencePath}/Scout* ${fMRIFolder}
 
-    log_Msg "$fMRIFolder/$OrigTCSName: xorient=${xorient}, yorient=${yorient}, zorient=${zorient}"
+        for simage in SBRef_nonlin SBRef_nonlin_norm
+        do
+            ${FSLDIR}/bin/imcp ${fMRIReferencePath}/"${fMRIReference}_${simage}" ${fMRIFolder}/"${NameOffMRI}_${simage}"
```
coalsont, line 721: Why does this no longer contain "_orig_"?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: Let's move the physical space output stuff to a separate pull request.

fMRISurface/scripts/GenerateFMRIScenes.sh
```diff
@@ -31,8 +31,8 @@ PARAMETERs are [ ] = optional; < > = user supplied value
 
 #arguments to opts_Add*: switch, variable to set, name for inside of <> in help text, description, [default value if AddOptional], [compatibility flag, ...]
 #help info for option gets printed like "--foo=<$3> - $4"
-opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
-opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all Sessions"
+opts_AddMandatory '--session' 'Session' 'Session ID' "Session"
```
glasserm, line 35: Should the last session be Subject for backwards compatibility?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -428,6 +435,21 @@ then
...
+
+if (( IsLongitudinal )); then
+    if [ ! -d "$Path/$SessionLong" ]; then
+        log_Err_Abort "the --longitudinal-session must be specified and folder must exist in longitudinal mode"
+    fi
+    T1wCross2LongXfm=$Path/$SessionLong/T1w/xfms/T1w_cross_to_T1w_long.mat
+    if [ ! -f "$T1wCross2LongXfm" ]; then 
+        log_Err_Abort "Longitudinal session $SessionLong: cross-sectional to longitudinal transform $T1wCross2LongXfm does not exist. Has longtudinal PostFreesurfer been run?"
+    fi 
+    if [ -n "$fMRIReference" ] && [ "$fMRIReference" != "NONE" ]; then
```
glasserm, line 449: @demsarjure @sivcek We don't know how the fMRIReference works. Can you assist with ensuring that that functionality works with longitudinal mode?  Perhaps it is relevant only to cross-sectional processing and can be ignored?

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -64,12 +64,17 @@ opts_AddMandatory '--oscout' 'ScoutOutput' 'image' "output transformed + distort
 opts_AddMandatory '--ojacobian' 'JacobianOut' 'image' "output transformed + distortion corrected Jacobian image"
```
glasserm, line 64: This seems to have a lot of native space changes in it that should be in another pull request.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -645,15 +674,18 @@ then
...
-    ReferenceImage=${SubjectFolder}/T1w/T1w_acpc_dc.nii.gz
+    #TODO: point to longitudinal
+    ReferenceImage=${SessionFolder}/T1w/T1w_acpc_dc.nii.gz
 
     Files="PhaseOne_gdc_dc PhaseTwo_gdc_dc SBRef_dc"
     for File in ${Files}
     do
         if ((UseJacobian))
         then
+            #TODO: correctly point to longitudinal output image
```
glasserm, line 685: Is this still at TODO?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -645,15 +674,18 @@ then
...
             ${FSLDIR}/bin/applywarp --interp=spline -i "${WD}/FieldMap/${File}_jac" -r ${ReferenceImage} --premat=${WD}/fMRI2str.mat -o ${WD}/${File}
         else
+            #TODO: correctly point to longitudinal output
```
glasserm, line 688: Is this still a TODO?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -663,9 +695,10 @@ then
         #final bias field computation
 
         #run bias field computation script, go ahead and reuse the same working dir as previous run
+        #TODO: correctly point to longitudinal output
```
glasserm, line 698: Is this still at TODO?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -677,18 +710,22 @@ then
         for File in ${Files}
         do
             #we need to apply the new bias field to them for output
+            #TODO: correctly point to longitudinal output
```
glasserm, line 713: Is this still a TODO?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -626,145 +649,168 @@ fi
...
-    ${RUN} ${FSLDIR}/bin/imcp "$fMRIFolder"/"$OrigScoutName" "$fMRIFolder"/"$ScoutName"_gdc
-    #make fake jacobians of all 1s, for completeness
-    ${RUN} ${FSLDIR}/bin/fslmaths "$fMRIFolder"/"$OrigScoutName" -mul 0 -add 1 "$fMRIFolder"/"$ScoutName"_gdc_warp_jacobian
-    ${RUN} ${FSLDIR}/bin/fslroi "$fMRIFolder"/"$NameOffMRI"_gdc_warp "$fMRIFolder"/"$NameOffMRI"_gdc_warp_jacobian 0 1
-    ${RUN} ${FSLDIR}/bin/fslmaths "$fMRIFolder"/"$NameOffMRI"_gdc_warp_jacobian -mul 0 -add 1 "$fMRIFolder"/"$NameOffMRI"_gdc_warp_jacobian
-fi
+fi # if (( ! IsLongitudinal ))
 
-#Split echos
+#Split echos.
```
glasserm, line 813: Only run in cross-sectional.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -872,7 +939,7 @@ log_Msg "mkdir -p ${fMRIFolder}/OneStepResampling"
 mkdir -p ${fMRIFolder}/OneStepResampling
 tscArgs="";sctArgs="";
 for iEcho in $(seq 0 $((nEcho-1))) ; do
-    ${RUN} ${PipelineScripts}/OneStepResampling.sh \
+    cmd=(${RUN} ${PipelineScripts}/OneStepResampling.sh \
```
glasserm, line 942: Why do we have this cmd here?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -940,7 +1010,7 @@ then
             ${FSLDIR}/bin/fslmaths ${ResultsFolder}/${NameOffMRI}_pseudo_transmit_raw.nii.gz -mas ${fMRIFolder}/${FreeSurferBrainMask}.${FinalfMRIResolution}.nii.gz ${ResultsFolder}/${NameOffMRI}_pseudo_transmit_raw.nii.gz
             ${FSLDIR}/bin/applywarp --interp=trilinear -i ${DCFolder}/ComputeSpinEchoBiasField/${NameOffMRI}_pseudo_transmit_field.nii.gz -r ${fMRIFolder}/${NameOffMRI}_SBRef_nonlin -w ${AtlasSpaceFolder}/xfms/${AtlasTransform} -o ${ResultsFolder}/${NameOffMRI}_pseudo_transmit_field.nii.gz
             ${FSLDIR}/bin/fslmaths ${ResultsFolder}/${NameOffMRI}_pseudo_transmit_field.nii.gz -mas ${fMRIFolder}/${FreeSurferBrainMask}.${FinalfMRIResolution}.nii.gz ${ResultsFolder}/${NameOffMRI}_pseudo_transmit_field.nii.gz
-        else
+        else #never triggered in Longitudinal mode
```
glasserm, line 1013: @demsarjure @sivcek This needs your attention.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: Also need @demsarjure @sivcek to review.

glasserm (changes requested):
I have made various comments.  Please address them and also @demsarjure @sivcek please assist with the fMRIReference stuff. Thanks!

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
demsarjure: @glasserm I already went over the code and provided some comments. I will probably have some more comments/issues to report once I start testing. If all goes to plan, should be sometime this week.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
demsarjure: Yes, I believe fMRIReference is irrelevant for longitudinal processing.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
demsarjure: Like mentioned above, I think this is fine as reference is not needed any more in longitudinal processing.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: In that case we don't need to have an error message here if this code won't run anyway in longitudinal mode.

fMRISurface/scripts/GenerateFMRIScenes.sh
mmilch01: Yes. Fixed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: I think it is fixed in the current version.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Diff here is extremely confusing. Longitudinal related changes are actually quite minimal and local. It may be best to compare these files side by side.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Sorry, could you expand on what you mean?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: I used it for debugging. Reverse?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: Reference need not change for longitudinal, as long as image space does not change between timepoints. As far as I know, HCP pipelines don't support varying timepoint resolution? Or should we?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: See above

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: See above

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: See above

fMRIVolume/scripts/OneStepResampling.sh
mmilch01: All space change code is commented out.

glasserm:
The fMRIReference code should not actually do anything in longitudinal mode if longitudinal mode is implemented correctly, so we don't need an error message here.

Matt.
________________________________

glasserm:
You should not run this block in longitudinal mode, so please if/then it accordingly.

Matt.


***@***.***?anonymous&ep=bwmEmailSignature>                Book time to meet with ***@***.***?anonymous&ep=bwmEmailSignature>
________________________________

glasserm:
You need to suppress white space changes and then it is not confusing.  I always forget to do that too.

Matt.
________________________________

glasserm:
Yes if we don't need it.

Matt.
________________________________

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: He is using it to print the full sub-script command before running it (without writing it twice).  `newopts` automatically logs its arguments (in `OneStepResampling`), though.

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
```diff
@@ -34,9 +34,9 @@ source "${HCPPIPEDIR}/global/scripts/processingmodecheck.shlib"  # Check process
 
 opts_SetScriptDescription "Run fMRISurface processing"
 
-opts_AddMandatory '--studyfolder' 'Path' 'path' "folder containing all subject" "--path"
+opts_AddMandatory '--studyfolder' 'Path' 'path' "folder containing all session" "--path"
 
-opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--session' 'Session' 'session ID' "--subject"
```
coalsont, line 39: This is missing a long description, so the intention of `--subject` to be a legacy flag will not currently work (wrong number of arguments).  Use a quoted empty string, if nothing else.
```suggestion
opts_AddMandatory '--session' 'Session' 'session ID' "" '--subject'
```

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
```diff
@@ -105,7 +105,7 @@ esac
 
 if [ "${RegName}" = "FS" ] ; then
     log_Warn "WARNING: FreeSurfer's surface registration (based on cortical folding) is deprecated in the"
-    log_Warn "         HCP Pipelines as it results in poorer cross-subject functional and cortical areal "
+    log_Warn "         HCP Pipelines as it results in poorer cross-session functional and cortical areal "
```
coalsont, line 108: This probably shouldn't have been find/replaced here.

fMRISurface/scripts/RibbonVolumeToSurfaceMapping.sh
```diff
@@ -54,7 +54,7 @@ log_Msg "START"
 
 WorkingDirectory="$1"
 VolumefMRI="$2"
-Subject="$3"
+Session="$3"
```
coalsont, line 57: When the internal logic of a script doesn't need to change, I'm not sure we need to go to the trouble of replacing the variable names.  I'm not saying undo this, but maybe don't do it for future scripts that don't have functionality changes.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -46,9 +46,9 @@ NONE_METHOD_OPT="NONE"
 # --------------------------------------------------------------------------------
 opts_SetScriptDescription "Run fMRIVolume processing"
 
-opts_AddMandatory '--studyfolder' 'Path' 'path' "folder containing all subject" "--path"
+opts_AddMandatory '--studyfolder' 'Path' 'path' "folder containing all sessions" "--path"
 
-opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--session' 'Session' 'Session ID' "--subject"
```
coalsont, line 51: Again, missing a long description argument, so it will not use the `--subject` argument as a legacy flag.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: It sounds like "not supported" is not the correct thing to tell users, it is more like "fmri reference is used during the initial cross-sectional run on a timepoint, but ignored during the subsequent longitudinal call on the same timepoint".  That suggests it should be a warning, not an error.

If this is how longitudinal is intended to be run (first with `--is-longitudinal=FALSE`, then later with `--is-longitudinal=TRUE` on the same session+fMRIName pair), then perhaps this should be explained in the usage somewhere (or users will think "yes, I collected longitudinal data, so I just run it once with `TRUE`, right?").

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -428,6 +435,21 @@ then
     log_Err_Abort "the --usejacobian option must be 'true' or 'false'"
 fi
 
+IsLongitudinal=$(opts_StringToBool "$IsLongitudinal")
+T1wCross2LongXfm="NONE"
+
+if (( IsLongitudinal )); then
+    if [ ! -d "$Path/$SessionLong" ]; then
+        log_Err_Abort "the --longitudinal-session must be specified and folder must exist in longitudinal mode"
```
coalsont, line 443: This suggests `--longitudinal-session` must be specified and non-empty whenever `--is-longitudinal=TRUE`.  So, do we really even need the `--is-longitudinal` option, or should it simply test if `--longitudinal-session` gave an empty string?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -129,13 +129,25 @@ opts_AddOptional '--preregistertool' 'PreregisterTool' "'epi_reg' OR 'flirt'" "'
...
 if ((pipedirguessed))
 then
     log_Err_Abort "HCPPIPEDIR is not set, you must first source your edited copy of Examples/Scripts/SetUpHCPPipeline.sh"
 fi
 
+IsLongitudinal=$(opts_StringToBool "$IsLongitudinal")
+
+if (( $IsLongitudinal )); then
+    if [ ! -f "$T1wCross2LongXfm" ]; then
+        log_Err_Abort "--t1w-cross2long-xfm must point to a valid file when --is-longitudinal is used"
```
coalsont, line 147: It could be argued that `--is-longitudinal` is redundant here also (you wouldn't supply `--t1w-cross2long-xfm` in cross-sectional mode).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -222,420 +233,438 @@ fi
...
+            # Test if Magnitude Brain and T1w Brain Are Similar in Size, if not, assume Magnitude Brain Extraction
+            # Failed and Must Be Retried After Removing Bias Field
+            MagnitudeBrainSize=$(${FSLDIR}/bin/fslstats ${WD}/Magnitude_brain -V | cut -d " " -f 2)
+            T1wBrainSize=$(${FSLDIR}/bin/fslstats ${WD}/${T1wBrainImageFile} -V | cut -d " " -f 2)
 
-        else
-            # register scout to T1w image using fieldmap
-            ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/Scout.nii.gz --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}${ScoutExtension} --fmap=${WD}/FieldMap.nii.gz --fmapmag=${WD}/Magnitude.nii.gz --fmapmagbrain=${WD}/Magnitude_brain.nii.gz --echospacing=${EchoSpacing} --pedir=${UnwarpDir}
+            if [[ X$(echo "if ( (${MagnitudeBrainSize} / ${T1wBrainSize}) > 1.25 ) {1}" | bc -l) = X1 || X$(echo "if ( (${MagnitudeBrainSize} / ${T1wBrainSize}) < 0.75 ) {1}" | bc -l) = X1 ]] ; then
```
coalsont, line 372:
```suggestion
            if [[ "$(echo "((${MagnitudeBrainSize} / ${T1wBrainSize}) > 1.25) || ((${MagnitudeBrainSize} / ${T1wBrainSize}) < 0.75)" | bc -l)" == 1 ]] ; then
```
As long as we are editing this line, the `X$something` construction isn't needed with `[[` (or simply with proper quoting).  `bc` can do logical or, instead of two separate conditionals (and without needing conditionals inside bc at all).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -222,420 +233,438 @@ fi
...
+                log_Err_Abort "Script programming error. Unhandled Distortion Correction Method: ${DistortionCorrection}"            
+            fi
 
-            ${FSLDIR}/bin/flirt -interp spline -dof 6 -in ${WD}/Scout.nii.gz -ref ${T1wImage} -omat "$WD"/Scout2T1w.mat -out ${WD}/Scout2T1w.nii.gz -searchrx -30 30 -searchry -30 30 -searchrz -30 30
-            ${FSLDIR}/bin/convert_xfm -omat "$WD"/T1w2Scout.mat -inverse "$WD"/Scout2T1w.mat
-            ${FSLDIR}/bin/applywarp --interp=nn -i ${WD}/${T1wBrainImageFile} -r ${WD}/Scout.nii.gz --premat="$WD"/T1w2Scout.mat -o ${WD}/Scout_brain_mask.nii.gz
-            fslmaths ${WD}/Scout_brain_mask.nii.gz -bin ${WD}/Scout_brain_mask.nii.gz
-            fslmaths ${WD}/Scout.nii.gz -mas ${WD}/Scout_brain_mask.nii.gz ${WD}/Scout_brain.nii.gz
+            cp ${ScoutInputName}.nii.gz ${WD}/Scout.nii.gz
+            #ScoutExtension="_undistorted" this is now initialized before case switch to accomodate longitudinal mode
```
coalsont, line 365: I would just delete these lines rather than commenting them out.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -222,420 +233,438 @@ fi
...
+                ${FSLDIR}/bin/applywarp --rel --interp=spline -i ${WD}/Scout_brain.nii.gz -r ${WD}/Scout_brain.nii.gz -w ${WD}/WarpField.nii.gz -o ${WD}/${ScoutInputFile}${ScoutExtension}
+            else
+                # create a spline interpolated image of scout (distortion corrected in same space)
+                log_Msg "create a spline interpolated image of scout (distortion corrected in same space)"
+                ${FSLDIR}/bin/applywarp --rel --interp=spline -i ${ScoutInputName} -r ${ScoutInputName} -w ${WD}/WarpField.nii.gz -o ${WD}/${ScoutInputFile}${ScoutExtension}
+            fi
 
-        #copy the initial registration into the final affine's filename, as it is pretty good
-        #we need something to get between the spaces to compute an initial bias field
-        cp "${WD}/${ScoutInputFile}${ScoutExtension}2T1w_init.mat" "${WD}/fMRI2str.mat"
```
coalsont, line 452: Deleting the copy command but not the comments is confusing.

coalsont: To be clear: there is a copy of just the comments on (currently) line 473.  Due whitespace changes, the edits tagged by my comment show them as being deleted, when they still exist in the file.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
glasserm: I don't think we edited that line.

fMRISurface/GenericfMRISurfaceProcessingPipeline.sh
glasserm: Agree

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: I think we should delete it if we don't need it.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: You need to suppress white space changes and then it is not confusing.  I always forget to do that too.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: You should not run this block in longitudinal mode, so please if/then it accordingly.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: The fMRIReference code should not actually do anything in longitudinal mode if longitudinal mode is implemented correctly, so we don't need an error message here.  It is fine if people used fMRIReference in cross-sectional mode.

fMRIVolume/scripts/OneStepResampling.sh
glasserm: I would prefer that they were deleted and just put in a separate commit/pull request to reduce the likelihood of confusion later.

fMRISurface/scripts/RibbonVolumeToSurfaceMapping.sh
glasserm: Agree, though perhaps not worth changing it back...

glasserm:
Since my attempt at e-mail-based comments didn't work, I deleted them and moved them to the appropriate threads.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Correct. But is-longitudinal option is standard way to point to longitudinal session across scripts, instead of forcing the user to guess it is not required here.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Changed to warning, adding Tim's suggested prompt.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: With regards to the original @coalsont comment about _orig_, I don't recall modifying this line, but I just reverted it to the master branch version.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: The block between lines (new) 683 and 809 is already only run in cross-sectional mode. Did you mean the next 'split echos' block? For that, I kept statements assigning [tcs/sct]Echoes[..] but put all wb_command commands under conditional.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: Ok

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: Same as above, trying to be consistent across longitudinal /cross-sectional parameters

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: ok

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: Ok

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: It appears to have been part of the conflict section I marked in this commit:

d5a2fe43c2cdbcdd185314b428a1951914e99ee1

So, what probably happened is when I didn't try to resolve this while rebasing the commits to a fresh branch, you saw that you had already seen that code, and deleted the lines that included the (subtle, recent) change to add `_orig_` to those copies.

fMRIVolume/scripts/OneStepResampling.sh
mmilch01: Ok

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -222,420 +233,433 @@ fi
...
-echo "$0 $@" >> $WD/log.txt
-echo "PWD = `pwd`" >> $WD/log.txt
-echo "date: `date`" >> $WD/log.txt
-echo " " >> $WD/log.txt
+case $DistortionCorrection in
+	${FIELDMAP_METHOD_OPT} | ${SIEMENS_METHOD_OPT} | ${GE_HEALTHCARE_LEGACY_METHOD_OPT} | ${GE_HEALTHCARE_METHOD_OPT} | ${PHILIPS_METHOD_OPT} | ${SPIN_ECHO_METHOD_OPT} ) ScoutExtension="_undistorted" 
+	;;
+	${NONE_METHOD_OPT}) ScoutExtension="_nosdc" 
+	;;
+esac
```
coalsont, line 243:
```suggestion
case $DistortionCorrection in
	${FIELDMAP_METHOD_OPT} | ${SIEMENS_METHOD_OPT} | ${GE_HEALTHCARE_LEGACY_METHOD_OPT} | ${GE_HEALTHCARE_METHOD_OPT} | ${PHILIPS_METHOD_OPT} | ${SPIN_ECHO_METHOD_OPT} )
		ScoutExtension="_undistorted" 
	;;
	${NONE_METHOD_OPT})
		ScoutExtension="_nosdc" 
	;;
esac
```
Hard to see the "do this" part when it is on the end of a long line.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
coalsont: At the moment, it will still print "splitting echoes" if multi echo is detected, even though it only sets some shell variables.  If we don't need any of the variables in longitudinal mode, I would put the longitudinal conditional around the entire block instead of buried in the loop, for readability (and avoiding wrong messages).

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -222,420 +233,435 @@ fi
...
+        ${NONE_METHOD_OPT})
 
-            # NOTE: To work with later code a uniform Jacobian is created.
+                # NOTE: To work with later code a uniform Jacobian is created.
 
-            log_Msg "---> No distortion correction"
+                log_Msg "---> No distortion correction"
 
-            ScoutExtension="_nosdc"
+                #ScoutExtension="_nosdc" #this is now initialized before case switch to accomodate longitudinal mode
```
coalsont, line 521: We deleted the other lines like this instead of commenting them out.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -142,11 +144,15 @@ verbose_echo "             --oscout: ${ScoutOutput}"
 verbose_echo "          --ojacobian: ${JacobianOut}"
 verbose_echo "        --fmrirefpath: ${fMRIReferencePath}"
 verbose_echo "         --fmrirefreg: ${fMRIReferenceReg}"
+#verbose_echo "          --t1-native: ${T1wImageNative}"
+#verbose_echo "   --biasfield-native: ${BiasFieldNative}"
+#verbose_echo "--freesruferbrainmask-native: $FreeSurferBrainMaskNative"
```
coalsont, line 149: This looks like more native-space-output stuff.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
mmilch01: Ok

fMRIVolume/scripts/OneStepResampling.sh
mmilch01: Removed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: We do need tcsEchoes* to be set for multi-echo for both longitudinal and non-longitudinal modes. The if clause inside the "for iEcho.." loop prevents re-running wb_command in longitudinal mode.

fMRIVolume/scripts/OneStepResampling.sh
```diff
@@ -144,9 +146,10 @@ verbose_echo "        --fmrirefpath: ${fMRIReferencePath}"
 verbose_echo "         --fmrirefreg: ${fMRIReferenceReg}"
 verbose_echo " "
 
-BiasFieldFile=`basename "$BiasField"`
-T1wImageFile=`basename $T1wImage`
-FreeSurferBrainMaskFile=`basename "$FreeSurferBrainMask"`
+BiasFieldFile=$(basename "$BiasField")
+T1wImageFile=$(basename $T1wImage)
+#T1wImageFileNative=$(basename $T1wImageNative)
```
coalsont, line 151: This line also looks like a native volume space change, commented out.

PostFreeSurfer/PostFreeSurferPipelineLongPrep.sh
```diff
@@ -394,19 +394,20 @@ if (( TemplateProcessing ==  1 )); then
...
-    
-    #finalize all TP's with template to MNI152 atlas transform    
+
+    #finalize all TP's with template to MNI152 atlas transform
     for tp in ${timepoints[@]}; do
         log_Msg "Timepoint $tp: applying MNI152 atlas transform and brain mask"
         #These variables are redefined
         Timepoint_long=$tp.long.$Template
         AtlasSpaceFolder_timepoint=$StudyFolder/$Timepoint_long/MNINonLinear
+        mkdir -p $AtlasSpaceFolder_timepoint
```
coalsont, line 404: With the rebase onto the fixes in master, this line is redundant.

PR #320, opened 2025-01-08
HCP HippUnfold and Post HippUnfold Implementations

glasserm:
This is an implementation of the current version of HippUnfold 1.5.2-pre.2: https://hippunfold.readthedocs.io/en/latest/.  This requires the current development version of Connectome Workbench to use.  The file organization is non-final and needs evaluation and comment. There are a variety of unfinished things marked with TODO.  Some of these are pending comments and some pending a new planned HippUnfold 2.0 release.


HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -0,0 +1,101 @@
...
+
+log_Msg "Running T1w HippUnfold for subject: $Subject"
+apptainer run --bind $StudyFolder -e $HIPPUNFOLDPATH $HippUnfoldT1wFolder $HippUnfoldT1wFolderOut participant --modality T1w --cores all --force-output --generate_myelin_map --output-density 0p5mm 1mm 2mm
+log_Msg "T1w HippUnfold completed."
+log_Msg "Running T2w HippUnfold for subject: $Subject"
+apptainer run --bind $StudyFolder -e $HIPPUNFOLDPATH $HippUnfoldT2wFolder $HippUnfoldT2wFolderOut participant --modality T2w --cores all --force-output --generate_myelin_map --output-density 0p5mm 1mm 2mm
+log_Msg "T2w HippUnfold completed."
+log_Msg "Running T1wT2w HippUnfold for subject: $Subject"
+apptainer run --bind $StudyFolder -e $HIPPUNFOLDPATH $HippUnfoldT1wT2wFolder $HippUnfoldT1wT2wFolderOut participant --modality T1w --cores all --force-output --generate_myelin_map --output-density 0p5mm 1mm 2mm --force-nnunet-model T1T2w
+log_Msg "T1wT2w HippUnfold completed."
```
demsarjure, line 99: I knew this day would come sooner or later. This will force us to run containers inside containers (HippUnfold inside QuNex), on paper it should work. I guess we will see in practice if there are any issues with this.

What we need to figure out is whether we should put the HippUnfold container inside the QuNex container (this adds a couple GBs to the QuNex size) or if we require users to download it manually. I am leaning towards the first option as it is more user friendly and nowadays size is not that important as we all have good connectivity and enough disk space. A case for having the container outside could be made by saying this is a pipeline that will not be executed by most of our users?

Need to think this through a bit.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+fi
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+
+opts_SetScriptDescription "Make some BIDS structures and run HippUnfold"
+
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
```
demsarjure, line 18: This was changed to Session during the refactoring?

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -0,0 +1,101 @@
...
+fi
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+
+
+opts_SetScriptDescription "Make some BIDS structures and run HippUnfold"
+
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
```
demsarjure, line 18: This was changed to Session during the refactoring?

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+${CARET7DIR}/wb_command -volume-resample ${PhysicalHippUnfoldFolder}/${Subject}.HippocampalSubfields.nii.gz ${AtlasFolder}/T1w_restore.nii.gz ENCLOSING_VOXEL ${AtlasHippUnfoldFolder}/${Subject}.HippocampalSubfields.nii.gz -warp ${AtlasFolder}/xfms/acpc_dc2standard.nii.gz -fnirt ${AtlasFolder}/T1w_restore.nii.gz
+${CARET7DIR}/wb_command -add-to-spec-file ${AtlasHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec INVALID ${AtlasHippUnfoldFolder}/${Subject}.HippocampalSubfields.nii.gz
+
+#NIFTI Input Volumes
+${CARET7DIR}/wb_command -add-to-spec-file ${PhysicalHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec INVALID ${T1wFolder}/T1w_acpc_dc_restore.nii.gz
+${CARET7DIR}/wb_command -add-to-spec-file ${PhysicalHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec INVALID ${T1wFolder}/T2w_acpc_dc_restore.nii.gz
+${CARET7DIR}/wb_command -add-to-spec-file ${AtlasHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec INVALID ${AtlasFolder}/T1w_restore.nii.gz
+${CARET7DIR}/wb_command -add-to-spec-file ${AtlasHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec INVALID ${AtlasFolder}/T2w_restore.nii.gz
+
+#TODO: Merge Native Meshes, anything with 0pt5mm meshes?
```
demsarjure, line 216: TODO or not TODO?

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+#$PhysicalHippUnfoldFolderOut/hippunfold/sub-${Subject}/surf/sub-${Subject}_space-${Modality}_den-${Mesh}_label-hipp_gyrification.dscalar.nii
+#$PhysicalHippUnfoldFolderOut/hippunfold/sub-${Subject}/surf/sub-${Subject}_space-${Modality}_den-${Mesh}_label-hipp_myelin.dscalar.nii
+#$PhysicalHippUnfoldFolderOut/hippunfold/sub-${Subject}/surf/sub-${Subject}_space-${Modality}_den-${Mesh}_label-hipp_surfarea.dscalar.nii #Not created by default
+#$PhysicalHippUnfoldFolderOut/hippunfold/sub-${Subject}/surf/sub-${Subject}_space-${Modality}_den-${Mesh}_label-hipp_thickness.dscalar.nii
+
+#CIFTILabels (recreated by the script)
+#$PhysicalHippUnfoldFolderOut/hippunfold/sub-${Subject}/surf/sub-${Subject}_space-${Modality}_den-${Mesh}_label-hipp_atlas-multihist7_subfields.dlabel.nii
+
+#VolumeLabels
+#$PhysicalHippUnfoldFolderOut/hippunfold/sub-${Subject}/anat/sub-${Subject}_hemi-${Hemisphere}_space-${Modality}_desc-subfields_atlas-multihist7_dseg.nii.gz
```
demsarjure, line 299: Cleanup these notes at the end?

demsarjure (commented):
I reviewed this and added some minor comments/discussion points.

Examples/Scripts/SetUpHCPPipeline.sh
```diff
@@ -26,6 +26,8 @@ export MATLAB_COMPILER_RUNTIME=/export/matlab/MCR/R2017b/v93
 # If a suitable version of wb_command is on your $PATH, CARET7DIR can be blank
 export CARET7DIR=
 export HCPCIFTIRWDIR="$HCPPIPEDIR"/global/matlab/cifti-matlab
+export HIPPUNFOLDPATH="${HOME}/pipeline_tools/HippUnfold/khanlab_hippunfold_latest.sif"
```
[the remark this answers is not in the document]
coalsont: Users should not add anything inside of the repo folder.  That way madness lies.

HippUnfoldHCP/HippUnfoldHCP.sh
coalsont: Do they provide their code such that it can be run without their container wrapping it?  I think of containers as a convenience of not having to manually configure a bunch of software, not as a replacement for being able to install someone's software.

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -0,0 +1,101 @@
...
+#cp "$T2wImage" "$HippUnfoldT2wFolder/${Subject}_T2w_acpc_dc_restore.nii.gz"
+#cp "$T1wImage" "$HippUnfoldT1wT2wFolder/${Subject}_T1w_acpc_dc_restore.nii.gz"
+#cp "$T2wImage" "$HippUnfoldT1wT2wFolder/${Subject}_T2w_acpc_dc_restore.nii.gz"
+
+cp "$T1wImage" "$HippUnfoldT1wFolder/sub-${Subject}_T1w.nii.gz"
+cp "$T2wImage" "$HippUnfoldT1wFolder/sub-${Subject}_T2w.nii.gz"
+cp "$T1wImage" "$HippUnfoldT2wFolder/sub-${Subject}_T1w.nii.gz"
+cp "$T2wImage" "$HippUnfoldT2wFolder/sub-${Subject}_T2w.nii.gz"
+cp "$T1wImage" "$HippUnfoldT1wT2wFolder/sub-${Subject}_T1w.nii.gz"
+cp "$T2wImage" "$HippUnfoldT1wT2wFolder/sub-${Subject}_T2w.nii.gz"
```
coalsont, line 80: Does it really need to create `T1w/T2w.nii.gz` and `T2w/T1w.nii.gz`?  Do the 3 calls even need 3 separate input folders (with currently the same initial contents)?  Why are there 3 hippunfold calls anyway?

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+if [ $Mesh = "2mm" ] ; then
+  ${CARET7DIR}/wb_command -spec-file-merge ${PhysicalHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec ${T1wFolder}/fsaverage_LR32k/${Subject}.MSMAll.32k_fs_LR.wb.spec ${PhysicalHippUnfoldFolder}/${Subject}.${Mesh}.MSMAll.32k.wb_spec #TODO: Don't Hardcode Cortex
+  ${CARET7DIR}/wb_command -spec-file-merge ${AtlasHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec ${AtlasFolder}/fsaverage_LR32k/${Subject}.MSMAll.32k_fs_LR.wb.spec ${AtlasHippUnfoldFolder}/${Subject}.${Mesh}.MSMAll.32k.wb_spec #TODO: Don't Hardcode Cortex
+elif [ $Mesh = "1mm" ] ; then
+  ${CARET7DIR}/wb_command -spec-file-merge ${PhysicalHippUnfoldFolder}/${Subject}.${Mesh}.wb_spec ${AtlasFolder}/${Subject}.MSMAll.164k_fs_LR.wb.spec ${PhysicalHippUnfoldFolder}/${Subject}.${Mesh}.MSMAll.164k.wb_spec #TODO: Don't Hardcode Cortex
+fi
+
+if [ ${Flag} = "On" ] ; then
+  Modality="T1wT2w"
+fi
```
coalsont, line 226: Would be simpler and less error-prone to declare the variables local at the start of the function (if you feel the need to reuse the variable names)...

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+if [ -z ${PhysicalHippUnfoldDIR} ] ; then
+  PhysicalHippUnfoldDIR="${T1wFolder}/HippUnfold"
+fi 
+
+if [ -z ${AtlasHippUnfoldDIR} ] ; then
+  AtlasHippUnfoldDIR="${AtlasFolder}/HippUnfold"
+fi 
+
+function PostHippUnfold {
+HippUnfoldFolderOut=${1}
```
coalsont, line 45: I would indent the entire function to make it obvious where it ends.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+
+if [ ${Modality} = "T1wT2w" ] ; then
+  Modality="T1w"
+  Flag="On"
+else
+  Flag="Off"
+fi
+
+function PALETTE {
+File=${1}
```
coalsont, line 62: Probably better to define this function outside the body of the other function.  Nesting like this doesn't remove the function once the outer function ends (so it doesn't have the usual benefits of nested functions), but it does redundantly redefine it each call.  It also increases the distance between the start and end of the outer function.

Also, the body isn't indented relative to the context.  I don't want matlab-style indentation to leak outside of matlab (I don't even like it in matlab...).

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+        elif [ ${Hemisphere} = "R" ] ; then
+          Expression="Var + 8"
+        fi
+        ${CARET7DIR}/wb_command -metric-math "${Expression}" ${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Label}.${Mesh}.shape.gii -var Var $HippUnfoldFolderOut/hippunfold/sub-${Subject}/surf/sub-${Subject}_hemi-${Hemisphere}_space-${Modality}_den-${Mesh}_label-${Structure}_${Label}.label.gii
+        ${CARET7DIR}/wb_command -metric-label-import ${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Label}.${Mesh}.shape.gii $HCPPIPEDIR/global/config/HippUnfoldLut.txt ${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Label}.${Mesh}.label.gii
+      elif [ ${Structure} = "dentate" ] ; then
+        if [ ${Hemisphere} = "L" ] ; then
+          Expression="Var*0+1*6"
+        elif [ ${Hemisphere} = "R" ] ; then
+          Expression="Var*0+1*6 + 8"
```
coalsont, line 162:
```suggestion
          Expression="6"
        elif [ ${Hemisphere} = "R" ] ; then
          Expression="6 + 8"
```
or `"14"`, if the construction of the value isn't particularly meaningful to understanding.  Just because you need an input file to set the output dimensions doesn't mean the expression has to use it.  There is a special case deliberately to not issue an "unused file" warning when there are no variables and only 1 input file.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+    done
+    
+    #GIFTI Labels
+    #Don't add GIFTI to Specs
+    for Label in $Labels ; do
+      Name=`echo $Label | cut -d "@" -f 2`
+      Label=`echo $Label | cut -d "@" -f 1`
+      if [ ${Structure} = "hipp" ] ; then
+        if [ ${Hemisphere} = "L" ] ; then
+          Expression="Var*1"
```
coalsont, line 152:
```suggestion
          Expression="Var"
```

HippUnfoldHCP/PostHippUnfoldHCP.sh
coalsont: Also, there is only one line in the code (in a loop) that calls this function.  Functions are mainly for when you want to call it from more than one line in the code (such as when you have to change the arguments in a case-specific way).

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+Subject=${6}
+Modality=${7}
+Mesh=${8}
+
+if [ ${Modality} = "T1wT2w" ] ; then
+  Modality="T1w"
+  Flag="On"
+else
+  Flag="Off"
+fi
```
coalsont, line 59: You could also consider this pattern:
```suggestion
EffectiveModality="$Modality"
if [ ${Modality} = "T1wT2w" ] ; then
  EffectiveModality="T1w"
fi
```

HippUnfoldHCP/HippUnfoldHCP.sh
jordandekraker: HippUnfold v2.0.0 will be available on conda (ETA ~1 month)
The current HippUnfold v1.5.2 can be installed with `git clone; pip install -e .`, however it also uses containers internally for dependencies like nighres, wb_command, etc. See [here](https://hippunfold.readthedocs.io/en/latest/contributing/contributing.html) (and note that poetry is not necessary, its just for python linting)

demsarjure: Thanks for the info. I felt like adding your whole container into our container would be suboptimal space wise as there should be some overlap between the tools we depend on. We indeed both use wb_command, so that one would be redundant for sure, there could be others. I will first try to do this optimally and install all the dependencies manually. If this does not work and is too much of a hassle, we can go the container route.

@glasserm What is your preferred timeline here? Would you like to wait for HippUnfold v2.0.0 and then start processing with this one. Sounds like a major upgrade, that is only a month or so away. Or is this urgent and waiting an extra month is not an option?

Examples/Scripts/SetUpHCPPipeline.sh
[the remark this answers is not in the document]
glasserm: This is the standard convention we have used for all of the HCP Pipelines so it is odd to be complaining about it here.  It is intended to be changed but is not supposed to be blank (unlike the CARET7DIR).

HippUnfoldHCP/HippUnfoldHCP.sh
glasserm: > Thanks for the info. I felt like adding your whole container into our container would be suboptimal space wise as there should be some overlap between the tools we depend on. We indeed both use wb_command, so that one would be redundant for sure, there could be others. I will first try to do this optimally and install all the dependencies manually. If this does not work and is too much of a hassle, we can go the container route.
> 
> @glasserm What is your preferred timeline here? Would you like to wait for HippUnfold v2.0.0 and then start processing with this one. Sounds like a major upgrade, that is only a month or so away. Or is this urgent and waiting an extra month is not an option?

This isn't intended to be merged until HippUnfold v2.0.0 is available and the PR reflects and changes needed for that.  I agree that ideally you would install HippUnfold + dependencies in QuNex rather than the container.

glasserm: > RE: #Couldn't get non-BIDS to work (wouldn't do anything) #Seriously: don't put a $ here...
> 
> The `Subject` wildcard should simply be `subject`. This can be used to avoid having to make a BIDS directory as input. Note that hippunfold expects to iterate over all possible completions of the wildcard `subject` (and can also iterate over other wildcards like `session` or `acq`, as per pyBIDS). Best parallelization can be achieved by running batches of subjects, but I understand you may not want to set up your pipelines that way.
I currently wrap in some BIDS because I couldn't get the non-BIDS options to work reliably.  If you guys provide an ability to simply specify paths to relevant inputs, I will remove the wrapped BIDS.

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: I think some of those changes might have been overzealous.  In any case, this is a cross-sectional pipeline at the moment and I have not thought at all about making it longitudinal.  That might be something that HippUnfold would need to internally do.  We rely on FreeSurfer to setup the longitudinal processing for cerebral cortex.

HippUnfoldHCP/HippUnfoldHCP.sh
glasserm: Same as the other comment.

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: Dependent on HippUnfold 2.0

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: They will likely stay as code documentation to help me remember what HippUnfold does.

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: That is in fact why it is a function, there are a bunch of variables that change depending upon which HippUnfold output is being used.  If we standardized on only one kind of HippUnfold inputs (e.g., both T1w and T2w) then I wouldn't need the function.  

As it stands it is most of the script and I didn't want to waste the space.  The function may not be needed in the end if we standardize on only one kind of HippUnfold inputs.

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: I don't think I was clear on what was inside or outside a function's scope, so I put things inside just to be sure.  Let's see what happens with the HippUnfold inputs...

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: I didn't know that was possible.  It is written in the fslmaths conventions.

HippUnfoldHCP/PostHippUnfoldHCP.sh
coalsont: Looks like `local` works to me:
```
$ function foo() { local bar=broken; }
tim@blackbox:~$ bar=fixed
tim@blackbox:~$ foo
tim@blackbox:~$ echo $bar
fixed
```

coalsont: But in general, these problems don't exist if you just use a new variable name rather than destroying a value you need again later.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -0,0 +1,300 @@
...
+    PhysicalHippUnfoldFolderOut="$PhysicalHippUnfoldDIR/${Modality}_hippunfold"
+    AtlasHippUnfoldolderOut="$AtlasHippUnfoldDIR/${Modality}_hippunfold"
+
+    PhysicalHippUnfoldFolder="$PhysicalHippUnfoldFolderOut/$Mesh"
+    AtlasHippUnfoldFolder="$AtlasHippUnfoldolderOut/$Mesh"
+    
+    mkdir -p ${PhysicalHippUnfoldFolder} ${AtlasHippUnfoldFolder}
+
+    log_Msg "Processing $Modality $Mesh"
+    PostHippUnfold $PhysicalHippUnfoldFolderOut $PhysicalHippUnfoldFolder $AtlasHippUnfoldFolder $T1wFolder $AtlasFolder $Subject $Modality $Mesh
```
coalsont, line 242:
```suggestion
for LoopModality in T1w T2w T1wT2w ; do 
  for Mesh in 0p5mm 1mm 2mm ; do #TODO: Native meshes not yet available from HippUnfold: "Native" Folder Name "native" Mesh Name 
    
    PhysicalHippUnfoldFolderOut="$PhysicalHippUnfoldDIR/${LoopModality}_hippunfold"
    AtlasHippUnfoldolderOut="$AtlasHippUnfoldDIR/${LoopModality}_hippunfold"

    PhysicalHippUnfoldFolder="$PhysicalHippUnfoldFolderOut/$Mesh"
    AtlasHippUnfoldFolder="$AtlasHippUnfoldolderOut/$Mesh"
    
    mkdir -p ${PhysicalHippUnfoldFolder} ${AtlasHippUnfoldFolder}

    log_Msg "Processing $LoopModality $Mesh"
    PostHippUnfold $PhysicalHippUnfoldFolderOut $PhysicalHippUnfoldFolder $AtlasHippUnfoldFolder $T1wFolder $AtlasFolder $Subject $LoopModality $Mesh
```
Here is another possibility to avoid the variable collision problem.

HippUnfoldHCP/PostHippUnfoldHCP.sh
coalsont: To be clear, the very first use of each variable in the function needs the local keyword to do this, but later uses don't:

```
$ function foo() { local bar=broken; bar=morebroken; }
$ bar=fixed
$ foo
$ echo $bar
fixed
```
If it isn't on the first one, you get a problem:
```
$ function foobar() { bar=oops; local bar=okay; }
$ bar=fixed
$ foobar
$ echo $bar
oops
```

coalsont:
I undid the "single loop body moved to function" and used a new variable for the `space` bids tag instead of clobbering `Modality`.  I also made the physical output folder variable name consistent.  Please test.

HippUnfoldHCP/PostHippUnfoldHCP.sh
coalsont: There is only one variable which behaves inconsistently between the cases, and you already put that conditional logic *into* the body.  Functions are for when you want to move the inconsistency *outside* the body, by manually calling the function more than once with different explicit patterns (instead of looping).  Then it's like copy-paste and edit, but readable.

HippUnfoldHCP/HippUnfoldHCP.sh
coalsont: Matt says the T1w image in the T2w folder may be needed for registration.  The T2w image in the T1w folder still seems strange to me.  Another question, can these be symlinks instead of actual copies?  It looks like the container calls bind mount the entire study folder, so I would think relative symlinks would work.

mharms:
Just getting around to taking a look at this.  So, is the next version of HippUnfold going to support direct specification of input files, rather than having to hack an immediate "BIDS-style" input directory?

In terms of organization, I was wondering if it wouldn't be helpful to segregate all the `*.surf.gii` output files for the different mesh resolutions into their own `surf` directory, and same for all the `*.shape.gii` files, so that the `*.dscalar.nii` files more readily stand out on their own.  (Yes, I know that would be different from the organization of the brain data, but this is a fresh chance to give a little more structure to the output organization).

Can you remind me, how does one actually view the tessellation of the surfaces?  I tried zooming in a lot, but I still wasn't getting to see the actual triangles.

glasserm:
1) That is the hope.
2) I prefer to maintain parallelism with existing HCP organizations and not generate a bunch of subfolders.
3) That is under the surface menu in Connectome Workbench.

mharms:
> 3\. That is under the surface menu in Connectome Workbench.

Ahh, found it.  `Surface -> Properties`, and then change the "Drawing Type" to "Links (Edges)"

glasserm:
Now supports and depends upon HippUnfold Version 2.0.0.

HippUnfoldHCP/HippUnfoldHCP.sh
glasserm: I got the non BIDS options to work this time around due to properly understanding them.

HippUnfoldHCP/PostHippUnfoldHCP.sh
glasserm: Done for Native Meshes.

HippUnfoldHCP/HippUnfoldHCP.sh
glasserm: Done

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -0,0 +1,78 @@
...
+apptainer run --bind $StudyFolder -e $HIPPUNFOLDPATH $HippUnfoldT2wFolder $HippUnfoldT2wFolder participant --modality T2w --path-T1w $HippUnfoldT2wFolder/s_{subject}_T1w_acpc_dc_restore.nii.gz --path-T2w $HippUnfoldT2wFolder/s_{subject}_T2w_acpc_dc_restore.nii.gz --cores all --force-output --generate_myelin_map --output-density native 512 2k 8k 18k
+log_Msg "T2w HippUnfold completed."
+log_Msg "Running T1wT2w HippUnfold for subject: $Subject"                                                                                                            #Seriously: don't put a $ here and don't capitalize S in subject...       #Seriously: don't put a $ here and don't capitalize S in subject...
+apptainer run --bind $StudyFolder -e $HIPPUNFOLDPATH $HippUnfoldT1wT2wFolder $HippUnfoldT1wT2wFolder participant --modality T2w --path-T1w $HippUnfoldT1wT2wFolder/s_{subject}_T1w_acpc_dc_restore.nii.gz --path-T2w $HippUnfoldT1wT2wFolder/s_{subject}_T2w_acpc_dc_restore.nii.gz --cores all --force-output --generate_myelin_map --output-density native 512 2k 8k 18k --force-nnunet-model T1T2w
+log_Msg "T1wT2w HippUnfold completed."
```
coalsont, line 76: Is someone going to edit this to have the possibility of running with `--use-conda` instead of their container?

The warnings probably deserve a real newline instead spaces after the `log_Msg`, and we may not need quite that many repetitions.

demsarjure:
Hi,

I already have this edited on my end to support containers and hippunfold
directly. Will commit once it is tested.

Jure

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -0,0 +1,91 @@
...
+
+log_Msg "Created folder structure under $HippUnfoldDIR and copied T1w and T2w images"
+log_Msg "Starting HippUnfold pipeline for subject: $Subject"
+
+log_Msg "Running T1w HippUnfold for subject: $Subject"                                                                                                      #Seriously: don't put a $ here and don't capitalize S in subject...    #Seriously: don't put a $ here and don't capitalize S in subject...
+if [ -z ${HIPPUNFOLDPATH} ] ; then
+  hippunfold $HippUnfoldT1wFolder $HippUnfoldT1wFolder participant --modality T1w --path-T1w $HippUnfoldT1wFolder/s_{subject}_T1w_acpc_dc_restore.nii.gz --path-T2w $HippUnfoldT1wFolder/s_{subject}_T2w_acpc_dc_restore.nii.gz --cores all --force-output --generate_myelin_map --output-density native 512 2k 8k 18k --use-conda
+else
+  apptainer run --bind $StudyFolder -e $HIPPUNFOLDPATH $HippUnfoldT1wFolder $HippUnfoldT1wFolder participant --modality T1w --path-T1w $HippUnfoldT1wFolder/s_{subject}_T1w_acpc_dc_restore.nii.gz --path-T2w $HippUnfoldT1wFolder/s_{subject}_T2w_acpc_dc_restore.nii.gz --cores all --force-output --generate_myelin_map --output-density native 512 2k 8k 18k
+fi
```
coalsont, line 72: If `--use-conda` doesn't need to be the last argument, then we should probably this kind of construct instead, so that we don't accidentally desync the arguments between container vs conda:

```
if [[ "${HIPPUNFOLDPATH:-}" == "" ]]
then
    hippcmd=(hippunfold --use-conda)
else
    hippcmd=(apptainer run --bind "$StudyFolder" -e "$HIPPUNFOLDPATH")
fi
"${hippcmd[@]}" "$HippUnfoldT1wFolder" "$HippUnfoldT1wFolder" participant --modality T1w --path-T1w "$HippUnfoldT1wFolder"/s_{subject}_T1w_acpc_dc_restore.nii.gz --path-T2w "$HippUnfoldT1wFolder"/s_{subject}_T2w_acpc_dc_restore.nii.gz --cores all --force-output --generate_myelin_map --output-density native 512 2k 8k 18k
"${hippcmd[@]}" <next command's arguments>
...
```

demsarjure: I agree, feel free to apply this edit over my latest changes. I think the script is not yet in a final state as there are also still some "cleanup" comments in there (`#Seriously: don't put a $ here and don't capitalize S in subject...`).

coalsont: I have made this edit.  I also noticed that the conditional for specifically the `T1T2w` nnunet command looked backwards (ran `apptainer` when the container path variable was empty, and vice versa).  Why did that pass your testing?

demsarjure: Tests were ran two times, once by using apptainer and once with hippunfold by commenting out the code. If statements were added post-festum. I plan to retest everything once we have official hippunfold 2.0 and a new wb_command and after that we can merge into master, both here and my interface on the QuNex side. So these were not final tests yet.

PR #322, opened 2025-02-19
Diffusion preprocessing longitudinal

mmilch01:
Longitudinal preprocessing saves time by using cross-sectional diffusion eddy_correct output. 
Requirements for the processing directory:
1) Structural timepoints with diffusion preprocessing
2) Longitudinal timepoints with all structural pipelines run

### DIFF
--- a/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
@@ -238,21 +238,21 @@ DiffRes=$(printf "%0.2f" ${DiffRes})
 
 log_Msg "Running Diffusion to Structural Registration"
 ${runcmd} ${HCPPIPEDIR_dMRI}/DiffusionToStructural.sh \
-	--t1folder="${T1wFolder}" \
-	--session="${Session}" \
-	--workingdir="${outdir}/reg" \
-	--datadiffdir="${outdir}/data" \
-	--t1="${T1wImage}" \
-	--t1restore="${T1wRestoreImage}" \
-	--t1restorebrain="${T1wRestoreImageBrain}" \
-	--biasfield="${BiasField}" \
-	--brainmask="${FreeSurferBrainMask}" \
-	--datadiffT1wdir="${outdirT1w}" \
-	--regoutput="${RegOutput}" \
-	--QAimage="${QAImage}" \
-	--dof="${DegreesOfFreedom}" \
-	--gdflag=${GdFlag} \
-	--diffresol=${DiffRes} \
+    --t1folder="${T1wFolder}" \
+    --session="${Session}" \
+    --workingdir="${outdir}/reg" \
+    --datadiffdir="${outdir}/data" \
+    --t1="${T1wImage}" \
+    --t1restore="${T1wRestoreImage}" \
+    --t1restorebrain="${T1wRestoreImageBrain}" \
+    --biasfield="${BiasField}" \
+    --brainmask="${FreeSurferBrainMask}" \
+    --datadiffT1wdir="${outdirT1w}" \
+    --regoutput="${RegOutput}" \
+    --QAimage="${QAImage}" \
+    --dof="${DegreesOfFreedom}" \
+    --gdflag=${GdFlag} \
+    --diffresol=${DiffRes} \
     --t1w-cross2long-xfm="$T1wCross2LongXfm"
 
 to_location="${outdirT1w}/eddylogs"
--- a/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
+++ b/DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
@@ -140,6 +140,9 @@ Defaults to 1" "1"
 
 opts_AddOptional '--printcom' 'runcmd' 'echo' 'to echo or otherwise  output the commands that would be executed instead of  actually running them. --printcom=echo is intended to  be used for testing purposes'
 
+#If set, longitudinal mode is triggered.
+opts_AddOptional '--t1w-cross2long-xfm' 'T1wCross2LongXfm' ".mat Affine transform from cross-sectional T1w_acpc_dc space to longitudinal template space. If set, longitudinal mode is triggered." ""
+
 
 opts_ParseArguments "$@"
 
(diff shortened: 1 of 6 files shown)


PostFreeSurfer/PostFreeSurferPipelineLongPrep.sh
```diff
@@ -365,37 +365,37 @@ if (( TemplateProcessing ==  1 )); then
...
-        --ot1=${AtlasSpaceFolder_template}/T1w \
-        --ot1rest=${AtlasSpaceFolder_template}/T1w_restore \
-        --ot1restbrain=${AtlasSpaceFolder_template}/T1w_restore_brain \
-        --ot2=${AtlasSpaceFolder_template}/T2w \
-        --ot2rest=${AtlasSpaceFolder_template}/T2w_restore \
-        --ot2restbrain=${AtlasSpaceFolder_template}/T2w_restore_brain \
-        --fnirtconfig=${FNIRTConfig}
+    log_Msg "template $Template: performing atlas registration from longitudinal template to MNI152 (FLIRT and FNIRT)"
+
+ #   ${HCPPIPEDIR_PreFS}/AtlasRegistrationToMNI152_FLIRTandFNIRT.sh \
```
glasserm, line 370: Why is this module commented out?

DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
```diff
@@ -253,11 +264,11 @@ from_files=$(ls ${from_directory}/eddy_unwarped_images.* | grep -v .nii)
 
 ${runcmd} mkdir -p ${to_location}
 for filename in ${from_files}; do
-	${runcmd} cp -p ${filename} ${to_location}
+	${runcmd} cp -p ${filename} ${to_location} || ${runcmd} cp ${filename} ${to_location}
```
coalsont, line 267: This should probably be `|| true`, because `cp -p` should already have copied the file if possible, and merely couldn't preserve all of the permissions/timestamps/etc.  Doing an extra `cp` will copy it again, and undo whatever `cp -p` was able to preserve.

PostFreeSurfer/PostFreeSurferPipelineLongPrep.sh
```diff
@@ -406,19 +406,24 @@ if (( TemplateProcessing ==  1 )); then
...
 
         #link altas transforms to the timepoint directory.
-        if [ -d "$AtlasSpaceFolder_timepoint/xfms" ]; then 
+        if [ -d "$AtlasSpaceFolder_timepoint/xfms" ]; then
             rm -rf "$AtlasSpaceFolder_timepoint/xfms"
-        fi                
-        ln -sf "${AtlasSpaceFolder_template}/xfms" "${AtlasSpaceFolder_timepoint}/xfms"
+        fi
+        mkdir -p ${AtlasSpaceFolder_timepoint}
+        ln -sf "${AtlasSpaceFolder_template}/xfms" "${AtlasSpaceFolder_timepoint}/"
```
coalsont, line 413: This edit is already in master, including it in this PR is causing it to have a merge conflict.

PostFreeSurfer/PostFreeSurferPipelineLongPrep.sh
```diff
@@ -406,19 +406,24 @@ if (( TemplateProcessing ==  1 )); then
...
-        
+
         #mask the acpc_dc space images.
         ${FSLDIR}/bin/fslmaths $T1w_dir_long/T1w_acpc_dc -mas "$Timepoint_brain_mask_acpc_dc" $T1w_dir_long/T1w_acpc_dc_brain
         ${FSLDIR}/bin/fslmaths $T1w_dir_long/T1w_acpc_dc_restore -mas "$Timepoint_brain_mask_acpc_dc" $T1w_dir_long/T1w_acpc_dc_restore_brain
 
+        # Bias field in MNI space.
+        ${FSLDIR}/bin/applywarp --rel --interp=spline -i $T1w_dir_long/BiasField_acpc_dc -r ${T1wTemplate} -w ${AtlasSpaceFolder_template}/$WARP \
+            -o ${AtlasSpaceFolder_timepoint}/BiasField
+
```
coalsont, line 426: This is also already in master, it should not have been repeated in a PR commit (don't fix problems that are already in master in an unrelated PR, keep PRs single-purpose).  If the only other edit (commenting out a stage) is also wrong, I suggest reverting this file to the version in master, this should do the trick (while on the PR branch):

`git checkout master -- PostFreeSurfer/PostFreeSurferPipelineLongPrep.sh`

Then stage (or it might already be staged) and commit.

coalsont: Err, scratch that, having a copy of the master version when the branch point is older is the problem here...

coalsont: I think I'll just remove the last commit and rebase.

coalsont: Please save any commits on this branch you haven't pushed (`git branch save-commits`), uncommitted work (`git stash`) and reset your diffusion branch to origin:

```
git checkout --detach
git remote update
git branch -f diffusion_preprocessing_longitudinal origin/diffusion_preprocessing_longitudinal
git checkout diffusion_preprocessing_longitudinal
```

DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
```diff
@@ -242,7 +252,8 @@ ${runcmd} ${HCPPIPEDIR_dMRI}/DiffusionToStructural.sh \
 	--QAimage="${QAImage}" \
 	--dof="${DegreesOfFreedom}" \
 	--gdflag=${GdFlag} \
-	--diffresol=${DiffRes}
+	--diffresol=${DiffRes} \
+    --t1w-cross2long-xfm="$T1wCross2LongXfm"
```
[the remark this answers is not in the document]
mmilch01: ok

DiffusionPreprocessing/DiffPreprocPipeline_PostEddy.sh
coalsont: I am aware.  However, if `cp -p` errors, the effects it did have are still better than a standard `cp`.  I suppose `|| true` does have the drawback that if it wasn't able to copy the file at all, the script will continue (while trying a bare `cp` would also fail and the script wouldn't continue).  However, the script just created that folder anyway, so unless the filesystem is full, it should have been able to copy it.

I guess the first question is, why did you make this change?

mmilch01: This was a fix for diffusion pipeline failing because of cp -p erroring on brainmappers when trying to change permissions. To me copying things twice is  safer than assuming that cp -p would gracefully copy given it throws an error. Are you sure that's the case?

PR #324, opened 2025-02-26
Force provided fMRIScout to be a single volume

mharms:


glasserm (approved):

coalsont:
How do you end up with a multi-frame scout file?  Would averaging be more appropriate?

PR #326, opened 2025-03-05
[Reclean] Fix the volume output name and adjust with CleanSubstring

cyang31:
To fix the reclean multi-run pipeline that previously only generated `*_clean.nii.gz` volume output.


mharms (commented):
That's basically what I was expecting would be necessary.

coalsont:
Seems reasonable.

glasserm (approved):

PR #329, opened 2025-03-13
Longitudinal ICAFix+fMRIVolume multi-echo fix

mmilch01:
I. Enables longitudinal mode in the ICAFIX pipeline. Only supports multi-run mode. To use:

1) Run all cross-sectional ICAFIX pipelines on each timepoint
2) Longitudinal timepoints require the same sequence of longitudinal pipelines to be run as a pre-requisite, i.e. longitudinal FreeSurfer, PostFreeSurfer, fMRIVolume, and fMRISurface. Then, run ReApplyFixMultiRun pipeline on each longitudinal timepoint, setting the optional --is-longitudinal and --longitudinal-session flags. 

II. Fixes issues to correctly run fMRIVolume pipeline in longitudinal mode on multi-echo runs.

### DIFF
--- a/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
+++ b/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
@@ -1,0 +1,247 @@
+#!/bin/bash
+#set -xv
+
+# Global default values
+DEFAULT_STUDY_FOLDER="${HOME}/data/Pipelines_ExampleData"
+DEFAULT_SUBJECT_LIST="HCA6002236"
+DEFAULT_ENVIRONMENT_SCRIPT="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh"
+DEFAULT_RUN_LOCAL="FALSE"
+DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
+DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3"
+
+#
+# Function Description
+#	Get the command line options for this script
+#
+# Global Output Variables
+#	${StudyFolder}			- Path to folder containing all subjects data in subdirectories named 
+#							  for the subject id
+#	${Subjlist}				- Space delimited list of subject IDs
+#	${EnvironmentScript}	- Script to source to setup pipeline environment
+#	${RunLocal}				- Indication whether to run this processing "locally" i.e. not submit
+#							  the processing to a cluster or grid
+#
+get_options() {
+	local scriptName=$(basename ${0})
+	local arguments=("$@")
+
+	# initialize global output variables
+	StudyFolder="${DEFAULT_STUDY_FOLDER}"
+	Subjlist="${DEFAULT_SUBJECT_LIST}"
+	EnvironmentScript="${DEFAULT_ENVIRONMENT_SCRIPT}"
+	RunLocal="${DEFAULT_RUN_LOCAL}"
+	
+	# make these into command-line options later if desired.
+	PossibleVisits=($DEFAULT_POSSIBLE_VISITS)
+	Templates=($DEFAULT_TEMPLATE_LIST)
+
+	# parse arguments
+	local index=0
+	local numArgs=${#arguments[@]}
+	local argument
+
+	while [ ${index} -lt ${numArgs} ]
+	do
+		argument=${arguments[index]}
+
+		case ${argument} in
+			--StudyFolder=*)
+				StudyFolder=${argument#*=}
+				index=$(( index + 1 ))
+				;;
+			--Subject=*)
+				Subjlist=${argument#*=}
+				index=$(( index + 1 ))
+				;;
+			--EnvironmentScript=*)
+				EnvironmentScript=${argument#*=}
+				index=$(( index + 1 ))
+				;;
+			--runlocal | --RunLocal)
+				RunLocal="TRUE"
+				index=$(( index + 1 ))
+				;;
+			*)
+				echo "ERROR: Unrecognized Option: ${argument}"
+				exit 1
+				;;
+		esac
+	done
+
+	# check required parameters
+	if [ -z ${StudyFolder} ]
+	then
+		echo "ERROR: StudyFolder not specified"
+		exit 1
+	fi
+
+	if [ -z "${Subjlist}" ]
+	then
+		echo "ERROR: Subjlist not specified"
+		exit 1
+	fi
+
+	if [ -z ${EnvironmentScript} ]
+	then
+		echo "ERROR: EnvironmentScript not specified"
+		exit 1
+	fi
+
+	if [ -z ${RunLocal} ]
+	then
+		echo "ERROR: RunLocal is an empty string"
+		exit 1
+	fi
+
+	# report options
+	echo "-- ${scriptName}: Specified Command-Line Options: -- Start --"
+	echo "   StudyFolder: ${StudyFolder}"
+	echo "   Subjlist: ${Subjlist}"
+	echo "	 Templates: ${Templates[@]}"
+	echo "	 Possible visits: ${PossibleVisits[@]}"
+	echo "   EnvironmentScript: ${EnvironmentScript}"
+	echo "   RunLocal: ${RunLocal}"
+	echo "-- ${scriptName}: Specified Command-Line Options: -- End --"
+}  # get_options()
+
+# Function description
+#
+# For the given subject, identify_timepoins creates a string listing @ separated visits/timepoints to process
+# Uses StudyFolder, ExcludeVisits, PossibleVisits global variables as input.
+# Subject must be supplied as the first argument. 
+
+function identify_timepoints
+{
+    local subject=$1
+    local tplist=""
+    local tp visit n
+
+    #build the list of timepoints
+    n=0
+    for visit in ${PossibleVisits[*]}; do
+        tp="${subject}_${visit}"
+        if [ -d "$StudyFolder/$tp" ] && ! [[ " ${ExcludeVisits[*]+${ExcludeVisits[*]}} " =~ [[:space:]]"$tp"[[:space:]] ]]; then
+             if (( n==0 )); then 
+                    tplist="$tp"
+             else
+                    tplist="$tplist@$tp"
+             fi
+        fi
+        ((n++))
+    done
+    echo $tplist
+}
+
+#
+# Function Description
+#	Main processing of this script
+#
+#	Gets user specified command line options and runs a batch of ReApplyFix processing
+#
+main() {
+	# get command line options
+	get_options "$@"
+
+	# set up pipeline environment variables and software
+	source "$EnvironmentScript"
+
+	# set list of fMRI on which to run ReApplyFixPipeline, separate MR FIX groups with %, use spaces (or @ like dedrift...) to otherwise separate runs
+	# ReApplyFixPipeline
+	fMRINames="rfMRI_REST1_AP@rfMRI_REST1_PA@tfMRI_VISMOTOR_PA@tfMRI_CARIT_PA@tfMRI_FACENAME_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"
+  
+	# specify the name of concatenated folder
+	# if run Multi-Run specify ConcatNames as null string
+	ConcatNames="fMRI_CONCAT_ALL"
+	#ConcatNames="tfMRI_WM_GAMBLING_MOTOR_RL_LR@tfMRI_LANGUAGE_SOCIAL_RELATIONAL_EMOTION_RL_LR"
+
+	# set highpass
+	highpass=2000
+	
+	#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
+	#DO NOT include "-q " at the beginning
+	#default to no queue, implying run local
+	#QUEUE=""
+	QUEUE="short.q"
+
+	# regression mode
+	RegName="NONE"
+
+	# low res mesh
+	LowResMesh=32
+
+	# matlab mode
+	# 0 = Use compiled MATLAB
+	# 1 = Use interpreted MATLAB
+	# 2 = Use interpreted Octave
+	MatlabMode=1
+
+	# motion regression or not
+	MotionReg=FALSE
+
+	# clean up intermediates
+	DeleteIntermediates=FALSE
+
+	#MR FIX config support for non-HCP settings
+	config=""
+	processingmode="HCPStyleData"
+	#uncomment the below two lines for legacy-style data
+	#config="$HCPPIPEDIR"/ICAFIX/config/legacy.conf
+	#processingmode="LegacyStyleData"
+
+	if [[ "${RunLocal}" == "TRUE" || "$QUEUE" == "" ]]; then
+		queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+	else
+		queuing_command=("${FSLDIR}/bin/fsl_sub" -q "$QUEUE")
+	fi
+
+	for Subject in ${Subjlist}; do
+		TemplateLong="${Templates[i]}"
+		Timepoint_list_cross_at_separated=$(identify_timepoints "$Subject")
+		IFS=@ read -r -a Timepoint_list_cross <<< "${Timepoint_list_cross_at_separated}"
+		
+		for TimepointCross in "${Timepoint_list_cross[@]}"; do
+			TimepointLong=${TimepointCross}.long.${TemplateLong}
+			if [ -z ${ConcatNames} ]; then
+					echo "Single run legacy processing is not supported in longitudinal mode."
+					exit -1
+			else # Multi-Run				
+				#need arrays to sanity check number of concat groups
+				IFS=' @' read -a concatarray <<< "${ConcatNames}"
+				IFS=% read -a fmriarray <<< "${fMRINames}"
+				
+				if ((${#concatarray[@]} != ${#fmriarray[@]})); then
+					echo "ERROR: number of names in ConcatNames does not match number of fMRINames groups"
+					exit 1
+				fi
+
+				for ((i = 0; i < ${#concatarray[@]}; ++i)); do
+					ConcatName="${concatarray[$i]}"
+					fMRINamesGroup="${fmriarray[$i]}"
+				
+					"${queuing_command[@]}" "$HCPPIPEDIR"/ICAFIX/ReApplyFixMultiRunPipeline.sh \
+						--path="$StudyFolder" \
+						--session="$TimepointCross" \
+						--fmri-names="$fMRINamesGroup" \
+						--high-pass="$highpass" \
+						--reg-name="$RegName" \
+						--concat-fmri-name="$ConcatName" \
+						--low-res-mesh="$LowResMesh" \
+						--matlab-run-mode="$MatlabMode" \
+						--motion-regression="$MotionReg" \
+						--config="$config" \
+						--processing-mode="$processingmode" \
+						--is-longitudinal="TRUE" \
+						--longitudinal-session="$TimepointLong"
+					
+					echo "${TimepointCross} ${TimepointLong} ${ConcatName}"					
+				done
+			fi
+		done #iterating over timepoints
+	done #iterating over subjects
+}  # main()
+
+#
+# Invoke the main function to get things started
+#
+main "$@"
+
--- a/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
+++ b/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
@@ -9,6 +9,7 @@ DEFAULT_RUN_LOCAL="FALSE"
 DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
 DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3"
 
+ExcludeVisits=()
 #
 # Function Description
 #	Get the command line options for this script
--- a/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
+++ b/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
@@ -1,13 +1,23 @@
 #!/bin/bash
 #set -xv
 
+
+# Naming assumptions
+# This script assumes that:
+# 1. The cross-sectional session dirs are named <Subject>_<PossibleVisit>, 
+# 2. Longitudinal session dirs are named <Subject>_<Possible_Visit>.long.<Template>,
+# 3. Both longitudinal and cross-sectional sessions exist in the study folder.
+
 # Global default values
 DEFAULT_STUDY_FOLDER="${HOME}/data/Pipelines_ExampleData"
-DEFAULT_SUBJECT_LIST="HCA6002236"
+DEFAULT_SUBJECT_LIST="HCA6002236 HCA6002237"
 DEFAULT_ENVIRONMENT_SCRIPT="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh"
 DEFAULT_RUN_LOCAL="FALSE"
+
 DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
-DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3"
+
+# The template list should include labels for the longitudinal template of each subject, similar to those used in longitudinal FreeSurfer.
+DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3 HCA6002237_V1_V2_V3"
 
 ExcludeVisits=()
 #
@@ -15,12 +25,12 @@ ExcludeVisits=()
 #	Get the command line options for this script
 #
 # Global Output Variables
-#	${StudyFolder}			- Path to folder containing all subjects data in subdirectories named 
-#							  for the subject id
-#	${Subjlist}				- Space delimited list of subject IDs
-#	${EnvironmentScript}	- Script to source to setup pipeline environment
-#	${RunLocal}				- Indication whether to run this processing "locally" i.e. not submit
-#							  the processing to a cluster or grid
+#	${StudyFolder}			-	Path to folder containing all subjects data in subdirectories named 
+#								for the subject id
+#	${Subjlist}				-	Space delimited list of subject IDs
+#	${EnvironmentScript}	-	Script to source to setup pipeline environment
+#	${RunLocal}				-	Indication whether to run this processing "locally" i.e. not submit
+#								the processing to a cluster or grid
 #
 get_options() {
 	local scriptName=$(basename ${0})
@@ -107,7 +117,7 @@ get_options() {
 
 # Function description
 #
-# For the given subject, identify_timepoins creates a string listing @ separated visits/timepoints to process
+# For the given subject, identify_timepoints creates a string listing @ separated visits/timepoints to process
 # Uses StudyFolder, ExcludeVisits, PossibleVisits global variables as input.
 # Subject must be supplied as the first argument. 
 
--- a/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
+++ b/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
@@ -22,7 +22,7 @@ DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3 HCA6002237_V1_V2_V3"
 ExcludeVisits=()
 #
 # Function Description
-#   Get the command line options for this script
+# Get the command line options for this script
 #
 # Global Output Variables
 #   ${StudyFolder}           - Path to folder containing all subjects data in subdirectories named 
--- a/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
+++ b/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
@@ -156,13 +156,13 @@ main() {
 	#ConcatNames="tfMRI_WM_GAMBLING_MOTOR_RL_LR@tfMRI_LANGUAGE_SOCIAL_RELATIONAL_EMOTION_RL_LR"
 
 	# set highpass
-	highpass=2000
+	highpass=0
 	
 	#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
 	#DO NOT include "-q " at the beginning
 	#default to no queue, implying run local
 	#QUEUE=""
-	QUEUE="short.q"
+	QUEUE="hcp_priority.q"
 
 	# regression mode
 	RegName="NONE"
--- a/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
+++ b/Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
@@ -22,15 +22,15 @@ DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3 HCA6002237_V1_V2_V3"
 ExcludeVisits=()
 #
 # Function Description
-#	Get the command line options for this script
+#   Get the command line options for this script
 #
 # Global Output Variables
-#	${StudyFolder}			-	Path to folder containing all subjects data in subdirectories named 
-#								for the subject id
-#	${Subjlist}				-	Space delimited list of subject IDs
-#	${EnvironmentScript}	-	Script to source to setup pipeline environment
-#	${RunLocal}				-	Indication whether to run this processing "locally" i.e. not submit
-#								the processing to a cluster or grid
+#   ${StudyFolder}           - Path to folder containing all subjects data in subdirectories named 
+#                              for the subject id
+#   ${Subjlist}              - Space delimited list of subject IDs
+#   ${EnvironmentScript}     - Script to source to setup pipeline environment
+#   ${RunLocal}              - Indication whether to run this processing "locally" i.e. not submit
+#                              the processing to a cluster or grid
 #
 get_options() {
 	local scriptName=$(basename ${0})
--- a/ICAFIX/ReApplyFixMultiRunPipeline.sh
+++ b/ICAFIX/ReApplyFixMultiRunPipeline.sh
@@ -4,7 +4,7 @@
 #
 # ## Copyright Notice
 #
-# Copyright (C) 2017-2019 The Human Connectome Project/Connectome Coordination Facility
+# Copyright (C) 2017-2025 The Human Connectome Project/Connectome Coordination Facility
 #
 # * Washington University in St. Louis
 # * University of Minnesota
@@ -14,6 +14,7 @@
 #
 # * Matthew F. Glasser, Department of Anatomy and Neurobiology, Washington University in St. Louis
 # * Timothy B. Brown, Neuroinformatics Research Group, Washington University in St. Louis
+# * Mikhail Milchenko (longitudinal patch), Computational Imaging Research Center, Washington University in St. Louis
 #
 # ## Product
 #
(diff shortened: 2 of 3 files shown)


ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	local StudyFolder="$1" SessionCross="$2" SessionLong="$3" fMRINames="$4" ConcatName="$5" HighPass="$6"
+	local fmri S T
+	#copy for concatentated run
+	S="$StudyFolder/$SessionCross/MNINonLinear/Results/$ConcatName"
+	T="$StudyFolder/$SessionLong/MNINonLinear/Results/$ConcatName"
+	local file files_to_copy="Movement_Regressors_demean.txt ReclassifyAsNoise.txt ReclassifyAsSignal.txt"
+	mkdir -p $T
+	for file in $files_to_copy; do
+		cp $S/$file $T/
+	done
```
coalsont, line 197:
```suggestion
	mkdir -p "$T"
	for file in $files_to_copy; do
		cp "$S"/"$file" "$T"/
	done
```
And similar below.  Commands need quoted variables to handle whitespace in folder/file names properly.

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	for file in $files_to_copy; do
+		cp $ICADir/filtered_func_data.ica/$file $ICADirLong/filtered_func_data.ica/
+	done	
+	
+	#copy for individual fMRI runs
+	for fmri in $fMRINames; do		
+		S="$StudyFolder/$SessionCross/MNINonLinear/Results/$fmri"
+		T="$StudyFolder/$SessionLong/MNINonLinear/Results/$fmri"
+		mkdir -p $T/${fmri}_hp$HighPass.ica/mc
+		cp $S/${fmri}_hp$HighPass.ica/mc/prefiltered_func_data_mcf.par $T/${fmri}_hp$HighPass.ica/mc/
```
coalsont, line 231: Underscores outside of quotes will be literal, so you can just quote each variable without using `{}`.
```suggestion
		mkdir -p "$T"/"$fmri"_hp"$HighPass".ica/mc
		cp "$S"/"$fmri"_hp"$HighPass".ica/mc/prefiltered_func_data_mcf.par "$T"/"$fmri"_hp"$HighPass".ica/mc/		
```

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -667,27 +667,27 @@ if (( ! IsLongitudinal )); then
         mkdir "$fMRIFolder"
     fi
 else
-    #copy directory structure and create symbolic link per each file under source. 
+    #copy directory structure. 
 	mkdir -p "$ResultsFolderLong"
-    if ! cp -rf "$ResultsFolder" "$ResultsFolderLong"
-    then
-        log_Err_Abort "Copying cross-sectional output $ResultsFolder to longitudinal session folder $ResultsFolderLong failed."
-    fi
```
coalsont, line 675: Is this deletion a bugfix, or a temporary debugging speed hack?

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,247 @@
...
+	# ReApplyFixPipeline
+	fMRINames="rfMRI_REST1_AP@rfMRI_REST1_PA@tfMRI_VISMOTOR_PA@tfMRI_CARIT_PA@tfMRI_FACENAME_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"
+  
+	# specify the name of concatenated folder
+	# if run Multi-Run specify ConcatNames as null string
+	ConcatNames="fMRI_CONCAT_ALL"
+	#ConcatNames="tfMRI_WM_GAMBLING_MOTOR_RL_LR@tfMRI_LANGUAGE_SOCIAL_RELATIONAL_EMOTION_RL_LR"
+
+	# set highpass
+	highpass=2000
```
glasserm, line 158: Let's set this default correctly to highpass=0

demsarjure (commented):
I have nothing to add on top of Matt and Tim. I plan to implement the QuNex interface for this and do some test runs later this week. Will let you know how those go.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: This is a bug fix. Results folder need not be copied.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mmilch01: ok

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,247 @@
...
+	
+	#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
+	#DO NOT include "-q " at the beginning
+	#default to no queue, implying run local
+	#QUEUE=""
+	QUEUE="short.q"
+
```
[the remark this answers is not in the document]
mmilch01: Ok. I assume it's hcp_priority.q.

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	mkdir -p "$T"
+	for file in $files_to_copy; do
+		cp "$S"/"$file" "$T"/
+	done
+	
+	local ICADir="$S/${ConcatName}_hp$HighPass.ica"
+	local ICADirLong="$T/${ConcatName}_hp$HighPass.ica"
+	mkdir -p "$ICADirLong"				
+	if [ -f "$ICADir/HandNoise.txt" ]; then 
+		cp $ICADir/HandNoise.txt $ICADirLong/
```
coalsont, line 203: It would be nice if you quoted the rest of these variables in the command lines.
```suggestion
		cp "$ICADir"/HandNoise.txt "$ICADirLong"/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -46,9 +47,9 @@ opts_SetScriptDescription "ReApplyFix Pipeline for MultiRun ICA+FIX"
 #WARNING: this "default" is also used to special case whether filenames have ".<num>k" added to their regstring
 G_DEFAULT_LOW_RES_MESH=32
 
-opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects" '--path'
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all sessions" '--path'
 
-opts_AddMandatory '--subject' 'Subject' 'subject ID' "(e.g. 100610)"
+opts_AddMandatory '--session' 'SessionCross' 'session ID' "(e.g. 100610)" "--subject"
```
mharms, line 52:
```suggestion
opts_AddMandatory '--session' 'SessionCross' 'session ID' "(e.g. 100610_V1)" "--subject"
```
I suggest we use an example that makes clearer it is a "session" not a "subject" ID.

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
-	[ -e "${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_hp${HighPass}.ica/HandNoise.txt" ]
+	[ -e "${StudyFolder}/${Session}/MNINonLinear/Results/${fMRIName}/${fMRIName}_hp${HighPass}.ica/HandNoise.txt" ]
+}
+
+# ------------------------------------------------------------------------------
+# Copy necessary files from cross-sectional to longitudinal target
+# ------------------------------------------------------------------------------
+function copy_to_longitudinal()
+{
+	local StudyFolder="$1" SessionCross="$2" SessionLong="$3" fMRINames="$4" ConcatName="$5" HighPass="$6"
```
mharms, line 188: Can we list one local variable per line, for consistency with other code? (e.g., `have_hand_reclassification` just above)?
```suggestion
	local StudyFolder="${1}" 
	local SessionCross="${2}" 
	local SessionLong="${3}" 
	local fMRINames="${4}" 
	local ConcatName="${5}" 
	local HighPass="${6}"
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	local fmri S T
+	#copy for concatentated run
+	S="$StudyFolder/$SessionCross/MNINonLinear/Results/$ConcatName"
+	T="$StudyFolder/$SessionLong/MNINonLinear/Results/$ConcatName"
+	local file files_to_copy="Movement_Regressors_demean.txt ReclassifyAsNoise.txt ReclassifyAsSignal.txt"
+	mkdir -p "$T"
+	for file in $files_to_copy; do
+		cp "$S"/"$file" "$T"/
```
mharms, line 196:
```suggestion
		cp -p "$S"/"$file" "$T"/
```
Please preserve time stamps when copying files, so that if they aren't subsequently modified, the original time stamp remains.

ICAFIX/ReApplyFixMultiRunPipeline.sh
mharms, line 203:
```suggestion
		cp -p $ICADir/HandNoise.txt $ICADirLong/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	local ICADir="$S/${ConcatName}_hp$HighPass.ica"
+	local ICADirLong="$T/${ConcatName}_hp$HighPass.ica"
+	mkdir -p "$ICADirLong"				
+	if [ -f "$ICADir/HandNoise.txt" ]; then 
+		cp $ICADir/HandNoise.txt $ICADirLong/
+	fi
+	files_to_copy="fix4melview_HCP_Style_Single_Multirun_Dedrift_thr10.txt fix4melview_HCP_Style_Single_Multirun_Dedrift_thr10.wb_annsub.csv hand_labels_noise.txt HandNoise.txt \
+	HandSignal.txt Noise.txt ReclassifyAsNoise.txt ReclassifyAsSignal.txt Signal.txt .fix"
+	for file in $files_to_copy; do
+		cp $ICADir/$file $ICADirLong/
```
mharms, line 208:
```suggestion
		cp -p $ICADir/$file $ICADirLong/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	done 
+	
+	mkdir -p $ICADirLong/mc
+	cp $ICADir/mc/prefiltered_func_data_mcf_conf.nii.gz $ICADirLong/mc/
```
mharms, line 212:
```suggestion
	cp -p $ICADir/mc/prefiltered_func_data_mcf_conf.nii.gz $ICADirLong/mc/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	cp $ICADir/mc/prefiltered_func_data_mcf_conf_hp.nii.gz $ICADirLong/mc/
```
mharms, line 213:
```suggestion
	cp -p $ICADir/mc/prefiltered_func_data_mcf_conf_hp.nii.gz $ICADirLong/mc/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	cp $ICADir/mc/prefiltered_func_data_mcf.par $ICADirLong/mc/
```
mharms, line 214:
```suggestion
	cp -p $ICADir/mc/prefiltered_func_data_mcf.par $ICADirLong/mc/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	
+	mkdir -p $ICADirLong/fix/
+	cp $ICADir/fix/features.csv $ICADirLong/fix/
```
mharms, line 217:
```suggestion
	cp -p $ICADir/fix/features.csv $ICADirLong/fix/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	
+	files_to_copy="eigenvalues_percent ICAVolumeSpace.txt melodic_FTmix melodic_FTmix.sdseries.nii \
+	melodic_ICstats melodic_mix melodic_mix.sdseries.nii melodic_Tmodes melodic_unmix"
+	mkdir -p $ICADirLong/filtered_func_data.ica
+	for file in $files_to_copy; do
+		cp $ICADir/filtered_func_data.ica/$file $ICADirLong/filtered_func_data.ica/
```
mharms, line 223:
```suggestion
		cp -p $ICADir/filtered_func_data.ica/$file $ICADirLong/filtered_func_data.ica/
```

ICAFIX/ReApplyFixMultiRunPipeline.sh
```diff
@@ -168,11 +173,63 @@ determine_old_or_new_fsl()
...
+	done	
+	
+	#copy for individual fMRI runs
+	for fmri in $fMRINames; do		
+		S="$StudyFolder/$SessionCross/MNINonLinear/Results/$fmri"
+		T="$StudyFolder/$SessionLong/MNINonLinear/Results/$fmri"
+		mkdir -p "$T"/"$fmri"_hp"$HighPass".ica/mc
+		cp "$S"/"$fmri"_hp"$HighPass".ica/mc/prefiltered_func_data_mcf.par "$T"/"$fmri"_hp"$HighPass".ica/mc/
```
mharms, line 231:
```suggestion
		cp -p "$S"/"$fmri"_hp"$HighPass".ica/mc/prefiltered_func_data_mcf.par "$T"/"$fmri"_hp"$HighPass".ica/mc/		
```

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,248 @@
...
+DEFAULT_RUN_LOCAL="FALSE"
+DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
+DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3"
+
+ExcludeVisits=()
+#
+# Function Description
+#	Get the command line options for this script
+#
+# Global Output Variables
```
mharms, line 17: Odd alignment through these comments.  Tab vs spaces...?

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,248 @@
...
+	echo "	 Templates: ${Templates[@]}"
+	echo "	 Possible visits: ${PossibleVisits[@]}"
+	echo "   EnvironmentScript: ${EnvironmentScript}"
+	echo "   RunLocal: ${RunLocal}"
+	echo "-- ${scriptName}: Specified Command-Line Options: -- End --"
+}  # get_options()
+
+# Function description
+#
+# For the given subject, identify_timepoins creates a string listing @ separated visits/timepoints to process
```
mharms, line 110:
```suggestion
# For the given subject, identify_timepoints creates a string listing @ separated visits/timepoints to process
```
Fix typo.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,248 @@
+#!/bin/bash
+#set -xv
+
+# Global default values
+DEFAULT_STUDY_FOLDER="${HOME}/data/Pipelines_ExampleData"
+DEFAULT_SUBJECT_LIST="HCA6002236"
```
mharms, line 6: Can this be more than a single subject?  If so, I'd make the "default" example represent that.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,248 @@
...
+DEFAULT_ENVIRONMENT_SCRIPT="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh"
+DEFAULT_RUN_LOCAL="FALSE"
+DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
+DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3"
```
mharms, line 10: Ditto.

mharms: Also,  comment on what TEMPLATE_LIST represents would be helpful, esp. considering that this is an "Example" to guide users.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
coalsont: Looks like section headings to me.  Probably deliberate.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
```diff
@@ -0,0 +1,248 @@
+#!/bin/bash
+#set -xv
+
+# Global default values
+DEFAULT_STUDY_FOLDER="${HOME}/data/Pipelines_ExampleData"
+DEFAULT_SUBJECT_LIST="HCA6002236"
+DEFAULT_ENVIRONMENT_SCRIPT="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh"
+DEFAULT_RUN_LOCAL="FALSE"
+DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
```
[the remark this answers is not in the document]
coalsont: This is an example launcher script, not a pipeline.  If someone's session naming convention is different, they are expected to edit the logic to construct their names correctly.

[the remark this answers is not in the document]
mharms: Yes, but no guidance is provided that, for example, session IDs are expected to be of the form: `<subjectID>_<visit-label>`

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mharms: I should have tagged more of the code here. Look at the alignment of the stuff *below* the "#Global Output Variables" line, which is inconsistent and doesn't look deliberate to me.

coalsont: Ah, I see, the description side is disordered.

demsarjure:
I tested everything on our end, my tests worked without issues.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mmilch01: Ok.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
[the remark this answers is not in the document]
mmilch01: Ok, I added the comment on naming assumptions.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mmilch01: Not 100% sure what the problem is, but tried to fix by replacing spaces before descriptions with tabs.

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mmilch01: Ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: Ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: Ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: Ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: Ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: ok

ICAFIX/ReApplyFixMultiRunPipeline.sh
mmilch01: ok

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mharms: Not fixed, when I just view the full within GitHub.

coalsont: Tabs are different widths depending on people's editor settings (and some editors by default won't let you type a tab after other characters, because they assume you want to change the leading indent instead).  This is why I generally prefer spaces, they always look the same to everyone.

glasserm:
Is this ready to merge?

Examples/Scripts/ReApplyFixMultiRunBatch-long.sh
mmilch01: Ok, I replaced tabs with spaces in this block.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -1112,22 +1116,27 @@ ${FSLDIR}/bin/imrm ${fMRIFolder}/${NameOffMRI}_nonlin_norm
...
+
+		   ${FSLDIR}/bin/imrm "${fMRIFolder}/${tcsEchoesGdc[iEcho]}"
+
+		   ${FSLDIR}/bin/imrm "${fMRIFolder}/${sctEchoesOrig[iEcho]}"
+		   ${FSLDIR}/bin/imrm "${fMRIFolder}/${sctEchoesGdc[iEcho]}"
+		   ${FSLDIR}/bin/imrm "${fMRIFolder}/${sctEchoesGdc[iEcho]}_mask"
+	   done
+	   ${FSLDIR}/bin/imrm "${tcsEchoes[@]}"
+	   ${FSLDIR}/bin/imrm "${tcsEchoesMu[@]}"
+	fi
```
coalsont, line 1139: This has both tabs and spaces for indentation (within a line).  Sorry for not noticing earlier.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -1112,22 +1116,27 @@ ${FSLDIR}/bin/imrm ${fMRIFolder}/${NameOffMRI}_nonlin_norm
...
-        ${FSLDIR}/bin/imrm "${fMRIFolder}/${tcsEchoesOrig[iEcho]}_SBRef_nonlin"
-
-        ${FSLDIR}/bin/imrm "${fMRIFolder}/${tcsEchoesGdc[iEcho]}"
-
-        ${FSLDIR}/bin/imrm "${fMRIFolder}/${sctEchoesOrig[iEcho]}"
-        ${FSLDIR}/bin/imrm "${fMRIFolder}/${sctEchoesGdc[iEcho]}"
-        ${FSLDIR}/bin/imrm "${fMRIFolder}/${sctEchoesGdc[iEcho]}_mask"
-    done
-    ${FSLDIR}/bin/imrm "${tcsEchoes[@]}"
-    ${FSLDIR}/bin/imrm "${tcsEchoesMu[@]}"
```
coalsont, line 1130: So, to be clear, to support the possibility of a session being part of longitudinal analysis, we NEVER clean these up in non-longitudinal runs?

coalsont: I'm not immediately seeing where these names (as opposed to a renamed copy) are needed specifically in longitudinal mode.  The logic I would expect is more like "clean them up when they are generated, and then maybe check if they exist in longitudinal (if they might have been copied from something that didn't get cleaned up) and if so, remove them there also".

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: No problem, fixed.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
mmilch01: > I'm not immediately seeing where these names (as opposed to a renamed copy) are needed specifically in longitudinal mode. The logic I would expect is more like "clean them up when they are generated, and then maybe check if they exist in longitudinal (if they might have been copied from something that didn't get cleaned up) and if so, remove them there also".

Neither do I, so to answer your question I need to revisit and see why this failed before with this cleanup enabled on cross-sectional prior to running longitudinal. I'm rerunning the longitudinal session without these files.

mmilch01: I've been gradually adding these files to cross-sectional and checking where the pipeline failed. Here is an incomplete analysis (still checking on things):
1. <Session>/<fMRIRun>/Scout_gdc_E<echo-number>.nii.gz used in DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh, line 740
2. <Session>/<fMRIRun>/<fMRIRun>_orig_E<echo-number>.nii.gz used in OneStepResampling.sh, line 312
3. <Session>/<fMRIRun>/Scout_orig_E<echo-number>.nii.gz used in OneStepResampling.sh, line 322

mmilch01: The files listed above are required for longitudinal multi-echo to succeed. So I added an option "--extra-multiecho-cleanup" which defaults to TRUE so all the multiecho files are cleaned as before. If set to FALSE, only these three types are kept.

coalsont: So, when run on a longitudinal subject, it first gets run on...the template?...in cross-sectional mode with that cleanup flag set to false, then run on a longitudinal timepoint in longitudinal mode which copies these temporaries, and then deletes the copies.  Is that currently correct?  That means the longitudinal template has these temporaries and the pipelines never delete them (and it includes effectively a copy of the timeseries, not just scouts).

It seems like the trouble may all come from skipping the splitting commands and using a copy instead.  It feels like it would just be simpler to rerun the -volume-merges that are currently skipped up at https://github.com/Washington-University/HCPpipelines/blob/3242b43792089dbdb415f626f5f5d79d3b7e7ef2/fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh#L824 , and have them always cleaned up inside the same pipeline call that created them, and then we wouldn't have this issue of copying files that should have been temporaries internal to another pipelines call, and therefore not knowing when they should be cleaned up.  The splitting commands shouldn't take much time.

glasserm: That option is fine to have (though I would probably default to cleaning up the split echos and the option would be to preserve them).  Tim's idea was to do the echo splits again in longitudinal mode, which, after discussion, does seem like a better idea.

mmilch01: > So, when run on a longitudinal subject, it first gets run on...the template?...in cross-sectional mode with that cleanup flag set to false, then run on a longitudinal timepoint in longitudinal mode which copies these temporaries, and then deletes the copies. Is that currently correct? 

No, fMRIVolume only gets run on timepoints, although I guess that doesn't affect your point.

coalsont: > No, fMRIVolume only gets run on timepoints, although I guess that doesn't affect your point.

If it is only run on longitudinal timepoints, why does it get run more than once on the same input in the longitudinal context?  Does it first resample with the cross-sectional transform (such that it isn't fully aligned with the longitudinal structural), then come back and resample with the longitudinal transform?  The longitudinal alignment is calculated before the fMRI data is touched, right?

Are we planning to always maintain separate "visit-specific longitudinal-ignorant" and "longitudinal" folders (and volume spaces) for the visits, both with complete copies of the processed fMRI data?

mmilch01: I reverted to the previous version where multi-echo splitting is done both in longitudinal and cross-sectional cases, and echoes cleaned up in the end. It finishes without error. The option for extra cleanup of split echoes is removed.

mmilch01: > > No, fMRIVolume only gets run on timepoints, although I guess that doesn't affect your point.
> 
> If it is only run on longitudinal timepoints, why does it get run more than once on the same input in the longitudinal context? Does it first resample with the cross-sectional transform (such that it isn't fully aligned with the longitudinal structural space), then come back and resample with the longitudinal transform? The longitudinal alignment is calculated before the fMRI data is touched, right?
> 
> Are we planning to always maintain separate "visit-specific longitudinal-ignorant" and "longitudinal timepoint" folders (and volume spaces) for the visits, both with complete copies of the processed fMRI data?

1. Longitudinal processing runs registration to structural, resampling and the following final steps. All previous output from cross-sectional run is reused. 
2. This is a good point, I think more fine-tuning could save space when accounting for identical files in both types of sessions. Symlinking all files from cross to long, however, doesn't work as some get overwritten in longitudinal mode.

mharms: Hopefully we aren't saving copies of the cross-sectional fMRI time series (volume or CIFTI) in the longitudinal folders...?  If we are, that "fine-tuning" should probably happen now.

mmilch01: > Hopefully we aren't saving copies of the cross-sectional fMRI time series (volume or CIFTI) in the longitudinal folders...? If we are, that "fine-tuning" should probably happen now.

I try to avoid that. I'll double-check. I can see that the the non-longitudinal code does create copies of the original fMRI time series from unprocessed folder, at least at some places - for example, line 689 of GenericfMRIVolumeProcessingPipeline.sh

mharms:
Yes, there are a couple files in the cross-sectional fMRIVolume code that could potentially be deleted.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -660,34 +661,51 @@ ResultsFolderCross="$AtlasSpaceFolder"/"$ResultsFolder"/"$NameOffMRI"
 ResultsFolderLong="$AtlasSpaceFolderLong"/"$ResultsFolder"/"$NameOffMRI"
 ResultsFolder=$ResultsFolderCross
 
-if (( ! IsLongitudinal )); then 
+function cp_link
+{
+    local src="$1" dst="$2"
+    if [ -L "$src" ]; then
+        cp -a $src $dst
```
coalsont, line 668: Do we know if this works correctly when `src` is a relative symlink (since `dst` is at a different level of folder)?  In what context is `src` expected to already be a link?

coalsont: To clarify, `-a` implies `--preserve=all`, which includes `--preserve=links`, and I don't know if cp preserves the link target string as-is (which would break relative links if the destination is either at a different folder depth, or if the relative link doesn't `..` up to a directory that is shared by `src` and `dst`), or tries to intelligently adjust the link target based on the paths of the arguments to result in a link that still functions.

coalsont: There aren't many pipelines that make links at the moment.  When they do, they don't test whether the file it will point to is already a link (they don't intentionally run `cp -a` on a link, and I'm not sure what its behavior is).  Linux allows 40 total symlink traversals while resolving a path before giving up, so you can just make the link without a special test for "file is a link" and it should work for nearly all users.

mharms: Even though we are using a link, are we still deleting it once we are done with it?

mmilch01: > Even though we are using a link, are we still deleting it once we are done with it?
As you wish. The original code has this line commented out, I uncommented it.

mmilch01: Ok. This line is commented out in the original pre-longitudinal code. I uncommented it.

mmilch01: > Even though we are using a link, are we still deleting it once we are done with it?
Ok. This isn't done in the original code for cross-sectional, but we're doing it now in longitudinal mode.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -660,34 +661,42 @@ ResultsFolderCross="$AtlasSpaceFolder"/"$ResultsFolder"/"$NameOffMRI"
...
+    for fd in "$fMRIFolder"/*; do
+        fname="$(basename "$fd")"
+        #create link to the original fMRI series
+        if [ "$fname" == "${NameOffMRI}_orig.nii.gz" ]; then
+            ln -sf ../../"$Session"/"${NameOffMRI}"/"$fname" "$fMRIFolderLong/$fname"
+        #skip large files that will be generated
+        elif [ "$fname" == "${NameOffMRI}_orig_nonlin.nii.gz" -o "$fname" == "${NameOffMRI}_nonlin.nii.gz" ]; then 
+        	continue
+        else
+            cp -r --preserve=timestamps "$fd" "$fMRIFolderLong/"
```
coalsont, line 686: Unfortunately, mac's `cp` doesn't have a `--preserve` option.  The only `cp --preserve` we currently have in the pipelines is in scripts used for the old freesurfer 5.3.0 pipeline (and we never tried to run that on mac, apparently).  We usually try to make things plausibly compatible with mac.  Timestamps may not be particularly important to keep.

mharms: What's the problem with the original `-p`?

coalsont: I would guess preserving ownership is the issue, the user running the processing may not have the permissions to do that.  I don't know offhand whether cp exits with error if it can't preserve ownership.

mharms: No. If `cp` can't preserve ownership or group status "as allowed by permissions" it assigns something sensible instead.  It doesn't exit with error.

mmilch01: > No. If `cp` can't preserve ownership or group status "as allowed by permissions" it assigns something sensible instead. It doesn't exit with error.

This may be true elsewhere, but on my dev machine - brainmappers2 with Ubuntu 22.04 cp -p exits with error, breaking the pipeline.

mmilch01: Please feel free to update cp with whatever would work best for the HCP users.  The latest code is tested to work today so please merge if possible.

PR #334, opened 2025-04-11
MSMAll longitudinal

mmilch01:
Longitudinal adaptation of the MSMAll pipeline and longitudinal batch file for the DeDriftAndResample pipeline.

### DIFF
--- a/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
+++ b/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
@@ -62,7 +62,7 @@ get_batch_options "$@"
 
 #Location of Subject folders (named by subjectID)
 StudyFolder="${HOME}/data/Pipelines_ExampleData"
-EnvironmentScript="${StudyFolder}/scripts/SetUpHCPPipeline-long-msmall.sh" #Pipeline environment script
+EnvironmentScript="${StudyFolder}/scripts/SetUpHCPPipeline.sh" #Pipeline environment script
 
 #list longitudinal template IDs, one per subject
 Templates=(HCA6002236_V1_V2_V3 HCA6002237_V1_V2_V3)
--- a/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
+++ b/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
@@ -1,0 +1,220 @@
+#!/bin/bash 
+
+get_batch_options() {
+    local arguments=("$@")
+
+    command_line_specified_study_folder=""
+    command_line_specified_subj=""
+    command_line_specified_run_local="FALSE"
+
+    local index=0
+    local numArgs=${#arguments[@]}
+    local argument
+
+    while [ ${index} -lt ${numArgs} ]; do
+        argument=${arguments[index]}
+
+        case "$argument" in
+            --StudyFolder=*)
+                command_line_specified_study_folder=${argument#*=}
+                index=$(( index + 1 ))
+                ;;
+            --Subjlist=*)
+                command_line_specified_subj=${argument#*=}
+                index=$(( index + 1 ))
+                ;;
+            --runlocal)
+                command_line_specified_run_local="TRUE"
+                index=$(( index + 1 ))
+                ;;
+	    *)
+		echo ""
+		echo "ERROR: Unrecognized Option: ${argument}"
+		echo ""
+		exit 1
+		;;
+        esac
+    done
+}
+function identify_timepoints
+{
+    local subject=$1
+    local tplist=""
+    local tp visit n
+
+    #build the list of timepoints
+    n=0
+    for visit in ${PossibleVisits[*]}; do
+        tp="${subject}_${visit}"
+        if [ -d "$StudyFolder/$tp" ] && ! [[ " ${ExcludeVisits[*]+${ExcludeVisits[*]}} " =~ [[:space:]]"$tp"[[:space:]] ]]; then
+             if (( n==0 )); then 
+                    tplist="$tp"
+             else
+                    tplist="$tplist@$tp"
+             fi
+        fi
+        ((n++))
+    done
+    echo $tplist
+}
+
+get_batch_options "$@"
+
+#Location of Subject folders (named by subjectID)
+StudyFolder="${HOME}/data/Pipelines_ExampleData"
+EnvironmentScript="${StudyFolder}/scripts/SetUpHCPPipeline-long-msmall.sh" #Pipeline environment script
+
+#list longitudinal template IDs, one per subject
+Templates=(HCA6002236_V1_V2_V3 HCA6002237_V1_V2_V3)
+
+#list visit/timepoint ID, cross-sectional processsing timepoint folder name under $StudyFolder is expected to follow the pattern: <Subject>_<Visit_ID>
+PossibleVisits=(V1_MR V2_MR V3_MR)
+Subjlist=(HCA6002236 HCA6002237) #List of subject IDs
+
+#Set up pipeline environment variables and software
+source "$EnvironmentScript"
+
+# Requirements for this script
+#  installed versions of: FSL, Connectome Workbench (wb_command)
+#  environment: HCPPIPEDIR, FSLDIR, CARET7DIR 
+
+# NOTE: this script will error on subjects that are missing some fMRI runs that are specified in the MR FIX arguments
+
+########################################## INPUTS ########################################## 
+
+#This example script is set up for a single subject from the CCF development project
+
+######################################### DO WORK ##########################################
+
+#Example of how CCF Development was run
+
+#Don't edit things between here and MRFixConcatNames unless you know what you are doing
+HighResMesh="164"
+LowResMesh="32"
+#Do not use RegName from MSMAllPipelineBatch.sh
+RegName="MSMAll_InitalReg_2_d40_WRN"
+DeDriftRegFiles="${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.L.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii@${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.R.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii"
+ConcatRegName="MSMAll"
+#standard maps to resample
+Maps=(sulc curvature corrThickness thickness)
+MyelinMaps=(MyelinMap SmoothedMyelinMap) #No _BC, this will be reapplied
+#MRFixConcatNames and MRFixNames must exactly match the way MR FIX was run on the subjects
+MRFixConcatNames="fMRI_CONCAT_ALL"
+#SPECIAL: if your data used two (or more) MR FIX runs (which is generally not recommended), specify them like this, with no whitespace before or after the %:
+#MRFixConcatNames=(concat12 concat34)
+#MRFixNames=(run1 run2%run3 run4)
+MRFixNames="rfMRI_REST1_AP@rfMRI_REST1_PA@tfMRI_VISMOTOR_PA@tfMRI_CARIT_PA@tfMRI_FACENAME_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"
+#fixNames are for if single-run ICA FIX was used (not recommended)
+fixNames=()
+#dontFixNames are for runs that didn't have any kind of ICA artifact removal run on them (very not recommended)
+dontFixNames=()
+SmoothingFWHM="2" #Should equal previous grayordinates smoothing (because we are resampling from unsmoothed native mesh timeseries)
+HighPass="0"
+MotionRegression=FALSE
+MatlabMode="1" #Mode=0 compiled Matlab, Mode=1 interpreted Matlab, Mode=2 octave
+
+#Example of how older HCP-YA results were originally run
+#These settings are no longer recommended - recommendations are to do MR FIX using all of a subject's runs, in the order they were acquired, no motion regression, HighPass 0
+#StudyFolder="/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final" #Location of Subject folders (named by subjectID)
+#Subjlist=(100307 101006) #List of subject IDs
+#HighResMesh="164"
+#LowResMesh="32"
+#RegName="MSMAll_InitalReg_2_d40_WRN"
+#DeDriftRegFiles="${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.L.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii@${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.R.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii"
+#ConcatRegName="MSMAll"
+#Maps=(sulc curvature corrThickness thickness)
+#MyelinMaps=(MyelinMap SmoothedMyelinMap) #No _BC, this will be reapplied
+#MRFixConcatNames=()
+#MRFixNames=()
+#fixNames=(rfMRI_REST1_LR rfMRI_REST1_RL rfMRI_REST2_LR rfMRI_REST2_RL)
+#dontFixNames=(tfMRI_EMOTION_LR tfMRI_EMOTION_RL tfMRI_GAMBLING_LR tfMRI_GAMBLING_RL tfMRI_LANGUAGE_LR tfMRI_LANGUAGE_RL tfMRI_MOTOR_LR tfMRI_MOTOR_RL tfMRI_RELATIONAL_LR tfMRI_RELATIONAL_RL tfMRI_SOCIAL_LR tfMRI_SOCIAL_RL tfMRI_WM_LR tfMRI_WM_RL)
+#SmoothingFWHM="2" #Should equal previous grayordinates smoothing (because we are resampling from unsmoothed native mesh timeseries)
+#HighPass="2000"
+#MotionRegression=TRUE
+#MatlabMode="1" #Mode=0 compiled Matlab, Mode=1 interpreted Matlab, Mode=2 octave
+
+MSMAllTemplates="${HCPPIPEDIR}/global/templates/MSMAll"
+MyelinTargetFile="${MSMAllTemplates}/Q1-Q6_RelatedParcellation210.MyelinMap_BC_MSMAll_2_d41_WRN_DeDrift.32k_fs_LR.dscalar.nii"
+
+if [ -n "${command_line_specified_study_folder}" ]; then
+    StudyFolder="${command_line_specified_study_folder}"
+fi
+
+if [ -n "${command_line_specified_subj}" ]; then
+    Subjlist="${command_line_specified_subj}"
+fi
+
+# Log the originating call
+echo "$0" "$@"
+
+#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
+#DO NOT include "-q " at the beginning
+#default to no queue, implying run local
+#QUEUE=""
+QUEUE="short.q"
+
+Maps=$(IFS=@; echo "${Maps[*]}")
+MyelinMaps=$(IFS=@; echo "${MyelinMaps[*]}")
+MRFixConcatNames=$(IFS=@; echo "${MRFixConcatNames[*]}")
+MRFixNames=$(IFS=@; echo "${MRFixNames[*]}")
+fixNames=$(IFS=@; echo "${fixNames[*]}")
+dontFixNames=$(IFS=@; echo "${dontFixNames[*]}")
+
+for Subject in "${Subjlist[@]}" ; do
+    echo "    ${Subject}"
+    TemplateLong="${Templates[i]}"
+    Timepoint_list_cross_at_separated=$(identify_timepoints "$Subject")
+    IFS=@ read -r -a Timepoint_list_cross <<< "${Timepoint_list_cross_at_separated}"    
+    if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+        echo "About to locally run ${HCPPIPEDIR}/DeDriftAndResample/DeDriftAndResamplePipeline.sh"
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/DeDriftAndResample/DeDriftAndResamplePipeline.sh"
+        queuing_command=("$FSLDIR"/bin/fsl_sub -q "$QUEUE")
+    fi
+    
+    #process timepoints
+    for TimepointCross in "${Timepoint_list_cross[@]}"; do
+        TimepointLong=${TimepointCross}.long.${TemplateLong}
+        "${queuing_command[@]}" "$HCPPIPEDIR"/DeDriftAndResample/DeDriftAndResamplePipeline.sh \
+            --path="$StudyFolder" \
+            --subject="$TimepointLong" \
+            --high-res-mesh="$HighResMesh" \
+            --low-res-meshes="$LowResMesh" \
+            --registration-name="$RegName" \
+            --dedrift-reg-files="$DeDriftRegFiles" \
+            --concat-reg-name="$ConcatRegName" \
+            --maps="$Maps" \
+            --myelin-maps="$MyelinMaps" \
+            --multirun-fix-concat-names="$MRFixConcatNames" \
+            --multirun-fix-names="$MRFixNames" \
+            --fix-names="$fixNames" \
+            --dont-fix-names="$dontFixNames" \
+            --smoothing-fwhm="$SmoothingFWHM" \
+            --high-pass="$HighPass" \
+            --matlab-run-mode="$MatlabMode" \
+            --motion-regression="$MotionRegression" \
+            --myelin-target-file="$MyelinTargetFile"
+    done
+    #process the template
+    "${queuing_command[@]}" "$HCPPIPEDIR"/DeDriftAndResample/DeDriftAndResamplePipeline.sh \
+        --path="$StudyFolder" \
+        --subject="$Subject.long.$TemplateLong" \
+        --high-res-mesh="$HighResMesh" \
+        --low-res-meshes="$LowResMesh" \
+        --registration-name="$RegName" \
+        --dedrift-reg-files="$DeDriftRegFiles" \
+        --concat-reg-name="$ConcatRegName" \
+        --maps="$Maps" \
+        --myelin-maps="$MyelinMaps" \
+        --fix-names="" \
+        --dont-fix-names="$dontFixNames" \
+        --smoothing-fwhm="$SmoothingFWHM" \
+        --high-pass="$HighPass" \
+        --matlab-run-mode="$MatlabMode" \
+        --motion-regression="$MotionRegression" \
+        --myelin-target-file="$MyelinTargetFile"
+    
+    
+done
+
--- a/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
+++ b/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
@@ -92,7 +92,7 @@ source "$EnvironmentScript"
 HighResMesh="164"
 LowResMesh="32"
 #Do not use RegName from MSMAllPipelineBatch.sh
-RegName="MSMAll_InitalReg_2_d40_WRN"
+RegName="MSMAll_InitialReg_2_d40_WRN"
 DeDriftRegFiles="${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.L.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii@${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.R.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii"
 ConcatRegName="MSMAll"
 #standard maps to resample
@@ -119,7 +119,7 @@ MatlabMode="1" #Mode=0 compiled Matlab, Mode=1 interpreted Matlab, Mode=2 octave
 #Subjlist=(100307 101006) #List of subject IDs
 #HighResMesh="164"
 #LowResMesh="32"
-#RegName="MSMAll_InitalReg_2_d40_WRN"
+#RegName="MSMAll_InitialReg_2_d40_WRN"
 #DeDriftRegFiles="${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.L.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii@${HCPPIPEDIR}/global/templates/MSMAll/DeDriftingGroup.R.sphere.DeDriftMSMAll.164k_fs_LR.surf.gii"
 #ConcatRegName="MSMAll"
 #Maps=(sulc curvature corrThickness thickness)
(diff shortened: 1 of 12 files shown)


MSMAll/MSMAllPipeline.sh
```diff
@@ -216,31 +327,43 @@ log_Msg "fMRIProcSTRING: ${fMRIProcSTRING}"
...
+
+#copy the registration result sphere from template back to timepoints.    
+if (( IsLongitudinal )); then
+    
+    for tp in ${SessionsLong[@]}; do
+        cp $StudyFolder/$TemplateSession/MNINonLinear/Native/$TemplateSession.SphericalDistortion.native.dscalar.nii \
+            $StudyFolder/$SessionLong/MNINonLinear/Native/$SessionLong.SphericalDistortion.native.dscalar.nii
+        for Hemisphere in L R; do
+            SessionLong=$tp.long.$TemplateLong
+            cp $StudyFolder/$TemplateSession/MNINonLinear/Native/$TemplateSession.$Hemisphere.sphere.${OutputRegName}_2_d40_WRN.native.surf.gii $StudyFolder/$SessionLong/MNINonLinear/Native/$SessionLong.$Hemisphere.sphere.${OutputRegName}_2_d40_WRN.native.surf.gii
```
glasserm, line 365: ${OutputRegName}_2_d40_WRN: the _2_d40_WRN of this should not be hard coded.

MSMAll/MSMAllPipeline.sh
```diff
@@ -216,31 +327,43 @@ log_Msg "fMRIProcSTRING: ${fMRIProcSTRING}"
...
+    --msm-all-templates="${MSMAllTemplates}" \
+    --use-ind-mean="${UseIndMean}" \
+    --matlab-run-mode="${MatlabRunMode}"
+
+#copy the registration result sphere from template back to timepoints.    
+if (( IsLongitudinal )); then
+    
+    for tp in ${SessionsLong[@]}; do
+        cp $StudyFolder/$TemplateSession/MNINonLinear/Native/$TemplateSession.SphericalDistortion.native.dscalar.nii \
+            $StudyFolder/$SessionLong/MNINonLinear/Native/$SessionLong.SphericalDistortion.native.dscalar.nii
```
glasserm, line 362: Why is this file being copied?  Is the spherical inflation distortion truly the same across timepoints?  It is possible that it isn't.

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,177 @@ log_File_Must_Exist "${TopographyTarget}"
...
 log_Msg "output_proc_string: ${output_proc_string}"
 
-expected_concatenated_output_dir="${StudyFolder}/${Subject}/MNINonLinear/Results/${OutputfMRIName}"
-expected_concatenated_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}${output_proc_string}.dtseries.nii"
-before_vn_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}_novn.dtseries.nii"
+IsLongitudinal=$(opts_StringToBool "$IsLongitudinal")
+
+if (( IsLongitudinal ));  then 
+    
+    if [ -z "$SubjectLong" -o -z "$Sessions" -o -z "TemplateLong" -o -z "${mrfixNamesToUse}" -o -z "$mrfixConcatName" ]; then
```
glasserm, line 154:
```suggestion
    if [[ -z "$SubjectLong" || -z "$Sessions" || -z "TemplateLong" || -z "${mrfixNamesToUse}" || -z "$mrfixConcatName" ]]; then 
```
Same comment as the other PR.

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,177 @@ log_File_Must_Exist "${TopographyTarget}"
...
+IsLongitudinal=$(opts_StringToBool "$IsLongitudinal")
+
+if (( IsLongitudinal ));  then 
+    
+    if [ -z "$SubjectLong" -o -z "$Sessions" -o -z "TemplateLong" -o -z "${mrfixNamesToUse}" -o -z "$mrfixConcatName" ]; then 
+        log_Err_Abort "--subject-long, --sessions-long, --template-long are mandatory in longitudinal mode"
+    fi            
+    if [ "$OutputRegName" != "MSMAll_InitalReg" ]; then 
+        log_Err_Abort "Currently, only MSM_All_InitialReg is supported as output reg name in longitudinal mode"
+    fi
```
glasserm, line 159: We need to support arbitrary RegNames.

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,177 @@ log_File_Must_Exist "${TopographyTarget}"
...
-before_vn_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}_novn.dtseries.nii"
+IsLongitudinal=$(opts_StringToBool "$IsLongitudinal")
+
+if (( IsLongitudinal ));  then 
+    
+    if [ -z "$SubjectLong" -o -z "$Sessions" -o -z "TemplateLong" -o -z "${mrfixNamesToUse}" -o -z "$mrfixConcatName" ]; then 
+        log_Err_Abort "--subject-long, --sessions-long, --template-long are mandatory in longitudinal mode"
+    fi            
+    if [ "$OutputRegName" != "MSMAll_InitalReg" ]; then 
+        log_Err_Abort "Currently, only MSM_All_InitialReg is supported as output reg name in longitudinal mode"
```
glasserm, line 158:
```suggestion
        log_Err_Abort "Currently, only MSMAll_InitialReg is supported as output reg name in longitudinal mode"
```

glasserm (changes requested):
Please see above.

MSMAll/MSMAllPipeline.sh
mmilch01: Ok. As long as the Reg name matches cross-sectional, removing this check should be ok.

MSMAll/MSMAllPipeline.sh
mmilch01: Ok

MSMAll/MSMAllPipeline.sh
glasserm: I guess I am a bit confused here.  Where is this map normally generated from in cross-sectional?  I am not sure that the spherical inflation is exactly the same across the longitudinal timepoints (which is what would be implied by copying this file).

MSMAll/MSMAllPipeline.sh
glasserm: Yes it will match.

MSMAll/MSMAllPipeline.sh
mmilch01: Ok. I added an option --de-drift-reg-name to put the reg sphere name suffix to.

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,174 @@ log_File_Must_Exist "${TopographyTarget}"
...
+    # MNINonLinear/Results/. Multiple configuration files may be used for the same 
+    # subject with different fMRI combinations or longitudinal templates.
+    conf_file="$ResultsTemplateDir/$OutConfig"
+    echo "${TemplateRunsStr/#@/}" > "$conf_file"
+    echo "${TimepointsStr/#@/}" >> "$conf_file"
+    echo "${fMRIRunsStr/#@/}" >> "$conf_file"
+    echo "${ConcatNamesStr/#@/}" >> "$conf_file"
+
+    # average myelin maps from all timepoints.
+    ${average_cmd}
```
coalsont, line 225: Whitespace safety in command building:

```suggestion
    average_cmd=("${CARET7DIR}/wb_command" -cifti-average \
        "$StudyFolder/$TemplateSession/MNINonLinear/Native/$TemplateSession.$NativeMyelinMap")
        
    for tp in "${SessionsLong[@]}"; do
        SessionLong=$tp.long.$TemplateLong                      #longitudinal session directory name
        average_cmd+=(-cifti "$StudyFolder/$SessionLong/MNINonLinear/Native/$SessionLong.$NativeMyelinMap")
            
        echo "searching $SessionLong for eligible fMRI runs"
        if [ ! -d "$StudyFolder/$SessionLong/MNINonLinear/Results" ]; then 
            log_Err_Abort "ICAFix output does not exist for longitudinal session $SessionLong in $StudyFolder"
        fi    
        ResultsTPLongDir=$StudyFolder/$SessionLong/MNINonLinear/Results
        # iterate over possible fMRI runs and build a list of found runs for this timepoint.
        # Found runs are copied to template directory and all relevant files/dirs are renamed 
        # for the run to be unique within subject.
        # template-based fMRI naming pattern, assuming original run <fMRIRun> label is: <Session>_<fMRIRun>, 
        # where <Session> is longitudinal session label. It is expected to be named <Subject>_<Visit_ID>.
        for fmriName in "${PossibleRuns[@]}"; do
            if [ -d "$ResultsTPLongDir/$fmriName" ]; then            
                TemplateRun=${tp}_${fmriName}
                echo "found $TemplateRun, copying"
                mkdir -p $ResultsTemplateDir/$TemplateRun
                #bulk copy
                cp -r "$ResultsTPLongDir/$fmriName"/* "$ResultsTemplateDir/$TemplateRun/"
                pushd "$ResultsTemplateDir/${TemplateRun}" &> /dev/null                
                for fd in ${fmriName}_* ${fmriName}.*; do
                    [ -e "$fd" ] || continue
                    mv "$fd" "${fd//$fmriName/$TemplateRun}"
                done
                popd &> /dev/null            
                TemplateRunsStr="${TemplateRunsStr}@$TemplateRun"
                TimepointsStr="${TimepointsStr}@$tp"
                fMRIRunsStr="${fMRIRunsStr}@$fmriName"
                ConcatNamesStr="${ConcatNamesStr}@$mrfixConcatName"
            fi
        done
    done
    # Arrays with template fMRI names, matching timepoint labels, original run labels, 
    # and per-timepoint concatenated fMRI names are stored in a configuration file under 
    # MNINonLinear/Results/. Multiple configuration files may be used for the same 
    # subject with different fMRI combinations or longitudinal templates.
    conf_file="$ResultsTemplateDir/$OutConfig"
    echo "${TemplateRunsStr/#@/}" > "$conf_file"
    echo "${TimepointsStr/#@/}" >> "$conf_file"
    echo "${fMRIRunsStr/#@/}" >> "$conf_file"
    echo "${ConcatNamesStr/#@/}" >> "$conf_file"

    # average myelin maps from all timepoints.
    "${average_cmd[@]}"
```

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,174 @@ log_File_Must_Exist "${TopographyTarget}"
...
+            if [ -d "$ResultsTPLongDir/$fmriName" ]; then            
+                TemplateRun=${tp}_${fmriName}
+                echo "found $TemplateRun, copying"
+                mkdir -p $ResultsTemplateDir/$TemplateRun
+                #bulk copy
+                cp -r "$ResultsTPLongDir/$fmriName"/* "$ResultsTemplateDir/$TemplateRun/"
+                pushd "$ResultsTemplateDir/${TemplateRun}" &> /dev/null                
+                for fd in ${fmriName}_* ${fmriName}.*; do
+                    [ -e "$fd" ] || continue
+                    mv "$fd" "${fd//$fmriName/$TemplateRun}"
```
coalsont, line 204: I prefer explicit `if` rather than short-circuit conditionals, as I have to think about (okay, or specifically short-circuits when...the first is true...so that makes this..."if false, then continue"...
```suggestion
                    if [[ -e "$fd" ]]
                    then
                        mv "$fd" "${fd//$fmriName/$TemplateRun}"
                    fi
```

Also, you are looping over glob expressions, the only way this won't exist is if there are no matches for one of them.  Setting `shopt -s nullglob` would mean that you don't need to test for this possibility, because it instead expands to zero strings, instead of the unmodified glob string.

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,174 @@ log_File_Must_Exist "${TopographyTarget}"
...
+    done
+    # Arrays with template fMRI names, matching timepoint labels, original run labels, 
+    # and per-timepoint concatenated fMRI names are stored in a configuration file under 
+    # MNINonLinear/Results/. Multiple configuration files may be used for the same 
+    # subject with different fMRI combinations or longitudinal templates.
+    conf_file="$ResultsTemplateDir/$OutConfig"
+    echo "${TemplateRunsStr/#@/}" > "$conf_file"
+    echo "${TimepointsStr/#@/}" >> "$conf_file"
+    echo "${fMRIRunsStr/#@/}" >> "$conf_file"
+    echo "${ConcatNamesStr/#@/}" >> "$conf_file"
```
coalsont, line 222: I had to look up this syntax to figure out it was "replace at beginning".  We already use the simpler "remove prefix" syntax elsewhere, so I think it is more recognizable.  If it were me, I would store them as arrays and use a subshell with `IFS` to make the string without an extra `@`...
```suggestion
    echo "${TemplateRunsStr#@}" > "$conf_file"
    echo "${TimepointsStr#@}" >> "$conf_file"
    echo "${fMRIRunsStr#@}" >> "$conf_file"
    echo "${ConcatNamesStr#@}" >> "$conf_file"
```
A demo of the `IFS` trick (note, the semicolon and `*` instead of `@` are both important - the subshell stops `IFS` from affecting other commands in the script):
```
foo=(a b c)
(IFS='+'; echo "${foo[*]}")
```

MSMAll/scripts/MSMAll.sh
```diff
@@ -19,10 +19,10 @@ opts_SetScriptDescription "implements MSM-All Registration main script"
 #mandatory
 #general inputs
 opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects" '--path'
-opts_AddMandatory '--subject' 'Subject' '100206' "one subject ID"
+opts_AddMandatory '--session' 'Session' '100206' "one session ID" "--subject"
```
coalsont, line 22: I don't see any changes to the logic in this script, only a ton of variable replacements.  If this is the case, it would have been simpler to leave the variable name alone:
```suggestion
opts_AddMandatory '--session' 'Subject' '100206' "one session ID" "--subject"
<no changes to the rest of the script>
```
If there are changes in logic, please point them out.

MSMAll/scripts/MSMAll.sh
```diff
@@ -54,6 +54,12 @@ opts_AddOptional '--matlab-run-mode' 'MatlabRunMode' '0, 1, or 2' "defaults to $
 0 = compiled MATLAB
 1 = interpreted MATLAB
 2 = Octave" "$g_matlab_default_mode"
+opts_AddOptional '--is-longitudinal' 'IsLongitudinal' 'TRUE or FALSE' 'Longitudinal mode' "FALSE"
+opts_AddOptional '--longitudinal-sessions' 'SessionsLong' '@ separated list' 'List of @ separated longitudinal sessions' ""
+opts_AddOptional '--longitudinal-template' 'TemplateLong' 'Template label' 'Longitudinal template ID' ""
+opts_AddOptional '--longitudinal-subject' 'SubjectLong' 'Subject ID' 'Subject ID for longtitudinal sessions' ""
```
coalsont, line 60: None of these appear to get used.

MSMAll/MSMAllPipeline.sh
```diff
@@ -94,6 +95,14 @@ opts_AddOptional '--matlab-run-mode' 'MatlabRunMode' '0, 1, or 2' "defaults to $
 0 = compiled MATLAB
 1 = interpreted MATLAB
 2 = Octave" "$g_matlab_default_mode"
+
+#longitudinal options
+opts_AddOptional '--is-longitudinal' 'IsLongitudinal' 'TRUE or FALSE' "longitudinal mode" "FALSE"
+opts_AddOptional '--subject-long' 'SubjectLong' 'id' "subject ID in longitudinal mode" ""
+opts_AddOptional '--sessions-long' 'Sessions' 'list' "@ separated list of longitudinal timepoints" ""
+opts_AddOptional '--template-long' 'TemplateLong' 'template_id' "longitudinal template ID" ""
+opts_AddOptional '--fmri-out-config-file' 'OutConfig' 'file name' 'Output file with detected fMRI run configuration [fmri_list.txt]' "fmri_list.txt"
```
coalsont, line 104: Maybe clarify that the user should provide only a bare filename with no path here (since it writes the file relative to MNINonLinear/Results).

Other places in the pipelines, we detect and allow absolute paths even when the default behavior is to assume the path is relative to a specific folder.

MSMAll/scripts/SingleSubjectConcat.sh
```diff
@@ -14,18 +14,23 @@ source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"         # Debugging functio
...
+opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to give to concatenated single session scan"
 opts_AddMandatory '--fmri-proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing already done, e.g. '_Atlas_hp0_clean'"
 opts_AddMandatory '--output-proc-string' 'OutputProcSTRING' 'string' "the output file name component, e.g. '_vn'"
 #optional inputs
 opts_AddOptional '--start-frame' 'StartFrame' 'integer' "the starting frame to choose from each fMRI run (inclusive), defaults to '1'" '1'
 opts_AddOptional '--end-frame' 'EndFrame' 'integer' "the ending frame to choose from each fMRI run (inclusive), defaults to '' which preserves the ending frame of every fMRI run from --fmri-names-list" ''
+opts_AddOptional '--is-longitudinal' 'IsLongitudinal' 'TRUE or FALSE' "longitudinal mode [FALSE]" "FALSE"
+opts_AddOptional '--subject-long' 'SubjectLong' 'id' "subject ID in longitudinal mode" ""
+opts_AddOptional '--fmri-config-long' 'fMRIConfigLong' 'file_name' "longitudinal runs configuration file [fmri_list.txt]" "fmri_list.txt"
```
coalsont, line 31: Similarly, clarify that the path should not be absolute, or consider whether to add detection and support of absolute paths.

MSMAll/MSMAllPipeline.sh
coalsont: Measuring the difference between a timepoint's midthickness and the single longitudinal native sphere can/should produce different results than the longitudinal template midthickness.  On the other hand, since the surface wasn't re-inflated to a sphere for longitudinal surfaces, it can be expected to contain some anatomical changes on top of the actual distortion seen by the spherical inflation algorithm.

glasserm: I don't think we know, but we can ask.

Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
demsarjure, line 1: @mmilch01 There are no "special" longitudinal flags for DeDriftAndResample, this is pretty much the same as before? Just checking whether I need to change the DeDriftAndResample interface along with the MSMAll one as well.

The pipeline here is that you run the longitudinal MSMAll, followed by "classic" DeDriftAndResample right?

MSMAll/scripts/MSMAll.sh
demsarjure: Yes, but the "new" variable naming conventions (session vs subject) are aligned with all other scripts. I think it is as good as a time as any to do this refactoring here as well.

demsarjure (commented):
@mmilch01 I added some minor comments as well. Please let me know once this is ready for testing. Thanks!

MSMAll/MSMAllPipeline.sh
mmilch01: Ok.

MSMAll/scripts/SingleSubjectConcat.sh
mmilch01: Ok

Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
mmilch01: It is true so far. There's a possibility however that we might need to add longitudinal options, depending on how this thread resolves:
https://github.com/Washington-University/HCPpipelines/pull/334#discussion_r2040738772

MSMAll/scripts/MSMAll.sh
mmilch01: I see logic in both approaches but since I already did the renaming, will keep it for now

MSMAll/MSMAllPipeline.sh
```diff
@@ -103,7 +103,7 @@ opts_AddOptional '--sessions-long' 'Sessions' 'list' "@ separated list of longit
 opts_AddOptional '--template-long' 'TemplateLong' 'template_id' "longitudinal template ID" ""
 opts_AddOptional '--fmri-out-config-file' 'OutConfig' 'file name' 'Output file with detected fMRI run configuration [fmri_list.txt].
 Specify file name only, without path. The file will be stored under [Session]/MNINonLinear/Results folder.' "fmri_list.txt"
-opts_AddOptional '--dedrift-reg-name' 'DeDriftRegName' 'MSMAll_2_d40_WRN' 'Name part for [Session].[L|R].sphere.[DeDriftRegName].native.surf.gii to be copied to timepoints, to be used in DeDriftAndResample' "MSMAll_2_d40_WRN"
+opts_AddOptional '--dedrift-reg-name' 'DeDriftRegName' 'MSMAll_2_d40_WRN' 'Name part for [Session].[L|R].sphere.[DeDriftRegName].native.surf.gii to be copied to timepoints, to be used in DeDriftAndResample' "MSMAll_InitalReg_2_d40_WRN"
```
glasserm, line 106: This variable should be able to be inferred from existing arguments in this code: _${RegNameStem}_${NumIterations}_d${ICAdim}_${Method} from MSMAll.sh without needing to specify another argument.

MSMAll/MSMAllPipeline.sh
```diff
@@ -230,7 +230,7 @@ if (( IsLongitudinal ));  then
     # variance normalize and concatenate individual runs
     # of all timepoints in template folder. 
     # This script reads the $conf_file.
-    "${HCPPIPEDIR}"/MSMAll/scripts/SingleSubjectConcat.sh \
+    "${HCPPIPEDIR}"/MSMAll/scripts/SingleSubjectConcat.sh
```
glasserm, line 233: Is this really correct?

MSMAll/MSMAllPipeline.sh
```diff
@@ -326,7 +326,7 @@ fMRIProcSTRING+="${output_proc_string}"
 log_Msg "fMRIProcSTRING: ${fMRIProcSTRING}"
 
 # run MSMAll
-"${HCPPIPEDIR}"/MSMAll/scripts/"${ModuleName}" \
+"${HCPPIPEDIR}"/MSMAll/scripts/"${ModuleName}"
```
glasserm, line 329: Is this really correct?

MSMAll/MSMAllPipeline.sh
```diff
@@ -354,12 +354,42 @@ log_Msg "fMRIProcSTRING: ${fMRIProcSTRING}"
...
+
+        ${CARET7DIR}/wb_command -cifti-convert-to-scalar ${NativeFolderTP}/${SessionLong}.SphericalDistortion.native.dtseries.nii ROW ${NativeFolderTP}/${SessionLong}.SphericalDistortion.native.dscalar.nii
+        ${CARET7DIR}/wb_command -set-map-name ${NativeFolderTP}/${SessionLong}.SphericalDistortion.native.dscalar.nii 1 ${SessionLong}_SphericalDistortion
+        ${CARET7DIR}/wb_command -cifti-palette ${NativeFolderTP}/${SessionLong}.SphericalDistortion.native.dscalar.nii MODE_USER_SCALE ${NativeFolderTP}/${SessionLong}.SphericalDistortion.native.dscalar.nii -pos-user 0 1 -neg-user 0 -1 -interpolate true -palette-name ROY-BIG-BL -disp-pos true -disp-neg true -disp-zero false
+        rm ${NativeFolderTP}/${SessionLong}.SphericalDistortion.native.dtseries.nii  
+
+        for Mesh in ${HighResMesh} ${LowResMesh} ; do
```
glasserm, line 385: I wonder if we are missing any other initial registration resamplings that currently happen in cross-sectional mode MSMAll at the end?  Alternatively, perhaps we should not resample in longitudinal mode and just take care of that in DeDriftAndResample?

Examples/Scripts/MSMAllPipelineBatch-long.sh
```diff
@@ -124,6 +125,8 @@ mrfixConcatName="fMRI_CONCAT_ALL"
 mrfixNamesToUse="rfMRI_REST1_AP@rfMRI_REST1_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"
 # FIX output concat name for this new MSMAll run of MR FIX
 OutfMRIName="rfMRI_REST"
+# Reg name file used by DeDriftAndResample
+DeDriftRegName="MSMAll_InitalReg_2_d40_WRN"
```
coalsont, line 129: We probably shouldn't put the `InitalReg` typo from YA into the MSMAll example batch (or at least not as the main example).

MSMAll/MSMAllPipeline.sh
mmilch01: Ok, fixed.

MSMAll/MSMAllPipeline.sh
mmilch01: Not sure what are you referring to?

MSMAll/MSMAllPipeline.sh
mmilch01: Ok, fixed \

Examples/Scripts/MSMAllPipelineBatch-long.sh
mmilch01: This typo is also present in both DeDriftAndResample long and cross batches, as well as MSMAllPipeline.sh MSMAllPipelineBatch.sh and MSMAllPipelineBatch-long.sh. I suspect fixing it at one place would trigger a storm of errors on already processed data. So it would probably be better to fix it everywhere. Or, leave it as is? @glasserm  @coalsont @mharms let me know what you think would be the least painful.

MSMAll/MSMAllPipeline.sh
mmilch01: Conceptually these two approaches are equivalent, we'll need to isolate all resampling code that needs to be run either in DeDrift.. or in MSMAllPipeline.sh under if (( IsLongitudinal )). DeDrift.. does more than just resampling so I guess it would depend on predominant workflow preference. Doing it in MSMAllPipeline has advantage that it's a clean cut for longitudinal mode, equivalent to what is now implemented cross-sectionally.

MSMAll/MSMAllPipeline.sh
coalsont: You need to remove the option flag, too.

Examples/Scripts/MSMAllPipelineBatch-long.sh
coalsont: Lifespan changed it to `Initial`.  Examples that use YA should stay `Inital`, but it would be nice to have lifespan examples as the primaries, and have all the typo examples commented out...but, this may be out of scope.  For this PR, though, it would make sense for the example to be lifespan, not YA (because YA wasn't really designed to collect longitudinal), which also means the example wouldn't need the typo.

MSMAll/MSMAllPipeline.sh
glasserm: I think we can not resample initials in the longitudinal timepoints in MSMAll and just resample finals in DeDriftAndResample.

Examples/Scripts/MSMAllPipelineBatch-long.sh
glasserm: I agree.

mharms: Also, this is just an example script.  There is no reason that it needs to contain a typo, even if that typo happens to have made it into the actual processing of the HCP-YA data.

glasserm: I agree.

MSMAll/MSMAllPipeline.sh
mmilch01: Thought I did. Done.

MSMAll/MSMAllPipeline.sh
mmilch01: > I wonder if we are missing any other initial registration resamplings that currently happen in cross-sectional mode MSMAll at the end? Alternatively, perhaps we should not resample in longitudinal mode and just take care of that in DeDriftAndResample?

Ok, to clarify, should we move all resampling from MSMAll to DeDrift.., or just do it in longitudinal case?

glasserm: Just longitudinal.

Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
mmilch01: > @mmilch01 There are no "special" longitudinal flags for DeDriftAndResample, this is pretty much the same as before? Just checking whether I need to change the DeDriftAndResample interface along with the MSMAll one as well.
> 
> The pipeline here is that you run the longitudinal MSMAll, followed by "classic" DeDriftAndResample right?

Another update: please hold with your tests for DeDriftAndResample. Based on discussion below, I'll create a separate longitudinal branch for DeDriftAndResample, there will be some modifications there for longitudinal mode.

glasserm: No modifications are needed.

MSMAll/MSMAllPipeline.sh
glasserm: All you need to do is run DeDriftAndResample after creating the unresampled spherical distortion maps.

glasserm: I think we need to discuss before you do anything.

glasserm: This was discussed over the phone and a plan will be implemented and tested.

Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
mmilch01: Yes, sorry for the confusion, the original plan to not modify DeDriftAndResample stays.

glasserm (approved):
@demsarjure can you run your tests now?

glasserm (changes requested):
Jure found bugs in Slack.

MSMAll/MSMAllPipeline.sh
```diff
@@ -138,75 +147,178 @@ log_File_Must_Exist "${TopographyTarget}"
 output_proc_string="_vn" #To VN only to indicate that we did not revert the bias field before computing VN
 log_Msg "output_proc_string: ${output_proc_string}"
 
-expected_concatenated_output_dir="${StudyFolder}/${Subject}/MNINonLinear/Results/${OutputfMRIName}"
-expected_concatenated_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}${output_proc_string}.dtseries.nii"
-before_vn_output_file="${expected_concatenated_output_dir}/${OutputfMRIName}${fMRIProcSTRING}_novn.dtseries.nii"
+IsLongitudinal=$(opts_StringToBool "$IsLongitudinal")
+NumIterations=$(echo "${Iterations}" | wc -w)
```
glasserm, line 151: `NumIterations=$(echo "${IterationModes}" | sed s/_/ /g | wc -w)`

glasserm, line 151:
```suggestion
NumIterations=$(echo "${IterationModes}" | sed s/_/ /g | wc -w)
```

glasserm (approved):

PR #337, opened 2025-04-28
pd bug fix

bqrosen:
Modified fix_3_clean.m and the wrappers that call it (hcp_fix_multi_run and ReApplyFixMultiRunPipeline.sh) take $hp as a string, and handle pd# (polynomial detrend) cases. I did not recompile the matlab, but the compiled mode sections of the wrappers do not need to be changed because in compiled mode all arguments are passed as strings anyway. @coalsont, could you please recompile the matlab, assuming all else looks good?

### DIFF
--- a/ICAFIX/ReApplyFixMultiRunPipeline.sh
+++ b/ICAFIX/ReApplyFixMultiRunPipeline.sh
@@ -727,7 +727,7 @@ case ${MatlabMode} in
 	1 | 2)
 		# Use interpreted MATLAB or Octave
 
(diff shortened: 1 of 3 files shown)


mharms:
I should probably take a look at this, since I wrote the original `pd` code long ago.  What is the actual "bug" that is needing fixing?

glasserm:
I need to discuss it with Burke as this is not what I thought needed to be done.

bqrosen:
Turned out to be simpler than I thought.

PR #338, opened 2025-04-28
Never run hp in fix_3_clean for multirun fix

bqrosen:
Replace my previous PR with this one. Somehow, when I made fix_3_clean stand alone, I fed it $hp rather than $AlreadyHP. Was not detected because it doesn't matter for hp=0 (typical use case). Now fix_3_clean.m never uses hp.


glasserm:
Is this tested to work?

bqrosen:
When tested, it failed failed with--reuse-existing-ica=TRUE because the $AlreadyHP is set during ICA, so I just set it again in the FIX part to resolve ($AlreadyHP is a constant). Could set it once at the beginning instead, but it's more readable I think to set it close to where it is used.

glasserm:
Is it now tested to work, or is more testing needed?

bqrosen:
Tested to work.

PR #342, opened 2025-05-08
Mixture model bugfix

bqrosen:
Fix for rare bug when for at least one component no voxel is above the mmthresh (0) , so melodic re-thresholds at 0.05 and 0.01, which results in two maps for the components output thresholded volume instead of the expected  one. This fix takes the first map and issues a warning.


global/matlab/mixtureModel.m
```diff
@@ -62,24 +66,34 @@ function mixtureModel(inFile,outFile,wbcmd,melocmd)
...
 %% run gaussian mixture modeling with melodic
 [stat(3),~] = system(sprintf('mkdir -p %s;echo "1" > %s/grot', tDir, tDir),'-echo');
-[stat(4),~] = system(sprintf(...
+[stat(4),stout] = system(sprintf(...
   '%s -i %s --ICs=%s --mix=%s/grot -o %s --Oall --report -v --mmthresh=0',... 
   melocmd,inFile, inFile, tDir, tDir),'-echo');
+% [stat(5),~] = system(sprintf(...
+%     'FSLOUTPUTTYPE=%s;fslmerge -t %s $(ls %s/stats/thresh_zstat* | sort -V);FSLOUTPUTTYPE=%s;',...
+%     FSLOUTPUTTYPE, outFile, tDir ,FSLOUTPUTTYPE0),'-echo');% assumes every output has one map
+if contains(stout,'2-sided null hypothesis test at 0.05') % use fslhd <tDir>/stats/thresh_zstat* | grep dim4 | head -1 > 1 instead?
```
coalsont, line 90: It seems a bit fragile to rely on the exact wording of the message.  Consider testing an output for whether it contains more than one map.

Also, these `system()` calls all hide the stdout from the log file, is that what we want?  How much text output does melodic produce?

bqrosen:
I agree with making the warning trigger more robust. Done. 

Melodic produces a fair bit (~700 lines), but I record it anyway. The system calls use the ‘-echo’ argument, so all the calls stout is sent to the matlab console which is then captured in the .o log as called in the pipelines. 

-Burke

global/matlab/mixtureModel.m
```diff
@@ -77,9 +84,18 @@ function mixtureModel(inFile,outFile,wbcmd,melocmd)
 [stat(4),~] = system(sprintf(...
   '%s -i %s --ICs=%s --mix=%s/grot -o %s --Oall --report -v --mmthresh=0',... 
   melocmd,inFile, inFile, tDir, tDir),'-echo');
+% [stat(5),~] = system(sprintf(...
+%     'FSLOUTPUTTYPE=%s;fslmerge -t %s $(ls %s/stats/thresh_zstat* | sort -V);FSLOUTPUTTYPE=%s;',...
+%     FSLOUTPUTTYPE, outFile, tDir ,FSLOUTPUTTYPE0),'-echo');% assumes every output has one map
+[~,mapCounts] = system(sprintf(...
+  'find "%s/stats/" -name "thresh_zstat*" -exec sh -c "fslhd \\"{}\\" | grep dim4 | head -1 | tail -c +5" sh {} \\;',tDir));
```
coalsont, line 91: Is "the first output filename" really so hard to predict that you need `find`?  I also suggested `fslval` rather than grepping a `fslhd`.

global/matlab/mixtureModel.m
coalsont: Ah, I didn't notice the `, '-echo'` on the end of everything.  I would generally use the single-return form of `system()` rather than specifically using `~` to invoke the two-return form and then using `'-echo'` to revert the behavior change that causes.
```
>> foo = system('ls');
AFI  MEG  MNINonLinear	release-notes  T1w  unprocessed
>> [foo, ~] = system('ls');
>> [foo, ~] = system('ls', '-echo');
AFI  MEG  MNINonLinear	release-notes  T1w  unprocessed
```
Additionally, since this file doesn't actually do any matrix manipulation or math (or even have inputs or outputs other than filenames), it could/should be a bash script, and then we wouldn't need to recompile the matlab every time it changes...

ICAFIX/hcp_fix_multi_run
```diff
@@ -1063,18 +1063,19 @@ sed '1d;$d' ${melview} > ${melview//.txt/.wb_annsub.csv} # create workbench form
 # running stages separately doesn't create .fix file with noise component indices, so create it from the melview file
```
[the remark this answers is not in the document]
coalsont: He appears to have reused a branch rather than making a new one (or resetting it to master first), and it still has the commits that were squashed into master when the PR was merged.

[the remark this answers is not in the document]
glasserm: These sorts of things make code review harder.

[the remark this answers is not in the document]
coalsont: I can separate the new commits, but he will need to resync his copy before pushing again (I could use a new branch name to not need resyncing, but that would make this PR defunct).

global/matlab/mixtureModel.m
bqrosen: I concede that last point, but it's called by RSNregression.m, so it makes sense for it to be in matlab also.

coalsont: RSNregression could use `system()` to run the bash script, or if nothing else, the `system()` call for the bash script can be in a simple matlab wrapper function (even if it did allow a matrix variable as an input).  Bash scripting also makes the commands being run more readable than `system(sprintf(..., ...))` (and has more convenient string matching syntax).

This may not be a good time, but I think it would be good to refactor this at some point.

global/matlab/mixtureModel.m
bqrosen: melodic spits out a volume file for each component and the script doesn't currently use the icadim so it would have to figure it out from inFile and build the all the file names. find -exec in a temporary directory is simpler. I switched fslhd | grep to fslval. Didn't know that one.

global/matlab/mixtureModel.m
glasserm: I agree that I would have implemented this in bash, but hopefully other than this bugfix we won't have to deal with this code often.  It is likely not worth the effort to rewrite it now.

ICAFIX/hcp_fix_multi_run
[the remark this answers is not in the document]
bqrosen: I created a new branch and don't see the offending change in git CLI or GitHub. Does this happen because I didn't run git fetch origin before creating the branch?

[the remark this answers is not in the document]
coalsont: Fetch doesn't change anything, it just updates your local clone's view of what has been updated on the server by other people.

Ah, I see the branch name was different than the PR those commits came from.  You must have done a "git branch" while still on the old PR branch, when you should have done `git checkout master` (and then `git pull --ff-only` to grab the current state after the PR merge) to get back to the main branch before making a new one.  `master` is just a branch like any other (the only reason it feels special is because git clone checks it out by default, which is actually a result of github telling git that it is the "active branch").  `git branch` starts the new branch at whatever commit you currently have checked out (by default).

global/matlab/mixtureModel.m
```diff
@@ -77,17 +84,26 @@ function mixtureModel(inFile,outFile,wbcmd,melocmd)
...
   melocmd,inFile, inFile, tDir, tDir),'-echo');
-[stat(5),~] = system(sprintf(...
-    'FSLOUTPUTTYPE=%s;fslmerge -t %s $(ls %s/stats/thresh_zstat* | sort -V);FSLOUTPUTTYPE=%s;',...
-    FSLOUTPUTTYPE, outFile, tDir ,FSLOUTPUTTYPE0),'-echo');
-[stat(6),~] = system(['rm -r ' tDir],'-echo');% clean up temporary files
+% [stat(5),~] = system(sprintf(...
+%     'FSLOUTPUTTYPE=%s;fslmerge -t %s $(ls %s/stats/thresh_zstat* | sort -V);FSLOUTPUTTYPE=%s;',...
+%     FSLOUTPUTTYPE, outFile, tDir ,FSLOUTPUTTYPE0),'-echo');% assumes every output has one map
+[stat(5),mapCounts] = system(sprintf(...
+  "find %s/stats/ -name 'thresh_zstat*' -exec sh -c 'fslval $0 dim4' {} \\;",tDir));
```
coalsont, line 91: `sh -c`?  $0?  That is not good practice ($0 is what `ps`/`top`/`htop` show as the process name).  See if this works.

```suggestion
  "find %s/stats/ -name 'thresh_zstat*' -exec fslval {} dim4 \\;",tDir));
```

coalsont: If `find -exec` ignores `$PATH` and can't find `fslval`, we can at least avoid legacy `sh` and abusing $0 for arguments:

```suggestion
  "find %s/stats/ -name 'thresh_zstat*' -exec bash -c 'fslval ""$1"" dim4' bash {} \\;",tDir));
```

global/matlab/mixtureModel.m
bqrosen: copilot gpt4.1 refactored the matlab to bash and I tested tested the function to work in isolation. I update RSNregression.m to call the bash version but did not test it.

ICAFIX/hcp_fix_multi_run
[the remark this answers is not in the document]
bqrosen: That's exactly what happened. Got it.

global/scripts/RSNregression.m
```diff
@@ -305,7 +305,8 @@ function RSNregression(InputFile, InputVNFile, Method, ParamsFile, OutputBeta, v
         Z(isnan(Z)) = 0;
         outTemplate.cdata = Z;
         ciftisavereset(outTemplate, optional.OutputZ, wbcommand);
-        mixtureModel(optional.OutputZ,optional.OutputZMM);
+        % mixtureModel(optional.OutputZ,optional.OutputZMM);
+        stat = system(sprintf('mixtureModel %s %s %s',optional.OutputZ,optional.OutputZMM),wbcommand);
```
coalsont, line 309: This is missing the `.sh` and the path based on `getenv(HCPPIPEDIR`.

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+
+  # convert input if needed
+  if [[ "$inFile0" == *dscalar.nii ]]; then
+    # convert input cifti to nifti
+    inFile="$(mktemp --suffix=.nii)"
+    $wbcmd -cifti-convert -to-nifti "$inFile0" "$inFile" -smaller-dims
+    # local dims
+    dims=$($wbcmd -file-information "$inFile" | grep Dimensions | awk -F': ' '{print $2}')
+    # shellcheck disable=SC2206
+    dims_arr=($dims)
```
coalsont, line 63: The `-file-information` output uses commas and spaces, while bash would need the comma removed.

It would be a little safer to use `grep '^Dimensions' to specifically require it to match the start of the line (if `-no-map-info` isn't used, and some goofball names one of their maps "Dimensions"...).

I generally recommend `-no-map-info` on `-file-information`, particularly for scripting.  On cifti files it avoids loading the data, but volume files don't support not loading the data, so it is mostly cutting down the amount of stuff grep might see.

`-file-information` probably won't give you a "1" for the number of maps if there is only one map.  Maybe that isn't the error condition you are looking for, though.

```suggestion
    "$wbcmd" -cifti-convert -to-nifti "$inFile0" "$inFile" -smaller-dims
    # local dims
    dims=$("$wbcmd" -file-information "$inFile" | grep Dimensions | awk -F': ' '{print $2}' | tr ',' ' ')
    # shellcheck disable=SC2206
    dims_arr=($dims)
```

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+    outFile="${outFile0%.nii.gz}"
+  else
+    echo "outFile is not a nifti or dscalar cifti?" >&2
+    return 1
+  fi
+
+  # run gaussian mixture modeling with melodic
+  mkdir -p "$tDir"
+  echo "1" > "$tDir/grot"
+  $melocmd -i "$inFile" --ICs="$inFile" --mix="$tDir/grot" -o "$tDir" --Oall --report -v --mmthresh=0
```
coalsont, line 105:
```suggestion
  "$melocmd" -i "$inFile" --ICs="$inFile" --mix="$tDir/grot" -o "$tDir" --Oall --report -v --mmthresh=0
```

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+
+  # check for multiple z-score maps
+  mapCounts=$(find "$tDir/stats/" -name 'thresh_zstat*' -exec bash -c 'fslval "$1" dim4' _ {} \;)
+  maxCount=0
+  for count in $mapCounts; do
+    if (( count > maxCount )); then maxCount=$count; fi
+  done
+  if (( maxCount > 1 )); then
+    echo "warning: At least one component returned two z-score maps (alpha = 0.05 and 0.01)! Using first map only." >&2
+  fi
```
coalsont, line 115: You don't need find to do globbing in bash (they aren't in an unpredictable folder name, right?), and you can stop early when you find one.  Can they ever be mismatched (does it ever need to check more than the first one)?

Side note, using "_" for $0 (the process name) in `bash -c` isn't very helpful.  I use "bash" for $0 because it is accurate, even if the rest of the arguments don't make much sense.
```suggestion
  for map in "$tDir"/stats/thresh_zstat*
  do
    mapCount=$(fslval "$map" dim4 | tr -d ' ')
    if ((mapCount > 1))
    then
      echo "warning: At least one component returned two z-score maps (alpha = 0.05 and 0.01)! Using first map only." 1>&2
      break
    fi
  done
```

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+
+  # concatenate volumes
+  $wbshrtctscmd -volume-concatenate -map 1 "${outFile}.${ext}" $(ls "$tDir"/stats/thresh_zstat* | sort -V)
+
+  # clean up temporary files
+  rm -r "$tDir"
+
+  # convert output to cifti, if needed
+  if [[ "$outFile0" == *dscalar.nii ]]; then
+    $wbcmd -cifti-convert -from-nifti "${outFile}.nii.gz" "$inFile0" "$outFile0"
```
coalsont, line 125:
```suggestion
  "$wbshrtctscmd" -volume-concatenate -map 1 "${outFile}.${ext}" $(ls "$tDir"/stats/thresh_zstat* | sort -V)

  # clean up temporary files
  rm -r "$tDir"

  # convert output to cifti, if needed
  if [[ "$outFile0" == *dscalar.nii ]]; then
    "$wbcmd" -cifti-convert -from-nifti "${outFile}.nii.gz" "$inFile0" "$outFile0"
```

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+    imrm "$inFile" "$outFile"
+  elif [[ "$inFile0" == *dscalar.nii ]]; then
+    imrm "$inFile"
+  fi
+
+  # restore FSLOUTPUTTYPE if needed
+  export FSLOUTPUTTYPE="$FSLOUTPUTTYPE0"
```
coalsont, line 132: `export` is only needed when the variable wasn't already exported.  If it belongs anywhere in the script, it should be the first time it is set (if someone doesn't have FSL set up properly, if this script doesn't export it, FSL would complain about the variable not being configured).

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+}
+# Call the function if the script is run directly
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+  mixtureModel "$@"
+fi
+# Usage example:
+# mixtureModel input.dscalar.nii output.dscalar.nii
```
coalsont, line 139: This is pythonish stuff to guard against an executable script being sourced (included) to use it as a function.  We don't need this, we separate our bash into scripts and libraries (and scripts can simply run other scripts, without being weird about it).  The implementation doesn't need to be a function at all.

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,139 @@
...
+#
+# ToDo:
+# Feed arbitrary melodic arguments.
+
+mixtureModel() {
+  local inFile="$1"
+  local outFile="$2"
+  local wbcmd="${3:-wb_command}"
+  local melocmd="${4:-melodic}"
+  local wbshrtctscmd="${5:-wb_shortcuts}"
```
coalsont, line 34: Scripts in `global/scripts` should probably be using `newopts.shlib`, not positional parameters.

global/scripts/mixtureModel.sh
bqrosen: A couple of fslvals is  probably the best solution. I guess I digitized what I as a human would literally do if I were checking manually ( -file-information  and look at the right spot in the output).

global/scripts/mixtureModel.sh
bqrosen: You can stop early, but you can't just check the first file. For test volume, the offending component is number 46 of 72.

global/scripts/mixtureModel.sh
bqrosen: FSLOUTPUTTYPE was only used for fslmerge no that we're using wb_shortcuts -volume-concatenate it doesn't need to be in the script at all.

global/scripts/mixtureModel.sh
```diff
@@ -0,0 +1,131 @@
...
+# See also:
+# https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;6e85d498.1607
+# https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/MELODIC#Using_melodic_for_just_doing_mixture-modelling
+# https://www.fmrib.ox.ac.uk/datasets/techrep/tr02cb1/tr02cb1.pdf
+#
+# Created by Burke Rosen
+# 2024-07-08
+# Dependencies: workbench, FSL
+# Written with workbench 2.0 and FSL 6.0.7.1
+# Refactored from matlab to bash 2025-05-19 by Copilot GPT-4.1
```
coalsont: On 2025-05-19, eh?  Interesting...

global/scripts/mixtureModel.sh
coalsont: Ah.  I re-added it thinking melodic would need to be told what to use, but forgot we delete the files melodic outputs anyway.

coalsont:
Okay, I think I'm done messing with it.  Please test whether it still works.

mharms:
I don't have anything to add to this.  The refactoring to use Bash rather than Matlab seems nice.

PR #345, opened 2025-05-28
Longitudinal tICA

mmilch01:
Longitudinal adaptation of tICA pipeline. The main difference with the base version is the new script tICAMakeCleanLongitudinalTemplate.sh called in the end of tICAPipeline.sh. This script extracts cleaned fMRI runs from each of the sessions, normalizes variance, re-concatenates across all runs and sessions, and multiplies by session-averaged sICA and tICA cleaned variance. The output is stored under the longitudinal template MNINonLinear/Results folder.

### DIFF
--- a/Examples/Scripts/RunTICAPipeline_PrecomputedTICA_SubjectCleanBatch-long.sh
+++ b/Examples/Scripts/RunTICAPipeline_PrecomputedTICA_SubjectCleanBatch-long.sh
@@ -139,7 +139,7 @@ function identify_timepoints
     for visit in ${PossibleVisits[*]}; do
         tp="${subject}_${visit}"
         if [ -d "$StudyFolder/$tp" ] && ! [[ " ${ExcludeVisits[*]+${ExcludeVisits[*]}} " =~ [[:space:]]"$tp"[[:space:]] ]]; then
-             if (( n==0 )); then 
+             if (( n==0 )); then
                     tplist="$tp"
              else
                     tplist="$tplist@$tp"
@@ -167,7 +167,7 @@ main() {
 
     # general settings
     # ICA mode
-    # ICA mode is hard-coded to be REUSE_ICA, which is mandatory in longitudinal mode.
+    # ICA mode is hard-coded to be REUSE_TICA, which is mandatory in longitudinal mode.
     # ICAmode="REUSE_TICA"
 
     # set how many subjects to do in parallel (local, not cluster-distributed) during individual projection and cleanup, defaults to all detected physical cores, '-1'
@@ -187,10 +187,10 @@ main() {
 
     # set list of fMRI on which ICA+FIX has been run, use @ to separate runs (not from the precomputed data)
     fMRINames="rfMRI_REST1_AP@rfMRI_REST1_PA@tfMRI_VISMOTOR_PA@tfMRI_CARIT_PA@tfMRI_FACENAME_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"
-    
+
     # fMRI names to extract. Required in longitudinal mode; must match (in longitudinal mode) cortical registration (MSMAll) extract names.
     extractfMRINames="rfMRI_REST1_AP@rfMRI_REST1_PA@rfMRI_REST2_AP@rfMRI_REST2_PA"
-    
+
     # fMRI name to extract. Required in longitudinal mode. Must match (in longitudinal mode) cortical registration (MSMAll) extract name.
     extractfMRIOut="rfMRI_REST"
 
@@ -255,7 +255,7 @@ main() {
         echo "queueing with fsl_sub to to $QUEUE"
         queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
     fi
-    
+
     #iterate over subjects
     for i in ${!Subjlist[@]}; do
         Subject="${Subjlist[i]}"
@@ -275,7 +275,7 @@ main() {
             --out-group-name="$GroupAverageName" \
             --fmri-resolution="$fMRIResolution" \
             --surf-reg-name="$RegName" \
-            --ica-mode="REUSE_ICA" \
+            --ica-mode="REUSE_TICA" \
             --num-wishart="$numWisharts" \
             --low-res="$LowResMesh" \
             --low-sica-dims="$LowsICADims" \
--- a/Examples/Scripts/RunTICAPipeline_PrecomputedTICA_SubjectCleanBatch-long.sh
+++ b/Examples/Scripts/RunTICAPipeline_PrecomputedTICA_SubjectCleanBatch-long.sh
@@ -1,0 +1,305 @@
+#!/bin/bash
+
+set -eu
+
+# this is an example script to run the tICA pipeline to clean a batch of subjects with group sICA and group tICA results that are both generated from a previous computation
+# steps that aren't needed for this mode are automaticaly skipped inside the pipeline
+# please make sure that precomputed group sICA and group tICA exist
+# please make sure that ICA-FIX, MSMAll and MakeAverageDataset are done properly matching the input arguments before running this tICA pipeline
+
+# Global default values
+DEFAULT_STUDY_FOLDER="${HOME}/data/Pipelines_ExampleData" # location of Subject folders (named by subjectID)
+# example subjects including 175 7T HCP subjects, separated by @
+# space separated subject list
+DEFAULT_SUBJECT_LIST="HCA6002236 HCA6002237"
+# list of longitudinal sessions suffices
+DEFAULT_POSSIBLE_VISITS="V1_MR V2_MR V3_MR"
+DEFAULT_TEMPLATE_LIST="HCA6002236_V1_V2_V3 HCA6002237_V1_V2_V3"
+DEFAULT_ENVIRONMENT_SCRIPT="$DEFAULT_STUDY_FOLDER/scripts/SetUpHCPPipeline-long.sh"
+DEFAULT_GROUP_NAME="HCA1798_MSMAll" # the group average name, which must be specified the same in MakeAverageDataset before running this tICA script
+DEFAULT_REG_NAME="MSMAll" # the registration string corresponding to the input files, which must be specified the same in MSMAll pipeline before running this tICA script
+DEFAULT_MATLAB_MODE=1 # MatlabMode
+DEFAULT_RUN_LOCAL=0
+DEFAULT_QUEUE="short.q"
+#DEFAULT_QUEUE="hcp_priority.q"
+
+get_options() {
+    local scriptName=$(basename "$0")
+    local arguments=("$@")
+
+    # initialize global variables
+    StudyFolder="${DEFAULT_STUDY_FOLDER}"
+    Subjlist=(${DEFAULT_SUBJECT_LIST})
+    TemplateList=($DEFAULT_TEMPLATE_LIST)
+    PossibleVisits=($DEFAULT_POSSIBLE_VISITS)
+    EnvironmentScript="${DEFAULT_ENVIRONMENT_SCRIPT}"
+    GroupAverageName="${DEFAULT_GROUP_NAME}"
+    RegName="${DEFAULT_REG_NAME}"
+    MatlabMode="${DEFAULT_MATLAB_MODE}"
+    RunLocal="${DEFAULT_RUN_LOCAL}"
+    QUEUE="${DEFAULT_QUEUE}"
+
+    # parse arguments
+    local index argument
+    local numArgs=${#arguments[@]}
+
+    for ((index = 0; index < ${#arguments[@]}; ++index))
+    do
+        argument="${arguments[index]}"
+
+        case "${argument}" in
+            --StudyFolder=*)
+                StudyFolder="${argument#*=}"
+                ;;
+            --Subject=*)
+                Subjlist="${argument#*=}"
+                ;;
+            --EnvironmentScript=*)
+                EnvironmentScript="${argument#*=}"
+                ;;
+            --GroupAverageName=*)
+                GroupAverageName="${argument#*=}"
+                ;;
+            --RegName=*)
+                RegName="${argument#*=}"
+                ;;
+            --MatlabMode=*)
+                MatlabMode="${argument#*=}"
+                ;;
+            --runlocal | --RunLocal)
+                RunLocal=1
+                ;;
+            --queue=*)
+                QUEUE="${argument#*=}"
+                ;;
+            *)
+                echo "ERROR: Unrecognized Option: ${argument}"
+                exit 1
+                ;;
+        esac
+    done
+
+    # check required parameters
+    if [[ "$StudyFolder" == "" ]]
+    then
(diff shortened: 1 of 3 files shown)


tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
```diff
@@ -0,0 +1,158 @@
...
+opts_AddMandatory '--session-list' 'SesslistRaw' 'HCA6002236_V1_MR@HCA6002236_V2_MR...' "list of longitudinal timepoint/session IDs separated by @s."
+opts_AddMandatory '--template-long' 'TemplateLong' 'ID' 'Longitudinal template ID'
+opts_AddMandatory '--extract-fmri-name-list' 'concatNamesToUse' 'name@name@name...' "list of fMRI run names to concatenate into the --extract-fmri-out output after tICA cleanup"
+opts_AddMandatory '--highpass' 'HighPass' 'hp value' 'High pass used with these data'
+opts_AddMandatory '--extract-fmri-name' 'extractNameOut' 'name' "fMRI name for concatenated extracted runs.
+Must match the one used in cortical registration."
+opts_AddOptional  '--reg-name' 'RegName' 'registration algorithm' "Cortical registration algorithm used [MSMAll]" "MSMAll"
+opts_AddOptional  '--fmri-name-concat-all' 'extractNameAll' 'name' "Concatenated output run label. If specified, must match the one used in multi-run FIX and cortical registration." ""
+opts_AddOptional  '--fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of all fmri run names separated by @. Required with --extract-fmri-out-all" ""
```
coalsont, line 29: `--extract-fmri-out-all` doesn't exist, is this what you meant?  Also, you don't need to specify the empty string as the default for optionals (unless there are compatibility switches).
```suggestion
opts_AddOptional  '--fmri-name-concat-all' 'extractNameAll' 'name' "Concatenated output run label. If specified, must match the one used in multi-run FIX and cortical registration.  Requires --fmri-names."
opts_AddOptional  '--fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "when using --fmri-name-concat-all, specify the list of MR FIX run names separated by @ here"
```

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
```diff
@@ -0,0 +1,158 @@
...
+        resultsDir="$StudyFolder"/"$sessionLong"/MNINonLinear/Results
+
+        # Average vn's in the atlas and native spaces. We assume that $nameOut exists in $sessionLong tICA output
+        vn_average_cifti_array+=(-cifti "$resultsDir/$nameOut/${nameOut}_Atlas_${RegName}_hp${HighPass}_clean_tclean_vn.dscalar.nii")
+        vn_average_nifti_array+=(-volume "$resultsDir/$nameOut/${nameOut}_hp${HighPass}_clean_tclean_vn.nii.gz")
+        
+        #reset session starting frame
+        frame_ind1=0
+        
+        for fmri in ${fMRIs[*]}; do
```
coalsont, line 79:
```suggestion
        for fmri in "${fMRIs[@]}"; do
```
If we want to support spaces as additional separators (and therefore can't have spaces in the fmri names), add it to the IFS above like `IFS='@ ' read -r -a ...`.

tICA/tICAPipeline.sh
```diff
@@ -667,3 +694,18 @@ then
     opts_conf_WriteConfig "$confoutfile"
 fi
 
+if (( IsLongitudinal )); then 
+    #Split, group variance normalize and concatenate cleaned timeseries across all sessions, storing in longitudinal template output.
+    #Also create averages across sessions for cleaned variance.
+    "$HCPPIPEDIR"/tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh \
+        --study-folder="$StudyFolder"       \
+        --subject="$Subject"                \
```
coalsont, line 702: You don't appear to be looping over...sessions...here.  I'm guessing that this is expected to fail when the user has told the tICA to stop before the CleanData step...so perhaps it should be in a second `par_addjob`/`par_runjobs` section of CleanData (to make sure all the tICACleanData.sh calls finish first)?

coalsont: On further inspection, if "longitudinal mode" can only handle the visits of a single subject, then maybe it should be a separate script, not awkwardly tacked on to the end of `tICAPipeline.sh`.

coalsont: In fact, the only changes appear to be handling the session list in longitudinal mode, the script call added to the end, and almost everything else is just renaming "subject" to session".  Everything else seems to be the arguments in the batch script (and the new extra script).

glasserm: Misha implemented what I asked him to.

coalsont: As long as tICAPipeline.sh supports doing the group stuff on large datasets, its implementation is not in fact "primarily a single-subject pipeline", because all of the complication to handle a group is necessary.  Users may primarily use it in single-subject mode, since we released a precomputed group for them to bootstrap from, but that is beside the point.  "Run tICA this particular way, then run this other script afterwards" doesn't need to add complication to the help info of `tICAPipeline.sh` itself.

I imagine bad things will happen if someone tries longitudinal in a mode other than REUSE_TICA, so the batch as written would need a warning about that.  If, however, we made a top-level script specifically for longitudinal like I suggested, we can hardcode this into its internal tICA call, and then the user doesn't need to think about it.

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
```diff
@@ -0,0 +1,158 @@
...
+    local fMRIStr="$1" sessionsStr="$2" template="$3" nameOut="$4" 
+    local fMRIs sessions
+    IFS=@ read -r -a fMRIs <<< "${fMRIStr}"
+    IFS=@ read -r -a sessions <<< "${sessionsStr}"
+    local numSessions="${#sessions[@]}"
+    local OutDir="$TemplateDir/MNINonLinear/Results/$nameOut"
+    mkdir -p "$OutDir"
+    
+    #per fMRI run lists for concat/merge commands
+    local vn_average_cifti_array
```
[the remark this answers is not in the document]
mmilch01: Ok

tICA/tICAPipeline.sh
```diff
@@ -118,12 +126,15 @@ fi
 opts_ShowValues
 
 #processing code goes here
-IFS='@' read -a Subjlist <<<"$SubjlistRaw"
```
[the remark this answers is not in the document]
mmilch01: Yes, it's initialized further down depending on longitudinal mode (line 173 currently)

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
mmilch01: Ok

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
```diff
@@ -0,0 +1,158 @@
...
+
+    # multiply by average _vn - nifti
+    wb_command -volume-math "TS * VN" "$OutDir/${nameOut}_hp${HighPass}_clean_tclean.nii.gz" \
+        -var TS "$OutDir/${nameOut}_hp${HighPass}_clean_tclean_v1.nii" \
+        -var VN "$OutDir/${nameOut}_hp${HighPass}_clean_tclean_vn.nii.gz" -repeat
+    
+    #clean up
+    rm -f "$OutDir/${nameOut}_Atlas_${RegName}_hp${HighPass}_clean_tclean_v1.dtseries.nii" \
+        "$OutDir/${nameOut}_hp${HighPass}_clean_tclean_v1.nii" \
+        "$OutDir/${nameOut}_hp${HighPass}_clean_tclean_vn_all.nii.gz"
```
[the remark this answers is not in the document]
mmilch01: Ok. I've made the requested changes and will commit after testing.

tICA/tICAPipeline.sh
coalsont: I am definitely not suggesting a modified copy of `tICAPipeline.sh`.

I am suggesting a new script that simply runs the existing tICAPipeline.sh with appropriate arguments (passing in the longitudinal session names as additional sessions, rather than needing to put that logic into tICAPipeline.sh), and then the does the extra stuff needed for longitudinal (directly, or by calling the script you wrote, depending on whether someone might need to (re)run the extra stuff without repeating the tICAPipeline.sh part).

mmilch01: I've added error checking and switched to tempfiles.shlib for temp file handling. This is tested to work as before, ready to review. Wrt Tim's suggestion of a separate tICAPipeline longitudinal script -- I see this design as more or less equivalent in complexity to what we have now. But, that's not how we implemented longitudinal pipelines previously.

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
```diff
@@ -0,0 +1,147 @@
...
+        resultsDir="$StudyFolder"/"$sessionLong"/MNINonLinear/Results
+
+        # Average vn's in the atlas and native spaces. We assume that $nameOut exists in $sessionLong tICA output
+        vn_average_cifti_array+=(-cifti "$resultsDir/$nameOut/${nameOut}_Atlas_${RegName}_hp${HighPass}_clean_tclean_vn.dscalar.nii")
+        vn_average_nifti_array+=(-volume "$resultsDir/$nameOut/${nameOut}_hp${HighPass}_clean_tclean_vn.nii.gz")
+        
+        #reset session starting frame
+        frame_ind1=0
+        
+        for fmri in ${fMRIs[@]}; do
```
coalsont, line 78: Quoting arrays with `[@]` is a good habit - that way, when you want whitespace safety in an array, it is already built into your habits.
```suggestion
        for fmri in "${fMRIs[@]}"; do
```

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
```diff
@@ -0,0 +1,147 @@
...
+    #per fMRI run lists for concat/merge commands
+    local vn_average_cifti_array=()
+    local vn_average_nifti_array=() #average of the _vn from all timepoints
+    local ts_concat_cifti_array=()
+    local ts_concat_nifti_array=() #average nifti timeseries
+    
+    #loop variables 
+    local session sessionLong resultsDir fmri fmri_rt frame_ind1 frame_ind2 NumTPs temp_demean1_v1 temp_demean1 temp_demean2_v1 temp_demean2
+    
+    for session in ${sessions[*]}; do
```
coalsont, line 66:
```suggestion
    for session in "${sessions[@]}"; do
```

tICA/tICAPipeline.sh
```diff
@@ -149,6 +160,25 @@ then
...
 fi
 
+if [[ "$IsLongitudinal" == "1" ]]; then
+    if [[ "$ICAmode" != "REUSE_ICA" ]]; then
+        log_Err "mode other than REUSE_ICA is not supported in longitudinal processing"
+    fi
+    if [[ "$TemplateLong" == "" || "$Subject" == "" || "$extractNameOut" == "" ]]; then 
+        log_Err_Abort "--extract-fmri-out, --longitudinal-template and --longitudinal-subject are required in longitudinal mode."
+    fi
+    if [[ "$ExtractAllRunsLong" == "1" ]]; then
```
coalsont, line 170: For style, I prefer handling booleans like this (this is why opts_StringToBool normalizes to a number instead of something like "true").
```suggestion
    if ((ExtractAllRunsLong)); then 
```

tICA/tICAPipeline.sh
```diff
@@ -149,6 +160,25 @@ then
...
+    fi
+    if [[ "$TemplateLong" == "" || "$Subject" == "" || "$extractNameOut" == "" ]]; then 
+        log_Err_Abort "--extract-fmri-out, --longitudinal-template and --longitudinal-subject are required in longitudinal mode."
+    fi
+    if [[ "$ExtractAllRunsLong" == "1" ]]; then 
+        extractNameAllLong="$MRFixConcatName"
+    fi
+    IFS='@' read -a SesslistCross <<<"$SesslistRaw"
+    Sesslist=()
+    for sess in ${SesslistCross[@]}; do
```
coalsont, line 175:
```suggestion
    for sess in "${SesslistCross[@]}"; do
```

glasserm (approved):

Examples/Scripts/RunTICAPipeline_PrecomputedTICA_SubjectCleanBatch-long.sh
```diff
@@ -0,0 +1,305 @@
...
+    # set up pipeline environment variables and software
+    source "${EnvironmentScript}"
+
+    # general settings
+    # set the ICA mode
+    # NEW - estimate a new sICA and a new tICA
+    # REUSE_SICA_ONLY - reuse an existing sICA and estimate a new tICA
+    # INITIALIZE_TICA - reuse an existing sICA and use an existing tICA to start the estimation
+    # REUSE_TICA - reuse an existing sICA and an existing tICA"
+    ICAmode="REUSE_TICA"
```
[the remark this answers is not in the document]
mmilch01: Ok

tICA/scripts/tICAMakeCleanLongitudinalTemplate.sh
mmilch01: Ok

tICA/tICAPipeline.sh
```diff
@@ -149,6 +160,25 @@ then
     signalTxtName="ReCleanSignal.txt"
 fi
 
+if [[ "$IsLongitudinal" == "1" ]]; then
+    if [[ "$ICAmode" != "REUSE_ICA" ]]; then
+        log_Err "mode other than REUSE_ICA is not supported in longitudinal processing"
```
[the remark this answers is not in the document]
mmilch01: Ok

tICA/tICAPipeline.sh
mmilch01: Ok. This is changed in a separate commit.

tICA/tICAPipeline.sh
mmilch01: Ok. Also in an outside commit.

tICA/tICAPipeline.sh
```diff
@@ -149,6 +160,25 @@ then
     signalTxtName="ReCleanSignal.txt"
 fi
 
+if [[ "$IsLongitudinal" == "1" ]]; then
```
coalsont: Consider similar boolean style (to match what you already did with `IsLongitudinal` at the end of the script).
```suggestion
if ((IsLongitudinal)); then
```

mmilch01:
> @mmilch01 is this ready to merge?

Yes from my perspective unless we want to wait on Jure to test in in QuNex.

PR #348, opened 2025-07-07
Automatic creation of individual GMWM templates in transmit bias

demsarjure:
If the user does not set the `--gmwm-template` parameter in transmit bias pipelines. The pipeline will not try to automatically create it before running `Phase3_IndividualAdjustment.sh`.

### DIFF
--- a/TransmitBias/RunIndividualOnly.sh
+++ b/TransmitBias/RunIndividualOnly.sh
@@ -21,7 +21,9 @@ AFI - actual flip angle sequence with two different echo times, requires --afi-i
 B1Tx - b1 transmit sequence magnitude/phase pair, requires --b1tx-magnitude, --b1tx-phase, and --group-corrected-myelin
 
 PseudoTransmit - use spin echo fieldmaps, SBRef, and a template transmit-corrected myelin map to derive empirical correction, requires --pt-fmri-names, --myelin-template, --group-uncorrected-myelin, and --reference-value"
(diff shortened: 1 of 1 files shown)


TransmitBias/RunIndividualOnly.sh
```diff
@@ -145,6 +147,28 @@ esac
...
     --old-myelin-mapping="$oldmappingStr"
 
+# if GMWMtemplate is not set, create the individual GMWM template
+if [[ -z "$GMWMtemplate" ]]
+then
+    # output GMWM template location
+    GMWMOut="${StudyFolder}/${Subject}/MNINonLinear/gmwm_template.nii.gz"
+
+    # generate
+    wb_command -volume-label-import \
```
coalsont: Might be better to just use the one already generated by Phase1:

https://github.com/Washington-University/HCPpipelines/blob/b971ad3677d430a52246833282e7ed637aa75906/TransmitBias/scripts/CreateTransmitBiasROIs.sh#L78

glasserm:
That's a good point, perhaps we can just make the parameter optional and use the file Tim pointed to if not provided?

demsarjure:
Fixed. Note that transmit bias script have a bunch of remarks/questions in the form of `#MFG:`. For example:

```
#could be optional?
#MFG: stay mandatory
```

This is something that could be probably cleaned up here?

PR #349, opened 2025-07-07
ReClean resolution handling

coalsont:
This needed a bit more than my initial optimistic commit.

It had an argument named `--fmri-resolution` which appears to have actually needed to match the grayordinate resolution.  So, I have done some renaming in addition to exposing the extra arguments.

It should still be compatible with the previous Batch (compatibility flag name, defaults for the added options).

It has not been tested.


mharms:
I was confused by the diffs initially, but now I see that this PR is tweaking a commit that you previously committed directly to the `master` branch.  Conceptually, this looks reasonable to me.

glasserm (approved):
This looks fine to me.  Shall we just merge it and have Jure build the container for Petra to test?

demsarjure (approved):

coalsont:
I usually recompile in the PR branch, in case it needs to be done more than once due to additional changes.

coalsont:
Do we want to test before merging?

demsarjure:
On our end, I can test together with all the latest merges before the next QuNex release. So probably the most efficient to just merge this.

PR #354, opened 2025-08-21
tICA for NHPs, including optionally hand selecting groupSICA components to send forward to groupTICA

bqrosen:


tICA/scripts/ComputeGroupTICA.m
```diff
@@ -230,29 +253,38 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
             ts.NtimepointsPerSubject = size(tICATCS.cdata, 2);
             tICASpectra.cdata = nets_spectra_sp(ts)';
-            %FIXME
-            ciftisave(tICASpectra, [SubjFolder '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' SubjectList{i} '.' OutString '_tICA' RegString '_spectra.' LowResMesh 'k_fs_LR.sdseries.nii'], 'wb_command');
+
+
+            tICASpectra.diminfo{1}.length = tICAdim;
+            tICASpectra.diminfo{1}.maps = tICASpectra.diminfo{1}.maps(sigIdx);
+            tICASpectra.diminfo{2}.length = size(tICASpectra.cdata,2);
+            cifti_write(tICASpectra, [SubjFolder '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' SubjectList{i} '.' OutString '_tICA' RegString '_spectra.' LowResMesh 'k_fs_LR.sdseries.nii']);%FIXME
```
coalsont: For future reference, there is a `cifti_write_sdseries` convenience function so you wouldn't need to figure out which dimension contains the map names.  Here though, you'd still need to pass the names of the selected set of maps into it, so it wouldn't be much simpler.

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -267,3 +299,11 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
     lines = array{1};
 end
 
+function C = filterCifti(C,dimIdx,filtIdx)
+    % subselect from one dim of cifti
+    C.diminfo{dimIdx}.length = numel(filtIdx);
+    C.diminfo{dimIdx}.maps = C.diminfo{dimIdx}.maps(filtIdx);
+    idx = repmat({':'}, 1, ndims(C.cdata));
+    idx{dimIdx} = filtIdx;
+    C.cdata = C.cdata(idx{:});
```
coalsont: Char array ':' for all indexes?  Do we know what minimum version of matlab is needed to support this?

tICA/scripts/ComputeGroupTICA.sh
```diff
@@ -64,7 +65,7 @@ fi
 opts_ShowValues
 
 #FIXME: hardcoded naming conventions, move these to high level script when ready
-OutputFolder="$OutGroupFolder/MNINonLinear/Results/$fMRIConcatName/tICA_d$sICAdim"
+OutputFolder="$OutGroupFolder/MNINonLinear/Results/$fMRIConcatName/tICA_d$tICAdim"
 
 TCSConcatName="$OutputFolder/sICA_TCS_$sICAdim.sdseries.nii"
```
coalsont: Is it correct to have all of these files named `.../tICA_d85/sICA_TCS_87.sdseries.nii` and similar?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -28,35 +29,55 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
         error('input lists are not the same length');
     end
-
+    
     TCSMask = reshape(TCSMaskConcat.cdata, [sICAdim, RunsXNumTimePoints, numsubj]);
 
-    numfullsubj = 0;
+    %% optionally, filter Group sICA components to hand picked list before running tICA
+    handSigFile = [OutputFolder '/../sICA/HandSignal.txt'];% assumes OutputFolder and /sICA are always in the same parent dir
+    if exist(handSigFile,'file')
```
coalsont: It may be more elegant to pass this filename as an option, rather than hardcoding changed behavior based on what filenames exist.  But, I suppose if this is the same file from the usual `ApplyHandReClassifications` codepath, it probably makes some sense to hardcode it.

tICA/scripts/GroupSICA.m
```diff
@@ -19,6 +19,7 @@ function GroupSICA(indata, indatavn, OutFolder, wfoutname, numWisharts, dimList,
     if status == 0
     	error(['failed to create ' OutFolder ' because of ' msg]);
     end
+    save([OutFolder '/Out.mat'], 'Out', '-v7.3');
```
coalsont, line 22: So, with this change, we save this with the calcDim still modified (if requested), without the dim in the filename, earlier in the process.  Why?  Is this for debugging failures or something?

tICA/tICAPipeline.sh
```diff
@@ -403,17 +410,17 @@ do
...
                 --icadim-override="$sicadimOverride" \
                 --matlab-run-mode="$MatlabMode"
-            sICAActualDim=$(cat "$sICAoutfolder/most_recent_dim.txt")
-            if [[ "$tICADim" == "" ]]
-            then
-                tICADim="$sICAActualDim"
+            ;;
+        (indProjSICA)
+            # now we have the dimensionality, set the sICA output strings
+            if [[ "$sicadimOverride" ]]; then
```
coalsont, line 413:
```suggestion
            if [[ "$sicadimOverride" != "" ]]; then 
```

tICA/tICAPipeline.sh
```diff
@@ -608,12 +646,19 @@ do
             par_runjobs "$parLimit"
             ;;
         (ComputeTICAFeatures)
+            # now we have the dimensionality, set the sICA and tICA output strings
+            sICAActualDim=$(cat "$sICAoutfolder/most_recent_dim.txt")
```
coalsont: Is it correct to ignore the provided `--sicadim-override` argument if someone starts the pipeline from this step?  If in doubt, the logic should probably match the other places that check this.

tICA/tICAPipeline.sh
```diff
@@ -666,7 +711,7 @@ do
                     --study-folder="$StudyFolder" \
                     --subject="$Session" \
                     --noise-list="$NuisanceListTxt" \
-                    --timeseries="${StudyFolder}/${Session}/MNINonLinear/fsaverage_LR32k/${Session}.${OutputString}_WR_tICA${RegString}_ts.32k_fs_LR.sdseries.nii" \
+                    --timeseries="${StudyFolder}/${Session}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Session}.${TICA}_WR_tICA${RegString}_ts.${LowResMesh}k_fs_LR.sdseries.nii" \
```
coalsont: I am guessing this was intended:
```suggestion
                    --timeseries="${StudyFolder}/${Session}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Session}.${OutputStringTICA}_WR_tICA${RegString}_ts.${LowResMesh}k_fs_LR.sdseries.nii" \
```

tICA/scripts/ComputeGroupTICA.sh
bqrosen: That's intended, because these are sICA files.

tICA/scripts/ComputeGroupTICA.m
bqrosen: It works in 2017b.

tICA/scripts/ComputeGroupTICA.m
bqrosen: It's not the same file, but it is analogous, it uses the same hardcodeded file detection for consistency.

tICA/scripts/GroupSICA.m
bqrosen: The requested dimensionality does not change this file, which just contains the output of icaDim().

tICA/tICAPipeline.sh
bqrosen: Nope, I missed this step when putting in the if statements. Fixed.

tICA/scripts/GroupSICA.m
coalsont: It can modify the .calcDim of the struct (just an integer), even though it doesn't change anything else (just like before).  Changing the naming conventions of an output generally deserves some thought.  Perhaps this file is only intended for subject-wise manual inspection, so that we can expect a human to notice which naming convention was used in a given dataset?

tICA/scripts/ComputeGroupTICA.sh
glasserm: Right.

tICA/scripts/ComputeGroupTICA.m
glasserm: Ok

glasserm: I was thinking we would remove the sICA components before they are weighted regressed into the individuals, not after.

tICA/scripts/GroupSICA.m
bqrosen: What I meant is that the subfunction myprocess() does not change Out.CalcDim to match its first argument, ICAdim, so the multiple Out_<#>.mat files saved in the `for i = dimList(:)'` loop have identical contents. An alternative solution would be to add `Out.CalcDim = ICAdim;` inside myprocess(). But it seems like a waste to save multiple copies of these large files to change a single integer field which I'm pretty sure is not used for anything.

coalsont: Ah, from the diff, I hadn't caught on that it was looping resulting in multiple copies.  As long as we never intend to use this output noninteractively (and would therefore want it to have a predictable name regardless of pipeline version), this seems reasonable.

glasserm:
So our goal is to have real structured artifacts removed by the initial sICA denoising at the individual level.  Autocorrelated unstructured noise is not "real" in the sense that it is something we need to attempt to remove with ICA-based denoising.  In fact, removing part of it while leaving part behind, may cause more problems than just ignoring the phenomenon.  I was going for an implementation that ignored the phenomenon, since we cannot easily fix it at this point (the fix would be an equal area surface on the midthickness).

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -118,13 +119,13 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
             normicasig = W * TCSFullConcat.cdata;
         elseif i == 1
             IT = [num2str(i)];
-            [iq, A, W, normicasig, sR] = icasso('both', TCSFullConcat.cdata, iterations, 'approach', 'symm', 'g', nlfunc, 'lastEig', sICAdim, 'numOfIC', tICAdim, 'maxNumIterations', 1000); %x1
+            [iq, A, W, normicasig, sR] = icasso('both', TCSFullConcat.cdata, iterations, 'approach', 'symm', 'g', nlfunc, 'lastEig', sICAdim, 'numOfIC', sICAdim, 'maxNumIterations', 1000); %x1
         elseif i > 5
```
glasserm, line 121: shouldn't this be tICAdim?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -118,13 +119,13 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
             IT = ['F'];
-            [normicasig, A, W] = fastica(TCSFullConcat.cdata, 'initGuess', A, 'approach', 'symm', 'g', nlfunc, 'lastEig', sICAdim, 'numOfIC', tICAdim, 'displayMode', 'off', 'maxNumIterations', 1000); %x1
+            [normicasig, A, W] = fastica(TCSFullConcat.cdata, 'initGuess', A, 'approach', 'symm', 'g', nlfunc, 'lastEig', sICAdim, 'numOfIC', sICAdim, 'displayMode', 'off', 'maxNumIterations', 1000); %x1
         else
```
glasserm, line 124: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -118,13 +119,13 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
             IT = [num2str(i)];
-            [iq, A, W, normicasig, sR] = icasso('bootstrap', TCSFullConcat.cdata, iterations, 'initGuess', A, 'approach', 'symm', 'g', nlfunc, 'lastEig', sICAdim, 'numOfIC', tICAdim, 'maxNumIterations', 1000); %x4
+            [iq, A, W, normicasig, sR] = icasso('bootstrap', TCSFullConcat.cdata, iterations, 'initGuess', A, 'approach', 'symm', 'g', nlfunc, 'lastEig', sICAdim, 'numOfIC', sICAdim, 'maxNumIterations', 1000); %x4
         end
```
glasserm, line 127: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -183,9 +184,9 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
         tICAVolMaps.cdata(:, [1:1:length(Is)]) = single(tICAVolMaps.cdata(:, Is));
 
         
-        nameParamPart = ['_' num2str(tICAdim) '_' nlfunc IT];
-        ciftisavereset(tICAMaps, [OutputFolder '/' tICAMapsNamePart nameParamPart '.dscalar.nii'], wbcommand);
-        ciftisavereset(tICAVolMaps, [OutputFolder '/' tICAVolMapsNamePart nameParamPart '.dscalar.nii'], wbcommand);
+        nameParamPart = ['_' num2str(sICAdim) '_' nlfunc IT];
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
 
         tICAtcs.cdata = tICAtcs.cdata';
 
-        ciftisavereset(tICAtcs, [OutputFolder '/' tICAtcsNamePart nameParamPart '.sdseries.nii'], wbcommand);
+        cifti_write(tICAtcs, [OutputFolder '/' tICAtcsNamePart nameParamPart '.sdseries.nii']);
```
glasserm: Does cifti_write handle reset behavior?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
 
-        tICAtcsAll = reshape(tICAtcs.cdata, tICAdim, RunsXNumTimePoints, numsubj);
+        tICAtcsAll = reshape(tICAtcs.cdata, sICAdim, RunsXNumTimePoints, numsubj);
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
 
         tICAtcsmean = AvgTCS;
-        tICAtcsmean.cdata = sum(tICAtcsAll .* single(TCSMask(1:tICAdim, :, :) == 1), 3) / numfullsubj;
+        tICAtcsmean.cdata = sum(tICAtcsAll .* single(TCSMask(1:sICAdim, :, :) == 1), 3) / numfullsubj;
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
         tICAtcsabsmean = AvgTCS;
-        tICAtcsabsmean.cdata = sum(abs(tICAtcsAll .* single(TCSMask(1:tICAdim, :, :) == 1)), 3) / numfullsubj;
+        tICAtcsabsmean.cdata = sum(abs(tICAtcsAll .* single(TCSMask(1:sICAdim, :, :) == 1)), 3) / numfullsubj;
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
 
-        ciftisavereset(tICAtcsmean, [OutputFolder '/' tICAtcsmeanNamePart nameParamPart '.sdseries.nii'], wbcommand);
-        ciftisavereset(tICAtcsabsmean, [OutputFolder '/' tICAtcsabsmeanNamePart nameParamPart '.sdseries.nii'], wbcommand);
+        cifti_write(tICAtcsmean, [OutputFolder '/' tICAtcsmeanNamePart nameParamPart '.sdseries.nii']);
+        cifti_write(tICAtcsabsmean, [OutputFolder '/' tICAtcsabsmeanNamePart nameParamPart '.sdseries.nii']);
```
glasserm: Reset?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
 
         tICAspectra = sICAAvgSpectra;
         tICAspectranorm = sICAAvgSpectra;
-        ts.Nnodes = tICAdim;
+        ts.Nnodes = sICAdim;
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
         ts.Nsubjects = numfullsubj;
         ts.NtimepointsPerSubject = RunsXNumTimePoints;
-        ts.ts = (((tICAtcs.cdata .* single(TCSMaskConcat.cdata(1:tICAdim, :) == 1)) ./ numfullsubj) .* numsubj)'; %compensate for the zeros by ratio of full-data subjects to all subjects?
+        ts.ts = (((tICAtcs.cdata .* single(TCSMaskConcat.cdata(1:sICAdim, :) == 1)) ./ numfullsubj) .* numsubj)'; %compensate for the zeros by ratio of full-data subjects to all subjects?
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
         tICAspectra.cdata = nets_spectra_sp(ts)';
         tICAspectranorm.cdata = nets_spectra_sp(ts, [], 1)';
 
-        ciftisavereset(tICAspectra, [OutputFolder '/' tICAspectraNamePart nameParamPart '.sdseries.nii'], wbcommand);
-        ciftisavereset(tICAspectranorm, [OutputFolder '/' tICAspectranormNamePart nameParamPart '.sdseries.nii'], wbcommand);
+        cifti_write(tICAspectra, [OutputFolder '/' tICAspectraNamePart nameParamPart '.sdseries.nii']);
+        cifti_write(tICAspectranorm, [OutputFolder '/' tICAspectranormNamePart nameParamPart '.sdseries.nii']);
```
glasserm: Reset?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
+            sICASpectra = cifti_read(SpectraList{i});
             tICATCS = sICATCS;
             %tICATCS.cdata = pinv(tICAmix) * sICATCS.cdata;
             tICATCS.cdata = squeeze(tICAtcsAll(:, std(tICAtcsAll(:, :, i), [], 1) > 0, i));
             tICASpectra = sICASpectra;
             SubjFolder = [StudyFolder '/' SubjectList{i} '/'];
-            %FIXME: how to deal with subject ID in this filename without hardcoding conventions?
-            ciftisave(tICATCS, [SubjFolder 'MNINonLinear/fsaverage_LR' LowResMesh 'k/' SubjectList{i} '.' OutString '_tICA' RegString '_ts.' LowResMesh 'k_fs_LR.sdseries.nii'], 'wb_command');
+
+            tICATCS.diminfo{1}.length = sICAdim;
```
glasserm: And this?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -209,50 +210,59 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
+            tICATCS.diminfo{1}.maps = tICATCS.diminfo{1}.maps;%(sigIdx);
```
glasserm: This seems like a simple equality, not what you intended?

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -267,3 +277,11 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
...
 end
 
+% function C = filterCifti(C,dimIdx,filtIdx)
+%     % subselect from one dim of cifti
+%     C.diminfo{dimIdx}.length = numel(filtIdx);
+%     C.diminfo{dimIdx}.maps = C.diminfo{dimIdx}.maps(filtIdx);
+%     idx = repmat({':'}, 1, ndims(C.cdata));
+%     idx{dimIdx} = filtIdx;
+%     C.cdata = C.cdata(idx{:});
+% end
```
glasserm: Commented out.

tICA/scripts/ConcatGroupSICA.m
```diff
@@ -76,12 +74,14 @@ function ConcatGroupSICA(TCSListName, MapListName, VolMapListName, SpectraListNa
 
     TCSFullConcat = TCSSub; % use the last good subject as a template
     TCSFullConcat.cdata = TCSFull;
-    ciftisavereset(TCSFullConcat, OutTCSName, wbcommand);
+    TCSFullConcat.diminfo{2}.length = size(TCSFullConcat.cdata,2);
+    cifti_write(TCSFullConcat, OutTCSName);
```
glasserm: Reset?

tICA/scripts/ConcatGroupSICA.m
```diff
@@ -76,12 +74,14 @@ function ConcatGroupSICA(TCSListName, MapListName, VolMapListName, SpectraListNa
...
     clear TCSFullConcat;
 
     TCSMaskConcat = TCSSub;
     TCSMaskConcat.cdata = TCSMask;
-    ciftisavereset(TCSMaskConcat, OutTCSMaskName, wbcommand);
+    TCSMaskConcat.diminfo{2}.length = size(TCSMaskConcat.cdata,2);
+    cifti_write(TCSMaskConcat, OutTCSMaskName);
```
glasserm: Reset?

tICA/scripts/ConcatGroupSICA.sh
```diff
@@ -59,7 +59,14 @@ opts_ShowValues
...
 
-OutputFolder="$OutGroupFolder/MNINonLinear/Results/$fMRIConcatName/tICA_d$sICAdim"
+# if Group sICA hand classifications exists, use it to filter the group sICA components before running tICA
+HandSignalFile="${OutGroupFolder}/MNINonLinear/Results/${fMRIConcatName}/sICA/HandSignal.txt" 
+if [ -e "${HandSignalFile}" ]; then
+    tICADim=$(wc -w < "${HandSignalFile}")
+else
+    tICADim="$sICAdim"
+fi
+OutputFolder="$OutGroupFolder/MNINonLinear/Results/$fMRIConcatName/tICA_d$tICADim"
```
glasserm, line 69: sICADim not tICADim here?

tICA/scripts/GroupSICA.m
```diff
@@ -32,8 +33,6 @@ function GroupSICA(indata, indatavn, OutFolder, wfoutname, numWisharts, dimList,
     newcii = cii;
     newcii.cdata = Out.data;
     ciftisave(newcii, wfoutname, 'wb_command');
-
-    mkdir(OutFolder);
```
glasserm, line 36: Why was this deleted?

tICA/scripts/GroupSICA.m
```diff
@@ -71,7 +89,7 @@ function myprocess(ICAdim, Out, vn, OutFolder, cii)
     
     save([OutFolder '/iq_' num2str(ICAdim)], 'iq', '-v7.3');
     save([OutFolder '/sR_' num2str(ICAdim)], 'sR', '-v7.3');
-    save([OutFolder '/Out_' num2str(ICAdim)], 'Out', '-v7.3');
+%     save([OutFolder '/Out_' num2str(ICAdim)], 'Out', '-v7.3');
```
glasserm, line 92: Why is this commented out?

tICA/tICAPipeline.sh
```diff
@@ -520,7 +561,7 @@ do
                         --out-folder="${StudyFolder}/${GroupAverageName}"
                         --fmri-concat-name="$OutputfMRIName"
                         --surf-reg-name="$RegName"
-                        --ica-dim="$tICADim"
+                        --sICA-dim="$tICADim"
```
glasserm: Should this be tICA dim?

glasserm (changes requested):
I made a variety of comments.  Please address them.

tICA/scripts/ComputeGroupTICA.m
bqrosen: Fixed. Note that this is just a variable name change, not a functionality change, because $tICAdim was being passed in as the --sICA-dim argument to ComputeGroupTICA.sh in tICAPipeline.sh

tICA/scripts/GroupSICA.m
bqrosen: It's a duplicate of line 18.

tICA/scripts/GroupSICA.m
bqrosen: I moved the saving of Out out of the myprocess() sub function (to line 22), because it was saving an identical Out file for each iteration of the dimList loop. myprocess() does not modify the contents of Out in anyway so I don't think there is any reason to save multiple copies of these large files.

tICA/scripts/ComputeGroupTICA.m
bqrosen: Silent mutation debris from moving the filtering upstream. Removed for cleanliness .

tICA/scripts/ComputeGroupTICA.m
bqrosen: cifti_write itself doesn't but I manually modified the diminfo header for sdseries where needed.  I now use cifti_diminfo_make_series and tcifti_diminfo_make_scalars to make the reset the diminfo more explicitly for sdseries ciftis.

coalsont: No.  The `ciftisavereset` wrapper will probably exist forever, so it can be used.  The current alternatives for sdseries are `cifti_write_sdseries` (which discards the map names, ~and doesn't currently have an option to set them~ unless passed explicitly as a `'namelist'` vararg), or to manually adjust the diminfo (`cifti_write_from_template` is currently specific to files with either brain models or parcels mappings).

Edit: `cifti_write_sdseries` can take a name list, I missed that it gets passed into the inner function as an "unrecognized vararg" from the writing function.

tICA/scripts/ComputeGroupTICA.m
```diff
@@ -184,8 +185,8 @@ function ComputeGroupTICA(StudyFolder, SubjListName, TCSListName, SpectraListNam
 
         
         nameParamPart = ['_' num2str(tICAdim) '_' nlfunc IT];
-        ciftisavereset(tICAMaps, [OutputFolder '/' tICAMapsNamePart nameParamPart '.dscalar.nii'], wbcommand);
-        ciftisavereset(tICAVolMaps, [OutputFolder '/' tICAVolMapsNamePart nameParamPart '.dscalar.nii'], wbcommand);
+        cifti_write(tICAMaps, [OutputFolder '/' tICAMapsNamePart nameParamPart '.dscalar.nii']);
+        cifti_write(tICAVolMaps, [OutputFolder '/' tICAVolMapsNamePart nameParamPart '.dscalar.nii']);
```
coalsont, line 187: This appears to expect the tICA mixing matrix to be square:

https://github.com/Washington-University/HCPpipelines/blob/308aada226ba1ae54e9444db64e52ca89320eef0/tICA/scripts/ComputeGroupTICA.m#L145-L146

tICA/scripts/ComputeGroupTICA.m
coalsont: Since the current solution here uses `cifti_diminfo_make_scalars` without a namelist, it would be equivalent to do these 3 lines as a single `cifti_write_sdseries` line.

bqrosen: changed to cifti_write_sdseries

tICA/scripts/ComputeGroupTICA.m
coalsont: Moot for now because this script generates the mixing matrix, and doesn't currently support reducing the number of components compared to ICA (and the sICA components were already removed).

PR #368, opened 2025-11-07
Bugfix longitudinal example

myersm0:
Fix undefined loop variable in longitudinal batch example.

The loop variable `i` was never defined when accessing Templates array, causing the script to fail or use incorrect templates for subjects.

This assumes Templates and Subjlist arrays are meant to be parallel (subject i uses template i). I believe this is the intention, by example of other scripts in the same directory.

### DIFF
--- a/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
+++ b/Examples/Scripts/DeDriftAndResamplePipelineBatch-long.sh
@@ -4,7 +4,6 @@ get_batch_options() {
     local arguments=("$@")
 
     command_line_specified_study_folder=""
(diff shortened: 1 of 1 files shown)


coalsont:
Looks like a good fix.

I also notice the `--Subjlist` argument parsing isn't going to work for more than one subject, as it doesn't result in an array, and would be largely useless without being able to set the Templates array in a similar fashion...maybe we should just remove all lines related to `--Subjlist` and `command_line_specified_subj`...

coalsont:
Please also remove any bits of code referencing the now-defunct `command_line_specified_subj` also.

PR #370, opened 2025-12-01
fMRISurface unified with NHP processing

mmilch01:
FMRISurface adapted for NHP processing.

### DIFF
--- a/Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
+++ b/Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
@@ -1,0 +1,166 @@
+#!/bin/bash 
+
+get_batch_options() {
+    local arguments=("$@")
+
+    command_line_specified_study_folder=""
+    command_line_specified_sess=""
+    command_line_specified_run_local="FALSE"
+
+    local index=0
+    local numArgs=${#arguments[@]}
+    local argument
+
+    while [ ${index} -lt ${numArgs} ]; do
+        argument=${arguments[index]}
+
+        case ${argument} in
+            --StudyFolder=*)
+                command_line_specified_study_folder=${argument#*=}
+                index=$(( index + 1 ))
+                ;;
+            --Session=*)
+                command_line_specified_sess=${argument#*=}
+                index=$(( index + 1 ))
+                ;;
+            --runlocal)
+                command_line_specified_run_local="TRUE"
+                index=$(( index + 1 ))
+                ;;
+            *)
+                echo ""
+                echo "ERROR: Unrecognized Option: ${argument}"
+                echo ""
+                exit 1
+            ;;
+        esac
+    done
+}
+
+
+#TODO NHP: check if -S <num> is NHP feature or legacy human feature (likely the latter)
+# Usage () {
+# echo ""
+# echo "Usage $0 <StudyFolder> <Subject1> [options]"
+# echo ""
+# echo "   Options:"
+# echo "       -S <num>             : specify Nth session. You need all sets of variables in hcppipe_conf.txt"
+# echo "                              (i.e., Tasklist, Taskreflist, TopupPositive, TopupNegative, PhaseEncodingList,"
+# echo "                              and DwellTime, TruePatientPosition [HFS, HFSx, FFSx], ScannerPatientPosition "
+# echo "                              [HFS, HFP, FFS, FFP])"
+# echo "       -d                   : dry run (print commands in the terminal but not run)" 
+# echo ""
+# exit 1;
+# }
+
+
+get_batch_options "$@"
+
+StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Session folders (named by SessionID)
+Sesslist="100307 100610" #Space delimited list of Session IDs
+EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script
+SPECIES="Human"
+
+if [ -n "${command_line_specified_study_folder}" ]; then
+    StudyFolder="${command_line_specified_study_folder}"
+fi
+
+if [ -n "${command_line_specified_sess}" ]; then
+    Sesslist="${command_line_specified_sess}"
+fi
+
+if [ "$SPECIES" = "" ] ; then 
+    echo "ERROR: please export SPECIES first to any of Macaque, MacaqueCyno, MacaqueRhesus, Marmoset, NightMonkey, Chimp, Human"
+    exit 1
+fi
+
+# Requirements for this script
+#  installed versions of: FSL, Connectome Workbench (wb_command)
+#  environment: HCPPIPEDIR, FSLDIR, CARET7DIR
+
+#Set up pipeline environment variables and software
+source "$EnvironmentScript"
+source "$HCPPIPEDIR"/Examples/Scripts/SetUpSPECIES.sh --species="$SPECIES"
+
+
+# Log the originating call
+echo "$@"
+
+#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
+#DO NOT include "-q " at the beginning
+#default to no queue, implying run local
+QUEUE=""
+#QUEUE="hcp_priority.q"
+
+########################################## INPUTS ##########################################
+
+#Scripts called by this script do assume they run on the outputs of the FreeSurfer Pipeline
+
+######################################### DO WORK ##########################################
+
+TaskList=()
+TaskList+=(rfMRI_REST1_RL)
+TaskList+=(rfMRI_REST1_LR)
+TaskList+=(rfMRI_REST2_RL)
+TaskList+=(rfMRI_REST2_LR)
+TaskList+=(tfMRI_EMOTION_RL)
+TaskList+=(tfMRI_EMOTION_LR)
+TaskList+=(tfMRI_GAMBLING_RL)
+TaskList+=(tfMRI_GAMBLING_LR)
+TaskList+=(tfMRI_LANGUAGE_RL)
+TaskList+=(tfMRI_LANGUAGE_LR)
+TaskList+=(tfMRI_MOTOR_RL)
+TaskList+=(tfMRI_MOTOR_LR)
+TaskList+=(tfMRI_RELATIONAL_RL)
+TaskList+=(tfMRI_RELATIONAL_LR)
+TaskList+=(tfMRI_SOCIAL_RL)
+TaskList+=(tfMRI_SOCIAL_LR)
+TaskList+=(tfMRI_WM_RL)
+TaskList+=(tfMRI_WM_LR)
+
+for Session in $Sesslist ; do
+    echo $Session
+
+    for fMRIName in "${TaskList[@]}" ; do
+        echo "  ${fMRIName}"
+        
+        #The following options are defined in SetUpSecies.sh
+        #LowResMesh="32" #Needs to match what is in PostFreeSurfer, 32 is on average 2mm spacing between the vertices on the midthickness
+        #FinalfMRIResolution="2" #Needs to match what is in fMRIVolume, i.e. 2mm for 3T HCP data and 1.6mm for 7T HCP data
+        #SmoothingFWHM="2" #Recommended to be roughly the grayordinates spacing, i.e 2mm on HCP data 
+        #GrayordinatesResolution="2" #Needs to match what is in PostFreeSurfer. 2mm gives the HCP standard grayordinates space with 91282 grayordinates.  Can be different from the FinalfMRIResolution (e.g. in the case of HCP 7T data at 1.6mm)
+        RegName="MSMSulc" #MSMSulc is recommended, if binary is not available use FS (FreeSurfer)
+
+        if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+            echo "About to locally run ${HCPPIPEDIR}/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh"
+            queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        else
+            echo "About to use fsl_sub to queue ${HCPPIPEDIR}/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh"
+            queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
+        fi
+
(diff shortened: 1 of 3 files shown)


Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
```diff
@@ -0,0 +1,166 @@
...
+    exit 1
+fi
+
+# Requirements for this script
+#  installed versions of: FSL, Connectome Workbench (wb_command)
+#  environment: HCPPIPEDIR, FSLDIR, CARET7DIR
+
+#Set up pipeline environment variables and software
+source "$EnvironmentScript"
+source "$HCPPIPEDIR"/Examples/Scripts/SetUpSPECIES.sh --species="$SPECIES"
```
coalsont: As long as SetUpSPECIES.sh is using `newopts.shlib`, we need to add the workarounds for the differences in how sourcing behaves.

```suggestion
source "$HCPPIPEDIR"/Examples/Scripts/SetUpSPECIES.sh --species="$SPECIES"
#HACK: work around the log tool name hack in the sourced script
#since debug.shlib will be active by default, set the log toolname back to the Batch script
log_SetToolName "$(basename -- "$0")"
```

fMRISurface/scripts/SubcorticalProcessing.sh
```diff
@@ -73,6 +73,8 @@ log_Msg "SmoothingFWHM: ${SmoothingFWHM}"
 BrainOrdinatesResolution="$7"
 log_Msg "BrainOrdinatesResolution: ${BrainOrdinatesResolution}"
 
+Species="$8"
```
coalsont, line 76: For completeness, log the parameter:

```suggestion
Species="$8"
log_Msg "Species: ${Species}"
```

fMRISurface/scripts/SubcorticalProcessing.sh
```diff
@@ -99,7 +104,12 @@ fi
 
 log_Msg "Dilating out zeros"
 #dilate out any exact zeros in the input data, for instance if the brain mask is wrong. Note that the CIFTI space cannot contain zeros to produce a valid CIFTI file (dilation also occurs below).
-${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii COLUMN 0 30 ${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii
+if (( NonHumanSpecies )); then 
+    ${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii COLUMN 0 10 ${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii -nearest -merged-volume
```
coalsont: `-nearest` will result in less-smooth values (with more exact duplicate values, but is faster when dilation has to actually fill a large distance).  10 was an old value of the dilation distance, but might be okay for brains smaller than human anyway.

I think the main important change here is the addition of `-merged-volume`, which makes the dilation ignore the segmentation of structures and use more naive "voxels are just voxels" logic.  Perhaps somehow a mask managed to completely exclude a subcortical structure in some subjects, and this was the fix.

A possibly better way to deal with an all-zero structure is to do the ordinary dilation first to let it fill within-structure whatever it can, then run a `-merged-volume` dilation afterward to only deal with all-zero subcortical structures.

@takuya-hayashi @glasserm Thoughts?

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
```diff
@@ -0,0 +1,166 @@
...
+# exit 1;
+# }
+
+
+get_batch_options "$@"
+
+StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Session folders (named by SessionID)
+Sesslist="100307 100610" #Space delimited list of Session IDs
+EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script
+SPECIES="Human"
```
glasserm: For NHP batch I would default to something other than human.  Perhaps Macaque.

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
```diff
@@ -0,0 +1,166 @@
...
+
+        if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+            echo "About to locally run ${HCPPIPEDIR}/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh"
+            queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        else
+            echo "About to use fsl_sub to queue ${HCPPIPEDIR}/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh"
+            queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
+        fi
+
+        "${queuing_command[@]}" "$HCPPIPEDIR"/fMRISurface/GenericfMRISurfaceProcessingPipeline.sh \
```
glasserm: Include $SPECIES in the commandline?

fMRISurface/scripts/SubcorticalProcessing.sh
glasserm: We can hear from Takuya why he made the -merged-volume change.  10mm dilation was the initial human default until we found some cases where it was not enough.

fMRISurface/scripts/SubcorticalProcessing.sh
```diff
@@ -126,11 +136,18 @@ rm -f ${ResultsFolder}/${NameOffMRI}_temp_template.dlabel.nii
...
+if (( NonHumanSpecies )); then 
+    ${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_atlas.dtseries.nii COLUMN 0 10 ${ResultsFolder}/${NameOffMRI}_temp_atlas_dilate.dtseries.nii -nearest -merged-volume
+else
+    ${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_atlas.dtseries.nii COLUMN 0 30 ${ResultsFolder}/${NameOffMRI}_temp_atlas_dilate.dtseries.nii
+fi    
 rm -f ${ResultsFolder}/${NameOffMRI}_temp_atlas.dtseries.nii
 
 #write output volume, delete temporary
 #NOTE: $VolumefMRI contains a path in it, it is not a file in the current directory
+#TODO check NHP: NHP version doesn't have '_dilate' suffix in the command below - supposedly because no dilation is done. Resolve before merging.
```
glasserm: I don't understand this comment.

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
```diff
@@ -0,0 +1,166 @@
...
+            --path="$StudyFolder" \
+            --session="$Session" \
+            --fmriname="$fMRIName" \
+            --lowresmesh="$LowResMesh" \
+            --fmrires="$FinalfMRIResolution" \
```
glasserm: Some of these parameters probably have species specific defaults in SetSpecies.sh (the two above the comment and the two below it).

fMRISurface/scripts/SubcorticalProcessing.sh
glasserm: Tim will revise to always do human dilation and then do NHP dilation only for NHPs.

fMRISurface/scripts/SubcorticalProcessing.sh
glasserm: Misha will elaborate and make changes if needed.

fMRISurface/scripts/SubcorticalProcessing.sh
```diff
@@ -99,7 +104,12 @@ fi
 
 log_Msg "Dilating out zeros"
 #dilate out any exact zeros in the input data, for instance if the brain mask is wrong. Note that the CIFTI space cannot contain zeros to produce a valid CIFTI file (dilation also occurs below).
-${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii COLUMN 0 30 ${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii
+if (( NonHumanSpecies )); then 
+    ${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii COLUMN 0 10 ${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii -nearest -merged-volume
+else
+    ${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii COLUMN 0 30 ${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii
+fi
```
coalsont: Here is my suggested improvement to do within-structure dilation first, so that any data that doesn't need `-merged-volume` to get rid of the zeros doesn't bleed data across structures.  I am leaving in `-nearest` under the assumption that it may have been done for speed (and because weighted wouldn't be substantially closer to salvaging a masked-out structure anyway).

I imagine we don't want this 2-step process in human, because if the brain mask excludes an entire subcortical structure, the subject's data quality needs to be checked.

There is a second dilation section below with similar logic, which should presumably be updated the same way.
```suggestion
if (( NonHumanSpecies )); then
    "$CARET7DIR"/wb_command -cifti-dilate "${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii" COLUMN 0 10 "${ResultsFolder}/${NameOffMRI}_temp_subject_dilate_init.dtseries.nii"
    "$CARET7DIR"/wb_command -cifti-dilate "${ResultsFolder}/${NameOffMRI}_temp_subject_dilate_init.dtseries.nii" COLUMN 0 10 "${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii" -nearest -merged-volume
    rm -f "${ResultsFolder}/${NameOffMRI}_temp_subject_dilate_init.dtseries.nii"
else
    "$CARET7DIR"/wb_command -cifti-dilate "${ResultsFolder}/${NameOffMRI}_temp_subject.dtseries.nii" COLUMN 0 30 "${ResultsFolder}/${NameOffMRI}_temp_subject_dilate.dtseries.nii"
fi
```

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
mmilch01: Addressed in fd586e5c

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
mmilch01: Addressed in fd586e5c

fMRISurface/scripts/SubcorticalProcessing.sh
mmilch01: This is resolved after applying Tim's suggestions in ca41bd2. Comment removed in fd586e5c.

fMRISurface/scripts/SubcorticalProcessing.sh
mmilch01: Addressed in ca41bd2

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
```diff
@@ -0,0 +1,171 @@
...
+# echo "       -d                   : dry run (print commands in the terminal but not run)" 
+# echo ""
+# exit 1;
+# }
+
+
+get_batch_options "$@"
+
+StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Session folders (named by SessionID)
+Sesslist="100307 100610" #Space delimited list of Session IDs
```
coalsont: Maybe we shouldn't put obviously human subject IDs into an NHP example launcher configured for Macaque.  I'm not sure what example subject identifiers to use, though.

fMRISurface/scripts/SubcorticalProcessing.sh
```diff
@@ -126,7 +139,12 @@ rm -f ${ResultsFolder}/${NameOffMRI}_temp_template.dlabel.nii
 #the standard space output cifti must not contain zeros (or correlation, ICA, variance normalization, etc will break), so dilate in case freesurfer was unable to segment something (may only be applicable for bad quality structurals)
 #NOTE: wb_command v1.4.0 and later should only output exact 0s past the edge of predilate, so this works as desired
 #earlier verions of wb_command may produce undesired results in the subjects that need this dilation
-${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_atlas.dtseries.nii COLUMN 0 30 ${ResultsFolder}/${NameOffMRI}_temp_atlas_dilate.dtseries.nii
+
+if (( NonHumanSpecies )); then 
+    ${CARET7DIR}/wb_command -cifti-dilate ${ResultsFolder}/${NameOffMRI}_temp_atlas.dtseries.nii COLUMN 0 10 ${ResultsFolder}/${NameOffMRI}_temp_atlas_dilate.dtseries.nii -nearest -merged-volume
+else
```
coalsont, line 147: Maybe this should match the other dilation section above.

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
glasserm: Agree

fMRISurface/scripts/SubcorticalProcessing.sh
glasserm: Agree

mmilch01: Addressed in 78613ea.

Examples/Scripts/GenericfMRISurfaceProcessingPipelineBatchNHP.sh
mmilch01: Addressed in 78613ea.

glasserm (approved):
Testing passed.

PR #376, opened 2026-01-04
Enable different set of bolds for each session in TransmitLong

sivcek:
Currently TransmitBiasLong assumes that all the sessions have the exact same set of fMRI images. This often is not the case in longitudinal datasets. This PR enables specifying different set of fMRI images for each session while keeping backward compatibility with old behavior. If a user desires to specify different set of fMRI images for each session, they list fMRI names for each session in the `--pt-fmri-names` argument, separating sessions with `%` character. If no `%` character is present, the same set of fMRI names is used for all sessions. If `%` character is present, the code checks whether the number of fMRI sets match the number of sessions, and uses the sets in the same order as the sessions are listed.

In the example:
```
--sessions="session1@session2@session2" \
--pt-fmri-names="rest@taskA@taskB%rest@taskA%rest@taskB"
```
the pairings are:

- `session1` – `rest`, `taskA`, and `taskB`, 
- `session2` – `rest`, `taskA`
- `session3` – `rest`, `taskB`


glasserm (approved):
This looks fine to me.

TransmitBias/TransmitBiasLong.sh
```diff
@@ -43,7 +43,7 @@ opts_AddOptional '--afi-angle' 'AFITargFlipAngle' 'number' "target flip angle of
 opts_AddOptional '--b1tx-phase-divisor' 'B1TxDiv' 'number' "what to divide the phase map by to obtain proportion of intended flip angle, default 800" '800'
 
 #PseudoTransmit-specific. Should be the same as in the corresponding cross-sectional calls.
-opts_AddOptional '--pt-fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "fmri runs to use SE/SBRef files from, separated by @"
+opts_AddOptional '--pt-fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "fmri runs to use SE/SBRef files from, separated by @. For per-session lists, separate session-specific sets using % (e.g., run1@run2%run2%run1@run3)."
```
mharms:
```suggestion
opts_AddOptional '--pt-fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "fmri runs to use SE/SBRef files from, separated by @. For per-session lists, separate session-specific sets using % (e.g., run1@run2%run2%run1@run3 to use run1 and run2 from session1, run2 from session2, and run1 and run3 from session3)."
```

mharms: Suggested a bit more wording to make the example very explicit.

glasserm:
@sivcek @coalsont @mharms can we get this merged?

mharms:
Looks like we are waiting for @sivcek to respond to my earlier suggestion, and for @coalsont to chime in...

coalsont:
Looks reasonable to me.

PR #378, opened 2026-01-07
PFMs

bqrosen:
Pipeline for PROFUMO decomposition. I started from Andrea's Yang's PFMs code, and finished generalizing it in the HCP Pipelines style. I tested in on Mac25Rhesus and Mac25Cyno, and it is working there. 

Known issues:
(1) This PR does not include the ./PFM/profumo_v2.sif singularity container, because it is 6.6G. 
(2) The full run example script uses Mac25Rhesus rather than something more standard like HCP1071 and has local paths in it.
(3) The matlab needs to be compiled


PFM/scripts/PFMNotes.sh
```diff
@@ -0,0 +1,122 @@
...
+            then
+                matlab_args+=", "
+            fi
+            matlab_args+="'$thisarg'"
+        done
+        matlabcode="
+            addpath('$HCPPIPEDIR/global/matlab');
+            addpath('$this_script_dir');
+            addpath('$HCPCIFTIRWDIR');
+            PFMNotesGroup($matlab_args);"
```
coalsont: Is this supposed to be `PFMNotes(...);`?

PFM/scripts/PFMNotes.m
```diff
@@ -0,0 +1,99 @@
+function PFMNotesGroup(StudyFolder, SubjListRaw, GroupAverageName, OutputfMRIName, PFMdimStr, OutputPrefix, RegString, LowResMesh, RunsXNumTimePointsStr, CIFTIVerticesStr, CIFTIVolumeStr, PFMFolder)
+    
+    % Parse string inputs
+    Subjlist = strsplit(SubjListRaw, '@');
+    PFMdim = str2double(PFMdimStr);
+    RunsXNumTimePoints = str2double(RunsXNumTimePointsStr);
+    CIFTI = str2double(CIFTIVerticesStr);
+    CIFTIVol = str2double(CIFTIVolumeStr);
```
coalsont: These don't appear to get used.  Hopefully we can remove these magic numbers as inputs to the pipeline, also, as it seems like they don't get passed from there to anything anyway.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,275 @@
...
+    done
+    perl -pi -e 'if (eof) { s/,$// }' $ProfumoConfig  # remove trailing comma
+    echo "}" >> $ProfumoConfig
+
+    # PFM pipeline execution
+    echo "Starting PFM postprocessing pipeline..."
+    echo "Data type: ${OutputfMRIName}"
+    echo "PFM dimension: ${PFMdim}"
+
+    "${queuing_command[@]}" "$HCPPIPEDIR"/PFM/PFMPipeline.sh \
```
coalsont: This file looks like it should be in `Examples/Scripts/`

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,307 @@
...
+opts_AddMandatory '--group-average-name' 'GroupAverageName' 'string' 'name to use for the group output folder'
+opts_AddMandatory '--pfm-dimension' 'PFMdim' 'integer' "PFM dimensionality (e.g., 76, 92, 65)"
+opts_AddMandatory '--pfm-folder' 'PFMFolder' 'path' "path to PFM results folder containing Results.ppp"
+opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files"
+opts_AddMandatory '--profumo-config' 'ProfumoConfig' 'path' "path to PROFUMO JSON configuration file"
+opts_AddMandatory '--profumo-tr' 'TR' "repetition time for PROFUMO analysis" '0.72'
+opts_AddMandatory '--ref-image' 'RefImage' 'path' "reference image for PROFUMO postprocessing"
+opts_AddMandatory '--runs-timepoints' 'RunsXNumTimePoints' "total timepoints across runs (e.g., 4800 for rest, 3880 for task)" '4800'
+opts_AddMandatory '--cifti-vertices' 'CIFTIVertices' "number of CIFTI vertices" '91282'
+opts_AddMandatory '--cifti-volume' 'CIFTIVolume' "number of CIFTI volume vertices" '283226'
```
coalsont: These appear to be unused.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,307 @@
...
+opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to use for PFM pipeline outputs"
+opts_AddMandatory '--proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing already done, e.g. '_Atlas_MSMAll_hp2000_clean_rclean_tclean'"
+opts_AddMandatory '--group-average-name' 'GroupAverageName' 'string' 'name to use for the group output folder'
+opts_AddMandatory '--pfm-dimension' 'PFMdim' 'integer' "PFM dimensionality (e.g., 76, 92, 65)"
+opts_AddMandatory '--pfm-folder' 'PFMFolder' 'path' "path to PFM results folder containing Results.ppp"
+opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files"
+opts_AddMandatory '--profumo-config' 'ProfumoConfig' 'path' "path to PROFUMO JSON configuration file"
+opts_AddMandatory '--profumo-tr' 'TR' "repetition time for PROFUMO analysis" '0.72'
+opts_AddMandatory '--ref-image' 'RefImage' 'path' "reference image for PROFUMO postprocessing"
+opts_AddMandatory '--runs-timepoints' 'RunsXNumTimePoints' "total timepoints across runs (e.g., 4800 for rest, 3880 for task)" '4800'
```
coalsont: Mandatory parameters can't have default values (that is effectively the definition of mandatory).  A few more of these are below, too.

We probably shouldn't bake highly HCP_YA-specific values into any pipeline's default values regardless (really, just about any scanner setting shouldn't have a default value in any pipeline).  The example launcher can have specific values from some HCP project, but not the pipeline itself.
```suggestion
opts_AddMandatory '--profumo-tr' 'TR' "repetition time for PROFUMO analysis"
opts_AddMandatory '--ref-image' 'RefImage' 'path' "reference image for PROFUMO postprocessing"
opts_AddMandatory '--runs-timepoints' 'RunsXNumTimePoints' "total timepoints across runs (e.g., 4800 for rest, 3880 for task)"
```

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,275 @@
+#!/bin/bash
+
+set -eu
+
+# This is an example script to run the full PFM postprocessing pipeline
```
glasserm, line 5: Probably this should be called PostPFMPipeline.sh if it doesn't actually run PFMs.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,275 @@
...
+    RefImage="${StudyFolder}/$GroupAverageName/MNINonLinear/Results/Mac25Rhesus_v5_BOLD_REST_CONCAT_MIGP/Mac25Rhesus_v5_BOLD_REST_CONCAT_MIGP_Atlas_hppd2_clean_meanvn.dscalar.nii"
+
+    # set the file name component representing the preprocessing already done
+    fMRIProcSTRING="hp${HighPass}_clean"
+
+    # set the mesh resolution, like '32' for 32k_fs_LR
+    LowResMesh="10"
+
+    # RSN regression settings
+    RSNMethod="dual"  # dual or single (default: dual)
```
glasserm: For PFMs, dual is the only available option.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,275 @@
...
+# This is an example script to run the full PFM postprocessing pipeline
+# Steps involved:
+# 1. ImportPFMNotes - Import PFM results and create time courses, spectra, and maps
+# 2. RSNRegression - Run RSN regression on PFM data for dual regression
+# 3. PFMNotesGroup - Generate group-level statistics and averages
+#
+# Please make sure that PROFUMO, ICA-FIX, MSMAll and MakeAverageDataset are done properly 
+# matching the input arguments before running this PFM pipeline
+
+# Global default values
```
glasserm: Honestly I prefer the way all of the non-sICA+FIX pipelines are organized without using these DEFAULT values.  I have never understood why those pipelines are different from the rest of the HCP Pipelines.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,275 @@
...
+    CIFTIVertices="27559" # of rows in cifti
+    CIFTIVolume="66480"
+
+    # Volume template file
+    VolumeTemplateCIFTI="${StudyFolder}/$GroupAverageName/MNINonLinear/Results/Mac25Rhesus_v5_BOLD_REST_CONCAT_WF8/tICA_d46/sICA_VolMaps_46.dscalar.nii"
+
+    # PROFUMO settings
+    ProfumoSingularity="/media/myelin/burke/projects/Mac25Rhesus/HCPpipelines/PFM/profumo_v2.sif" 
+    ProfumoConfig="${PFMFolder}/dataLocations.json"  
+    TR="1.0"
```
glasserm: Was that actually the TR for the macaque data?  I thought it was less than 1s.

Examples/Scripts/RunPFMpipelineFullRun.sh
glasserm: Perhaps this code can be used to run weighted as well?  I think that could be done with RSNRegression alone too.

Examples/Scripts/RunPFMpipelineFullRun.sh
glasserm: Alternatively, the pipeline could include the code to build the JSON for PFMs and actually run PFMs as well.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,307 @@
...
+                "${ProfumoSingularity}" \
+                /opt/fsl/fslpython/envs/profumo/bin/python3 /opt/profumo/Python/postprocess_results.py \
+                --web-report \
+                "${PFM_PATH}" \
+                "${RESULTS_PATH}" \
+                "${REAL_REF_IMAGE}"
+            ;;
+        (ImportPFMNotes)
+            log_Msg "Running ImportPFMNotes step"
+            "$HCPPIPEDIR"/PFM/scripts/ImportPFMNotes.sh \
```
glasserm: Notes are proto-pipelines.  We should discard the "Notes" tag once something is formally a Pipeline module.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,307 @@
...
+                done
+                
+                if [[ "$fMRINamesForSub" == "" ]]
+                then
+                    log_Warn "No valid fMRI runs found for subject $Subject, skipping"
+                    continue
+                fi
+                
+                # Set maps and timeseries based on method (following your example exactly)
+                if [[ "$RSNMethod" == "single" ]]
```
glasserm: Inconsistency between whether the options are dual and single or weighted.  I believe only dual and weighted end up being useful. Single may have been there for testing purposes and was found to not work as intended.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,307 @@
...
+                    --fix-legacy-bias="$FixLegacyBias"
+                    --scale-factor="$ScaleFactor"
+                )
+                
+                # Add conditional arguments based on method
+                if [[ "$RSNMethod" == "dual" ]]
+                then
+                    rsn_cmd+=(--group-maps="$GroupMaps")
+                fi
+                if [[ "$RSNMethod" == "single" ]]
```
glasserm: See above about single.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,307 @@
...
+                # Queue parallel job
+                par_addjob "${rsn_cmd[@]}"
+            done
+            
+            # Run the jobs
+            par_runjobs "$parLimit"
+            ;;
+        (PFMNotesGroup)
+            log_Msg "Running PFMNotesGroup step"
+            "$HCPPIPEDIR"/PFM/scripts/PFMNotesGroup.sh \
```
glasserm: See above about Notes.

PFM/scripts/ImportPFMNotes.m
```diff
@@ -0,0 +1,72 @@
...
+            % PFMSpectra.diminfo{1,2} = sICASpectra.diminfo{1,2};
+
+            ciftisave(PFMTCSorig, [StudyFolder '/' Subjlist{s} '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' Subjlist{s} '.' OutputPrefix RegString '_ts_orig.' LowResMesh 'k_fs_LR.sdseries.nii'], wbcommand);
+            ciftisave(PFMSpectraorig, [StudyFolder '/' Subjlist{s} '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' Subjlist{s} '.' OutputPrefix RegString '_spectra_orig.' LowResMesh 'k_fs_LR.sdseries.nii'], wbcommand);
+
+            ciftisave(PFMTCS, [StudyFolder '/' Subjlist{s} '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' Subjlist{s} '.' OutputPrefix RegString '_ts.' LowResMesh 'k_fs_LR.sdseries.nii'], wbcommand);
+            ciftisave(PFMSpectra, [StudyFolder '/' Subjlist{s} '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' Subjlist{s} '.' OutputPrefix RegString '_spectra.' LowResMesh 'k_fs_LR.sdseries.nii'], wbcommand);
+
+            unix(['cp ' PFMFolder '/Results.ppp/Maps/sub-' Subjlist{s} '.dscalar.nii ' StudyFolder '/' Subjlist{s} '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' Subjlist{s} '.' OutputPrefix RegString '_origmaps.' LowResMesh 'k_fs_LR.dscalar.nii']);
```
glasserm: Does this need to be converted to system?

PFM/scripts/PostPROFUMO.sh
```diff
@@ -0,0 +1,122 @@
...
+opts_AddMandatory '--fmri-names' 'fMRIListRaw' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' 'list of fmri run names separated by @s'
+opts_AddMandatory '--proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing"
+opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to use for PFM pipeline outputs"
+opts_AddMandatory '--output-string' 'OutputSTRING' 'string' "output string identifier"
+opts_AddMandatory '--output-prefix' 'OutputPrefix' 'string' "output prefix for files"
+opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "the registration string"
+opts_AddMandatory '--low-res-mesh' 'LowResMesh' 'string' "mesh resolution"
+opts_AddMandatory '--pfm-folder' 'PFMFolder' 'path' "path to PFM results folder"
+opts_AddMandatory '--concat-name' 'ConcatName' 'string' "concatenated fMRI name if using multi-run data"
+opts_AddMandatory '--profumo-tr' 'TR' "repetition time for PROFUMO analysis" '0.72'
```
glasserm, line 30: See Tim's concern elsewhere about this.

PFM/scripts/PFMNotes.m
```diff
@@ -0,0 +1,99 @@
+function PFMNotesGroup(StudyFolder, SubjListRaw, GroupAverageName, OutputfMRIName, PFMdimStr, OutputPrefix, RegString, LowResMesh, RunsXNumTimePointsStr, CIFTIVerticesStr, CIFTIVolumeStr, PFMFolder)
```
glasserm: Need to clarify PFMNotes vs PFMNotesGroup.

glasserm (changes requested):
Please see my above comments.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,317 @@
...
+opts_AddMandatory '--subject-list' 'SubjlistRaw' '100206@100307...' "list of subject IDs separated by @s"
+opts_AddMandatory '--fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of fmri run names separated by @s"
+opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to use for PFM pipeline outputs"
+opts_AddMandatory '--proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing already done, e.g. '_Atlas_MSMAll_hp2000_clean_rclean_tclean'"
+opts_AddMandatory '--group-average-name' 'GroupAverageName' 'string' 'name to use for the group output folder'
+opts_AddMandatory '--pfm-dimension' 'PFMdim' 'integer' "PFM dimensionality (e.g., 76, 92, 65)"
+opts_AddMandatory '--pfm-folder' 'PFMFolder' 'path' "path to PFM results folder containing Results.ppp"
+opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files"
+opts_AddMandatory '--profumo-config' 'ProfumoConfig' 'path' "path to PROFUMO JSON configuration file"
+opts_AddMandatory '--profumo-tr' 'TR' "repetition time for PROFUMO analysis" "repetition time for PROFUMO analysis in seconds"
```
coalsont: I think this would look better:
```
   --profumo-tr=<seconds> - repetition time for PROFUMO analysis in seconds
```
than this:
```
   --profumo-tr=<repetition time for PROFUMO analysis> - repetition time for PROFUMO analysis in seconds
```
```suggestion
opts_AddMandatory '--profumo-tr' 'TR' "seconds" "repetition time for PROFUMO analysis in seconds"
```

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,263 @@
...
+
+    # initialize variables
+    StudyFolder="/media/myelin/brainmappers/BICAN/Macaque/MacaqueRhesus"
+    Subjlist="$(find ${StudyFolder} -maxdepth 1 -type d -name "A*" -exec basename {} \; | paste -sd@ -)"
+    EnvironmentScript="/media/myelin/burke/projects/Mac25Rhesus/scripts/Mac25Rhesus_v5_SetUpHCPPipeline.sh"
+    GroupAverageName="Mac25Rhesus_v5"
+    RegName=""
+    MatlabMode=1
+    RunLocal=0
+    QUEUE="matlabparallelhigh.q@brainmappers-desktop3"
```
coalsont: For your testing purposes, please make a copy of the batch rather than pushing your settings (and paths), even to a branch.

I'm not sure we have any other batch scripts that submit a job that is expected to go highly parallel on each job (the non-longitudinal tICA batch examples are configured to run locally).

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,263 @@
...
+    # set the start step beginning from RunPROFUMO which is by default the first step
+    # StartStep="RunPROFUMO"
+    # StopStep="RunPROFUMO"  
+    # StartStep="PostPROFUMO"
+    # StopStep="PostPROFUMO"  
+    # StartStep="PostPROFUMO"
+    # StartStep="RSNRegression"  
+    # StopStep="RSNRegression"
+    # StartStep="RunPROFUMO"
+    # StopStep=""
```
coalsont: I think these lines may confuse users, but a list or explanation of the possibilities might be good.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,263 @@
...
+
+    StartStep="PostPROFUMO"
+    StopStep="GroupPFMs"
+
+    # set how many subjects to do in parallel (local, not cluster-distributed) during RSN regression, defaults to all detected physical cores, '-1'
+    parLimit=5
```
coalsont: If `-1` isn't expected to work for most users, explain what the operative constraint is (memory?) so they can take a guess for their own data+hardware.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,293 @@
...
+            apptainer exec --bind $(dirname "${StudyFolder}") \
+                --env PROFUMODIR=/opt/profumo \
+                "${ProfumoSingularity}" \
+                /opt/fsl/fslpython/envs/profumo/bin/python3 /opt/profumo/Python/postprocess_results.py \
+                --web-report \
+                "${PFM_PATH}" \
+                "${RESULTS_PATH}" \
+                "${REAL_REF_IMAGE}"
+            ;;
+        (PostPROFUMO)
```
coalsont: Are there reasons users would want to rerun `PostPROFUMO` without rerunning `RunPROFUMO`, or is this separation just for debugging `PostPROFUMO`?

bqrosen: The RunPROFUMO step can take a long time, while PostPROFUMO is very fast. If a user makes a mistake in setting up the PostPROFUMO parameters (e.g., paths), it's nice to be able to rerun it separately.

Examples/Scripts/RunPFMpipelineFullRun.sh
bqrosen: Yeah the batch script needs to be still needs to be overhauled.. It should probably use  HCP_YA task data like fMRIStats does.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,263 @@
...
+                                    --output-string="$OutputSTRING" \
+                                    --proc-string="$fMRIProcSTRING" \
+                                    --group-average-name="$GroupAverageName" \
+                                    --pfm-dimension="$PFMdim" \
+                                    --pfm-folder="$PFMFolder" \
+                                    --surf-reg-name="$RegName" \
+                                    --concat-name="$ConcatName" \
+                                    --low-res-mesh="$LowResMesh" \
+                                    --runs-timepoints="$subjectExpectedTimepoints" \
+                                    --low-dims="$LowDims" \
```
coalsont: This looks left over (and its variable above).

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,345 @@
...
+            cat "${ProfumoConfig}" | while IFS= read -r line; do
+                # Only process lines that contain .nii" (file paths)
+                if [[ "$line" != *'.nii"'* ]]; then continue;fi
+                
+                # Extract the file path from JSON value (text between colons and quotes)
+                # Pattern: "RUNNAME": "/path/to/file.nii"
+                filePath=$(echo "$line" | sed -E 's/^[[:space:]]*"[^"]*"[[:space:]]*:[[:space:]]*"([^"]*)".*/\1/')
+                
+                # Check CIFTI version 
+                ciftiVersion=$(wb_command -nifti-information "$filePath" -print-xml -version 2 2>/dev/null | grep -oP 'CIFTI.*Version="\K[0-9]' || echo "")
```
coalsont: `-version 2` will cause it to never say "Version=1":

https://github.com/Washington-University/workbench/blob/051afca921c91af65e1c88f882a739c7ea7374a7/src/Cifti/CiftiXML.cxx#L755

It is unfortunate that an implementation detail in Profumo is being worked around through the oddity of an obsolete format version (I suspect the cifti version may be a red herring, and the difference is simply "written by wb_command" or not).  The data offset alignment of newly-written files should be the same regardless of XML version (though the header dimensions are backwards in cifti-1), though for a while cifti-matlab used 8-byte alignment instead of 16 (due to a misplaced `+8`, it was never 16-aligned).  I just checked, and the pipelines haven't updated `cifti-matlab` to include the 16-byte fix.

PFM/scripts/GroupPFMs.sh
```diff
@@ -0,0 +1,114 @@
...
+
+Usage: $log_ToolName PARAMETER...
+
+PARAMETERs are [ ] = optional; < > = user supplied value
+"
+    #automatic argument descriptions
+    opts_ShowArguments
+    
+    #do not use exit, the parsing code takes care of it
+}
```
coalsont: Use the newer usage style (please copy from the top of `newopts.shlib` when making new scripts):
```suggestion
opts_SetScriptDescription "Import PFM notes and create time courses, spectra, and maps"
```

PFM/scripts/PostPROFUMO.sh
```diff
@@ -0,0 +1,121 @@
...
+
+Usage: $log_ToolName PARAMETER...
+
+PARAMETERs are [ ] = optional; < > = user supplied value
+"
+    #automatic argument descriptions
+    opts_ShowArguments
+    
+    #do not use exit, the parsing code takes care of it
+}
```
coalsont:
```suggestion
opts_SetScriptDescription "Import PFM notes and create time courses, spectra, and maps"
```

PFM/scripts/GroupPFMs.sh
coalsont: I also suggest updating this style in whatever you copied it from (whether it is in the repo or not), so it doesn't get copied again.

PFM/PFMPipeline.sh
coalsont: ~Note, the pipelines `cifti-matlab` also contains an added hack to allow "hippocampus dentate", so it isn't simply a drop-in replacement.~ scratch that, it hasn't been committed to master.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,345 @@
...
+do
+    stepName="${pipelineSteps[stepInd]}"
+    case "$stepName" in
+        (RunPROFUMO)
+            log_Msg "Running PROFUMO analysis step"
+            
+            # Validate required PROFUMO parameters
+            if [[ "$ProfumoConfig" == "" ]]
+            then
+                log_Err_Abort "PROFUMO config file must be specified with --profumo-config"
```
coalsont, line 148: It would appear that the user has to write the config file, which must contain the list of subjects (among other things), since it doesn't appear to get passed to profumo in another way.  How complicated (or version-dependent) is the config format, and should we consider having the pipeline script generate it?

PFM/PFMPipeline.sh
coalsont: I have updated master with a new release of cifti-matlab that fixes the alignment (and should also support hippocampus dentate).  Of course, this won't "fix" existing files written with previous pipeline versions to make SIMD on `mmap`ed files need less care in coding (in my opinion, `mmap` is more trouble than it's worth, and needs another workaround for things like int16-with-scl_slope cifti anyway).

Instead of converting to cifti-1, I recommend just rewriting the file unconditionally (as cifti-2) and see if the SIMD stuff works (`-file-convert -cifti-version-convert` seems like the best choice for a nothing-but-rewrite command, since -cifti-math now resets label to scalar).

PFM/PFMPipeline.sh
bqrosen: It's written by the batch script, which I think offers a little more flexibility for the user than putting it directly in the pipeline script. but I can see the argument for moving it into the pipeline script.

glasserm:
@bqrosen @coalsont What is pending on this PR?

bqrosen:
Andrea has a PR on to PFMs branch with some outstanding comments
https://github.com/Washington-University/HCPpipelines/pull/392

Also, maybe we should wait for the Wishart filtering to be implemented before merging the branch (to master). Should I close PR 378 until the PFMs branch is in a more ready state and then start a new one later,  leave PR 378 hanging and update it with newer modifications to the PFMs branch as we go, or merge PR 378 and then submit a second PR down the road with more changes?

-Burke

coalsont:
It doesn't yet support compiled matlab mode, I'd rather not merge it until that is ready.  Is Andrea available to finish the edits in the other PR, or do we need a different path forward?

glasserm:
That's fine.  We can add it here.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,264 @@
...
+
+    # RSN regression settings
+    FixLegacyBiasString="NO"
+    ScaleFactor="0.01"
+
+    # Volume template file
+    VolumeTemplateCIFTI="${HOME}/data/HCPpipelines_ExampleData/${GroupAverageName}/MNINonLinear/${GroupAverageName}_CIFTIVolumeTemplate_${OutputfMRIName}.${fMRIResolution}.dscalar.nii"
+ 
+    # PROFUMO settings
+    ProfumoSingularity="$HCPPIPEDIR/PFM/profumo_v2.sif"
```
coalsont: I guess we still need to decide how to instruct users to obtain this.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,455 @@
...
+opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "the registration string corresponding to the input files"
+opts_AddMandatory '--profumo-config' 'ProfumoConfig' 'path' "path to PROFUMO JSON configuration file"
+opts_AddMandatory '--profumo-tr' 'TR' "seconds" "repetition time for PROFUMO analysis"
+opts_AddMandatory '--ref-image' 'RefImage' 'path' "reference image for PROFUMO postprocessing"
+opts_AddMandatory '--runs-timepoints' 'RunsXNumTimePoints' "total timepoints across runs" "total timepoints across runs"
+opts_AddMandatory '--concat-name' 'ConcatName' "concatenated fMRI name if using multi-run data" ''
+opts_AddMandatory '--volume-template-file' 'VolumeTemplateFile' "volume template file path" ''
+
+#PROFUMO specific parameters
+opts_AddOptional '--profumo-threads' 'ProfumoThreads' 'integer' "number of threads for PROFUMO" '25'
```
coalsont: Should this be inferred from `--parallel-limit` instead of being an option?  Note, we have functions for physical and logical core count, to handle the "-1" (automatic) case:

https://github.com/Washington-University/HCPpipelines/blob/a4d3ebd6b8ffc10ccf8cc27df3153b85d8873345/global/scripts/parallel.shlib#L132

PFM/scripts/ApplyWFProfumo.m
```diff
@@ -0,0 +1,26 @@
...
+    outList = strsplit(outputFiles, ',');
+
+    allData = [];
+    tpCounts = [];
+    runMeans = {};
+    for i = 1:length(inList)
+        cii = ciftiopen(strtrim(inList{i}), 'wb_command');
+        tpCounts(i) = size(cii.cdata, 2);
+        runMeans{i} = mean(cii.cdata, 2);
+        allData = [allData (cii.cdata - runMeans{i})];
```
coalsont: Please note the comments @glasserm left here:

https://github.com/Washington-University/HCPpipelines/pull/392#pullrequestreview-4667792222

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,455 @@
...
+                --multiStartIterations "${MultiStartIterations}" ${LoadSequentiallyArg} ${InitialMapsArg}
+            apptainer exec --bind $(dirname "${StudyFolder}") \
+                --env PROFUMODIR=/opt/profumo \
+                --env PYTHONNOUSERSITE=1 \
+                "${ProfumoSingularity}" \
+                /opt/profumo/C++/PROFUMO "${ProfumoConfigToUse}" \
+                "${PFMdim}" "${PFM_PATH}" \
+                --useHRF "${TR}" --covModel "${CovModel}" --dofCorrection "${DOFCorrection}" \
+                --nThreads "${ProfumoThreads}" --lowRankData "${LowRankData}" --randomSeed "${RandomSeed}" \
+                --multiStartIterations "${MultiStartIterations}" ${LoadSequentiallyArg} ${InitialMapsArg}
```
coalsont: Don't keep two copies of the same command in the code:

EDIT: in fact, I initially missed "--randomSeed" because it wasn't in the echo statement.

```suggestion
            cmd=(apptainer exec --bind $(dirname "${StudyFolder}") \
                --env PROFUMODIR=/opt/profumo \
                --env PYTHONNOUSERSITE=1 \
                "${ProfumoSingularity}" \
                /opt/profumo/C++/PROFUMO "${ProfumoConfigToUse}" \
                "${PFMdim}" "${PFM_PATH}" \
                --useHRF "${TR}" --covModel "${CovModel}" --dofCorrection "${DOFCorrection}" \
                --nThreads "${ProfumoThreads}" --lowRankData "${LowRankData}" --randomSeed "${RandomSeed}" \
                --multiStartIterations "${MultiStartIterations}" ${LoadSequentiallyArg} ${InitialMapsArg})
            log_Msg "running command: ${cmd[*]}"
            "${cmd[@]}"
```

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,455 @@
...
+            PFM_PATH="${PFMFolder}/Analysis.pfm"
+            RESULTS_PATH="${PFMFolder}/Results.ppp"
+            REAL_REF_IMAGE=$(readlink -f "${RefImage}")
+
+            # Remove any existing Results.ppp directory (and ++ variants created by re-runs)
+            # so postprocess_results.py writes fresh output to Results.ppp
+            if [[ -d "${PFMFolder}/Results.ppp" ]]
+            then
+                log_Warn "Results.ppp folder ${PFMFolder}/Results.ppp(+) already exists, clearing before postprocessing"
+                rm -rf "${PFMFolder}"/Results.ppp* 2>/dev/null || true # ignore errors due to nfs silly renamed files, or similar
```
coalsont: Why do we need to remove everything matching `Results.ppp*`?  What if the user has done a `cp -r Results.ppp Results.ppp-save-finally-success`?

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,455 @@
...
+                "${REAL_REF_IMAGE}"
+            apptainer exec --bind $(dirname "${StudyFolder}") \
+                --env PROFUMODIR=/opt/profumo \
+                --env PYTHONNOUSERSITE=1 \
+                "${ProfumoSingularity}" \
+                /opt/fsl/fslpython/envs/profumo/bin/python3 /opt/profumo/Python/postprocess_results.py \
+                --web-report \
+                "${PFM_PATH}" \
+                "${RESULTS_PATH}" \
+                "${REAL_REF_IMAGE}"
```
coalsont:
```suggestion
            cmd=(apptainer exec --bind $(dirname "${StudyFolder}") \
                --env PROFUMODIR=/opt/profumo \
                --env PYTHONNOUSERSITE=1 \
                "${ProfumoSingularity}" \
                /opt/fsl/fslpython/envs/profumo/bin/python3 /opt/profumo/Python/postprocess_results.py \
                --web-report \
                "${PFM_PATH}" \
                "${RESULTS_PATH}" \
                "${REAL_REF_IMAGE}")
            log_Msg "Running command: ${cmd[*]}"
            "${cmd[@]}"
```

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
```diff
@@ -55,29 +55,16 @@ source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
...
-    
-    #do not use exit, the parsing code takes care of it
-}
+# -------------------------------------------
+#  Usage Description
+# -------------------------------------------
 
-# ------------------------------------------------------------------------------
+#description of this script to use in usage
+opts_SetScriptDescription "Run TaskfMRIAnalysis pipeline for a subject. Pipeline will run Level1 (scan-level) analyses, and Level2 (single subject-level) analysis as specified."
```
coalsont, line 63: These edits are not exactly related to PFMs...but I guess they may reduce the copy+paste propagation of the older usage style.

tICA/scripts/tICACleanData.sh
```diff
@@ -13,20 +13,8 @@ source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
...
-
-Usage: $log_ToolName PARAMETER...
-
-PARAMETERs are [ ] = optional; < > = user supplied value
-"
-    #automatic argument descriptions
-    opts_ShowArguments
-}
+#description of this script to use in usage
+opts_SetScriptDescription "regresses noise group temporal ICA components out of CIFTI and optionaly volume timeseries data and optionally correct the bias legacy field"
```
coalsont: If we are editing this line, might as well fix this...

```suggestion
opts_SetScriptDescription "regresses noise group temporal ICA components out of CIFTI and optionally volume timeseries data and optionally correct the bias legacy field"
```

TaskfMRIAnalysis/scripts/makeSubjectTaskSummary.sh
bqrosen: Indeed, see https://github.com/Washington-University/HCPpipelines/pull/378#discussion_r2850095140

PFM/PFMPipeline.sh
bqrosen: For the test data, RSNRegression uses significantly more memory per thread than RunPROFUMO. I suspect that for most users --profumo-threads should be the number of physical cores and --parallel-limit will be a 5 or fewer. I'll change the default for --profumo-threads to -1 and handle it with par_numphys().

PFM/PFMPipeline.sh
bqrosen: If there is exists Results.ppp<*> then PROFUMO will write its results to Results.ppp<*>+ and downstream processing won't know where to find them. If you prefer, instead of removing  Results.ppp* we could do a log_Error and crash out, with a message asking the user to remove them manually.

Examples/Scripts/RunPFMpipelineFullRun.sh
bqrosen: I added it to balsashare

PFM/PFMPipeline.sh
coalsont: So to be clear, if "Results.ppp" does not exist, but "Results.ppp+" or "Results.ppp-save" exist, profumo *won't* write its results to "Results.ppp"?  That is surprising.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,268 @@
...
+
+#
+# Function Description
+#	Main processing of this script
+#
+#	Gets user specified command line options and runs PFM postprocessing pipeline 
+#   (please make sure the PROFUMO, ICA-FIX, MSMAll and MakeAverageDataset are finished before running this script)
+#
+
+main() {
```
coalsont: It is generally more straightforward in bash to not put the main stuff in a function (it is only "called" once, immediately after it is defined).

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,268 @@
...
+    DOFCorrection="0.5"
+    CovModel="Subject"
+    nStarts="5" # number of multi-start iterations for PROFUMO
+    RandomSeed="$randSeed" # random seed for PROFUMO reproducibility
+    # RefImage will be auto-set based on data type below
+
+    # Download the PROFUMO Singularity image from the following link and place it in the PFM folder, 
+    # or change the path below to point to your own copy of the image
+    # https://balsa.wustl.edu/myelin/download?dirName=public&filepath=profumo_v2.sif
+    ProfumoSingularity="$HCPPIPEDIR/PFM/profumo_v2.sif"
```
coalsont: Since this is basically the first thing a new profumo user will need to do, I suggest putting it near the top of the settings the user is expected to read/edit.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,268 @@
...
+    else
+      echo "queueing with fsl_sub to $QUEUE"
+      queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
+    fi
+
+    # general settings
+    # set the start step beginning from RunPROFUMO which is by default the first step
+    StartStep="RunPROFUMO"
+    StopStep="GroupPFMs"
+    NumWishart="5"
```
glasserm: I think we ended up with 6 for humans.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -0,0 +1,268 @@
...
+
+    # set temporal highpass full-width (2*sigma) used in preprocessing
+    HighPass="2000"
+
+    #set fMRIResolution of data, like '2','1.60' or '2.40'
+    fMRIResolution="2.0"  
+
+    # PFM settings for REST data
+    # set the PFM dimensionality
+    PFMdim="99"
```
glasserm: We ended up with 76 for human 3T resting state.

Examples/Scripts/RunPFMpipelineFullRun.sh
glasserm: Should it be in the setup script like this sort of thing often is?

PFM/scripts/ApplyWFProfumo.m
glasserm: Yes, we need to do this correctly as I stated there.

PFM/scripts/ApplyWFProfumo.m
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    Out = icaDim(allData, 0, 1, -1, numWisharts);
+
+    startTP = 1;
+    for i = 1:length(outList)
+        endTP = startTP + tpCounts(i) - 1;
+        cii = ciftiopen(strtrim(inList{i}), 'wb_command');
+        cii.cdata = Out.data(:, startTP:endTP) + runMeans{i};
```
glasserm: Comment above also applies to deconcatination.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,445 @@
...
+#description to use in usage
+opts_SetScriptDescription "implements complete PFM pipeline with four main steps: Run PROFUMO, Post-PROFUMO, RSN Regression, and Group PFM processing"
+
+#mandatory parameters
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder that contains all subjects"
+opts_AddMandatory '--subject-list' 'SubjlistRaw' '100206@100307...' "list of subject IDs separated by @s"
+opts_AddMandatory '--fmri-names' 'fMRINames' 'rfMRI_REST1_LR@rfMRI_REST1_RL...' "list of fmri run names separated by @s"
+opts_AddMandatory '--output-fmri-name' 'OutputfMRIName' 'rfMRI_REST' "name to use for PFM pipeline outputs"
+opts_AddMandatory '--output-string' 'OutputSTRING' 'string' "output string for individual subject files (typically includes dimension, group name, and seed)"
+opts_AddMandatory '--proc-string' 'fMRIProcSTRING' 'string' "file name component representing the preprocessing already done, e.g. '_Atlas_MSMAll_hp2000_clean_rclean_tclean'"
```
glasserm: I'd rather use a more forward-looking example: "_Atlas_MSMAll_hp0_clean_tclean"

PFM/PFMPipeline.sh
glasserm: I think Tim wants you to clear only "Results.ppp" and not anything else.  You could also rename it instead of deleting it.

glasserm:
Once the above issues are addressed, I would be interested in looking at the test outputs of this pipeline.

Examples/Scripts/RunPFMpipelineFullRun.sh
coalsont: It is specific to only this pipeline, so I don't think it needs to be in global setup (users that don't want to do PFMs would need to skip over it if it is in a global location).

PFM/.gitignore
```diff
@@ -0,0 +1 @@
+profumo_v2.sif
```
coalsont: More future-proof against new profumo versions:

```suggestion
*.sif
```

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -108,161 +108,141 @@ get_options() {
...
+  queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+else
+  echo "queueing with fsl_sub to $QUEUE"
+  queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
+fi
+
+# Download the PROFUMO Singularity image from the following link and place it in the PFM folder, 
+# or change the path below to point to your own copy of the image
+# https://balsa.wustl.edu/myelin/download?dirName=public&filepath=profumo_v2.sif
+ProfumoSingularity="$HCPPIPEDIR/PFM/profumo_v2.sif"
```
glasserm, line 128: Usually stuff like this is handled in the HCP Pipelines setup script.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -108,161 +108,141 @@ get_options() {
...
+#set fMRIResolution of data, like '2','1.60' or '2.40'
+fMRIResolution="2.0"  
+
+# PFM settings for REST data
+# set the PFM dimensionality
+PFMdim="76" 
+
+PFMFolder=${StudyFolder}/$GroupAverageName/MNINonLinear/Results/${OutputfMRIName}_PFM_d${PFMdim}
+# Reference image for PROFUMO
+RefImage="${StudyFolder}/$GroupAverageName/MNINonLinear/Results/${OutputfMRIName}/${OutputfMRIName}_Atlas_MSMAll_hp${HighPass}_clean_tclean_meanvn.dscalar.nii"
```
glasserm: Is this intended to match HCP-YA or later data?  HCP-YA is _clean_rclean_tclean

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -108,161 +108,141 @@ get_options() {
...
+
+OutputfMRIName="rfMRI_REST"
+# set the MR concat fMRI name, if multi-run FIX was used, leave empty for single runs
+ConcatName=""
+
+# set the output spectra size for individual projection, RunsXNumTimePoints    #subjectExpectedTimepoints="3655"
+subjectExpectedTimepoints="4800"
+
+# set temporal highpass full-width (2*sigma) used in preprocessing
+HighPass="0"
```
glasserm: HCP-YA is 2000, later is 0.

Examples/Scripts/RunPFMpipelineFullRun.sh
coalsont: As I commented earlier, unless we are expecting every user to be required to run PFMs, then downloading and specifying the location of the profumo container seems specific to only this pipeline, unlike FSL or wb_command, which are used by nearly all pipelines.  Even for freesurfer, `bbregister` is used in multiple pipelines, which is why it makes sense to have in the global setup.

arkky68: I m thinking if we can make it overridable rather than required in the global setup:

```bash
ProfumoSingularity="${ProfumoSingularity:-$HCPPIPEDIR/PFM/profumo_v2.sif}"
```

That way users who don't run PFMs never have to set anything,  but anyone running this regularly can export it from their setup script if they prefer.

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: I think it is to match HCP-YA, updated in the pull request

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: Updated in the new pull request

Examples/Scripts/RunPFMpipelineFullRun.sh
coalsont: I don't expect all end users to feel comfortable editing a line that has `${something:-` in it, but I do expect some of them to download it to somewhere other than the pipelines folder.  Someone who is comfortable with bash syntax can easily set things up some other way anyway.

This particular container appears to be single-purpose for a single pipeline.  As long as that is the case, I don't think it belongs in a centralized configuration location (where people who aren't running PFMs would have to read it and understand that the setting isn't relevant to them).

PFM/PFMPipeline.sh
```diff
@@ -198,7 +197,7 @@ do
                                 for fMRIName in "${fMRINamesArray[@]}"
                                 do
                                     origFile="${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas${RegString}_${fMRIProcSTRING}.dtseries.nii"
-                                    VN="${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas${RegString}_${fMRIProcSTRING}_vn.dscalar.nii"
+                                    VN="${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas${RegString}_${fMRIProcSTRING%%_*}_vn.dscalar.nii" # not terribly robust
```
glasserm: @coalsont is this globbing okay with you?  I'm not clear on why we need it either.

Examples/Scripts/RunPFMpipelineFullRun.sh
glasserm: With regard to putting it in the setup script, if we make it optional in there, isn't that fine?  I think it might be best to keep all external dependencies referenced in there instead of scattered across the pipelines.  I don't think we currently have any external dependencies that are not in the setup script.

PFM/PFMPipeline.sh
bqrosen: Yeah, as I commented, it's a little fragile.  I think it's needed because $fMRIProcSTRING contains both the high pass substring and the "_clean" substring,  but the VN file only has the highpass substring without the "_clean" substring. And the script doesn't have access to the separate substrings that make up $fMRIProcSTRING. Would explicitly removing the "_clean" substring  be better? I'm not sure if it is always "_clean".

coalsont: I'm not sure that a script should ever be splitting procstrings apart.  ~If the logic is otherwise correct, then it might be better to make a copy (or symlink) of the `_vn` file for each procstring (should be small) in order to make scripting easier (and whichever pipeline is responsible for this situation, should be made to do this in the future).~  With the apparent exception of a "_clean_vn.dscalar.nii" (which probably isn't specific to one cleaning type either), there shouldn't be separate _vn files for each cleaning type.

Examples/Scripts/RunPFMpipelineFullRun.sh
coalsont: If we expect nearly all users to run PFMs, it might make sense.  For any user that isn't going to run PFMs, having it in the global setup would just be clutter that they have to read, understand, and then skip over.  We can move it to a global location later if it becomes something that an overwhelming majority of users will need to do.  Keeping the global setup short by only containing things that nearly all users need to do is a benefit to the user experience and gets them started faster before asking them to make judgement calls.

"External" isn't the reason I put things in a global location, it is whether they are used in more than one pipeline (or aren't actually associated with any specific pipeline, like `VolumeToCIFTI.sh`).  MSM, while also being a single file, is used in both PostFS and MSMAll, hence it would not be a good fit for being in a single pipeline's batch, as then users would need to configure it twice.  That does not currently apply to profumo, and I suspect it may never apply.  The hippunfold container likewise will probably only be used in one pipeline, and it isn't clear what percentage of users will want to use that, either, so it would also be a candidate for batch script configuration, regardless of being external.

PFM/PFMPipeline.sh
glasserm: There are two different _vn files.  The one without the ${fMRIProcSTRING} is generated upon initial concatenation.  It has some deleterious properties; however, and is not actually what we want to use when want to variance normalize data.  One just needs to ensure that if we need to use both _vn files that they are both used appropriately.  The one with ${fMRIProcSTRING} is intended for general use.  It is created in a different manner than the original file and does not cause issues with downstream analyses.  Perhaps we need to review in person what we are doing and therefore what the correct usage should be.

Examples/Scripts/RunPFMpipelineFullRun.sh
glasserm: In the long run, I think HippUnfold will be run by most/all users (as the future CIFTI grayordinates space will use hippocampal surfaces rather than hippocampal volumes).  Also, in the long run, I think PFMs will be the primary functional network analysis tool.  I am leaning strongly towards putting them in the setup script for that reason.

PFM/PFMPipeline.sh
coalsont: To clarify a misunderstanding I had: this `%%` is actually just retrieving the "hp" section of the filename, which currently is not a separate argument to the script (it is part of the procstring).  If we need to build a filename without a "clean" in it, I think highpass should probably be a separate argument.

PFM/PFMPipeline.sh
```diff
@@ -181,7 +186,7 @@ do
                         then
                             # Use already concatenated file
                             concatFile="${StudyFolder}/${Subject}/MNINonLinear/Results/${ConcatName}/${ConcatName}_Atlas${RegString}_${fMRIProcSTRING}.dtseries.nii"
-                            VNA="${StudyFolder}/${Subject}/MNINonLinear/Results/${ConcatName}/${ConcatName}_Atlas${RegString}_${fMRIProcSTRING}_vn.dscalar.nii"
+                            clean_VN="${StudyFolder}/${Subject}/MNINonLinear/Results/${ConcatName}/${ConcatName}_Atlas${RegString}_${fMRIProcSTRING}_vn.dscalar.nii"
```
coalsont, line 191: Should this be hardcoded to `_clean_vn` rather than using `$fMRIProcSTRING`?

EDIT: I guess we might still need the highpass value to build it that way.  I was under the impression that the "clean VN" isn't currently written out for all new procstrings.

PFM/PFMPipeline.sh
```diff
@@ -197,22 +202,50 @@ do
...
+                                            -var clean_VN ${clean_VN} -select 1 1 -repeat
+                                        fi
+
+                                        if [[ "$VAweightBool" == 1 ]];then
+                                          log_Msg "Weighting data by average vertex areas"
+                                          VA=${StudyFolder}/${Subject}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Subject}.midthickness_va.${LowResMesh}k_fs_LR.dscalar.nii
+                                          VAgray=${StudyFolder}/${Subject}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Subject}.midthickness_va.grayordinates.${LowResMesh}k_fs_LR.dscalar.nii
+                                          ATLASroiL=${StudyFolder}/${Subject}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Subject}.L.atlasroi.${LowResMesh}k_fs_LR.shape.gii
+                                          ATLASroiR=${StudyFolder}/${Subject}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Subject}.R.atlasroi.${LowResMesh}k_fs_LR.shape.gii
+                                          if [[ ! -f "$VAgray" ]]; then
```
coalsont: Please try not to use "if file exists, don't remake it" constructs, they make debugging more confusing, as files with bad contents will refuse to fix themselves when rerun after the otherwise correct editing of the code.

PFM/scripts/PostPROFUMO.m
```diff
@@ -89,6 +90,20 @@ function PostPROFUMO(StudyFolder, SubjListRaw, fMRIListRaw, ConcatName, fMRIProc
     ts.NtimepointsPerSubject = size(TCS, 1);
     PFMSpectra = cifti_struct_create_sdseries(nets_spectra_sp(ts)','step',1/TR);
 
+    %% divide out vertex area weights
+    if VAweightBool
+      fprintf('Dividing out vertex area weights\n');
+      VAgray = ciftiopen([StudyFolder '/' Subjlist{s} '/MNINonLinear/fsaverage_LR' LowResMesh 'k/' Subjlist{s} '.midthickness_va.grayordinates.' LowResMesh 'k_fs_LR.dscalar.nii'], wbcommand);
```
coalsont: Maybe take an argument of the VA file, instead of a boolean, so the filename doesn't get hardcoded into the compiled matlab.

PFM/PFMPipeline.sh
glasserm: ProcSTRING will vary.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,520 @@
...
+                                        
+                                        ## The "*_clean*" volume and CIFTI timeseries have no mean, i.e. they are intensity normalized.
+                                        ## However, timeseries files without "*_vn*"" retain run variances, i.e. they are not variance normalized.
+                                        ## To expound, for concatenated (non "_vn*") timeseries, the indiviudal run variances are divided out, data concatenated and the mean of run variances (VNA) multiplied back in.
+                                        ## Therefore, naive splitting of the concatenated timeseries will result the data retaining this run-average variance. 
+                                        ## In addition, the tICA pipeline creates a concat "*_clean_vn*" dscalar file, this contains the pooled variance across runs, rather than average of run variances. 
+                                        ## The variance normalization option below divides each run (after splitting) by the clean_VN map to yield variance normalized (post-wishart) single run timeseries.
```
glasserm:
```suggestion
                                        ## The concatenated timeseries are intensity normalized and differences in unstructured noise variance between runs and the means have been removed.  For model-free analyses (i.e., not task-GLM) that prefer single runs, it is best to simply deconcatenate the runs.  If variance normalization is desired, use the same _clean_vn file from the concatenated folder for each run, rather than the original _vn files, which may cause extreme values in areas of little or no signal and require more complex handling.
```

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,520 @@
...
+                                        if [[ "$VarNormBool" == 1 ]];then
+                                          wb_command -cifti-math "(TCS / clean_VN)" ${outFile} \
+                                            -var TCS ${outFile} \
+                                            -var clean_VN ${clean_VN} -select 1 1 -repeat
+                                        fi
+
+                                        if [[ "$VAweightBool" == 1 ]];then
+                                          log_Msg "Weighting data by average vertex areas"
+                                          VA=${StudyFolder}/${Subject}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Subject}.midthickness_va.${LowResMesh}k_fs_LR.dscalar.nii
+                                          VAgray=${StudyFolder}/${Subject}/MNINonLinear/fsaverage_LR${LowResMesh}k/${Subject}.midthickness_va.grayordinates.${LowResMesh}k_fs_LR.dscalar.nii
```
glasserm: Need to include ${RegSTRING}

I think you want this: ${StudyFolder}/${Subject}/T1w/fsaverage_LR32k/${Subject}.midthickness${RegSTRING}_va_norm.32k_fs_LR.dscalar.nii

I don't know what all those lines below are doing.  Are they creating the above file?

bqrosen: I'm not certain what ${StudyFolder}/${Subject}/T1w/fsaverage_LR32k/${Subject}.midthickness${RegSTRING}_va_norm.32k_fs_LR.dscalar.nii is, but the files in YA_HCP don't have volume data in them so I think this needs a distinct name. I've added $RegSTRING.

PFM/PFMPipeline.sh
bqrosen: It was shorthand to create it once per subject instead of once per fMRIName. I've now done the proper thing and moved the VAgray file creation  out into the subject loop, while leaving its application in the fMRIName loop.

PFM/scripts/PostPROFUMO.m
bqrosen: It would have to be a filename for each subject. The output paths are constructed the same way, is is that bad to have the paths in the compiled matlab?

coalsont: Ah right, it is a groupwise matlab call.

For established files used as input, or for the output files, having the naming conventions built into matlab is okay, since they are unlikely to change.  For intermediates like this that we might want to swap for something else later while leaving the original file as-is, it is nice to pass in the filename as an argument.  But, since that is more difficult in a groupwise script, it is okay.  I have previously used text files to pass in lists of files, but this isn't that important, I just thought it would be an easier change than that.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,518 @@
...
+                                    tempfiles_create "tmp_lab_XXXXXX.nii.gz" tmp_lab_file
+                                    tempfiles_create "tmp_Lva_XXXXXX.shape.gii" tmp_Lva_file
+                                    tempfiles_create "tmp_Rva_XXXXXX.shape.gii" tmp_Rva_file
+                                    wb_command -cifti-separate ${outFile} COLUMN -volume-all "$tmp_jnk_file" -roi "$tmp_roi_file" -label "$tmp_lab_file"
+                                    wb_command -cifti-separate ${VA} COLUMN -metric CORTEX_LEFT "$tmp_Lva_file" -metric CORTEX_RIGHT "$tmp_Rva_file"
+                                    mean_VA=$(wb_command -cifti-stats ${VA} -reduce MEAN) # $VA is a dscalar cifti already masked by ATLASroi
+                                    wb_command -volume-math "(ROI * $mean_VA)" "$tmp_roi_file" -var ROI "$tmp_roi_file"
+                                    wb_command -cifti-create-dense-scalar ${VAgray} -volume "$tmp_roi_file" "$tmp_lab_file" \
+                                      -left-metric "$tmp_Lva_file" -roi-left $ATLASroiL -right-metric "$tmp_Rva_file" -roi-right $ATLASroiR
+                              fi
```
coalsont: Indentation is wrong here.
```suggestion
                                fi
```

PFM/PFMPipeline.sh
```diff
@@ -196,6 +196,26 @@ do
                                     --output="$concatOutFile" \
                                     --num-wishart="$NumWishart" \
                                     --matlab-run-mode="$MatlabMode"
+
```
glasserm, line 201: Is this even needed with the file I pointed you to that had normalized vertex areas and 1s for voxels?

bqrosen: va_norm files don't have volume data in them. I believe they are created DeDriftAndResamplePipeline.sh and volume data isn't added there. That said, should I normalize the areas to 1 before using them as weights?

coalsont: Normalizing the vertex areas to average to 1 (and using voxels set to 1) is equivalent to setting the voxels to the average of the vertex areas.  Do one or the other, but not both.

PFM/PFMPipeline.sh
```diff
@@ -0,0 +1,516 @@
...
+                                "$HCPPIPEDIR"/PFM/scripts/ApplyWFProfumo.sh \
+                                    --input="$concatFile" \
+                                    --output="$concatOutFile" \
+                                    --num-wishart="$NumWishart" \
+                                    --matlab-run-mode="$MatlabMode"
+
+                                if [[ "$VAweightBool" == 1 ]]; then
+                                    # create VA_norm cifti with volume grayordinates filled with ones areas for weighting 
+                                    VAnorm=${StudyFolder}/${Subject}/T1w/fsaverage_LR${LowResMesh}k/${Subject}.midthickness${RegString}_va_norm.${LowResMesh}k_fs_LR.dscalar.nii
+                                    VAgray=${StudyFolder}/${Subject}/T1w/fsaverage_LR${LowResMesh}k/${Subject}.midthickness${RegString}_va_norm.grayordinates.${LowResMesh}k_fs_LR.dscalar.nii
```
bqrosen: This file has normalized vertex areas from /T1w space but the subcortical voxels (which are filled with 1's) are in /MNINonlinear space. Should the file be saved in /T1w or /MNINonLinear? Or should it be a temporary file like the one created in RSNRegression? One issue with it being a temporary file is that it would need to be  created twice (in RunPROFUMO step and again in the postPROFUMO step).

glasserm: For fMRI don't we use the MNINonLinear vertex areas elsewhere?  I think this is correct given that the fMRI data are mapped onto the surface in MNI space.  That would solve this delemma...

coalsont: The surface used for mapping is irrelevant to the choice of vertex areas, because both the surfaces and volume are warped with the same warp (and thus the results are effectively identical to doing it in native space, as long as the jacobians aren't extreme - the choice of using MNI for mapping cortical data was purely for convenience and storage efficiency).  The question is what they are being used for, and whether that should use the subject's original brain shape or not.

bqrosen: RSNRegression uses normalized the /T1w areas and creates a temporary file with the volume components filled with ones. So I've now done the same in both the runPROFUMO and postPROFUMO steps.

PR #364, opened 2025-10-20
Unified NHP FreeSurferPipeline

TakJim:
This pull request adds the current NHP version of the FreeSurfer processing pipeline scripts.

Files Added
- Examples/Scripts/FreeSurferPipelineBatchNHP.sh – Batch execution script for NHP pipeline
- FreeSurfer/FreeSurferPipelineNHP.sh – Main entry point of the NHP FreeSurfer pipeline
- FreeSurfer/custom/SetUpFSNHP.sh – Environment setup script for NHP FreeSurfer processing
- FreeSurfer/custom/conf2hiresNHP
- FreeSurfer/custom/recon-all.v6.hiresNHP
- FreeSurfer/scripts/FreeSurferHiresPialNHP.sh
- FreeSurfer/scripts/FreeSurferHiresWhiteNHP.sh

PR #387, opened 2026-03-03
Add NHP template files

TakJim:
Add NHP template files and adjust parameters in some scripts.

Related to PR #364 review feedback.


.gitattributes
```diff
@@ -0,0 +1 @@
+*.gca filter=lfs diff=lfs merge=lfs -text
```
coalsont, line 1: Note, we haven't used git-lfs so far.

.gitignore
```diff
@@ -14,3 +14,7 @@
 
 # nfs temporaries
 .nfs*
+*.gca
+*0.25mm*.nii.gz
+*0.15mm*.nii.gz
+*0.1mm*.nii.gz
```
coalsont: It seems wrong to .gitignore file patterns like this.  Even with git-lfs handling .gca files, is it really necessary to also .gitignore them?

glasserm: Let's remove this.

.gitattributes
glasserm: Tim to work on git-lfs.

.gitignore
TakJim: Fixed it in 927bc9e.

.gitattributes
coalsont: Github charges for downloads of LFS data beyond 10GB/month (both LFS-enabled clones and LFS-included release zips count).  Given the large amount of monthly clones the pipelines get, we may not want to do it this way.

If the templates aren't useful for their intended purpose without the .gca files, then that suggests that the entirety of the NHP templates should be a separate download, to avoid confusing users into thinking they have the templates when they are missing a key piece (this also means that people using the pipelines for human data won't have to wait for it to download the NHP template data).

coalsont:
I have put these files (and the .gca files that wouldn't fit) into a zip and created a simple download script:

https://github.com/Washington-University/HCPpipelines/commit/3d5f67c46844f44ab5fda2206a94c19ae592d670

I'll also add a check to tell the users to run that script into FreeSurferPipelineNHP.  I think this is better for users (particularly for human data) than substantially increasing the size of the repo (when it needs an extra step either way, due to the uncompressed .gca files being too large for github to swallow), so I am tentatively closing this PR.

PR #372, opened 2025-12-15
NHP fMRIVolume

TakJim:
This PR applies fMRIVolume to NHP data.

Some additional modifications are still required. The parts that I have not been able to address yet are listed below.

### TODO
- [x] In GenericfMRIVolumeProcessingPipelineBatchNHP.sh, accept arguments via opts_AddMandatory / opts_AddOptional
- [x] In BatchNHP.sh, reorganize the parameters defined in hcppipe_conf.txt and allow users to specify them directly within BatchNHP.sh
- [x] Apply the IsLongitudinal=TRUE processing logic to PipelineBatchNHP.sh

PR #390, opened 2026-03-31
initial implementation of TOPUP_MISMATCHED

demsarjure:
This is the initial implementation that I made based on your descriptions. Maybe I missed something. I started running some tests and so far things look OK.


glasserm (changes requested):
This looks very nice and took about as long as I expected for the implementation by a skilled practitioner.  My main comment is that these changes will interact with the NHP changes and will need to be reconciled with them: https://github.com/Washington-University/HCPpipelines/pull/372

demsarjure:
My tests worked fine, so this seems like it is good to go. I will wait for NHP pipelines to be merged into master and then revise so ther are no conflicts before merging this as well.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -512,18 +528,242 @@ if (( ! IsLongitudinal )); then
...
+            # 1/ Convert fieldmap from Hz to rad/s (FUGUE/epi_reg expect rad/s)
+            log_Msg "Converting fieldmap from Hz to rad/s"
+            ${FSLDIR}/bin/fslmaths ${InhomFieldMap} -mul 6.2832 ${WD}/FieldMap_rads_orig -odt float
+
+            # 2/ Register fieldmap to T1w space
+            # Use abs(fieldmap) as a registerable proxy (has tissue contrast from B0 inhomogeneity pattern)
+            log_Msg "Registering fieldmap to T1w space"
+            ${FSLDIR}/bin/fslmaths ${WD}/FieldMap_rads_orig -abs ${WD}/FieldMap_abs
+            ${FSLDIR}/bin/bet ${WD}/FieldMap_abs ${WD}/FieldMap_abs_brain -f 0.35 -m
+            ${FSLDIR}/bin/flirt -interp spline -dof 6 -in ${WD}/FieldMap_abs_brain -ref ${WD}/${T1wBrainImageFile} -omat ${WD}/fmap2T1w.mat -out ${WD}/FieldMap_abs2T1w -searchrx -30 30 -searchry -30 30 -searchrz -30 30
```
coalsont: Fieldmaps are generally derived from images, so there should usually be some image that is already in the same position as the fieldmap.  I'm not sure that registering the absolute field error to the T1w_brain with flirt is the approach we would use.

demsarjure: Hi @coalsont 

I added this here since one of our users urgently needed it, I am building my dev containers of this branch so I just put it here to make my life easier. We can also remove this functionality if it is not something you find useful later before I merge into master.

The full story is that some datasets ship with inhomogenity field maps (topup output essentially). One such example being UKB which one of our "premium" users needs to process promptly for some of their research work.

In UKB case, the fieldmaps are derived from diffusion. Specifically, the fieldmap used for fMRI distortion correction in UKB was generated using TOPUP from a reverse-direction pair of diffusion B0 images and is the "real" fieldmap that shows the inhomogeneity at each voxel.

The process above mimicks what UKB did in their processing I believe. The image that is in the same space as the fieldmap is diffusion. I think using that for registration would be a bit too convoluted?

coalsont: I think diffusion B0s should have some anatomical contrast (and are probably more recognizably brainlike than abs field error).

I think it is up to @glasserm as to whether we put this in the "legacy processing" category and/or modify it to register differently.  However, he is on vacation at the moment.

demsarjure: Agreed, I will let it sit on this branch and then either delete it or modify it if you think this would be a useful addition for the master branch.

mharms: While not our standard HCP acquisition, its still a TOPUP-based fieldmap, so I wouldn't relegate this option to "legacy processing" mode.

mharms: Why not use an actual anatomical image for the image that is used for registration (in identical space to the derived fieldmap, so that the anatomical image and fieldmap can be assumed to be in exact registration for any subsequent operations)?

demsarjure: Yes, the more I think about what I did in the registration step, the less I like it. Let me revisit this and then we can discuss.

glasserm: This shouldn't need to be legacy mode.

glasserm:
@demsarjure @coalsont is this mainly pending the NHP stuff, or is there more to do?

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -512,18 +530,243 @@ if (( ! IsLongitudinal )); then
...
+                    --fmriname="$NameOffMRI" \
+                    --corticallut="$HCPPIPEDIR/global/config/FreeSurferCorticalLabelTableLut.txt" \
+                    --subcorticallut="$HCPPIPEDIR/global/config/FreeSurferSubcorticalLabelTableLut.txt" \
+                    --smoothingfwhm="2" \
+                    --inputdir="$WD"
+            fi
+
+            ;;
+
+        ${REAL_FIELDMAP_METHOD_OPT})
```
glasserm: There is no Jacobian support here.

glasserm: I guess we don't support Jacobian for any fieldmap method...

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
demsarjure: Agreed, both new DC correction methods `TOPUP_MISMATCHED` and `REAL_FIELDMAP` are non-legacy ones.

demsarjure:
> @demsarjure @coalsont is this mainly pending the NHP stuff, or is there more to do?

Yes, this is mainly pending NHP stuff. The only thing left to resolve are your comments around Jacobian.

demsarjure:
@glasserm  For some reason, I can again see your comments around Jacobian only in email... I am not sure what is happening here, maybe you need to click the `Submit Review` button on top or sth. Weird things...

Anyhow, the comment was:

```
+            if [[ "$BiasCorrection" == "SEBASED" ]]
+            then
+                mkdir -p "$WD/ComputeSpinEchoBiasField"
+                "${HCPPIPEDIR_fMRIVol}/ComputeSpinEchoBiasField.sh" \
+                    --workingdir="$WD/ComputeSpinEchoBiasField" \
+                    --subjectfolder="$SessionFolder" \
+                    --fmriname="$NameOffMRI" \
+                    --corticallut="$HCPPIPEDIR/global/config/FreeSurferCorticalLabelTableLut.txt" \
+                    --subcorticallut="$HCPPIPEDIR/global/config/FreeSurferSubcorticalLabelTableLut.txt" \
+                    --smoothingfwhm="2" \
+                    --inputdir="$WD"
+            fi
+
+            ;;
+
+        ${REAL_FIELDMAP_METHOD_OPT})

Matt: I guess we don't support Jacobian for any fieldmap method...
```

Do you want me to add/change something here? I think the current state is consistent with what you had before. UseJacobian was relevant only for `SPIN_ECHO_METHOD_OPT` I now also added this for `TOPUP_MISMATCHED`, as this is basically "mismatched SE variant".

glasserm:
@demsarjure @coalsont can we finalize this so that Takuya can make his improved NHP topup PR?

demsarjure:
@glasserm I resolved the conflicts, so I think this is ready to be merged.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -662,18 +679,243 @@ if (( ! IsLongitudinal )); then
...
+            fmap_median=$(${FSLDIR}/bin/fslstats ${WD}/FieldMap_rads2T1w_unmasked -k ${WD}/FieldMap_T1w_brain_mask -P 50)
+            ${FSLDIR}/bin/fslmaths ${WD}/FieldMap_rads2T1w_unmasked -sub ${fmap_median} ${WD}/FieldMap
+
+            # 5/ Copy scout image
+            cp ${ScoutInputName}.nii.gz ${WD}/Scout.nii.gz
+
+            # Test if T1w Brain is much larger, suggesting poor fieldmap-based brain extraction;
+            # check if NHP (FreeSurferNHP) — if so, create brain-extracted scout for registration
+            MagnitudeBrainSize=$(${FSLDIR}/bin/fslstats ${WD}/${T1wBrainImageFile} -V | cut -d " " -f 2)
+            if [ -e ${FreeSurferSubjectFolder}/${FreeSurferSubjectID}_1mm ] ; then
```
coalsont: This looks like a copy of old NHP code, I don't think we do this in the new NHP code.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
```diff
@@ -662,18 +679,243 @@ if (( ! IsLongitudinal )); then
...
+
+                # register scout to T1w using fieldmap (fieldmap already in T1w space, so --nofmapreg)
+                # Use magnitude image registered to T1w as --fmapmag/--fmapmagbrain (same space as fieldmap after registration)
+                ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/Scout_brain.nii.gz --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}${ScoutExtension} --fmap=${WD}/FieldMap.nii.gz --fmapmag=${WD}/RealFmapMag2T1w --fmapmagbrain=${WD}/RealFmapMag2T1w --echospacing=${EchoSpacing} --pedir=${UnwarpDir} --nofmapreg
+            else
+                # register scout to T1w using fieldmap (fieldmap already in T1w space, so --nofmapreg)
+                # Use magnitude image registered to T1w as --fmapmag/--fmapmagbrain (same space as fieldmap after registration)
+                ${HCPPIPEDIR_Global}/epi_reg_dof --dof=${dof} --epi=${WD}/Scout.nii.gz --t1=${T1wImage} --t1brain=${WD}/${T1wBrainImageFile} --out=${WD}/${ScoutInputFile}${ScoutExtension} --fmap=${WD}/FieldMap.nii.gz --fmapmag=${WD}/RealFmapMag2T1w --fmapmagbrain=${WD}/RealFmapMag2T1w --echospacing=${EchoSpacing} --pedir=${UnwarpDir} --nofmapreg
```
coalsont: The other fieldmap code above appears to do a `flirt` registration of the scout to T1w before `epi_reg_dof`.  I don't know the history of this, and we use spin echo in all the HCP data, so I don't know what the impact is.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
demsarjure: Fixed

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
demsarjure: If I am not mistaken `epi_reg_dof` always does its own internal FLIRT pre-alignment before BBR (line 334 of the script), so no external pre-registration is missing here. The external FLIRT steps seen in other methods serve different purposes (brain extraction in the GE fallback, or NHP-specific multi-stage registration).

glasserm (approved):

glasserm:
If this is tested and Tim is satisfied, this can be merge.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
coalsont: I should have been more clear: the `_1mm` is no longer how we test for NHP.  We do have an NHP section in the topup case:

https://github.com/Washington-University/HCPpipelines/blob/6801b5f67258e9dbd7950d8c4f7c0289b1596432/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh#L497

Perhaps we should add a warning to the topup_mismatched mode when the species isn't "Human", if we don't have a dataset for testing mismatched in NHP?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
```diff
@@ -83,6 +85,21 @@ opts_AddMandatory '--dcmethod' 'DistortionCorrection' 'method' "Which method to
...
+             --seunwarpdir in addition to the standard SE parameters. This should be the
+             polarity of the positive SE image.
+             NOTE: Less accurate than '${SPIN_ECHO_METHOD_OPT}' when SE and fMRI parameters
+             differ substantially (especially PE direction axis or limited SE spatial coverage).
+
+        '${REAL_FIELDMAP_METHOD_OPT}'
+             use a pre-computed real fieldmap in Hz (e.g., from TOPUP --fout on diffusion B0
+             images, UKB style). The fieldmap is registered to T1w space and used directly
+             for distortion correction via epi_reg. No phase images or deltavTE are needed.
+             Requires --inrealfmap.
```
coalsont: I'm not sure "real" is the best name for this, since it is (presumably) the same underlying measurement as one of the other methods (are we calling those non-real?).  Maybe we should use "precomputed" as part of the name instead.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
coalsont: Does that mean we are registering an unmasked scout to a brain-masked T1w here?

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
demsarjure: I Had it named differently, but then Matt suggested we changed into REAL FIELDMAP. I have no preferences and can go with precomputed as well.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
demsarjure: I added a warning, and left a TODO note for future work.

We could create a dataset for mismatched NHP by simply changing the resolution of SE. We can do this later if needed. Even the topup mismatch is quite an edge case, not sure if wasting our bandwidth on imeplementing this for NHP as well is worthwile at the moment.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
demsarjure: Isn't this the standard `epi_reg` usage? It expects an unmasked EPI and handles masking internally via the boundary-based registration cost function against --t1brain? So, the unmasked scout is intentional here.

I reviewed the use of `epi_reg_dof` at other locations and the masked scout seems to be used exclusively exclusively in NHP (`$SPECIES != "Human"`) or when a brain size mismatch triggers a fallback.

For all standard Human paths, including the real-fieldmap case, the scout is unmasked and consistent with the use here.

Am I missing something?

glasserm: This is my recollection as well.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
glasserm: I don't recall the prior name now, but "real" refers to the fact that the fieldmap is in the correct units.  Precomputed seems fine too though.

coalsont: If we wanted to say it was in Hertz, "FIELDMAP_HERTZ" would be a better way to say that than "real".  To the user, I think the fact that it is precomputed rather than the raw fieldmap sequence images seems more relevant, so something like "PRECOMPUTED_FIELDMAP" is what I would call it.

fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh
coalsont: I was looking at this line, where in the fieldmap images method, it passes the newly-masked scout to `epi_reg_dof`:

https://github.com/Washington-University/HCPpipelines/blob/b3c53049574d3a0b9c588bd484c70acc5f3add14/fMRIVolume/scripts/DistortionCorrectionAndEPIToT1wReg_FLIRTBBRAndFreeSurferBBRbased.sh#L456

But I see there is an `else` that passes the unmasked scout instead.  The conditional appears to only do this when the brain masking of the magnitude image appears to have failed, so the `else` is actually the expected case instead.

fMRIVolume/GenericfMRIVolumeProcessingPipeline.sh
demsarjure: OK, will rename in my next commit.

demsarjure:
I renamed REAL to PRECOMPUTED. @coalsont I think you can click the merge button here.

coalsont:
@demsarjure What is the status of testing this?

demsarjure:
> @demsarjure What is the status of testing this?

This is tested. There are a couple of QuNex users waiting for this to run at scale, so if some bugs surface then I will fix them promptly.

PR #392, opened 2026-04-10
Pfm Update with human subjects

arkky68:
1 update the example script ( with human subject ) 
2 update with $ConcatName != "" situation
3 add fmri-resolution parameter


Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -17,14 +17,15 @@ get_options() {
...
 
     # initialize variables
-    StudyFolder="/media/myelin/brainmappers/BICAN/Macaque/MacaqueRhesus"
-    Subjlist="$(find ${StudyFolder} -maxdepth 1 -type d -name "A*" -exec basename {} \; | paste -sd@ -)"
-    EnvironmentScript="/media/myelin/burke/projects/Mac25Rhesus/scripts/Mac25Rhesus_v5_SetUpHCPPipeline.sh"
-    GroupAverageName="Mac25Rhesus_v5"
-    RegName=""
+    StudyFolder="/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final"
+    Subjlist="100610@102311"
+    EnvironmentScript="/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/Scripts/SetUpHCPPipeline_PFM.sh"
```
bqrosen:
```suggestion
    EnvironmentScript="${HOME}/projects/HCPpipelines/Examples/Scripts/SetUpHCPPipeline.sh"
```

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -17,14 +17,15 @@ get_options() {
     local arguments=("$@")
 
     # initialize variables
-    StudyFolder="/media/myelin/brainmappers/BICAN/Macaque/MacaqueRhesus"
-    Subjlist="$(find ${StudyFolder} -maxdepth 1 -type d -name "A*" -exec basename {} \; | paste -sd@ -)"
-    EnvironmentScript="/media/myelin/burke/projects/Mac25Rhesus/scripts/Mac25Rhesus_v5_SetUpHCPPipeline.sh"
-    GroupAverageName="Mac25Rhesus_v5"
-    RegName=""
+    StudyFolder="/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final"
```
bqrosen, line 20:
```suggestion
    StudyFolder="${HOME}/data/HCPpipelines_ExampleData"
```

PFM/scripts/PostPROFUMO.m
```diff
@@ -64,7 +64,7 @@ function PostPROFUMO(StudyFolder, SubjListRaw, fMRIListRaw, ConcatName, fMRIProc
       runAmp = load([PFMFolder '/Results.ppp/Amplitudes/sub-' Subjlist{s} '_run-' subfMRINames{r} '.csv']);
 
       origTCS = [origTCS ; runTCS];
-      TCS = [TCS ; runTCS .* repmat(runAmp', numel(runTCS), 1)];
+      TCS=[TCS ; runTCS.*repmat(runAmp',length(runTCS),1)];
```
bqrosen, line 67: length() in matlab is ambiguous for n-dimensional arrays. It's better to use either numel() or size() with a dimension argument.

PFM/PFMPipeline.sh
```diff
@@ -271,10 +279,12 @@ do
                 rsn_cmd=("$HCPPIPEDIR"/global/scripts/RSNregression.sh
                     --study-folder="$StudyFolder"
                     --subject="$Subject"
-                    --subject-timeseries="$ConcatName" # "$fMRINamesForSub"
+                    --subject-timeseries="$fMRINamesForSub" # "$fMRINamesForSub"
                     --surf-reg-name="$RegName"
+                    --low-ica-dims="$LowDims"
                     --low-res="$LowResMesh"
                     --proc-string="_$fMRIProcSTRING"
+                    --low-ica-template-name="$LowDimTemplate"
```
bqrosen, line 287: For the PFM pipeline, the only RSNregression method supported is 'dual'. low-dims and low-ica-template parameters are for the 'weighted' method. Because 'weighted' is not supported, there is no need to include low-dim related parameters.

PFM/PFMPipeline.sh
```diff
@@ -59,6 +60,8 @@ opts_AddOptional '--low-res-mesh' 'LowResMesh' 'string' "mesh resolution, like '
 #RSN regression specific parameters
 opts_AddOptional '--fix-legacy-bias' 'FixLegacyBias' 'YES or NO' 'whether the input data used legacy bias correction' 'NO'
 opts_AddOptional '--scale-factor' 'ScaleFactor' 'float' 'scale factor for RSN regression' '0.01'
+opts_AddOptional '--low-dims' 'LowDims' 'dims' "low ICA dimensions separated by @s for RSN regression" ''
```
bqrosen, line 63: For the PFM pipeline, the only RSNregression method supported is 'dual'. low-dims and low-ica-template parameters are for the 'weighted' method. Because 'weighted' is not supported, there is no need to include low-dim related parameters.

PFM/PFMPipeline.sh
```diff
@@ -239,25 +242,30 @@ do
             log_Msg "Running RSNRegression step"
             
             # Set up template paths
-            # LowDimTemplate="${StudyFolder}/${GroupAverageName}/MNINonLinear/Results/${OutputfMRIName}/sICA/melodic_oIC_${PFMdim}.dscalar.nii"
+            LowDimTemplate="${StudyFolder}/${GroupAverageName}/MNINonLinear/Results/${OutputfMRIName}/sICA/melodic_oIC_${PFMdim}.dscalar.nii"
```
bqrosen, line 245: For the PFM pipeline, the only RSNregression method supported is 'dual'. low-dims and low-ica-template parameters are for the 'weighted' method. Because 'weighted' is not supported, there is no need to include low-dim related parameters.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -135,78 +136,75 @@ main() {
...
     # RSN regression settings
-    LowDims="6"
     FixLegacyBiasString="NO"
     ScaleFactor="0.01"
+    LowDims="7@8@9@10@11@12@13@14@15@16@17@18@19@20@21"
 
     # Volume template file
-    VolumeTemplateCIFTI="/media/myelin/brainmappers/BICAN/Macaque/MacaqueRhesus/Mac25Rhesus_v5/MNINonLinear/Results/Mac25Rhesus_v5_VolMaps_16_template.dscalar.nii"
-
+    VolumeTemplateCIFTI="/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/S1200_MSMAll7T175/MNINonLinear/${GroupAverageName}_CIFTIVolumeTemplate_${OutputfMRIName}.${fMRIResolution}.dscalar.nii"
```
bqrosen, line 188: We'll figure out what to do with the VolumeTemplate and .sif local paths later.

bqrosen (changes requested):

glasserm:
@arkky68 there are some comments here.

PFM/scripts/WishartFilter.m
arkky68, line 1: New Wisharfilter.m creates

PFM/scripts/PostPROFUMO.m
arkky68: runTCS is a 2D matrix (timepoints × components), and numel(runTCS) would return the total number of elements, not the number of rows. So I changed to size(runTCS, 1) .

arkky68:
I am wondering if we should separate PROFUMO postprocessing into its own pipeline step rather than keeping it inside RunPROFUMO. I've run into situations where postprocessing needs to be rerun separately.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -135,78 +136,75 @@ main() {
...
-
+    
     # build Profumo data location json
     mkdir -p $PFMFolder
     echo '{' > $ProfumoConfig
     for Subject in $(echo $Subjlist | tr "@" "\n"); do
         echo -e "\t\"$Subject\": {" >> $ProfumoConfig
         for fMRIName in $(echo $fMRINames | tr "@" "\n"); do
-            runFile="${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas_${fMRIProcSTRING}.dtseries.nii"
+            runFile="${StudyFolder}/${Subject}/MNINonLinear/Results/${fMRIName}/${fMRIName}_Atlas_${RegName}_${fMRIProcSTRING}.dtseries.nii"
```
bqrosen: This needs to work if RegName="". Maybe a little if statement that adds an _ if RegName isn't ""

PFM/scripts/WishartFilter.m
bqrosen: It is great that you've pulled this out of icaDIm and made it its own module. Can you put it in global/matlab/icaDim instead and then edit icaDim.m to call it. That way, if we ever want to tweak the Wishart filtering in the future, we won't need to make the edits in two places. The PFM scripts will still have access to it there.

PFM/scripts/WishartFilter.sh
```diff
@@ -0,0 +1,85 @@
...
+then
+    pipedirguessed=1
+    export HCPPIPEDIR="$(dirname -- "$0")/../.."
+fi
+
+source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
+source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
+g_matlab_default_mode=1
+
+opts_SetScriptDescription "applies Wishart filter to a single CIFTI file using icaDim"
```
bqrosen: Specifically, a CIFTI dtseries file, right? Also it doesn't use icaDim.m anymore.

PFM/scripts/WishartFilter.sh
```diff
@@ -0,0 +1,85 @@
...
+
+opts_AddMandatory '--input' 'inputFile' 'file' "the input dtseries file"
+opts_AddMandatory '--output' 'outputFile' 'file' "the output wishart-filtered dtseries file"
+opts_AddMandatory '--num-wishart' 'numWisharts' 'integer' "how many wisharts to use in icaDim"
+opts_AddMandatory '--pfm-dimension' 'PFMdim' 'integer' "the PFM dimension to use for the low-dim template in icaDim"
```
bqrosen: I don't think this description is correct.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -135,78 +136,75 @@ main() {
...
 
     # RSN regression settings
-    LowDims="6"
     FixLegacyBiasString="NO"
     ScaleFactor="0.01"
 
     # Volume template file
-    VolumeTemplateCIFTI="/media/myelin/brainmappers/BICAN/Macaque/MacaqueRhesus/Mac25Rhesus_v5/MNINonLinear/Results/Mac25Rhesus_v5_VolMaps_16_template.dscalar.nii"
-
+    VolumeTemplateCIFTI="${HOME}/data/HCPpipelines_ExampleData/${GroupAverageName}/MNINonLinear/${GroupAverageName}_CIFTIVolumeTemplate_${OutputfMRIName}.${fMRIResolution}.dscalar.nii"
```
bqrosen, line 189: It looks like fMRI Resolution is only used to build the path to the VolumeTemplateCIFTI. If that is the case, it would be better to make VolumeTemplateCIFTI a mandatory argument, instead of fMRIResolution.

bqrosen:
> I am wondering if we should separate PROFUMO postprocessing into its own pipeline step rather than keeping it inside RunPROFUMO. I've run into situations where postprocessing needs to be rerun separately.

Assuming it execute quickly, I think it makes sense to move it from the RunPROFUMO to postPROFUMO which is very short. In the situations where you needed to run postprocessing separately, you also needed to rerun the postPROFUMO step, right?

PFM/scripts/PostPROFUMO.m
```diff
@@ -64,7 +64,7 @@ function PostPROFUMO(StudyFolder, SubjListRaw, fMRIListRaw, ConcatName, fMRIProc
       runAmp = load([PFMFolder '/Results.ppp/Amplitudes/sub-' Subjlist{s} '_run-' subfMRINames{r} '.csv']);
 
       origTCS = [origTCS ; runTCS];
-      TCS = [TCS ; runTCS .* repmat(runAmp', numel(runTCS), 1)];
+      TCS = [TCS ; runTCS.*repmat(runAmp',size(runTCS),1)];
```
coalsont: `size(array)` returns a vector, not a scalar, so `repmat(var, vector, 1)` is probably going to error.  You probably meant `size(runTCS, 1)`.

Examples/Scripts/RunPFMpipelineFullRun.sh
coalsont: Here is how I handled this:

https://github.com/Washington-University/HCPpipelines/blob/34dbcc8c63e8deab9bd73e920d2f0b273db72abb/TransmitBias/Phase3_IndividualAdjustment.sh#L68-L73

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -135,78 +136,75 @@ main() {
...
+ 
     # PROFUMO settings
-    ProfumoSingularity="/media/myelin/burke/projects/Mac25Rhesus/HCPpipelines/PFM/profumo_v2.sif" 
+    ProfumoSingularity="$HCPPIPEDIR/PFM/profumo_v2.sif" 
     ProfumoConfig="${PFMFolder}/dataLocations.json"  
-    TR="0.702"
+    TR="1.0"
```
glasserm: TR for HCP-YA rsting state is 0.72.

PFM/scripts/WishartFilter.m
glasserm: I'm not sure I like this.  We now have two code paths that do the same thing and would have to be maintained in parallel.  I would prefer we didn't duplicate code like that.  Why can't we use the existing icaDim for this?

PFM/scripts/WishartFilter.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
glasserm, line 1: Wrapping IcaDim with a shell script would be fine; however.

PFM/PFMPipeline.sh
```diff
@@ -142,6 +143,59 @@ do
                 log_Err_Abort "Reference image must be specified with --ref-image"
             fi
             
+            ProfumoConfigToUse="${ProfumoConfig}"
```
glasserm, line 146: Is it going to be necessary to create an entire copy of the fMRI timeseries on the file system to use this functionality?

PFM/scripts/WishartFilter.m
bqrosen: If WishartFilter.m is put in global/matlab/icaDim/ and icaDim.m is edited to call it instead of an internal subfunction, then there will still only be one copy of the code and one wouldn't have to redo the dimensionality estimation if you had already assessed the dimensionality earlier. An alternative to functionalizing the Wishart filtering would be to add a dimOveride argument to icaDim.m, but I think separating the dimensionally estimation from the Wishart filtering (while maintaining a single copy) is cleaner.

PFM/PFMPipeline.sh
bqrosen: I don't think there is a way to get around it, unless you count a RAMdisk. The real solution is to implement Wishart filtering In PROFUMO itself. Since PROFUMO already uses python for a function called estimate_smoothness.py, I would predict that it would be reasonably straightforward for Rezvan to incorporate it if we translate WishartFilter.m into python.

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: Changed to $RegString

PFM/scripts/WishartFilter.sh
arkky68: fixed icaDim.m and specific on CIFTI dtseries

PFM/scripts/WishartFilter.sh
arkky68: Updated the descriptions to match the usage

PFM/scripts/PostPROFUMO.m
arkky68: fixed to size(runTCS, 1). Also, MATLAB R2016b+ supports implicit expansion, so runTCS .* runAmp' would work without repmat. Should we use that for simplicity, or keep repmat "?

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: updated TR to 0.72

PFM/scripts/WishartFilter.sh
arkky68: Could you clarify what you had in mind after 'however'?

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: delete fMRIResolution from mandatory arguments and keep volume template cifti as mandatory argument

PFM/scripts/WishartFilter.m
arkky68: I would restructure so there's one shared `WishartFilter.m` in `global/matlab/icaDim/` with the core filtering math. `icaDim.m` will call it instead of having the math inline. The PFM wrapper will be renamed to `ApplyWishartFilterProfumo.m` and also call `WishartFilter.m` for the usage. Then extract `FitWishart`, `Smooth`, and `lpdist` from the bottom of `icaDim.m` into standalone `.m` files in the same directory, so there's a single copy of everything shared by both code paths.

PFM/scripts/PostPROFUMO.m
coalsont: I think repmat is more obvious (as long as the memory and time costs aren't significant).  If we do use the dot-math auto-expansion in the future, we may want to add a comment to any line that uses it, so we know it isn't a mistake that the sizes are different.

PFM/scripts/WishartFilter.m
glasserm: I'm a bit confused here. We always want to estimate the Dim for Wishart Filtering.  That is a per dataset value that is not predictable in advance.  There are ways to short circuit icaDim (that are used for legacy data where it doesn't work as well), but I don't think we want to do that here.

PFM/scripts/WishartFilter.sh
glasserm: In contrast to my not liking messing with icaDim code.

PFM/PFMPipeline.sh
glasserm: I guess the question is if PROFUMO reads the data multiple times or not.  If it doesn't, perhaps the Wishart Filtering could be "just in time" as it is for other code.

bqrosen: I think saving an at least temporary copy to disk is always going to be needed if the Wishart filtering is done in matlab and the subsequent step like melodic or profumo is not.

PFM/scripts/WishartFilter.m
bqrosen: Ah, the calcDim from ICAFIX is for whole concat dataset, whereas we need the Wishart filter each run separately for PFMs, so the previously calculated Dim isn't usable here. That makes sense. You're right just need a wrapper around icaDim, not a separate Wishart filtering function.

coalsont: Should we be Wishart filtering the per-subject concat timeseries and then splitting back into runs?

PFM/scripts/ApplyWFProfumo.m
```diff
@@ -0,0 +1,24 @@
...
+    numWisharts = str2double(numWisharts);
+    inList = strsplit(inputFiles, ',');
+    outList = strsplit(outputFiles, ',');
+
+    allData = [];
+    tpCounts = [];
+    for i = 1:length(inList)
+        cii = ciftiopen(strtrim(inList{i}), 'wb_command');
+        tpCounts(i) = size(cii.cdata, 2);
+        allData = [allData cii.cdata];
```
glasserm: Concatenating fMRI data is typically more complicated than this, including requiring demeaning, potentially detrending (if not already done) and variance normalizing.  Would it be better to use the already concatenated data or stick with however sICA+FIX was run (use concatenated data if MR+FIX and not concatenated data otherwise)?

PFM/PFMPipeline.sh
```diff
@@ -142,6 +143,78 @@ do
...
+            if [[ "$NumWishart" -gt 0 ]]
+            then
+                WFDir="${PFMFolder}/WishartFilter_WF${NumWishart}"
```
glasserm, line 149: Interested in @coalsont's thoughts on this.  We could make one copy of the WF data, but creating and deleting it multiple times might be even worse that making one copy and keeping it.  If we make one copy, we probably want to put it in the HCP Pipelines main folder structure for each individual.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -135,78 +136,76 @@ main() {
...
+    #set fMRIResolution of data, like '2','1.60' or '2.40'
+    fMRIResolution="2.0"  
 
     # PFM settings for REST data
-    PFMdim="16"  # set the PFM dimensionality
-    PFMFolder=${StudyFolder}/$GroupAverageName/MNINonLinear/Results/${GroupAverageName}_${OutputfMRIName}_PFM_d${PFMdim}_s${randSeed}_test
-    # Reference image for PROFUMO
-    RefImage="${StudyFolder}/$GroupAverageName/MNINonLinear/Results/Mac25Rhesus_v5_BOLD_REST_CONCAT_MIGP/Mac25Rhesus_v5_BOLD_REST_CONCAT_MIGP_Atlas_hppd2_clean_meanvn.dscalar.nii"
+    # set the PFM dimensionality
+    PFMdim="99"
```
glasserm, line 168: Why did you choose this number?

PFM/PFMPipeline.sh
```diff
@@ -173,33 +245,55 @@ do
...
-            ## Rewrite input files with wb_command, as a failsafe to avoid SIMD buffer alignment issue in PROFUMO if 
-            ## files were written with an older version of cifti-matlab with 8-byte instead of 16-byte alignment. 
-            cat "${ProfumoConfig}" | \
-                while IFS= read -r line; do
-                    if [[ "$line" != *'.nii"'* ]]; then continue;fi # Only process lines that contain .nii"
-                    filePath=$(echo "$line" | sed -E 's/^[[:space:]]*"[^"]*"[[:space:]]*:[[:space:]]*"([^"]*)".*/\1/')
-                    wb_command -file-convert -cifti-version-convert "$filePath" 2 "$filePath"
-                done           
-            
+            # files were written with an older version of cifti-matlab with 8-byte instead of 16-byte alignment.
```
coalsont, line 284: Does this fix need to stay in the released version of the pipeline?

PFM/PFMPipeline.sh
coalsont: I'm not sure I have the context for this.  If files are generated, used, and immediately deleted once per filename, that is the kindest to snapshots (assuming you don't otherwise want the files, and assuming they have to be written to a snapshotted filesystem).  I'm not sure in what context (other than testing) you'd repeatedly run the PFM pipeline with the same wishart filtering settings (there is an option that can be specified to keep the wishart filtered file in that case), which is the only way you'd get the same folder name here (and therefore same absolute filename).

If the subject group is large, since none of them are deleted until after everything is done (probably unavoidable), the length of runtime does increase the chance that they will be "caught" by a daily (or longer) snapshot.  But, as long as we don't want to keep them for other purposes, and don't regenerate and redelete them again, that is still preferable to letting them sit there through multiple long-term snapshots.

If we ever wanted to keep these subject-specific files long-term (not for testing purposes), we would normally put them somewhere in the subjects' normal folder.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -135,78 +136,76 @@ main() {
...
-    # StartStep="RSNRegression"  
-    # StopStep="RSNRegression"
-    # StartStep="RunPROFUMO"
-    # StopStep=""
-
-    StartStep="PostPROFUMO"
+    StartStep="RunPROFUMO"
     StopStep="GroupPFMs"
+    NumWishart="5"
+    KeepWishartFiles="YES"
```
coalsont: It is a bit odd for the example to use the opposite of the default.  The example (and the default for any new parameter) are generally what we expect other users to run "in production".

PFM/PFMPipeline.sh
```diff
@@ -173,33 +245,55 @@ do
...
+                /opt/profumo/C++/PROFUMO "${ProfumoConfigToUse}" \
                 "${PFMdim}" "${PFM_PATH}" \
                 --useHRF "${TR}" --covModel "${CovModel}" --dofCorrection "${DOFCorrection}" \
                 --nThreads "${ProfumoThreads}" --lowRankData "${LowRankData}" --randomSeed "${RandomSeed}" \
                 --multiStartIterations "${MultiStartIterations}" ${LoadSequentiallyArg} ${InitialMapsArg}
+            
+            #Cleanup WF files
+            if [[ "$NumWishart" -gt 0 ]]
+            then
+                KeepWishartBool=$(opts_StringToBool "$KeepWishartFiles")
```
coalsont: Please move this line closer to the argument parsing, so that its sanity checking that the value is recognizable as a boolean happens before any processing, rather than giving an error message and aborting after the PFM completes.  See:

https://github.com/Washington-University/HCPpipelines/blob/6801b5f67258e9dbd7950d8c4f7c0289b1596432/global/scripts/RSNregression.sh#L68-L73

PFM/scripts/WishartFilter.m
glasserm: Actually I think we should match how sICA+FIX was run.  That will produce the "stickiest" dimensionality estimation.

PFM/PFMPipeline.sh
glasserm: Group PFMs probably requires all dtseries on disk while it is running over several days.

PFM/PFMPipeline.sh
glasserm: It is probably required for HCP-YA data compatibility.

coalsont: I see, I hadn't noticed that this was for the case of not doing wishart filtering (thus, the config may not be referring to freshly-written files).

PFM/PFMPipeline.sh
coalsont: The "--keep-wishart-files" argument is already intended for any cases where specifically PFM may be run again with the same wishart filtering, though it does require the user to notice it and realize its relevance to their plans and whether their storage has automatic snapshots (but it isn't clear to me whether other users will typically run PFM more than once).

If the context is "users may want wishart filtered data for other purposes (and possibly have already generated it)", that is at a broader scope than the PFM pipeline itself, and may suggest a global script and naming convention for just computing WF results, which would then be used by the PFM pipeline, instead of having its own internal WF-computing code and conventions for temporary files.

Examples/Scripts/RunPFMpipelineFullRun.sh
```diff
@@ -246,11 +245,13 @@ main() {
...
                                     --profumo-singularity="$ProfumoSingularity" \
                                     --profumo-tr="$TR" \
+                                    --keep-wishart-files="$KeepWishartFiles" \
                                     --profumo-threads="$ProfumoThreads" \
                                     --profumo-dof-correction="$DOFCorrection" \
                                     --profumo-cov-model="$CovModel" \
-                                    --profumo-multi-start-iterations="$nStarts"\
+                                    --profumo-multi-start-iterations="$nStarts" \
                                     --profumo-random-seed="$RandomSeed" \
+                                    --num-wishart="$NumWishart" \
```
bqrosen: This argument is duplicated.

PFM/scripts/ApplyWFProfumo.m
arkky68: when a concat file exists (via --concat-name), we now use that directly instead of concatenating in MATLAB. When no concat file is available, each run is demeaned before concatenation and the mean is added back after filtering.

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: This is the dimensionality I used for HCP-YA, so i directly used it in the example file, or do you have a suggestion on this part, or just leave it blank?

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: updated to default(NO)

PFM/PFMPipeline.sh
arkky68: Moved   after `FixLegacyBiasBool=$(opts_StringToBool  #"$FixLegacyBias")`

Examples/Scripts/RunPFMpipelineFullRun.sh
arkky68: deleted the duplicate

Examples/Scripts/RunPFMpipelineFullRun.sh
glasserm: For which type of data? We were using d=76 for 3T resting state.

PFM/scripts/ApplyWFProfumo.m
glasserm: Actually, there should be concatenated tICA cleaned fMRI data that we can use even for single run sICA+FIX runs.  I wouldn't support unconcatenated data unless you are going to do it exactly how it is done in sICA+FIX.

coalsont:
This appears to be requesting changes to `PFMs`, which is itself the branch for #378.  There also appears to be a `pr-392` branch that has these commits plus one more.  Perhaps @bqrosen should update #378's `PFMs` branch to contain all of the proposed changes?

bqrosen:
I've merged PR 392 into 378 (branch PFMs), so now everything should be together. 

-Burke

PFM/scripts/ApplyWFProfumo.m
```diff
@@ -0,0 +1,26 @@
...
+    outList = strsplit(outputFiles, ',');
+
+    allData = [];
+    tpCounts = [];
+    runMeans = {};
+    for i = 1:length(inList)
+        cii = ciftiopen(strtrim(inList{i}), 'wb_command');
+        tpCounts(i) = size(cii.cdata, 2);
+        runMeans{i} = mean(cii.cdata, 2);
+        allData = [allData (cii.cdata - runMeans{i})];
```
glasserm, line 13: As I stated elsewhere, this is not sufficient for concatenating data.  We need to both demean and variance normalize.

PFM/scripts/ApplyWFProfumo.m
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    Out = icaDim(allData, 0, 1, -1, numWisharts);
+
+    startTP = 1;
+    for i = 1:length(outList)
+        endTP = startTP + tpCounts(i) - 1;
+        cii = ciftiopen(strtrim(inList{i}), 'wb_command');
+        cii.cdata = Out.data(:, startTP:endTP) + runMeans{i};
```
glasserm, line 22: Need to undemean and unvariance normalize here.

PR #395, opened 2026-05-05
FS8 support

demsarjure:
PR for easy comparison of FS8 required changes


FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -150,6 +142,11 @@ opts_AddOptional '--conf2hires' 'conf2hiresString' 'TRUE/FALSE' "Indicates that
 
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData or LegacyStyleData' "Controls whether the HCP acquisition and processing guidelines should be treated as requirements.  'HCPStyleData' (the default) follows the processing steps described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion.  'LegacyStyleData' allows additional processing functionality and use of some acquisitions that do not conform to 'HCP-Style' expectations.  In this script, it allows not having a high-resolution T2w image." "HCPStyleData"
 
+opts_AddOptional '--high-myelin' 'HighMyelin' "High Myelin" 'Value of the high myeling extra recon-all parameter. Set to NONE to disable the use of this parameter.' "NONE"
```
coalsont: In addition to the typo/wording, I suggest using empty string (`""`) instead of `NONE` as the indication not to use the flag.  I don't like the `NONE` convention and try to avoid even supporting it except for parameters with existing history.

Is this just a convenience shortcut for `--extra-reconall-arg=-high-myelin --extra-reconall-arg=<whatever>`?

```suggestion
opts_AddOptional '--high-myelin' 'HighMyelin' "High Myelin" 'Use this to pass -high-myelin with the provided argument as an extra recon-all parameter.'
```

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -566,7 +540,17 @@ recon_all_cmd+=(${extra_reconall_args[@]+"${extra_reconall_args[@]}"})
 # The -conf2hires flag should come after the ${extra_reconall_args[@]} array, since it needs
 # to have the "final say" over a couple settings within recon-all
 if ((conf2hires)); then
-    recon_all_cmd+=(-conf2hires)
+   recon_all_cmd+=(-conf2hires)
+fi
+
+# HighMyelin
+if [[ "${HighMyelin}" != "NONE" ]]; then
```
coalsont:
```suggestion
if [[ "${HighMyelin}" != "" ]]; then
```

FreeSurfer/FreeSurferPipeline.sh
demsarjure: I fixed the typo and set the default value to 0.3. Which is what seems to work best for HCP data and FS8. It is not only a flag to pass to extra recon all, but flag + value.

FreeSurfer/FreeSurferPipeline.sh
demsarjure: Fixed.

demsarjure:
@coalsont @glasserm 

The script should now support both FS6 and 7/8. Please review, I made the changes we discussed on Monday.

FreeSurfer/FreeSurferPipeline.sh
coalsont: > I fixed the typo and set the default value to 0.3. Which is what seems to work best for HCP data and FS8.

If we continue supporting FS6 with existing behavior and want to also do this by default, sounds like the script will have to check it for being unset, check the FS version, and then manually implement a different "default".

> It is not only a flag to pass to extra recon all, but flag + value.

My question included both the flag and value.  Sounds like the answer is yes.

coalsont: > If we continue supporting FS6 with existing behavior and want to also do this by default, sounds like the script will have to check it for being unset, check the FS version, and then manually implement a different "default".

Looks like the script forcibly erases the value in this variable (and `gpu`) for FS6.  Is -high-myelin a valid flag in FS6?

coalsont: To clarify, I think anything that is specified by the user to a pipeline should either be used per the user's intentions, or should create an error.  If they manually specify `--high-myelin=0.3 --gpu=true` with FS6, it should pass those flags if it can, and if not, error and tell the user not to specify them for FS6.  At the least, I think FreeSurfer has had gpu support in some steps for a long time.

This actually presents a challenge for `--high-myelin`, because if not specifying it means "turn it on only in FS8", then for the user to be able to turn it *off* in FS8, we have to be able to check that as distinct from the newopts default (and by convention, passing an empty string to a numeric option means "off").  So, if this is the behavior we want the pipeline to have, the default value should probably be a special string that isn't otherwise valid, perhaps "auto", which we use to detect that the user has not specified the option (or has manually requested this default automatic FS8 detection behavior).

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -102,53 +102,19 @@ log_Check_Env_Var FREESURFER_HOME
...
-    log_Warn "we added ${HCPPIPEDIR}/FreeSurfer/custom to the beginning of the PATH."
-    log_Warn ""
-    log_Warn "If you intended to use some other version of these tools, please configure"
-    log_Warn "your PATH before invoking this script, such that the tools you intended to"
-    log_Warn "use can be found on the PATH."
-    log_Warn ""
-    log_Warn "PATH set to: ${PATH}"
-  fi
-}
-
```
glasserm, line 138: Why is all this being deleted?

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -102,53 +102,19 @@ log_Check_Env_Var FREESURFER_HOME
...
 show_tool_versions()
 {
   # Show recon-all version
-  log_Msg "Showing recon-all.v6.hires version"
-  local which_recon_all=$(which recon-all.v6.hires)
+  log_Msg "Showing recon-all version"
+  local which_recon_all=$(which recon-all)
   log_Msg ${which_recon_all}
-  recon-all.v6.hires -version
+  recon-all -version
```
glasserm: Same with these changes.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -102,53 +102,19 @@ log_Check_Env_Var FREESURFER_HOME
...
   
   # Show tkregister version
-  log_Msg "Showing tkregister version"
-  which tkregister
-  tkregister -version
+  log_Msg "Showing tkregister2 version"
+  which tkregister2
+  tkregister2 -version
```
glasserm: And these.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -207,9 +173,6 @@ validate_freesurfer_version()
   fi
 }
 
-# Configure the use of FreeSurfer v6 custom tools
-configure_custom_tools
-
```
glasserm, line 212: And these.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -280,7 +243,7 @@ if (( start_stage < 1 )); then
   log_Msg "Creating the base template: ${TemplateID}"
   # ----------------------------------------------------------------------
 
-  recon_all_cmd="recon-all.v6.hires"
+  recon_all_cmd="recon-all"
```
glasserm: And these.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -312,7 +275,7 @@ if (( end_stage > 0 )); then
   # ----------------------------------------------------------------------
   for Session in ${Sessions} ; do
     log_Msg "Running longitudinal recon all for session: ${Session}"
-    recon_all_cmd="recon-all.v6.hires"
+    recon_all_cmd="recon-all"
```
glasserm: And these.

glasserm (changes requested):
I didn't understand some of these commits noted above.  They would seem to break longitudinal processing.

coalsont:
> I didn't understand some of these commits noted above. They would seem to break longitudinal processing.

I think he intends to do more stuff before a full review.

demsarjure:
> I didn't understand some of these commits noted above. They would seem to break longitudinal processing.

@glasserm My bad. I did not fix the longitudinal script, longitudinal script is still in the FS8 only mode. Let me fix this today.

FreeSurfer/FreeSurferPipeline.sh
demsarjure: OK, let met set it to "auto" and then set appropriately for FS6 and FS8. I am not sure if FS6 had a GPU flag? If it did we were not using it by default.

demsarjure: @coalsont I added the "AUTO" value as the default.

This sets HighMyelin to 0.3 for FS7/8 and does not set for FS6. And sets the use of GPU to True for FS7/8 and False for FS6.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -311,26 +335,46 @@ validate_freesurfer_version()
...
+        use_fs6=1
+        # validate that unsupported parameters are not set for FS6
+        if [[ "${HighMyelin}" != "AUTO" && "${HighMyelin}" != "" ]]; then
+            log_Err_Abort "FreeSurfer 6 does not support the --high-myelin parameter. Do not set --high-myelin when using FS6."
+        fi
+        HighMyelin=""
+        if [[ "${gpu}" == "AUTO" ]]; then
+            gpu=0
+        elif ((gpu)); then
+            log_Err_Abort "FreeSurfer 6 does not support GPU-accelerated recon-all. Do not set --gpu=TRUE when using FS6."
```
coalsont: Is this really the case?  I have an "ernie.gpu" folder timestamped 2012, which appears to be from v5.1.0.  Did they change the flag?

I also thought fs8 was doing more deep-learning stuff, is it valid to do a cpu-only fs8 run?  Or is the deep-learning stuff proposed for fs9?

coalsont: For reference, from my ancient freesurfer install:

```
$ recon-all --help
...
 -use-gpu : use GPU versions of mri_em_register, mri_ca_register,
             mris_inflate and mris_sphere
...
```

Looks like the flag is `-gpu` in fs8 (according to your edits below).  We should at least alter our message to say that *this pipeline* doesn't support gpu-accelerated fs6 (if it is anything like fs5, only a few steps have any gpu code, and they don't save much time, so we may not want to bother supporting it).

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -170,6 +174,14 @@ extra_reconall_args=(${extra_reconall_args_manual[@]+"${extra_reconall_args_manu
 flair=$(opts_StringToBool "$flairString")
 existing_session=$(opts_StringToBool "$existing_sessionString")
 conf2hires=$(opts_StringToBool "$conf2hiresString")
+if [[ "$gpuString" == "AUTO" ]]; then
+    gpu="AUTO"
+else
```
coalsont: I would delay any handling of `gpuString` until after you know the freesurfer version, so you can do all the logic in a single block (rather than temporarily combining string and integer possibilities in a single variable).

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -566,7 +620,17 @@ recon_all_cmd+=(${extra_reconall_args[@]+"${extra_reconall_args[@]}"})
 # The -conf2hires flag should come after the ${extra_reconall_args[@]} array, since it needs
 # to have the "final say" over a couple settings within recon-all
 if ((conf2hires)); then
-    recon_all_cmd+=(-conf2hires)
+   recon_all_cmd+=(-conf2hires)
```
coalsont: Three spaces is a weird indentation, and appears to exist only on this line and 633.

```suggestion
    recon_all_cmd+=(-conf2hires)
```

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -140,15 +138,27 @@ configure_custom_tools()
...
+  if ((use_fs6)); then
+    log_Msg "Showing recon-all.v6.hires version"
+    local which_recon_all=$(which recon-all.v6.hires || true)
+    log_Msg ${which_recon_all}
+    recon-all.v6.hires -version
+
+    # Show tkregister version
+    log_Msg "Showing tkregister version"
+    which tkregister
+    tkregister -version
```
coalsont: Since you are editing virtually every line of this function anyway, I suggest reindenting it to 4-space conventions, like you did for the function below.  I prefer 4-space, though not to the extent of reindenting entire files (which would make "git blame" less useful).

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -337,54 +337,62 @@ for Hemisphere in L R ; do
...
         for Surface in sphere.reg sphere ; do
-        mris_convert "$FreeSurferFolder"/surf/"$hemisphere"h."$Surface" "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$Surface".native.surf.gii
-        ${CARET7DIR}/wb_command -set-structure "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$Surface".native.surf.gii ${Structure} -surface-type SPHERICAL
-        done
-        ${CARET7DIR}/wb_command -add-to-spec-file "$AtlasSpaceFolder"/"$NativeFolder"/"$Session".native.wb.spec $Structure "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".sphere.native.surf.gii
+            mris_convert "$FreeSurferFolder"/surf/"$hemisphere"h."$Surface" "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$Surface".native.surf.gii
+            ${CARET7DIR}/wb_command -set-structure "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$Surface".native.surf.gii ${Structure} -surface-type SPHERICAL
+            done
+            ${CARET7DIR}/wb_command -add-to-spec-file "$AtlasSpaceFolder"/"$NativeFolder"/"$Session".native.wb.spec $Structure "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".sphere.native.surf.gii
```
coalsont: Indentation fix went two lines too far.  To be clear, I do support fixing ambiguous/broken indentation like you did in this file.

```suggestion
        done
        ${CARET7DIR}/wb_command -add-to-spec-file "$AtlasSpaceFolder"/"$NativeFolder"/"$Session".native.wb.spec $Structure "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere".sphere.native.surf.gii
```

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -337,54 +337,62 @@ for Hemisphere in L R ; do
...
+            i=$(( i+1 ))
+            done
 
-        #Create midthickness by averaging white and pial surfaces and use it to make inflated surfacess
-        for Folder in "$T1wFolder" "$AtlasSpaceFolder" ; do
-        ${CARET7DIR}/wb_command -surface-average "$Folder"/"$NativeFolder"/"$Session"."$Hemisphere".midthickness.native.surf.gii -surf "$Folder"/"$NativeFolder"/"$Session"."$Hemisphere".white.native.surf.gii -surf "$Folder"/"$NativeFolder"/"$Session"."$Hemisphere".pial.native.surf.gii
-        ${CARET7DIR}/wb_command -set-structure "$Folder"/"$NativeFolder"/"$Session"."$Hemisphere".midthickness.native.surf.gii ${Structure} -surface-type ANATOMICAL -surface-secondary-type MIDTHICKNESS
-        ${CARET7DIR}/wb_command -add-to-spec-file "$Folder"/"$NativeFolder"/"$Session".native.wb.spec $Structure "$Folder"/"$NativeFolder"/"$Session"."$Hemisphere".midthickness.native.surf.gii
+            #Create midthickness by averaging white and pial surfaces and use it to make inflated surfacess
+            for Folder in "$T1wFolder" "$AtlasSpaceFolder" ; do
```
coalsont: The `done` and `for` (and comment) are indented too much.

FreeSurfer/FreeSurferPipeline.sh
demsarjure: Agreed, I changed the error message.

Yes FS8 works without a GPU, but the accelerated steps are supposed to be much slower.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -150,6 +150,10 @@ opts_AddOptional '--conf2hires' 'conf2hiresString' 'TRUE/FALSE' "Indicates that
 
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData or LegacyStyleData' "Controls whether the HCP acquisition and processing guidelines should be treated as requirements.  'HCPStyleData' (the default) follows the processing steps described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion.  'LegacyStyleData' allows additional processing functionality and use of some acquisitions that do not conform to 'HCP-Style' expectations.  In this script, it allows not having a high-resolution T2w image." "HCPStyleData"
 
+opts_AddOptional '--high-myelin' 'HighMyelin' "High Myelin" 'Value of the high myelin extra recon-all parameter, relevant if using FreeSurfer 7 and above. By default it will be automatically set to 0.3 for FS7 and FS8 and disbaled for FS6.' "AUTO"
```
coalsont:
```suggestion
opts_AddOptional '--high-myelin' 'HighMyelin' "High Myelin" 'Value of the high myelin extra recon-all parameter, relevant if using FreeSurfer 7 and above. By default it will be automatically set to 0.3 for FS7 and FS8 and disabled for FS6.' "AUTO"
```

glasserm:
Is this PR fully tested?

demsarjure:
> Is this PR fully tested?

Yes, this is tested. All results for Erin's review were produced with this branch. Do we need to wait for any of the NHP things before merging into the master?

coalsont:
Since it contains a change to a PostFS script, it should wait until NHP PostFS (which should be close).  It will probably say there are merge conflicts, but your edits to that script are just renaming one output file and reindentation, so it would be best to handle the conflicts in this PR.

demsarjure:
> Since it contains a change to a PostFS script, it should wait until NHP PostFS (which should be close). It will probably say there are merge conflicts, but your edits to that script are just renaming one output file and reindentation, so it would be best to handle the conflicts in this PR.

Agreed. Let me wait until NHP PostFS makes it into the master then.

glasserm (approved):

coalsont:
The web editor showed only one conflict, so I went ahead and fixed it, plus some stray indentation from NHP PostFS.  Looks like NHP already fixed a lot of the indentation that was in this PR, which is why it no longer shows in the diff.  Do we want to test NHP against fs8, or is the existing human testing enough to justify merging?

demsarjure:
If needed, I can test FS8 with NHP as well. Let me know. I believe I have one monkey dataset lying around from some previous tests. I will be away for a week, can test once I return.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -311,26 +330,51 @@ validate_freesurfer_version()
...
 
 # Show tool versions
 show_tool_versions
 
-# Validate version of FreeSurfer in use
-validate_freesurfer_version
+# Configure the use of FreeSurfer v6 custom tools
+if ((use_fs6)); then
+    configure_custom_tools
+fi
```
coalsont, line 363: `configure_custom_tools` modifies `$PATH`, and used to be before `show_tool_versions`, which looks for the custom `recon-all.v6.hires` on `$PATH`.  I think FS6 will fail with it in this order.

demsarjure: Fixed.

For NHP we support only FS6 as there is another custom tooling that is NHP specific, do we want to also ad FS7 and 8 support in there?

glasserm:
Did we sort out the cause of the problem with the data sent to Erin?

demsarjure:
> Did we sort out the cause of the problem with the data sent to Erin?

I will share the data that was processed with the latest version of master + fs8 branch today, so you can check.

FreeSurfer/FreeSurferPipeline.sh
coalsont: Are we planning to test fs7 and 8 on NHP?

demsarjure: @glasserm do we leave NHP support for the custom FS6 tooling only, or do we also want to test 7 and 8? This is important because I also need to adapt the new NHP FS script in this case.

glasserm: This is a question for @takuya-hayashi.

takuya-hayashi: You can test it if you want, although I don't think it is very easy to adapt the new FS. Let me know if you need the data.

demsarjure: If we expect that non adapted results are worse than the custom fs6 ones, then we can print a warning in case NHP is executed with FS7/8? Similar to the topup mismatch functionality.

glasserm: I think that we probably should disable it unless it is thoroughly vetted.

demsarjure: OK, let me code this solution then.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -288,13 +288,30 @@ log_Msg "Preparing the folder structure"
...
+  session_count=${#SessionsArray[@]}
+
+  if (( session_count > 1 )); then
+    # Build a common T2w reference for multiple sessions.
+    "$HCPPIPEDIR/FreeSurfer/scripts/MakeAverageT2w.sh" "$StudyFolder" "$SubjectID" "$Sessions" "$TemplateID"
+    bootstrap_t2w_image="${StudyFolder}/${SubjectID}.long.${TemplateID}/T2w/bootstrap_average.nii.gz"
+  else
+    # With one session, use that session's T2w directly.
+    bootstrap_t2w_image="${StudyFolder}/${SessionsArray[0]}/T1w/T2w_acpc_dc_restore.nii.gz"
+  fi
```
mharms: Did @mmilch01 review this?  In general, in the longitudinal stream, "single timepoints" are supposed to go through all the same processing and transforms, to avoid introducing any bias between the cross-sectional and longitudinal processing for any singleton sessions.

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -150,6 +150,10 @@ opts_AddOptional '--conf2hires' 'conf2hiresString' 'TRUE/FALSE' "Indicates that
 
 opts_AddOptional '--processing-mode' 'ProcessingMode' 'HCPStyleData or LegacyStyleData' "Controls whether the HCP acquisition and processing guidelines should be treated as requirements.  'HCPStyleData' (the default) follows the processing steps described in Glasser et al. (2013) and requires 'HCP-Style' data acquistion.  'LegacyStyleData' allows additional processing functionality and use of some acquisitions that do not conform to 'HCP-Style' expectations.  In this script, it allows not having a high-resolution T2w image." "HCPStyleData"
 
+opts_AddOptional '--high-myelin' 'HighMyelin' "High Myelin" 'Value of the high myelin extra recon-all parameter, relevant if using FreeSurfer 7 and above. By default it will be automatically set to 0.3 for FS7 and FS8 and disabled for FS6.' "AUTO"
+
+opts_AddOptional '--gpu' 'gpuString' 'Boolean' "Specify whether to use the GPU-enabled version of recon-all. Defaults to True for FS7 and FS8 and False for FS6." "AUTO"
```
mharms: Is GPU relevant for FS7?  Looking at the `recon-all` script for 7.4.1, I'm not finding a single mention of "gpu".

FreeSurfer/FreeSurferPipeline.sh
```diff
@@ -569,6 +623,16 @@ if ((conf2hires)); then
...
 fi
 
+# HighMyelin
+if [[ "${HighMyelin}" != "" ]]; then
+    recon_all_cmd+=(-high-myelin "${HighMyelin}")
+fi
+
+# gpu or not?
+if ((gpu)); then
+    recon_all_cmd+=(-gpu)
```
mharms: Am I missing something?  I'm not finding a `case: "-gpu":` statement in `recon-all` of 8.2.0, so what is this doing?

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -59,6 +59,8 @@ opts_AddMandatory '--sessions' 'Sessions' 'sessions' "@ separated list of sessio
 opts_AddMandatory '--longitudinal-template' 'TemplateID' 'template-id' "Longitudinal template label"
 opts_AddOptional '--use-T2w' 'UseT2wString' 'boolean' "Set to 0/false/no for no T2-weighted processing [1]" "1"
 opts_AddOptional '--seed' 'recon_all_seed' "Seed" "recon-all seed value"
+opts_AddOptional '--high-myelin' 'HighMyelin' "High Myelin" 'Value of the high myelin extra recon-all parameter, relevant if using FreeSurfer 7 and above. By default it will be automatically set to 0.3 for FS7 and FS8 and disabled for FS6.' "AUTO"
+opts_AddOptional '--gpu' 'gpuString' 'Boolean' "Specify whether to use the GPU-enabled version of recon-all. Defaults to True for FS7 and FS8 and False for FS6." "AUTO"
```
mharms: Ditto earlier comment (no "gpu" mentioned anywhere in `recon-all` from FS 7.4.1)

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -284,9 +338,11 @@ if (( start_stage < 1 )); then
   log_Msg "Creating the base template: ${TemplateID}"
   # ----------------------------------------------------------------------
 
-  # note that -conf2hires option is not supported for longitudinal base template as of Freesurfer 6,
-  # although it is supported for longitudinal timepoints
-  recon_all_cmd="recon-all.v6.hires"
+  if ((use_fs6)); then
```
mharms, line 284: Please add @mmilch01 's comment back in, which is important relevant info.

mharms: (i.e., regarding -conf2hires not being supported for longitudinal base template)

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -298,6 +354,14 @@ if (( start_stage < 1 )); then
     recon_all_cmd+=" -norandomness -rng-seed ${recon_all_seed}"
   fi
 
+  if [[ "${HighMyelin}" != "" ]]; then
+    recon_all_cmd+=" -high-myelin ${HighMyelin}"
+  fi
+
+  if ((gpu)); then
+    recon_all_cmd+=" -gpu"
```
mharms: Ditto earlier comment.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -312,40 +376,52 @@ if (( start_stage < 1 )); then
...
-    recon_all_cmd="recon-all.v6.hires"
+    if ((use_fs6)); then
+      recon_all_cmd="recon-all.v6.hires"
+    else
+      recon_all_cmd="recon-all"
+    fi
     recon_all_cmd+=" -sd ${LongDIR}"
-    recon_all_cmd+=" -long ${Session} ${TemplateID} -all -conf2hires"
-    
+    recon_all_cmd+=" -long ${Session} ${TemplateID} -all"
```
mharms: Why did we drop `-conf2hires` here from the FS6 runs?

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -312,40 +376,52 @@ if (( start_stage < 1 )); then
...
+      exit -1
     fi
-            
+
+    if [[ "${HighMyelin}" != "" ]]; then
+      recon_all_cmd+=" -high-myelin ${HighMyelin}"
+    fi
+
+    if ((gpu)); then
+      recon_all_cmd+=" -gpu"
```
mharms: Ditto earlier.

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -339,8 +339,16 @@ for Hemisphere in L R ; do
             fsname=$(echo $Map | cut -d "@" -f 1)
             wbname=$(echo $Map | cut -d "@" -f 2)
             mapname=$(echo $Map | cut -d "@" -f 3)
-            
+
             mris_convert -c "$FreeSurferFolder"/surf/"$hemisphere"h."$fsname" "$FreeSurferFolder"/surf/"$hemisphere"h.white "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$wbname".native.shape.gii
+
+            # minor output incompatibility between fs6 and fs7+
```
mharms:
```suggestion
            # mris_convert in fs7+ adds an extra '?h.' to the front of the file name in the converted output
```

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: Without this patch running longitudinal on a single timepoint crashed. Per FreeSurfer guidance you should run single timepoint sessions through long freesurfer to avoid additional bias, this is why this was added. But if there is a cleaner option here, I will change the code per @mmilch01's guidance.

FreeSurfer/FreeSurferPipeline.sh
demsarjure: See below.

FreeSurfer/FreeSurferPipeline.sh
demsarjure: There are a bunch of options for FS/recon-all that are not documented, or documenter poorly. GPU and high-myelin are two of them that we use, I believe Doug helped us with the implementation here about a year ago when we started this.

I believe that even FS6 recon-all has a (hidden) gpu flag. @coalsont dug this up in some old documentation somewhere and should be in one of the threads around here. You are just not using it as benefits are slim. This one definitely works, as processing crashes if you do not have CUDA support on the system and you use the -gpu flag. Since there is no documentation, let me test this through experimentation for both fs7 and fs8 and find a way that works. I will probably do this after 15th of June when I return from my vacation.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: done

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: resolving this to keep a single gpu thread going

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: @mmilch01 This question is for you, I did not change this line, it was changed by you 19 months ago.

mharms:
Big picture question:  Don't we generally want to be using `-hires` with FS 8 (and probably FS 7 as well)?  According to Gemini, if we don't include that, and we have acquired higher than 1.0 mm resolution, then the very first `mri_convert` step still does the classical "conform" to 1 mm step, and we lose our sub-millimeter resolution immediately and permanently.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
mharms: Exactly, but the change here is NOT sending the single T2w time point through the `MakeAverageT2w.sh` script.  @mmilch01 : Is that something we can meaningfully apply to a single timepoint, that would mimic the operations in that script?

FreeSurfer/FreeSurferPipeline.sh
mharms: `recon-all` in FS 6 has a flag `-use-gpu` (which yes, we didn't use).  My point though is that `recon-all` in FS8 never introduces a `-gpu` flag that I can find.  And per the FS Release notes, GPUs are not supposed to be required to make use of the deep learning modules/features. That's not to say that GPUs on FS 8 might not be an _option_ to accelerate things. Indeed, `recon-all` in FS 8 has a variable named `$UseGPU`, which when set then adds `--gpu` to the _call to specific commands_ within `recon-all`.  (I haven't been able to quite figure out how `UseGPU` gets set, since it isn't set directly in `recon-all` itself.  I see that they have introduced a new `recon-config.yaml` config file however, so perhaps that is part of the puzzle...)

FWIW, `-high-myelin` is just a regular option/flag within `recon-all`:
```
   case "-high-myelin":
      if($#argv < 1) goto arg1err;
      set HighMyelinFactor = $argv[1]; shift
      set UseHighMyelin = 1
      breaksw
```

FreeSurfer/LongitudinalFreeSurferPipeline.sh
mharms: ?? `-conf2hires` is in the current master (which is why it is showing up in the diff).

FreeSurfer/FreeSurferPipeline.sh
mharms: Ahh, I think I see it.  The `.yaml` file has the following:

```
UseGPU:
    value: 0
    flags: -gpu
    action: enable
    descr: Use GPU for those GPU-enabled programs
```
So, the definition of `-gpu` as a valid flag to use in `recon-all`, and the linkage of that flag to the setting of the `$UseGPU` variable used internally in `recon-all`, appears to be specified in the `.yaml` config file.

mharms:
Also, from what I can tell, what we've set up as the default for FS 8 in this PR is just a "standard" `recon-all` run, and doesn't take advantage of the newer deep learning-based tools (e.g., SynthMorph, SynthSeg).  Is that the intent?

Note that there is a new `recon-all-clinical.sh` script, which defaults to using the DL tools, and is described as:

> "Recon-all-like stream for clinical scans of arbitrary orientation/resolution/contrast"

FreeSurfer/LongitudinalFreeSurferPipeline.sh
coalsont: It looks like the merge commit on the `fs8` branch on May 7, bf1a78f60cd456507f84a76e382217729990de43, should have picked up the reinstatement of `-conf2hires` from master, but didn't.  Please make a commit that changes this line to match `master` (possibly without the trailing whitespace).  This is speculation, but perhaps an edit that removed trailing whitespace conflicted with the `-conf2hires` coming back in, and the conflict was resolved incorrectly.

I tend to try to use rebase to pick up changes from master, because the resulting branch can remain linear (merge commits make the history harder to understand).  But, I use 'squash' for merging PRs anyway, so at least the history of master is linear (and doesn't contain as many broken commits from work in progress edits, in case we ever needed to bisect to hunt down something).

coalsont:
> Also, from what I can tell, what we've set up as the default for FS 8 in this PR is just a "standard" `recon-all` run, and doesn't take advantage of the newer deep learning-based tools (e.g., SynthMorph, SynthSeg). Is that the intent?
> 
> Note that there is a new `recon-all-clinical.sh` script, which defaults to using the DL tools, and is described as:
> 
> > "Recon-all-like stream for clinical scans of arbitrary orientation/resolution/contrast"

In this context, "clinical" may mean "low quality".  The context I am familiar with for SynthSeg is dealing with low-quality data, such as portable scanners, brain-extracted postmortem, etc.

I assume the "standard recon-all run" remains the MGH-recommended approach for high-quality data (if something is an improvement on high-quality data, they would change it in recon-all, right?).

coalsont:
> Don't we generally want to be using `-hires` with FS 8 (and probably FS 7 as well)?

If I recall, that existed in fs6 as well, and didn't perform very well (and it has an effect on native mesh resolution, which may have caused things like inflated and/or sulc to be different for non-anatomical reasons).  We would need to test `-hires` versus `-conf2hires` again, which may only be worth the effort if MGH thinks it has made substantial improvements to how it works.

mharms:
> I assume the "standard recon-all run" remains the MGH-recommended approach for high-quality data (if something is an improvement on high-quality data, they would change it in recon-all, right?).

Not necessarily.  I don't know what their current recommendation is for "high-quality" data, but they could have opted to maintain the "traditional" approach as the default for now at least, solely for backwards-consistency considerations.

mharms:
> > Don't we generally want to be using `-hires` with FS 8 (and probably FS 7 as well)?
> 
> If I recall, that existed in fs6 as well, and didn't perform very well (and it has an effect on native mesh resolution, which may have caused things like inflated and/or sulc to be different for non-anatomical reasons). We would need to test `-hires` versus `-conf2hires` again, which may only be worth the effort if MGH thinks it has made substantial improvements to how it works.

Per the usage notes, `-hires` was completely revamped going from FS 6 to 7.  But not having either `-hires` or `-conf2hires` in FS 8 doesn't strike me as what we want as our default, from what I've read.

FreeSurfer/FreeSurferPipeline.sh
mharms: Regarding the defaults: Wouldn't it be more appropriate for the default for the GPU flag to be FALSE, since GPUs are not needed, even when using the new DL modules (which we are by default not using anyway).

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: Resolved.

FreeSurfer/FreeSurferPipeline.sh
demsarjure: I removed the GPU flag, so by default this is disabled and consitent with the FS6 implementation. Users can turn on GPU mode through extra recon all flags. This is more robust and works across all FS versions since the user needs to adapt the parameter value to the FS version. I think the GPU the flag changed a bit, it was --use-cuda or sth at one point and is now --gpu.

demsarjure:
> > > Don't we generally want to be using `-hires` with FS 8 (and probably FS 7 as well)?
> > 
> > 
> > If I recall, that existed in fs6 as well, and didn't perform very well (and it has an effect on native mesh resolution, which may have caused things like inflated and/or sulc to be different for non-anatomical reasons). We would need to test `-hires` versus `-conf2hires` again, which may only be worth the effort if MGH thinks it has made substantial improvements to how it works.
> 
> Per the usage notes, `-hires` was completely revamped going from FS 6 to 7. But not having either `-hires` or `-conf2hires` in FS 8 doesn't strike me as what we want as our default, from what I've read.

I believe I resolved all issues above. The only "open" question is the use of `-hires` or `-conf2hires` in fs7 and 8. I do not have much experience here so I will defer to your guidance.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: Isn't `MakeAverageT2w.sh` exactly what it says? An average across multiple time T2w time points. Pretty hard to do anything meaningful with a single data point in that context.

mharms: I'm not sure exactly what @mmilch01 implemented here, but for the "robust template" generation in the "normal" FS approach with a single timepoint, I think they do something like put the image into a "midpoint" space defined by the timepoint and the L/R reversed version of itself.

mharms: Yes, I was roughly correct with my description above of how FS handles this sitution.
From https://www.ncbi.nlm.nih.gov/pubmed/23123680:

> To address this issue we modified FreeSurfer's longitudinal framework to process
subjects with a single time point in the following manner: we created
a pose normalized (upright) version of the input images by symmetrically
registering it with its left–right reversed image into a mid-space
(Reuter et al., 2010), we then processed it as the subject-specific template
and used it for the initialization of subsequent image processing
steps, such as skull stripping. This ensures the input image from a subject with
a single scan undergoes the same processing and interpolation
steps as serial images in the longitudinal stream and thus makes results
comparable (see Supplementary Material).

coalsont:
The question of what to do about "longitudinal" T2w with a single session also remains.

demsarjure:
> The question of what to do about "longitudinal" T2w with a single session also remains.

True, waiting for Misha's input here.

glasserm:
With regard to the above comments:
1) recon-all clinical is not relevant to the HCP Pipelines at this time.  This is, for example, what we use in HMBA to make surfaces from lowres hyperfine scans.
2) We need to test all the permutations (standard recon-all, -conf-to-hires, -hires
3) With regard to a single session and longitudinal mode, I assume we would just use the T2w data that is available and let FreeSurfer do whatever it is going to do.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
demsarjure: This will be fixed on Misha's branch now. Resolving here.

FreeSurfer/LongitudinalFreeSurferPipeline.sh
```diff
@@ -287,18 +310,33 @@ if (( start_stage < 1 )); then
...
   done
   recon_all_cmd+=" -all"
 
+  if ((use_fs6)); then
+    recon_all_cmd="recon-all.v6.hires"
+  else
+    recon_all_cmd="recon-all"
+  fi
+  recon_all_cmd+=" -sd ${TemplateT1wDir}"
+
```
coalsont: This looks like a bad merge, `-base` and all `-tp`s get erased by resetting `recon_all_cmd` to the executable name only.

coalsont: ~Also it sets `-sd` to something different.~  Looks like the `-sd` change is intentional, it matches master.

demsarjure: Nice catch. Fixed!

PostFreeSurfer/scripts/FreeSurfer2CaretConvertAndRegisterNonlinear.sh
```diff
@@ -339,19 +345,38 @@ for Hemisphere in L R ; do
...
             mris_convert -c "$FreeSurferFolder"/surf/"$hemisphere"h."$fsname" "$FreeSurferFolder"/surf/"$hemisphere"h.white "$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$wbname".native.shape.gii
+
+            # mris_convert in fs7+ adds an extra '?h.' to the front of the file name in the converted output
+            new_native_shape="$AtlasSpaceFolder"/"$NativeFolder"/"$hemisphere"h."$Session"."$Hemisphere"."$wbname".native.shape.gii
+            if [[ -e "$new_native_shape" ]]; then
+                legacy_native_shape="$AtlasSpaceFolder"/"$NativeFolder"/"$Session"."$Hemisphere"."$wbname".native.shape.gii
+                if [[ -e "$legacy_native_shape" ]]; then
+                    rm "$legacy_native_shape"
+                fi
+                mv "$new_native_shape" "$legacy_native_shape"
```
coalsont: In general, just use `-f` in scripts to deal with safeguards designed for interactive use, for `rm`, `mv`, `cp`, `ln`, etc, rather than making "if exists" checks all over the place.
```suggestion
                mv -f "$new_native_shape" "$legacy_native_shape"
```

glasserm:
@coalsont this is ready for merge.  After this, all components for the HCP Pipelines release are complete and we should work on making that release.

FreeSurfer/FreeSurferPipelineNHP.sh
```diff
@@ -680,35 +644,36 @@ if [ "$RunMode" -lt 2 ] ; then
...
-        rm -f "$SubjectDIR"/"$SubjectID"/mri/transforms/"$T2Type"raw.lta # remove this otherwise conf2hires will not update this - TH
-    else
-        T2Type="NONE"
-    fi
+			recon_all_initrun+=(-T2 "$(remove_ext "$T2wImage")_scaled.nii.gz")
+			T2Type=T2
+		fi
+		rm -f "$SubjectDIR"/"$SubjectID"/mri/transforms/"$T2Type"raw.lta # remove this otherwise conf2hires will not update this - TH
+	else
+		recon_all_pial=""
```
coalsont: I don't think this line was needed, it should already get set in every condition inside the only section that uses it.

global/scripts/CorrThick/metric_regression.py
```diff
@@ -304,7 +307,7 @@ def process_roi(args):
     )
     with threadpool_limits(limits=1, user_api="blas"):
         t_m = t - np.matmul(coeff[0 : len(coeff) - 1], curv_matrix)
-    index = np.where(region == j)
+    index = np.where(region == j)[0][0]
```
coalsont, line 310: @nagedem is this change reasonable?

FreeSurfer/FreeSurferPipelineNHP.sh
coalsont: After a bit of digging, it looks like this merge:

https://github.com/Washington-University/HCPpipelines/commit/b38bad0111fb8ee778b70d9193426c2ca8e25db5?w=1#diff-4291baad5461be97401b2c973c18f83078bb2d9c104fba2cf4036319ff730d64R701

(NOTE: thanks to new github styling and the size of the merge commit, the highlighted change is actually inside a collapsed section by default, so expand to reveal line 701 on the post-change line numbers)

Didn't fully apply this change:

https://github.com/Washington-University/HCPpipelines/commit/5e055f47527f09575bc6b053e53e03d8b4ecc501

global/scripts/CorrThick/metric_regression.py
demsarjure: @coalsont 

In Python `np.where` with a single argument doesn't return an index, it returns a tuple of index arrays, one array per dimension of the input:

``` python
region = np.array([5, 7, 7, 9])

np.where(region == 7)          # (array([1, 2]),) - tuple, length 1
np.where(region == 7)[0]       # array([1, 2]) - the row indices
np.where(region == 7)[0][0]    # 1 - plain int (what we need)
```

Where returns all indexes of the element, above those are 1 and 2. In our case we have only one such index, but it is still an array with a single element, so we need to take the first item only.

Are there any other things for me to take a look at, I think you resolved some things on your own?

coalsont: By email, she agreed with the change.

PR #399, opened 2026-06-23
Adapt longitudinal stream to handle single time point.

mmilch01:
Adapt longitudinal stream to handle single time point.

### DIFF
--- a/FreeSurfer/scripts/MakeAverageT2w.sh
+++ b/FreeSurfer/scripts/MakeAverageT2w.sh
@@ -75,20 +75,37 @@ cmd_lta=()
 
 IFS='@' read -a SessionsArray <<<"$Sessions"
 
-for session in "${SessionsArray[@]}"; do
-    cmd_mov+=("$StudyFolder/$session/T1w/$session/mri/orig/T2raw.mgz")
-    cmd_lta+=("$TemplateDir/T2w/xfms/${session}_t2w2bootstrap_average.lta")
-done
+nSessions=${#SessionsArray[@]}
+if (( nSessions < 1 )); then 
+    log_Err_Abort "At least one longitudinal session is required"
 
-echo "${cmd_mov[@]}"
+elif (( nSessions > 1 )); then #normal case.
+    for session in "${SessionsArray[@]}"; do
+        cmd_mov+=("$StudyFolder/$session/T1w/$session/mri/orig/T2raw.mgz")
+        cmd_lta+=("$TemplateDir/T2w/xfms/${session}_t2w2bootstrap_average.lta")
+    done
 
-cmd+=(--mov "${cmd_mov[@]}")
-cmd+=(--template "$TemplateDir/T2w/bootstrap_average.nii.gz" --iscale --satit)
-cmd+=(--lta "${cmd_lta[@]}")
+    echo "${cmd_mov[@]}"
 
-mkdir -p "$TemplateDir/T2w/xfms"
+    cmd+=(--mov "${cmd_mov[@]}")
+    cmd+=(--template "$TemplateDir/T2w/bootstrap_average.nii.gz" --iscale --satit)
+    cmd+=(--lta "${cmd_lta[@]}")
 
-echo "${cmd[@]}"
-"${cmd[@]}"
+    mkdir -p "$TemplateDir/T2w/xfms"
 
+    echo "${cmd[@]}"
+    "${cmd[@]}"
+
+else #exactly one session, use single T2w to create template.
+    session=${SessionsArray[0]}
+    mgz="$StudyFolder/$session/T1w/$session/mri/orig/T2raw.mgz"
+    nii="$StudyFolder/$session/T2w/T2w.nii.gz"
+    mkdir -p "$TemplateDir/T2w/xfms"
+    cp -f "$nii" "$TemplateDir/T2w/bootstrap_average.nii.gz"
+    tmp=$(mktemp identity_fslXXX.mat)
+    trap 'rm -f "$tmp"' EXIT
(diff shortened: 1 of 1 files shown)


FreeSurfer/scripts/MakeAverageT2w.sh
```diff
@@ -75,20 +75,37 @@ cmd_lta=()
...
+    cmd+=(--mov "${cmd_mov[@]}")
+    cmd+=(--template "$TemplateDir/T2w/bootstrap_average.nii.gz" --iscale --satit)
+    cmd+=(--lta "${cmd_lta[@]}")
+
+    mkdir -p "$TemplateDir/T2w/xfms"
+
+    echo "${cmd[@]}"
+    "${cmd[@]}"
+
+else #exactly one session, use single T2w to create template.
```
mharms, line 99: Maybe add a more detailed comment here to capture our thinking about this?
```suggestion
else #exactly one session, use single T2w to create template.
# This reflects a subtle difference in the longitudinal handling of the 'base creation for a single-session.
# BUT, the T2 gets interpolated already at least once anyway (to register to the T1),
# so there is always at least 1 interpolation applied to the T2.
# Plus, to the extent there is a difference, it is only in the creation of the 'base', and not
# in the longitudinal processing of the individual time points themselves.
```

FreeSurfer/scripts/MakeAverageT2w.sh
```diff
@@ -75,20 +75,37 @@ cmd_lta=()
...
+    session=${SessionsArray[0]}
+    mgz="$StudyFolder/$session/T1w/$session/mri/orig/T2raw.mgz"
+    nii="$StudyFolder/$session/T2w/T2w.nii.gz"
+    mkdir -p "$TemplateDir/T2w/xfms"
+    cp -f "$nii" "$TemplateDir/T2w/bootstrap_average.nii.gz"
+    tmp=$(mktemp identity_fslXXX.mat)
+    trap 'rm -f "$tmp"' EXIT
```
coalsont: Instead of manually using `trap` and `mktemp`, source `tempfiles.shlib` at the top, then do this (note, minimum of 6 X's, if I recall that is for mac compatibility):

```suggestion
    tempfiles_create identity_fsl_XXXXXX.mat tmp
```

Example:

https://github.com/Washington-University/HCPpipelines/blob/1c597afdb188c4067707537db25868ac79a9b41d/global/scripts/VolumeToCIFTI.sh#L13
https://github.com/Washington-University/HCPpipelines/blob/1c597afdb188c4067707537db25868ac79a9b41d/global/scripts/VolumeToCIFTI.sh#L179-L183

coalsont: You also need to add this near the top:

```
source "$HCPPIPEDIR/global/scripts/tempfiles.shlib" "$@"
```

mmilch01: Fixed this in 5df4ebfa730003e24ef6035664780e3056d73bdd

FreeSurfer/scripts/MakeAverageT2w.sh
```diff
@@ -75,20 +76,41 @@ cmd_lta=()
...
+# so there is always at least 1 interpolation applied to the T2.
+# Plus, to the extent there is a difference, it is only in the creation of the 'base', and not
+# in the longitudinal processing of the individual time points themselves.
+    session=${SessionsArray[0]}
+    mgz="$StudyFolder/$session/T1w/$session/mri/orig/T2raw.mgz"
+    nii="$StudyFolder/$session/T2w/T2w.nii.gz"
+    mkdir -p "$TemplateDir/T2w/xfms"
+    cp -f "$nii" "$TemplateDir/T2w/bootstrap_average.nii.gz"
+    tempfiles_create identity_fsl_XXXXXX.mat tmp
+    printf '1 0 0 0\n0 1 0 0\n0 0 1 0\n0 0 0 1\n' > "$tmp"
```
coalsont: I hadn't twigged that this is just an identity matrix.  We often use this instead of making one from scratch:

https://github.com/Washington-University/HCPpipelines/blob/9d740bdb0c265f40ad368533a4788d8e4e14acef/PreFreeSurfer/scripts/AnatomicalAverage.sh#L113

mmilch01: I changed the code to use the stock FSL identity matrix in 64ed905bc9c019baa9bddf7a8eaeca0b01797aa6.

coalsont:
Code looks reasonable to me.  I don't fully follow the consequences of longitudinal freesurfer's behavior of doing a verticalization to "match the number of resamplings" or something, versus us copying the cross-sectional file.

Does this need testing?

FreeSurfer/scripts/MakeAverageT2w.sh
```diff
@@ -75,20 +75,39 @@ cmd_lta=()
...
+    cmd+=(--template "$TemplateDir/T2w/bootstrap_average.nii.gz" --iscale --satit)
+    cmd+=(--lta "${cmd_lta[@]}")
+
+    mkdir -p "$TemplateDir/T2w/xfms"
+
+    echo "${cmd[@]}"
+    "${cmd[@]}"
+
+else #exactly one session, use single T2w to create template.
+    # This reflects a subtle difference in the longitudinal handling of the 'base creation for a single-session.
```
mharms:
```suggestion
    # This reflects a subtle difference in the longitudinal handling of the 'base' creation for a single-session.
```

mmilch01:
> Code looks reasonable to me. I don't fully follow the consequences of longitudinal freesurfer's behavior of doing a verticalization to "match the number of resamplings" or something, versus us copying the cross-sectional file.
> 
> Does this need testing?

The current revision is tested to run without error.

PR #401, opened 2026-07-15
Add support for Arena areal classifier

YuboW0:


ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+    echo "
+$log_ToolName: classifies functional areas in a single subject enabled by multiple versions of the areal classifier, including the mlp classifier designed in Glasser 2016, DOI 10.1038/nature18933
+
+Usage: $log_ToolName PARAMETER...
+
+PARAMETERs are [ ] = optional; < > = user supplied value
+"
+    #automatic argument descriptions
+    opts_ShowArguments
+}
```
coalsont: Use the new usage description style:

```suggestion
opts_SetScriptDescription "classifies functional areas in a single subject enabled by multiple versions of the areal classifier, including the mlp classifier designed in Glasser 2016, DOI 10.1038/nature18933"
```

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+        ;;
+    (*)
+        log_Err_Abort "unrecognized matlab mode: $matlab_mode"
+        ;;
+esac
+
+# check RSN type
+if [[ "$RegressMethod" != "PFM_weighted" ]]; then
+    log_Err_Abort "Only weighted regression is supported!"
+fi
```
coalsont, line 75: Then why does the option exist?

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+    "$HCPPIPEDIR"/global/scripts/download_areal_classifier_dependencies.sh "MMP_ROIs"
+fi
+
+# download model weights if necessary
+if [ "$ClassifierVersion" == "MATLAB" ]; then
+    filecount_model_weights=$(find "$HCPPIPEDIR"/ArealClassifier/data/mlp_classifier -maxdepth 1 -type f | wc -l)
+    if [[ "$filecount_model_weights" -lt 368 || "$ForceDownloadWeight" -ne 0 ]]; then
+        "$HCPPIPEDIR"/global/scripts/download_areal_classifier_dependencies.sh "MATLAB"
+    fi
+elif [ "$ClassifierVersion" == "ARENA_v1" ]; then
```
coalsont, line 96: Consider making a `--force-download=TRUE` flag in `download_areal_classifier_dependencies.sh`, and having it internally check whether they are already downloaded (if not requested to be forced, which would save a small amount of time and IOPS).  This would put the details of the check for completeness into the same file as the URL, so if we ever needed to modify the zips, both changes would be in a single script.

You could also have it internally unconditionally look at the ROIs.  Then here, you could simply do:

```
"$HCPPIPEDIR"/global/scripts/download_areal_classifier_dependencies.sh --model="$ClassifierVersion" --force-download="$ForceDownloadWeight"
```
and get both pieces.

For that specific download script, we could skip the `if ((pipedirguessed))` error of the usual newopts template, so that it can be called outside of a pipeline.  In some setups (probably including qunex), the pipelines folder isn't writable while the pipelines are running, so making the download script work without the full SetUpHCP... (so that it can easily be done in advance) is important.  If we want qunex to support all 3 models, then we might want a `--model=ALL` possibility, to simplify the qunex build process (and make it future-proof).

coalsont: In fact, if someone just goes ahead and launches cluster jobs, but their pipelines clone doesn't have these (and is writable), then they will immediately have jobs stomping on each other's downloads, in a folder (and probably filesystem) that isn't where the imaging data is.  Maybe the pipelines shouldn't be trying to automatically trigger the download by default.

coalsont: ...which also means that a check for whether to tell the user to go download them should be separate from the download script after all...

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+opts_AddMandatory '--regress-outstring' 'RegressOutString' 'string' "part of filename for regression intermediates"
+opts_AddOptional '--regress-method' 'RegressMethod' 'string' "PFM_weighted"
+opts_AddOptional '--regress-spectra' 'OutputSpectra' 'number' 'number of frequency bins for spectra (usually number of input timepoints)'
+opts_AddOptional '--regress-template-folder-pattern' 'RegressTemplatePattern' 'prefix@postfix' "string to set the folder names to inside of the regress template folder, use @ instead of the dimensionality" 'rfMRI_REST_Atlas_MSMAll_hp2000_clean_rclean_tclean_PCA_WF6.ica_d@_ROW_vn'
+opts_AddOptional '--output-suffix' 'FeatureOutputName' 'name' "suffix for output files, like NoTask"
+opts_AddMandatory '--classifier-training' 'TrainedFolder' 'path' "classifier training weights"
+opts_AddMandatory '--classifier-version' 'ClassifierVersion' 'string' "either the matlab version 'MATLAB' or the python version 'ARENA_v1', 'ARENA_v2'"
+# opts_AddOptional '--python-singularity' 'PythonSingularity' 'string' "the file path of the singularity" "$HCPPIPEDIR/ArealClassifier/singularity/hcp_python_singularity.simg"
```
coalsont, line 39: If we aren't going to use a singularity container within this pipeline, we should probably remove this and all related stuff.

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+AreaName="V1_ROI"
+AreaIINames="NONE"
+AxisOneBorder="Meridians3New_line_Gradient_ROI" #In ROIs Folder
+AxisTwoBorder="Eccentricity3New_line_Gradient_ROI" #In ROIs Folder
+AxisOneLeftParameters="-3@-2@-1%(-Var-3)"
+AxisOneRightParameters="3@2@1%(-Var+3)"
+AxisTwoLeftParameters="1@1.5@2%(Var-1)"
+AxisTwoRightParameters="1@1.5@2%(Var-1)"
+LinearGradientSmoothingFWHM="2"
+GradientSmoothingFWHM=`echo "1 * ( 2 * ( sqrt ( 2 * l ( 2 ) ) ) )" | bc -l` #Default is Sigma=1mm
```
coalsont:
```suggestion
GradientSmoothingFWHM=`echo "1 * (2 * (sqrt(2 * l(2))))" | bc -l` #Default is Sigma=1mm
```
I deliberately edited that formatting out of all of the pipelines...

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+            else
+                interpreter=(octave-cli -q --no-window-system)
+            fi
+            mPath="${HCPPIPEDIR}/ArealClassifier/scripts"
+            mGlobalPath="${HCPPIPEDIR}/global/matlab"
+
+            #maybe _$RegName should be $RegString, but needs fixing in combine, too
+            #CHECK INPUT NAME, currently assuming TopographicRegression magically makes it
+            matlabCode="addpath '$mPath'; addpath '$mGlobalPath';
+            computeMisc('${inputdtseriestxt}', \
```
coalsont, line 282: This is not a particularly good function name.

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+            log_Err_Abort "unrecognized matlab mode: $matlab_mode"
+            ;;
+    esac
+fi
+#average dropouts across runs, map to surface, put in cifti, since nothing else uses them this way
+#these temp names are a bit messy, but who cares
+volmerged="$(mktemp --tmpdir XXXXXX.dropouts.nii.gz)"
+leftdropouts="$volmerged.left.func.gii"
+rightdropouts="$volmerged.right.func.gii"
+tempfiles_add "$volmerged" "$leftdropouts" "$rightdropouts"
```
coalsont:
```suggestion
tempfiles_create XXXXXX.dropouts.nii.gz volmerged
leftdropouts="$volmerged.left.func.gii"
rightdropouts="$volmerged.right.func.gii"
tempfiles_add "$leftdropouts" "$rightdropouts"
```

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+        --hemisphere="L" \
+        --out-metric="$leftdropouts"
+    "${basecommand[@]}" \
+        --hemisphere="R" \
+        --out-metric="$rightdropouts"
+    wb_command -cifti-create-dense-from-template \
+        "$HCPPIPEDIR"/global/templates/91282_Greyordinates/91282_Greyordinates.dscalar.nii \
+        "${ResultsFolder}/${OutrfMRIName}/${OutrfMRIName}_Atlas_MSMAll_dropouts.dscalar.nii" \
+        -metric CORTEX_LEFT "$leftdropouts" \
+        -metric CORTEX_RIGHT "$rightdropouts"
```
coalsont, line 340: Looks like the cifti is the only important output here, consider replacing most of this (`MapVolumeToSurface.sh` until here) with a call to https://github.com/Washington-University/HCPpipelines/blob/master/global/scripts/VolumeToCIFTI.sh instead.

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+        if [ "$LabelTableString" = "Subregions" ]; then
+            LabelTable="$HCPPIPEDIR/ArealClassifier/data/$Hem.${parcel_type}_table.txt"
+        fi
+
+        InputDilROIs="$HCPPIPEDIR/ArealClassifier/data/ROIs_dil.$Hem.32k_fs_LR.dscalar.nii"
+        #InputFeatureTypes="$HCPPIPEDIR/ArealClassifier/data/feature_types.$Hem.txt" # raw version in Glasser 2016
+        InputFeatureTypes="$HCPPIPEDIR/ArealClassifier/data/feature_types.${Hem}_90.txt"
+
+        AreaNamesTemp="$(mktemp --tmpdir XXXXXX.areanames.txt)"
+        tempfiles_add "$AreaNamesTemp"
```
coalsont, line 406:
```suggestion
        tempfiles_create areanames_XXXXXX.txt AreaNamesTemp
```

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,536 @@
...
+        sed -n 'p;n' "$LabelTable" > "$AreaNamesTemp"
```
coalsont, line 408:
```suggestion
        #get the first line of each pair of lines
        sed -n 'p;n' "$LabelTable" > "$AreaNamesTemp"
```

ArealClassifier/scripts/ClassificationCleanup.sh
```diff
@@ -0,0 +1,295 @@
...
+
+#Find The Biggest (FTB) area, make initial label file, and convert labels to ROIs
+wb_command -cifti-reduce "$workdir/${tempbase}_AreaClasses.$dscalarend" INDEXMAX "$workdir/${tempbase}_AreaClasses_ftb.$dscalarend"
+wb_command -cifti-label-import "$workdir/${tempbase}_AreaClasses_ftb.$dscalarend" ${AreaColorsFile} "$workdir/${tempbase}_AreaClasses_ftb.$dlabelend"
+wb_command -cifti-all-labels-to-rois "$workdir/${tempbase}_AreaClasses_ftb.$dlabelend" 1 "$workdir/${tempbase}_Areas_raw.$dscalarend"
+
+#Erode everything a little to make "core areas" 2mm
+wb_command -cifti-math "(var - 1) * -1" "$workdir/${tempbase}_Areas_rawinv.$dscalarend" -var var "$workdir/${tempbase}_Areas_raw.$dscalarend" 
+wb_command -cifti-dilate "$workdir/${tempbase}_Areas_rawinv.$dscalarend" COLUMN ${ErodeDilateDistance} ${ErodeDilateDistance} "$workdir/${tempbase}_Areas_core_inv.$dscalarend" -${hemisphereword}-surface "${T1wDownSampleFolder}/${Subject}.${Hem}.midthickness${RegSTRING}.${LowResMesh}k_fs_LR.surf.gii" -nearest
+wb_command -cifti-math "(var - 1) * -1" "$workdir/${tempbase}_Areas_core.$dscalarend" -var var "$workdir/${tempbase}_Areas_core_inv.$dscalarend"
```
coalsont, line 133: This will need testing, but avoids a few bits of silliness (at the time, some of this may have been needed due to a bug in `-cifti-erode`).
```suggestion
wb_command -cifti-erode "$workdir/${tempbase}_Areas_raw.$dscalarend" COLUMN ${ErodeDilateDistance} ${ErodeDilateDistance} "$workdir/${tempbase}_Areas_core.$dscalarend" -${hemisphereword}-surface "${T1wDownSampleFolder}/${Subject}.${Hem}.midthickness${RegSTRING}.${LowResMesh}k_fs_LR.surf.gii"
```

ArealClassifier/scripts/ClassificationCleanup.sh
```diff
@@ -0,0 +1,295 @@
...
+wb_command -cifti-math "(var - 1) * -1" "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))inv.$dscalarend" -var var "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2)).$dscalarend" 
+wb_command -cifti-dilate "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))inv.$dscalarend" COLUMN $((${ErodeDilateDistance}*2)) $((${ErodeDilateDistance}*2)) "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))inv_ero$((${ErodeDilateDistance}*2)).$dscalarend" -${hemisphereword}-surface "${T1wDownSampleFolder}/${Subject}.${Hem}.midthickness${RegSTRING}.${LowResMesh}k_fs_LR.surf.gii" -nearest
+wb_command -cifti-math "(var - 1) * -1" "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))_ero$((${ErodeDilateDistance}*2)).$dscalarend" -var var "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))inv_ero$((${ErodeDilateDistance}*2)).$dscalarend"
```
coalsont, line 146: Similar, could convert to `-cifti-erode`.

ArealClassifier/scripts/ClassificationCleanup.sh
```diff
@@ -0,0 +1,295 @@
...
+
+#Mask out core ROI from all areas and then add core back in for each area
+wb_command -cifti-math "(dilero * (((coreroi > 0) - 1) * -1)) + (raw * (coreroi > 0))" "$workdir/${tempbase}_Areas_joined.$dscalarend" -var raw "$workdir/${tempbase}_Areas_raw.$dscalarend" -var dilero "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))_ero$((${ErodeDilateDistance}*2)).$dscalarend" -var coreroi "$workdir/${tempbase}_Areas_core_roi.$dscalarend" -select 1 1 -repeat
```
coalsont, line 149:
```suggestion
wb_command -cifti-math '(dilero * (!(coreroi > 0)) + (raw * (coreroi > 0))' "$workdir/${tempbase}_Areas_joined.$dscalarend" -var raw "$workdir/${tempbase}_Areas_raw.$dscalarend" -var dilero "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2))_ero$((${ErodeDilateDistance}*2)).$dscalarend" -var coreroi "$workdir/${tempbase}_Areas_core_roi.$dscalarend" -select 1 1 -repeat
```

ArealClassifier/scripts/ClassificationCleanup.sh
```diff
@@ -0,0 +1,295 @@
...
+wb_command -cifti-reduce "$workdir/${tempbase}_Areas_core.$dscalarend" MAX "$workdir/${tempbase}_Areas_core_roi.$dscalarend"
+
+#Find clusters at 25mm on Raw, find number of clusters
+wb_command -cifti-find-clusters "$workdir/${tempbase}_Areas_raw.$dscalarend" 0.99 25 0.99 125 COLUMN "$workdir/${tempbase}_Areas_raw_fc.$dscalarend" -${hemisphereword}-surface "${T1wDownSampleFolder}/${Subject}.${Hem}.midthickness${RegSTRING}.${LowResMesh}k_fs_LR.surf.gii"
+
+#Mask Raw clusters with Raw Clusters FC
+wb_command -cifti-math "var * (ROI > 0)" "$workdir/${tempbase}_Areas_raw_masked.$dscalarend" -var var "$workdir/${tempbase}_Areas_raw.$dscalarend" -var ROI "$workdir/${tempbase}_Areas_raw_fc.$dscalarend"
+
+#Dilate and Erode to connect to connect things 4mm
+wb_command -cifti-dilate "$workdir/${tempbase}_Areas_raw.$dscalarend" COLUMN $((${ErodeDilateDistance}*2)) $((${ErodeDilateDistance}*2)) "$workdir/${tempbase}_Areas_dil$((${ErodeDilateDistance}*2)).$dscalarend" -${hemisphereword}-surface "${T1wDownSampleFolder}/${Subject}.${Hem}.midthickness${RegSTRING}.${LowResMesh}k_fs_LR.surf.gii" -nearest
```
coalsont, line 143: `$((${ErodeDilateDistance}*2))` only works as expected for integer distances (which are in mm).  This should also probably be stored in another variable, rather than recomputed in lots of filenames.

ArealClassifier/scripts/ClassificationCleanup.sh
```diff
@@ -0,0 +1,295 @@
...
+wb_command -metric-math "Output > Input" "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$metricend" -var Input "$workdir/${tempbase}_Areas_raw_fcsmall.$metricend" -var Output "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$metricend"
+wb_command -cifti-create-dense-from-template "$workdir/${tempbase}_Areas_raw_fcsmall.$dscalarend" "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$dscalarend" -metric ${Structure} "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$metricend"
+wb_command -cifti-reduce "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$dscalarend" MAX "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$dscalarend"
+
+#find parts that were zeroed by find clusters
+wb_command -cifti-math "((Var > 0) - (Varfcsmall > 0)) * Fillable" "$workdir/${tempbase}_Areas_raw_fcsmallholesall.$dscalarend" -var Var "$workdir/${tempbase}_Areas_raw.$dscalarend" -var Varfcsmall "$workdir/${tempbase}_Areas_raw_fcsmall.$dscalarend" -var Fillable "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$dscalarend" -select 1 1 -repeat
```
coalsont, line 169: Since `_fcsmall` can't be nonzero where `_Areas_raw` is zero, and `_fillableholes` is binary, this should equal:

```suggestion
wb_command -cifti-math '((Var > 0) && !(Varfcsmall > 0)) && Fillable' "$workdir/${tempbase}_Areas_raw_fcsmallholesall.$dscalarend" -var Var "$workdir/${tempbase}_Areas_raw.$dscalarend" -var Varfcsmall "$workdir/${tempbase}_Areas_raw_fcsmall.$dscalarend" -var Fillable "$workdir/${tempbase}_Areas_raw_fcsmall_fillableholes.$dscalarend" -select 1 1 -repeat
```

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,524 @@
...
+            echo
+            ;;
+        (*)
+            log_Err_Abort "unrecognized matlab mode: $matlab_mode"
+            ;;
+    esac
+fi
+#average dropouts across runs, map to surface, put in cifti, since nothing else uses them this way
+#these temp names are a bit messy, but who cares
+tempfiles_create XXXXXX.dropouts.nii.gz volmerged
```
coalsont, line 298: Use a bit more descriptive name for temporary files, they end up in provenance, and make them more tab-complete friendly (for debugging) by not starting with the random characters.
```suggestion
tempfiles_create classifier_dropouts_XXXXXX.nii.gz volmerged
```

glasserm:
The merge of RSNRegression.sh and RSNRegression.m is removing features that we want to keep, rather than only adding new things we want (like PFM Weighted).  This includes removing support for mixture modeling, matlab compatibility flags, and z-transformation fixes.  @YuboW0 @bqrosen please fix this.

glasserm:
With regard to map volume to surface, I thought @coalsont that we already had that in the repo somewhere.

ArealClassifier/ArealClassifier.sh
glasserm: I agree, we should support PFM_DR too.

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,524 @@
...
+        ;;
+esac
+
+# check RSN type
+if [[ "$RegressMethod" != "PFM_weighted" ]]; then
+    log_Err_Abort "Only weighted regression is supported!"
+fi
+
+# check classifier version
+if [[ "$ClassifierVersion" == "MATLAB" || "$ClassifierVersion" == "ARENA_v1" || "$ClassifierVersion" == "ARENA_v2" ]]; then
```
glasserm, line 78: I wasn't planning to release the MATLAB classifier since there is complexity with its source code.

ArealClassifier/ArealClassifier.sh
glasserm: Agree.

ArealClassifier/ArealClassifier.sh
glasserm: Agree.

ArealClassifier/ArealClassifier.sh
glasserm: Agree.

ArealClassifier/ArealClassifier.sh
glasserm: Agree

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,524 @@
...
+if [[ "$filecount_HCP_MMP_ROIs" -lt 2024 || "$ForceDownloadWeight" -ne 0 ]]; then
+    "$HCPPIPEDIR"/global/scripts/download_areal_classifier_dependencies.sh "MMP_ROIs"
+fi
+
+# download model weights if necessary
+if [ "$ClassifierVersion" == "MATLAB" ]; then
+    filecount_model_weights=$(find "$HCPPIPEDIR"/ArealClassifier/data/mlp_classifier -maxdepth 1 -type f | wc -l)
+    if [[ "$filecount_model_weights" -lt 368 || "$ForceDownloadWeight" -ne 0 ]]; then
+        "$HCPPIPEDIR"/global/scripts/download_areal_classifier_dependencies.sh "MATLAB"
+    fi
```
glasserm, line 95: See above about the matlab classifier.

ArealClassifier/ArealClassifier.sh
```diff
@@ -0,0 +1,524 @@
...
+# check RSN type
+if [[ "$RegressMethod" != "PFM_weighted" ]]; then
+    log_Err_Abort "Only weighted regression is supported!"
+fi
+
+# check classifier version
+if [[ "$ClassifierVersion" == "MATLAB" || "$ClassifierVersion" == "ARENA_v1" || "$ClassifierVersion" == "ARENA_v2" ]]; then
+    log_Msg "ClassifierVersion is ${ClassifierVersion}."
+else
+    log_Err_Abort "ClassifierVersion must be one of the followings, 'MATLAB', 'ARENA_v1', 'ARENA_v2' but gets ${ClassifierVersion}"
```
glasserm, line 81: See above about matlab classifier.

PR #407, opened 2026-08-19
Integrate ClassifyTICA

arkky68:
Implement the tica Classify model by Alex. The model is shipped as ONNX. ClassifyTICA.sh reads ComputeTICAFeatures' features.csv and writes Noise.txt for CleanData.


tICA/tICAPipeline.sh
```diff
@@ -116,11 +119,14 @@ opts_AddOptional '--matlab-run-mode' 'MatlabMode' '0, 1, or 2' "defaults to $g_m
...
 # if Group sICA hand classifications exists, use it to filter the group sICA components before projecting to individuals
-HandSignalFile="${StudyFolder}/${GroupAverageName}/MNINonLinear/Results/${OutputfMRIName}/sICA/HandSignal.txt" 
+HandSignalFile="${StudyFolder}/${GroupAverageName}/MNINonLinear/Results/${OutputfMRIName}/sICA/HandSignal.txt"
 if [ -e "${HandSignalFile}" ]; then
     # Import the contents of $HandSignalFile as an array
     read -a sigIdx < "${HandSignalFile}"
     tICADim="${#sigIdx[@]}"
+elif [[ "$sicadimOverride" ]]; then
+
+    tICADim="$sicadimOverride"
```
glasserm, line 129: Was this a bug that needed fixing?

glasserm:
Thanks.  I would be interested in what @coalsont has to say about this.

tICA/tICAPipeline.sh
coalsont: Most of the other stages currently handle it like this:

https://github.com/Washington-University/HCPpipelines/blob/e2b78a665607e55d25769029d178e03ccd5611cb/tICA/tICAPipeline.sh#L611

Which we should do for the classify stage also, so that `tICADim` doesn't end up empty when using the "most recent dimensionality" logic:

https://github.com/Washington-University/HCPpipelines/blob/e2b78a665607e55d25769029d178e03ccd5611cb/tICA/tICAPipeline.sh#L418

Additionally, this entire conditional block shouldn't be above `opts_showValues` (it reads from a text file, so it can produce an error, so we want to show the arguments first).  Looks like the block was added in #354.  It would be good to move it below the `#processing code goes here` comment.

tICA/scripts/ClassifyTICA.sh
```diff
@@ -0,0 +1,88 @@
...
+opts_SetScriptDescription "classifies tICA components as signal or noise using a trained model, writing Noise.txt for CleanData"
+
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
+opts_AddMandatory '--out-group-name' 'GroupAverageName' 'string' 'name to use for the group output folder'
+opts_AddMandatory '--fmri-output-name' 'OutputfMRIName' 'string' "name for the output fMRI data, like 'rfMRI_REST_7T'"
+opts_AddMandatory '--ica-dim' 'tICAdim' 'integer' "number of temporal ICA components"
+
+opts_AddOptional '--python-singularity' 'PythonSingularity' 'string' "the file path of the singularity, specify empty string to use native environment instead" ""
+opts_AddOptional '--python-singularity-mount-path' 'PythonSingularityMountPath' 'string' "the file path of the mount path for singularity" ""
+opts_AddOptional '--python-interpreter' 'PythonInterpreter' 'string' "the python interpreter path" ""
```
coalsont, line 23: Handle static defaults early:

```suggestion
opts_AddOptional '--python-interpreter' 'PythonInterpreter' 'string' "the python executable, default 'python3' (from PATH)" "python3"
```

tICA/scripts/ClassifyTICA.sh
```diff
@@ -0,0 +1,88 @@
...
+    fi
+    if ! command -v singularity &> /dev/null; then
+        log_Err_Abort "Singularity is not installed or not in PATH."
+    fi
+    UseLocalPython="FALSE"
+    singularity_command=(singularity exec --bind "$PythonSingularityMountPath" "$PythonSingularity" python3)
+else
+    if [ ! -f "$PythonInterpreter" ]; then
+        PythonInterpreter="python3"
+    fi
```
coalsont, line 55: Simplify out the conditional in the later code, and don't silently ignore and continue if the user gave an incorrect path:

```suggestion
    PythonCmd=(singularity exec --bind "$PythonSingularityMountPath" "$PythonSingularity" python3)
else
    PythonCmd=("$PythonInterpreter")
```

tICA/scripts/ClassifyTICA.sh
```diff
@@ -0,0 +1,88 @@
...
+    --threshold="$Threshold"
+    --output_noise_file="$NoiseFile"
+    --output_proba_csv="$ProbaCsv"
+)
+
+if [ "$UseLocalPython" = "FALSE" ]; then
+    PythonLaunchCommand=("${singularity_command[@]}" "${pythonCode[@]}")
+else
+    PythonLaunchCommand=("${PythonInterpreter}" "${pythonCode[@]}")
+fi
```
coalsont, line 82: My earlier suggestions allow this simplification:

```suggestion
PythonLaunchCommand=("${PythonCmd[@]}" "${pythonCode[@]}")
```

tICA/scripts/ClassifyTICA.sh
```diff
@@ -0,0 +1,88 @@
...
+
+opts_SetScriptDescription "classifies tICA components as signal or noise using a trained model, writing Noise.txt for CleanData"
+
+opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
+opts_AddMandatory '--out-group-name' 'GroupAverageName' 'string' 'name to use for the group output folder'
+opts_AddMandatory '--fmri-output-name' 'OutputfMRIName' 'string' "name for the output fMRI data, like 'rfMRI_REST_7T'"
+opts_AddMandatory '--ica-dim' 'tICAdim' 'integer' "number of temporal ICA components"
+
+opts_AddOptional '--python-singularity' 'PythonSingularity' 'string' "the file path of the singularity, specify empty string to use native environment instead" ""
+opts_AddOptional '--python-singularity-mount-path' 'PythonSingularityMountPath' 'string' "the file path of the mount path for singularity" ""
```
coalsont, line 22: The user doesn't need to know the container is just for python3 and some python libraries, but they do need to know it is our custom container (I assume), not some standard container from the python.org website or something.

```suggestion
opts_AddOptional '--singularity-image' 'PythonSingularity' 'string' "the file path of the classifier singularity container, instead of using native python"
opts_AddOptional '--singularity-mount-path' 'PythonSingularityMountPath' 'string' "the --bind argument to get the data mounted into singularity"
```

tICA/tICAPipeline.sh
```diff
@@ -86,7 +86,10 @@ opts_AddOptional '--low-sica-dims' 'LowsICADims' 'num@num@num...' "the low sICA
 opts_AddOptional '--reclean-mode' 'RecleanModeString' 'YES or NO' 'whether the data should use ReCleanSignal.txt for DVARS' 'NO'
 
 #tICA Component Classification
-#not integrated yet
+opts_AddOptional '--python-singularity' 'PythonSingularity' 'string' "the file path of the singularity, specify empty string to use native environment instead" ""
+opts_AddOptional '--python-singularity-mount-path' 'PythonSingularityMountPath' 'string' "the file path of the mount path for singularity" ""
+opts_AddOptional '--python-interpreter' 'PythonInterpreter' 'string' "the python interpreter path" ""
```
coalsont, line 91: Edit these similar to my suggestions on the Classify script.

tICA/tICAPipeline.sh
arkky68: I added it so tICADim wouldn't be empty in the ClassifyTICA stage. I've removed it and used the same tICADim fallback the other stages use instead, per Tim's suggestion. Also moved the HandSignal block below #processing code goes here (it's after opts_ShowValues now as well).

tICA/tICAPipeline.sh
```diff
@@ -86,7 +86,10 @@ opts_AddOptional '--low-sica-dims' 'LowsICADims' 'num@num@num...' "the low sICA
 opts_AddOptional '--reclean-mode' 'RecleanModeString' 'YES or NO' 'whether the data should use ReCleanSignal.txt for DVARS' 'NO'
 
 #tICA Component Classification
-#not integrated yet
+opts_AddOptional '--singularity-image' 'PythonSingularity' 'string' "the file path of the classifier singularity container, instead of using native python" ""
+opts_AddOptional '--singularity-mount-path' 'PythonSingularityMountPath' 'string' "the --bind argument to get the data mounted into singularity" ""
```
coalsont: We should probably give an example of how to set up the `--bind` here.  It looks like the script adds the "--bind" part, so users need to know not to include that.

```suggestion
opts_AddOptional '--singularity-mount-path' 'PythonSingularityMountPath' 'hostpath:containedpath' "only the argument to --bind, to get your data folder mounted into singularity, like '/data:/data'" ""
```

PR #408, opened 2026-08-19
Implement single-run WF concat/deconcat, fix Analysis.pfm mkdir 

arkky68:


PFM/PFMPipeline.sh
```diff
@@ -245,57 +254,97 @@ do
...
-                            #         outFile="${WFDir}/${Subject}/${fMRIName}_Atlas${RegString}_${fMRIProcSTRING}_WF.dtseries.nii"
-                            #         wb_command -cifti-merge "$outFile" -direction ROW -cifti "$concatOutFile" -index "$startIdx" -up-to "$endIdx"
-                                    
-                            #         # un-variance normalize and un-demean each post-WF run
-                            #         wb_command -cifti-math "(TCS / VN) + MEAN" "$outFile" -var TCS "$outFile" -var MEAN "$meanFile" -var VN "$vnScalarFile" -select 1 1 -repeat
-                                  
-                            #         cumTP=$endIdx
-                            #     fi
-                            # done
+                            # No pre-existing concat file, so build one ourselves.
```
glasserm, line 257: Single runs also need to be variance normalized using the ${ProcString}_vn file before concatenating and then unvariance normalized by the average of those ${ProcString}_vn files.  If PFMs are to be run in variance normalized mode (which we have determined in macaques is a good idea) then we need to reapply the average variance normalization before running PFMs.

PR #409, opened 2026-08-24
Update HippUnfold HCP pipeline

orenpoliva:
Main changes for RunHippUnfold, HippUnfold, PostHippUnfold, RunPostHippUnfold:
-Updated the HippUnfold command to use the Maguire T2w model, including --force-nnunet-model maguire_T2w and  inner/outer registration settings of 0.
-Changed the output/file structure from supporting separate T1w, T2w, and T1wT2w models to using the Maguire T2w model. Because only one model is now used, the model-specific T1w, T2w, and T1wT2w subdirectories were removed.
-Log files are now written to a dedicated logging directory within each subject's T1w/HippUnfold directory.
-In HippUnfoldHCP.sh, imported runlocal from RunHippUnfoldHCP, and added separate per-job cache and Conda directories when runlocal=FALSE. This prevents simultaneous subjects from sharing these directories, which had caused failures when multiple HippUnfold jobs were run in parallel. These temporary per-job directories are removed after successful completion.
-In PostHippUnfoldHCP.sh, one use of ${Scalar} in the dentate-label generation section was replaced with an explicit thickness.shape.gii input. ${Scalar} is likely a leftover from the preceding scalar-processing loop. The metric is used as the template for generating the constant-valued dentate label; the thickness values themselves are not used in the calculation.
-Updated RunHippUnfoldHCP.sh and RunPostHippUnfoldHCP.sh for the current HCP dataset/batch-processing setup, including the study-folder, subject-list, environment-script, and queue configuration.

Two scripts were created for fMRI analysis of the hippocampus:
-Added fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh, which is comparable to the existing GenericfMRISurfaceProcessingPipeline.sh, the wrapper script used for cortical fMRI surface analysis.
-Added fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh to map fMRI data from the volume to the HippUnfold hippocampal and dentate surfaces at native, 512, 2k, 8k, and 18k mesh densities. This script is a modification of RibbonVolumeToSurfaceMapping.sh, which performs the corresponding volume-to-surface mapping for the cortex.

### DIFF
--- a/Examples/Scripts/RunHippUnfoldHCP.sh
+++ b/Examples/Scripts/RunHippUnfoldHCP.sh
@@ -1,4 +1,4 @@
-#!/bin/bash 
+#!/bin/bash
 
 get_batch_options() {
     local arguments=("$@")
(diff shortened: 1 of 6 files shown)


Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -27,21 +27,24 @@ get_batch_options() {
...
 
 get_batch_options "$@"
 
-StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID) 
-Subjlist="100307 100610" #Space delimited list of subject IDs 
-EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+StudyFolder="/media/myelin/brainmappers/Connectome_Project/YA_HCP_ReTest_Final"
+
+#sub 103818 was removed
+Subjlist="105923 111312 114823 115320 122317 125525 130518 135528 137128 139839 143325 144226 146129 149337 149741 151526 158035 169343 172332 175439 177746 185442 187547 192439 194140 195041 200109 200614 204521 250427 287248 341834 433839 562345 599671 601127 627549 660951 662551 783462 859671 861456 877168 917255"
```
coalsont, line 45: Try not to commit the copy of the batch file you actually use to run things.  We prefer to use placeholders in the example (for instance, to not put our filesystem layout into our examples), and users are encouraged to make a copy of the example before editing it.

glasserm:
Please re-upload addressing Tim's concern and please separate HippUnfold/PostHippUnfold from fMRIHippUnfold and put them in separate pull requests.

PR #410, opened 2026-08-25
Update HippUnfold HCP processing

orenpoliva:
Main changes for RunHippUnfold, HippUnfold, PostHippUnfold, RunPostHippUnfold: 

-Updated the HippUnfold command to use the Maguire T2w model, including --force-nnunet-model maguire_T2w and inner/outer registration settings of 0. 

-Changed the output/file structure from supporting separate T1w, T2w, and T1wT2w models to using the Maguire T2w model. Because only one model is now used, the model-specific T1w, T2w, and T1wT2w subdirectories were removed. 

-Log files are now written to a dedicated logging directory within each subject's T1w/HippUnfold directory. 

-In HippUnfoldHCP.sh, imported runlocal from RunHippUnfoldHCP, and added separate per-job cache and Conda directories when runlocal=FALSE. This prevents simultaneous subjects from sharing these directories, which had caused failures when multiple HippUnfold jobs were run in parallel. These temporary per-job directories are removed after successful completion. 

-In PostHippUnfoldHCP.sh, one use of ${Scalar} in the dentate-label generation section was replaced with an explicit thickness.shape.gii input. ${Scalar} is likely a leftover from the preceding scalar-processing loop. The metric is used as the template for generating the constant-valued dentate label; the thickness values themselves are not used in the calculation. 

-Updated RunHippUnfoldHCP.sh and RunPostHippUnfoldHCP.sh for generic study-folder, subject-list, environment-script.

### DIFF
--- a/Examples/Scripts/RunHippUnfoldHCP.sh
+++ b/Examples/Scripts/RunHippUnfoldHCP.sh
@@ -6,7 +6,8 @@ get_batch_options() {
     command_line_specified_study_folder=""
     command_line_specified_subject=""
     command_line_specified_run_local="FALSE"
-
+    command_line_specified_isolate_cache="FALSE"
+    
     local index=0
     local numArgs=${#arguments[@]}
     local argument
@@ -27,6 +28,10 @@ get_batch_options() {
                 command_line_specified_run_local="TRUE"
                 index=$(( index + 1 ))
                 ;;
+            --isolate-cache)
+                command_line_specified_isolate_cache="TRUE"
+                index=$(( index + 1 ))
+                ;;
             *)
                 echo ""
                 echo "ERROR: Unrecognized Option: ${argument}"
@@ -39,11 +44,10 @@ get_batch_options() {
 
 get_batch_options "$@"
 
-StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID)
-
-Subjlist="100307 100610" #Space delimited list of subject IDs
-
-EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script
+StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID) 
+Subjlist="100307 100610" #Space delimited list of subject IDs 
+#EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+EnvironmentScript="/media/myelin/oren/HippUnfoldTesting/HCPpipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment
 
 if [ -n "${command_line_specified_study_folder}" ]; then
     StudyFolder="${command_line_specified_study_folder}"
@@ -53,8 +57,14 @@ if [ -n "${command_line_specified_subject}" ]; then
     Subjlist="${command_line_specified_subject}"
 fi
 
+if [ -n "${command_line_specified_isolate_cache}" ]; then
+    IsolateCache="${command_line_specified_isolate_cache}"
+fi
+
 # Set up pipeline environment variables and software
 source "$EnvironmentScript"
+
+HippUnfoldCacheDIR="${StudyFolder}/HippUnfold/cache"
 
 # Log the originating call
 echo "$@"
@@ -75,7 +85,8 @@ for Subject in $Subjlist ; do
 
     LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/HippUnfoldHCP"
     mkdir -p "$LogDir"
-
+    cd "$LogDir"
+    
     if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
         echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
         queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
@@ -90,7 +101,8 @@ for Subject in $Subjlist ; do
         "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
         --study-folder="$StudyFolder" \
         --subject="$Subject" \
-        --runlocal="$RunLocal"
+        --isolate-cache="$IsolateCache" \
+    	--hippunfold-cache-dir="$HippUnfoldCacheDIR"
 
     # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
 
--- a/Examples/Scripts/RunHippUnfoldHCP.sh
+++ b/Examples/Scripts/RunHippUnfoldHCP.sh
@@ -1,4 +1,4 @@
-#!/bin/bash 
+#!/bin/bash
 
 get_batch_options() {
     local arguments=("$@")
@@ -27,21 +27,23 @@ get_batch_options() {
                 command_line_specified_run_local="TRUE"
                 index=$(( index + 1 ))
                 ;;
-	    *)
-		echo ""
-		echo "ERROR: Unrecognized Option: ${argument}"
-		echo ""
-		exit 1
-		;;
+            *)
+                echo ""
+                echo "ERROR: Unrecognized Option: ${argument}"
+                echo ""
+                exit 1
+                ;;
         esac
     done
 }
 
 get_batch_options "$@"
 
-StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID) 
-Subjlist="100307 100610" #Space delimited list of subject IDs 
-EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID)
+
+Subjlist="100307 100610" #Space delimited list of subject IDs
+
+EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script
 
 if [ -n "${command_line_specified_study_folder}" ]; then
     StudyFolder="${command_line_specified_study_folder}"
@@ -51,44 +53,51 @@ if [ -n "${command_line_specified_subject}" ]; then
     Subjlist="${command_line_specified_subject}"
 fi
 
-#Set up pipeline environment variables and software
+# Set up pipeline environment variables and software
 source "$EnvironmentScript"
 
 # Log the originating call
 echo "$@"
 
-#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
-#DO NOT include "-q " at the beginning
-#default to no queue, implying run local
+# NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
+# DO NOT include "-q " at the beginning
 QUEUE=""
 #QUEUE="hcp_priority.q"
 
-########################################## INPUTS ########################################## 
+########################################## INPUTS ##########################################
 
-#Scripts called by this script do assume they run on the outputs of the PostFreeSurfer Pipeline
+# Scripts called by this script do assume they run on the outputs of the PostFreeSurfer Pipeline
 
 ######################################### DO WORK ##########################################
 
 for Subject in $Subjlist ; do
-  echo $Subject
+    echo "$Subject"
 
-  if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
-      echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
-      queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
-  else
-      echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
-      queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
-  fi
+    LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/HippUnfoldHCP"
+    mkdir -p "$LogDir"
 
-  "${queuing_command[@]}" "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
-      --study-folder="$StudyFolder" \
-      --subject="$Subject" \
-      
-  # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
+    if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+        echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        RunLocal="TRUE"
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        RunLocal="FALSE"
+        queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
+    fi
 
-  echo "set -- --study-folder=$StudyFolder \
-      --subject=$Subject" \
+    "${queuing_command[@]}" \
+        "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
+        --study-folder="$StudyFolder" \
+        --subject="$Subject" \
+        --runlocal="$RunLocal"
 
-  echo ". ${EnvironmentScript}"
+    # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
+
+    echo "set --study-folder=$StudyFolder --subject=$Subject --runlocal=$RunLocal"
+
+    echo ". ${EnvironmentScript}"
 
 done
+
+
--- a/Examples/Scripts/RunPostHippUnfoldHCP.sh
+++ b/Examples/Scripts/RunPostHippUnfoldHCP.sh
@@ -27,21 +27,24 @@ get_batch_options() {
                 command_line_specified_run_local="TRUE"
                 index=$(( index + 1 ))
                 ;;
-	    *)
-		echo ""
-		echo "ERROR: Unrecognized Option: ${argument}"
-		echo ""
-		exit 1
-		;;
+            *)
+                echo ""
+                echo "ERROR: Unrecognized Option: ${argument}"
+                echo ""
+                exit 1
+                ;;
         esac
     done
 }
 
 get_batch_options "$@"
 
-StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID) 
-Subjlist="100307 100610" #Space delimited list of subject IDs 
-EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID)
+
+Subjlist="100307 100610" #Space delimited list of subject IDs
+
+EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script
+
 
 if [ -n "${command_line_specified_study_folder}" ]; then
     StudyFolder="${command_line_specified_study_folder}"
@@ -54,41 +57,44 @@ fi
 #Set up pipeline environment variables and software
 source "$EnvironmentScript"
 
-# Log the originating call
+#Log the originating call
 echo "$@"
 
 #NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
 #DO NOT include "-q " at the beginning
-#default to no queue, implying run local
 QUEUE=""
 #QUEUE="hcp_priority.q"
 
 ########################################## INPUTS ########################################## 
 
-#Scripts called by this script do assume they run on the outputs of the PostFreeSurfer Pipeline
+# Scripts called by this script do assume they run on the outputs of the PostFreeSurfer Pipeline
 
 ######################################### DO WORK ##########################################
 
 for Subject in $Subjlist ; do
   echo $Subject
 
+  LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/PostHippUnfoldHCP"
+  mkdir -p "$LogDir"
+
   if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
       echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/PostHippUnfoldHCP.sh"
       queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
   else
       echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/PostHippUnfoldHCP.sh"
-      queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE")
+      queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
   fi
 
   "${queuing_command[@]}" "$HCPPIPEDIR"/HippUnfoldHCP/PostHippUnfoldHCP.sh \
       --study-folder="$StudyFolder" \
-      --subject="$Subject" \
-      
+      --subject="$Subject"
+
   # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
 
   echo "set -- --study-folder=$StudyFolder \
-      --subject=$Subject" \
+      --subject=$Subject"
 
   echo ". ${EnvironmentScript}"
 
 done
+
--- a/Examples/Scripts/RunPostHippUnfoldHCP.sh
+++ b/Examples/Scripts/RunPostHippUnfoldHCP.sh
@@ -1,5 +1,3 @@
-#!/bin/bash 
-
 get_batch_options() {
     local arguments=("$@")
 
@@ -45,7 +43,6 @@ Subjlist="100307 100610" #Space delimited list of subject IDs
 
 EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script
 
-
 if [ -n "${command_line_specified_study_folder}" ]; then
     StudyFolder="${command_line_specified_study_folder}"
 fi
@@ -54,47 +51,51 @@ if [ -n "${command_line_specified_subject}" ]; then
     Subjlist="${command_line_specified_subject}"
 fi
 
-#Set up pipeline environment variables and software
+# Set up pipeline environment variables and software
 source "$EnvironmentScript"
 
-#Log the originating call
+# Log the originating call
 echo "$@"
 
-#NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
-#DO NOT include "-q " at the beginning
+# NOTE: syntax for QUEUE has changed compared to earlier pipeline releases,
+# DO NOT include "-q " at the beginning
 QUEUE=""
 #QUEUE="hcp_priority.q"
 
-########################################## INPUTS ########################################## 
+########################################## INPUTS ##########################################
 
 # Scripts called by this script do assume they run on the outputs of the PostFreeSurfer Pipeline
 
 ######################################### DO WORK ##########################################
 
 for Subject in $Subjlist ; do
-  echo $Subject
+    echo "$Subject"
 
-  LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/PostHippUnfoldHCP"
-  mkdir -p "$LogDir"
+    LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/HippUnfoldHCP"
+    
+    mkdir -p "$LogDir"
+    cd "$LogDir"
+    
+    if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+        echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        RunLocal="TRUE"
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        RunLocal="FALSE"
+        queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
+    fi
 
-  if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
-      echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/PostHippUnfoldHCP.sh"
-      queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
-  else
-      echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/PostHippUnfoldHCP.sh"
-      queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
-  fi
+    "${queuing_command[@]}" \
+        "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
+        --study-folder="$StudyFolder" \
+        --subject="$Subject" \
+        --runlocal="$RunLocal"
 
-  "${queuing_command[@]}" "$HCPPIPEDIR"/HippUnfoldHCP/PostHippUnfoldHCP.sh \
-      --study-folder="$StudyFolder" \
-      --subject="$Subject"
+    # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
 
-  # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
+    echo "set --study-folder=$StudyFolder --subject=$Subject --runlocal=$RunLocal"
 
-  echo "set -- --study-folder=$StudyFolder \
-      --subject=$Subject"
-
-  echo ". ${EnvironmentScript}"
+    echo ". ${EnvironmentScript}"
 
 done
-
--- a/HippUnfoldHCP/HippUnfoldHCP.sh
+++ b/HippUnfoldHCP/HippUnfoldHCP.sh
@@ -1,22 +1,25 @@
 #!/bin/bash
 set -eu
+
 pipedirguessed=0
+
 if [[ "${HCPPIPEDIR:-}" == "" ]]
 then
-  # pipedirguessed=1
-   #fix this if the script is more than one level below HCPPIPEDIR
-   export HCPPIPEDIR="$(dirname -- "$0")/.."
+    # pipedirguessed=1
+
+    # Fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
 fi
 
 source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
 source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
-
 
 opts_SetScriptDescription "Make some BIDS structures and run HippUnfold"
 
 opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
 opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
 opts_AddOptional '--hippunfold-dir' 'HippUnfoldDIR' 'path' "location of HippUnfold outputs"
+opts_AddOptional '--runlocal' 'RunLocal' 'TRUE|FALSE' "whether running locally" "FALSE"
 
 opts_ParseArguments "$@"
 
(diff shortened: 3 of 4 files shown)


HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -1,22 +1,25 @@
...
 set -eu
+
 pipedirguessed=0
+
 if [[ "${HCPPIPEDIR:-}" == "" ]]
 then
-  # pipedirguessed=1
-   #fix this if the script is more than one level below HCPPIPEDIR
-   export HCPPIPEDIR="$(dirname -- "$0")/.."
+    # pipedirguessed=1
```
coalsont: This code is intentional, to make it more likely the user actually uses a SetUpHCP... script if they want to do more than see the help information.
```suggestion
    pipedirguessed=1
```

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -1,22 +1,25 @@
...
 source "$HCPPIPEDIR/global/scripts/newopts.shlib" "$@"
 source "$HCPPIPEDIR/global/scripts/debug.shlib" "$@"
 
-
 opts_SetScriptDescription "Make some BIDS structures and run HippUnfold"
 
 opts_AddMandatory '--study-folder' 'StudyFolder' 'path' "folder containing all subjects"
 opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
 opts_AddOptional '--hippunfold-dir' 'HippUnfoldDIR' 'path' "location of HippUnfold outputs"
+opts_AddOptional '--runlocal' 'RunLocal' 'TRUE|FALSE' "whether running locally" "FALSE"
```
coalsont: This option should probably be something like `--isolate-cache`.  It is possible to attempt to run multiple subjects locally at the same time, which would likely have the same issue.

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -27,96 +30,142 @@ fi
...
 
+
+# ---------------------------------------------------------------------
+# HippUnfold cache
+# ---------------------------------------------------------------------
+
+if [[ -z "${HIPPUNFOLD_CACHE_DIR:-}" ]]
+then
+    log_Err_Abort "HIPPUNFOLD_CACHE_DIR is not set"
+fi
```
coalsont: I prefer not to use environment variables, particularly for "where to write stuff".  Make a parameter for this.

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -27,96 +30,142 @@ fi
...
 
 if [[ "${HIPPUNFOLDPATH:-}" == "" ]]
 then
     hippcmd=(hippunfold --use-conda)
 else
-    hippcmd=(apptainer run --bind "$StudyFolder" -e "$HIPPUNFOLDPATH")
+    if [[ "$RunLocal" == "FALSE" ]]
+    then
+    	# Separate cache and conda directory for each subject to avoid crash due to parallel use
+	JobCache="${HIPPUNFOLD_CACHE_DIR}/job-cache/${JOB_ID:-local}_${Subject}"
```
coalsont: Maybe I should implement support for folders in `tempfiles.shlib`, so we wouldn't need this hack just to avoid collision on something we presumably don't need to preserve.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -1,16 +1,17 @@
...
+
 pipedirguessed=0
 if [[ "${HCPPIPEDIR:-}" == "" ]]
 then
-  # pipedirguessed=1
-   #fix this if the script is more than one level below HCPPIPEDIR
-   export HCPPIPEDIR="$(dirname -- "$0")/.."
+# pipedirguessed=1
+#fix this if the script is more than one level below HCPPIPEDIR
+export HCPPIPEDIR="$(dirname -- "$0")/.."
```
coalsont: In addition to restoring the function of `pipedirguessed`, please restore some indentation.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,51 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        RunLocal="FALSE"
+        queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
+    fi
+
+    "${queuing_command[@]}" \
+        "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
+        --study-folder="$StudyFolder" \
+        --subject="$Subject" \
+        --runlocal="$RunLocal"
```
coalsont, line 93: PostHippUnfold doesn't accept (or need) this argument.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +57,59 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+    mkdir -p "$LogDir"
+    cd "$LogDir"
+    
+    if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+        echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        RunLocal="TRUE"
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        RunLocal="FALSE"
```
coalsont, line 96: If nothing else, we should probably force "isolate cache" in job queueing mode.  I don't think the batch needs to accept an argument to set it independently of running locally, either.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +57,59 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+    "${queuing_command[@]}" \
+        "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
+        --study-folder="$StudyFolder" \
+        --subject="$Subject" \
+        --isolate-cache="$IsolateCache" \
+    	--hippunfold-cache-dir="$HippUnfoldCacheDIR"
+
+    # The following lines are used for interactive debugging to set the positional parameters: $1 $2 $3 ...
+
+    echo "set --study-folder=$StudyFolder --subject=$Subject --runlocal=$RunLocal"
```
coalsont, line 109: When this type of line is present, it is supposed to be the same as the job arguments.  It should also use `--` to prevent attempted parsing for the other things the `set` builtin can do:

```suggestion
    echo "set -- --study-folder="$StudyFolder" \
        --subject="$Subject" \
        --isolate-cache="$IsolateCache" \
    	--hippunfold-cache-dir="$HippUnfoldCacheDIR"
```

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,51 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        RunLocal="TRUE"
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+        RunLocal="FALSE"
+        queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
+    fi
+
+    "${queuing_command[@]}" \
+        "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
```
coalsont, line 90: Why is `RunPostHippUnfold...` launching `HippUnfold...`?

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -41,7 +46,8 @@ get_batch_options "$@"
 
 StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID) 
 Subjlist="100307 100610" #Space delimited list of subject IDs 
-EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+#EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+EnvironmentScript="/media/myelin/oren/HippUnfoldTesting/HCPpipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment
```
glasserm, line 50: Please don't commit our file paths to the repo.  I recommend having the Batch script you use be separate from the Batch script in your repo that you develop on.

Examples/Scripts/RunHippUnfoldHCP.sh
glasserm: Why not always isolate cache?

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -1,5 +1,3 @@
-#!/bin/bash 
-
```
glasserm, line 2: Please always include the shabang.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,51 @@ if [ -n "${command_line_specified_subject}" ]; then
...
-  echo ". ${EnvironmentScript}"
+    echo "$Subject"
+
+    LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/HippUnfoldHCP"
+    
+    mkdir -p "$LogDir"
+    cd "$LogDir"
+    
+    if [[ "${command_line_specified_run_local}" == "TRUE" || "$QUEUE" == "" ]] ; then
+        echo "About to locally run ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
```
glasserm, line 80: This seems like a mistake.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,51 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+        queuing_command=("$HCPPIPEDIR"/global/scripts/captureoutput.sh)
+        RunLocal="TRUE"
+    else
+        echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
```
glasserm, line 84: Same here.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,51 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+        RunLocal="FALSE"
+        queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogDir")
+    fi
+
+    "${queuing_command[@]}" \
```
glasserm, line 89: And here

glasserm (changes requested):
Given the really extensive PostHippUnfold changes, I would like to review at test dataset.

HippUnfoldHCP/HippUnfoldHCP.sh
orenpoliva: While the diff shows differences here, I’m not seeing any actual changes to the code when I compare the original and recent versions.

HippUnfoldHCP/HippUnfoldHCP.sh
orenpoliva: Done

HippUnfoldHCP/HippUnfoldHCP.sh
orenpoliva: Done

HippUnfoldHCP/HippUnfoldHCP.sh
coalsont: Matt's version may have incorrectly commented this out, fix it anyway.

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -27,96 +29,123 @@ fi
...
+	JobCondaPrefix="${JobCache}/snakemake-conda"
+	JobCondaPkgs="${JobCache}/conda-pkgs"
+
+	mkdir -p "$JobCache"
+	mkdir -p "$JobCondaPrefix"
+	mkdir -p "$JobCondaPkgs"
+	
+	CondaPrefix="$JobCondaPrefix"
+
+        hippcmd=(
```
coalsont, line 102: Please use consistent indentation, these lines contain a mix of tabs and spaces.  I prefer 4 spaces, but at the very least it should be consistent within each indentation block.

```suggestion
    then
        # Separate cache and conda directory for each subject to avoid crash due to parallel use
        JobCache="${HippUnfoldCacheDIR}/job-cache/${JOB_ID:-local}_${Subject}"
        JobCondaPrefix="${JobCache}/snakemake-conda"
        JobCondaPkgs="${JobCache}/conda-pkgs"

        mkdir -p "$JobCache"
        mkdir -p "$JobCondaPrefix"
        mkdir -p "$JobCondaPkgs"
        
        CondaPrefix="$JobCondaPrefix"

        hippcmd=(
```

Examples/Scripts/RunHippUnfoldHCP.sh
coalsont: That would be fine.

orenpoliva:
HippUnfold

1. HippUnfoldHCP.sh
Identified a cache cleanup problem: subject-specific job cache directories were not always being deleted, causing the cache directory to grow to over 400 GB. To address this, an EXIT trap was added to HippUnfoldHCP.sh so that the subject-specific job cache is removed whenever the script exits, including following an error.
Added a Snakemake unlock step to remove stale locks that may remain after incomplete or failed previous runs and prevent subsequent runs from proceeding.

Added the --latency-wait 60 and --rerun-incomplete flags to the HippUnfold command to improve robustness against common causes of failed or interrupted runs.

2. RunPostHippUnfoldHCP.sh
Added a check to verify that HippUnfold completed successfully for each subject before initiating post-processing. Subjects without a successfully completed HippUnfold run are excluded from post-processing, and the user is notified.

3. PostHippUnfoldHCP.sh
Changed the output directory name from native to Native.
Completed the spec-file merge section. Previously, merged cortical + hippocampal spec files were created only for native (T1w + MNI), 512 (T1w + MNI), and 2k (T1w only). The missing merges were added for the remaining meshes and spaces.

Potentially fixed a bug in the transformation of hippocampal surfaces from T1w/physical space to MNI space. The wb_command -surface-apply-warpfield command supplied an MRI image (T1w_restore.nii.gz) as the argument to -fnirt, rather than the corresponding FNIRT transformation field. The original command contained:
wb_command -surface-apply-warpfield \
    <physical surface> \
    "${AtlasFolder}/xfms/standard2acpc_dc.nii.gz" \
    <MNI surface> \
    -fnirt "${AtlasFolder}/T1w_restore.nii.gz"
The -fnirt argument was changed to:
-fnirt "${AtlasFolder}/xfms/acpc_dc2standard.nii.gz"
Deleted and recreated all HippUnfold directories for the 45 subjects in both YA_HCP_Final and YA_HCP_ReTest_Final using the current version of the script. As a result, spec files generated by older versions of the script, which included additional NIfTI files such as T1w or T2w images without the _restore suffix, are no longer present.
PostHippUnfoldHCP.sh previously used two-space indentation. To conform to Tim's coding standard, the entire script was reformatted to use four-space indentation.

Questions:

1. Should QUEUE be defined in SetUpHCPPipeline.sh rather than directly in the run script?

2. In PostHippUnfoldHCP.sh, for the 2k, 8k, and 18k hippocampal meshes, I merged the hippocampal spec files with the 164k cortical MSMAll spec file. Is 164k the correct cortical spec file to use for all three hippocampal mesh resolutions?

3. Matt — you said that only the T1w and T2w NIfTI files should be included. I also see T1wDividedByT2w.nii.gz and T1wDividedByT2w_ribbon.nii.gz in the spec files. Should these also be removed?

coalsont:
> 1. Should QUEUE be defined in SetUpHCPPipeline.sh rather than directly in the run script?

No.  Different pipelines are (somewhat) expected to use different queues because they have different levels of internal parallelism and memory requirements.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +51,53 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+
+        if ! "$HCPPIPEDIR"/global/scripts/captureoutput.sh \
+            "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
+            --study-folder="$StudyFolder" \
+            --subject="$Subject"
+        then
+            echo "ERROR: HippUnfold failed for subject: $Subject"
+            exit 1
+        fi
+
```
coalsont, line 86: I'm not sure we want local runs to stop at the first failed subject, as it doesn't necessarily indicate that later subjects will also fail.  I'd suggest changing all this back to the normal `queueing_command` pattern.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,183 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+for Subject in $Subjlist ; do
+    HippLogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/HippUnfoldHCP"
+    LatestLog=$(ls -t "${HippLogDir}"/*.o* 2>/dev/null | head -1)
+
+    if [[ -n "$LatestLog" ]] && grep -qF "HippUnfold pipeline completed successfully for subject: ${Subject}" "$LatestLog" ; then
+        SuccessfulSubjects+=" ${Subject}"
+    else
+        FailedSubjects+=" ${Subject}"
+    fi
+done
```
coalsont, line 79: I don't think we want to use log files to decide whether to try to run the Post pipeline, particularly not by checking for a highly specific and otherwise cosmetic message (that any developer might edit or remove without much thought).  We could test for the existence of an important data file, but only if we really need the launcher to not even attempt some subjects.  I think it is rare that we have a launcher try to do this at all, rather than simply trying every subject in the user-provided list.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +51,53 @@ if [ -n "${command_line_specified_subject}" ]; then
...
-
-  echo "set -- --study-folder=$StudyFolder \
-      --subject=$Subject" \
-
-  echo ". ${EnvironmentScript}"
+    echo "$Subject"
+
+    LogDir="${StudyFolder}/${Subject}/T1w/HippUnfold/logs/HippUnfoldHCP"
+    mkdir -p "$LogDir"
+    cd "$LogDir"
```
coalsont, line 72: Putting the job logs into subject-specific folders is not how we usually do things, and makes it harder to grep all the logs from a batch at once.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,183 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+#StudyFolder="${HOME}/projects/Pipelines_ExampleData" #Location of Subject folders (named by SubjectID) 
+#Subjlist="100307 100610" #Space delimited list of subject IDs 
+#EnvironmentScript="${HOME}/projects/Pipelines/Examples/Scripts/SetUpHCPPipeline.sh" #Pipeline environment script 
+
+StudyFolder="/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/" 
+Subjlist="103818 105923 111312 114823 115320 122317 125525 130518 135528
+137128 139839 143325 144226 146129 149337 149741 151526 158035
+169343 172332 175439 177746 185442 187547 192439 194140 195041
+200109 200614 204521 250427 287248 341834 433839 562345 599671
+601127 627549 660951 662551 783462 859671 861456 877168 917255"
```
coalsont, line 163: Please make a copy with a different name or path (for instance, appending your initials, or somewhere outside the repo entirely) and don't commit it, rather than committing your testing paths and subject list to the example.

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +51,183 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+
+if [[ -n "$FailedSubjects" ]] ; then
+    echo ""
+    echo "Subjects${FailedSubjects} contained incomplete or unsuccessful HippUnfold outputs and were removed from processing."
+fi#!/bin/bash
+
+get_batch_options() {
+    local arguments=("$@")
+
+    command_line_specified_study_folder=""
```
coalsont, line 118: You appear to have pasted an extra copy of the example script into itself.

HippUnfoldHCP/HippUnfoldHCP.sh
```diff
@@ -3,15 +3,27 @@ set -eu
...
+# Remove subject specific cache directory even in case of script failure
+cleanup() {
+    if [[ -n "${JobCache:-}" && -d "$JobCache" ]]; then
+        log_Msg "Removing job cache: $JobCache"
+        rm -rf "$JobCache"
+        log_Msg "Temporary cache directory deleted"
+    fi
+}
+
+trap cleanup EXIT
```
coalsont, line 24: If anyone tries to use `tempfiles.shlib`, this `trap ... EXIT` will prevent it from working.  We can probably leave this as a "remove if completed successfully" rather than worrying about it.  Or, I can add folder support to `tempfiles.shlib`, to also handle the collision avoidance without expecting `JOB_ID` to be a robust indicator.

Examples/Scripts/RunHippUnfoldHCP.sh
orenpoliva: If we don’t put the log files in the subject-specific directories, would it be a good idea to include the subject ID in the log filename?

Examples/Scripts/RunPostHippUnfoldHCP.sh
orenpoliva: After noticing I did this, I committed the scripts again with empty paths. Did you not receive the updated commit?

Examples/Scripts/RunHippUnfoldHCP.sh
coalsont: I'm not sure that we can set the name of the logfile.  The queuing system typically generates it from the name of the executable and the job number (a counter that never resets, even through reboots).

Examples/Scripts/RunPostHippUnfoldHCP.sh
coalsont: See my comment on line 113, you appear to have tried to replace the contents of the file, but left the entire old version on the end of it.

orenpoliva:
I uploaded a new version that addresses Tim’s comments. It also includes several additions that I thought would improve the scripts:

1. Added options for specifying the environment setup script (SetUpHCPPipeline.sh) and execution queue.

2. Per Tim’s request, moved the log files from the individual subject directories back to a common directory. I also added an option for specifying the log directory, which defaults to StudyFolder.

3. Added a cortical-mesh option to RunPostHippUnfold and PostHippUnfold. It accepts 32 for 3T data and 59 for 7T data, with 32 as the default. In PostHippUnfold, the selected cortical mesh is merged into the hippocampal spec files at each standard mesh resolution (512, 2k, 8k, and 18k). The native hippocampal mesh is instead merged with the native cortical mesh.

The pipeline output can be inspected for subject 103818 at:

/media/myelin/oren/HippUnfo

coalsont:
> 2\. I also added an option for specifying the log directory, which defaults to StudyFolder.

Please not `StudyFolder`.  The usual default for queueing tools is to log to the current directory (`.`), which isn't amazing for organization if you are lazy, but can be controlled if you plan ahead and `cd` somewhere appropriate.  I don't want the pipelines to specifically advocate for "spam your data directory with log files, in the same place as your subject folders".  If the user doesn't specify a folder, the best idea is not to provide the log folder argument to the command (maybe fsl_sub has a configuration file option for this, it would be a shame to break it if so).

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +83,51 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+    queuing_command=("$HCPPIPEDIR/global/scripts/captureoutput.sh")
+
+else
+
+    echo "About to use fsl_sub to queue ${HCPPIPEDIR}/HippUnfoldHCP/HippUnfoldHCP.sh"
+    queuing_command=("$FSLDIR/bin/fsl_sub" -q "$QUEUE" -l "$LogFolder")
+
+fi
+
+"${queuing_command[@]}" \
```
coalsont, line 125: This is in a loop, it should be indented.

Examples/Scripts/RunPostHippUnfoldHCP.sh
glasserm: Launcher code like that is typically something the users add after the fact for convenience.

Examples/Scripts/RunHippUnfoldHCP.sh
glasserm: Let's treat logs as other pipelines do.

glasserm:
> > 2. I also added an option for specifying the log directory, which defaults to StudyFolder.
> 
> Please not `StudyFolder`. The usual default for queueing tools is to log to the current directory (`.`), which isn't amazing for organization if you are lazy, but can be controlled if you plan ahead and `cd` somewhere appropriate. I don't want the pipelines to specifically advocate for "spam your data directory with log files, in the same place as your subject folders". If the user doesn't specify a folder, the best idea is not to provide the log folder argument to the command (maybe fsl_sub has a configuration file option for this, it would be a shame to break it if so).

I agree.

glasserm:
> HippUnfold
> 
>     1. HippUnfoldHCP.sh
>        Identified a cache cleanup problem: subject-specific job cache directories were not always being deleted, causing the cache directory to grow to over 400 GB. To address this, an EXIT trap was added to HippUnfoldHCP.sh so that the subject-specific job cache is removed whenever the script exits, including following an error.
>        Added a Snakemake unlock step to remove stale locks that may remain after incomplete or failed previous runs and prevent subsequent runs from proceeding.
> 
> 
> Added the --latency-wait 60 and --rerun-incomplete flags to the HippUnfold command to improve robustness against common causes of failed or interrupted runs.
> 
>     2. RunPostHippUnfoldHCP.sh
>        Added a check to verify that HippUnfold completed successfully for each subject before initiating post-processing. Subjects without a successfully completed HippUnfold run are excluded from post-processing, and the user is notified.
> 
>     3. PostHippUnfoldHCP.sh
>        Changed the output directory name from native to Native.
>        Completed the spec-file merge section. Previously, merged cortical + hippocampal spec files were created only for native (T1w + MNI), 512 (T1w + MNI), and 2k (T1w only). The missing merges were added for the remaining meshes and spaces.
> 
> 
> Potentially fixed a bug in the transformation of hippocampal surfaces from T1w/physical space to MNI space. The wb_command -surface-apply-warpfield command supplied an MRI image (T1w_restore.nii.gz) as the argument to -fnirt, rather than the corresponding FNIRT transformation field. The original command contained: wb_command -surface-apply-warpfield "${AtlasFolder}/xfms/standard2acpc_dc.nii.gz" -fnirt "${AtlasFolder}/T1w_restore.nii.gz" The -fnirt argument was changed to: -fnirt "${AtlasFolder}/xfms/acpc_dc2standard.nii.gz" Deleted and recreated all HippUnfold directories for the 45 subjects in both YA_HCP_Final and YA_HCP_ReTest_Final using the current version of the script. As a result, spec files generated by older versions of the script, which included additional NIfTI files such as T1w or T2w images without the _restore suffix, are no longer present. PostHippUnfoldHCP.sh previously used two-space indentation. To conform to Tim's coding standard, the entire script was reformatted to use four-space indentation.
> 
> Questions:
> 
>     1. Should QUEUE be defined in SetUpHCPPipeline.sh rather than directly in the run script?
> 
>     2. In PostHippUnfoldHCP.sh, for the 2k, 8k, and 18k hippocampal meshes, I merged the hippocampal spec files with the 164k cortical MSMAll spec file. Is 164k the correct cortical spec file to use for all three hippocampal mesh resolutions?
> 
>     3. Matt — you said that only the T1w and T2w NIfTI files should be included. I also see T1wDividedByT2w.nii.gz and T1wDividedByT2w_ribbon.nii.gz in the spec files. Should these also be removed?

3. Those can stay.

glasserm:
I would pair 8k and 18k with 164k meshes, since they are of more similar sampling density.

Otherwise things looked good for my review of the test dataset.

glasserm:
The PostHippUnfold diff is still full of non-changes.  Can you clean it up so that I can see only the stuff that actually changed?

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -33,317 +31,275 @@ log_Msg "Starting PostHippUnfold pipeline for subject: $Subject"
...
+                cp "${RawHippUnfoldFolder}/metric/sub-${Subject}_hemi-${Hemisphere}_den-${Mesh}_label-${Structure}_${Scalar}.shape.gii" "${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii"
+
+                wb_command -set-structure "${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii" "${HemiStructure}"
+
+                PALETTE "${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii" "${Color}" metric wb_command
+
+                wb_command -set-map-names "${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii" -map 1 "${Subject}_${Name}"
+
+                if [[ "${Scalar}" = "surfarea" ]] ; then
+                    wb_command -surface-vertex-areas "${AtlasHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_midthickness.${Mesh}.surf.gii" "${AtlasHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii"
```
coalsont, line 177: Consider if we want to use "_va" or otherwise reference "vertex" in at least the filename, to match the existing HCP conventions.

glasserm: I guess we went with the HippUnfold convention here.

orenpoliva:
Recreated PostHippUnfoldHCP without the indentations fix so all changes should now be visible. I made a list here of all the changes done to this script from the beginning, in case you weren't able to see them:

1. Once we updated to the Maguire workflow, the pipeline no longer loops through separate T1w, T2w, and T1wT2w HippUnfold output folders. It now uses a single HippUnfold output directory.

2. Per Matt’s request, renamed the native-resolution folder from native to Native to match the folder naming used elsewhere in the HCP pipeline.

3. Fixed a bug in FNIRT surface-warp command. wb_command -surface-apply-warpfield requires the forward warpfield after -fnirt. The original command used T1w_restore.nii.gz there; instead, it now uses acpc_dc2standard.nii.gz.

4. In the command that creates the dentate label file, I replaced the $Scalar variable with _thickness in the dentate metric filename. Because $Scalar was left over from the previous loop, this change prevents future modifications to the scalar loop from unexpectedly affecting dentate label generation.

5. In the GIFTI metric and label loops, I renamed the loop variables from Scalar and Label to ScalarEntry and LabelEntry. This distinguishes the complete list entries from the individual values extracted from them and makes the code easier to follow.

6. Added the user-defined --hippunfold-mesh option. Users can choose which HippUnfold mesh resolutions and folders to generate. The list is separated by @, and the default includes all supported meshes: native@512@2k@8k@18k.

7. Added the user-defined --brain-mesh option. Users can choose which whole-brain cortical mesh is paired with each HippUnfold mesh in the merged spec files. The two lists are matched by position. The default brain-mesh list is native@32k@32k@164k@164k (this pairs native with native, 512 with 32k, 2k with 32k, 8k with 164k, and 18k with 164k). The script stops with an error if the two lists have different numbers of entries or contain an unsupported mesh. For 7T data, users can also select the 59k brain mesh instead of 32k.

8. Added HippUnfold outputs in MMORF space. The pipeline now creates a HippUnfold folder under MMORFNonLinear and generates MMORF-space surfaces, metrics, CIFTIs, label volumes, and spec files. This follows the same general process as the MNINonLinear output generation, but uses the MMORF inverse warp, mmorf2acpc_dc.nii.gz, instead of the FNIRT warp files.

9. Replaced the deprecated Workbench command -add-to-spec-file with -spec-file-modify.

glasserm:
@coalsont what else is needed here besides another test run?

coalsont:
> 3\. Fixed a bug in FNIRT surface-warp command. wb_command -surface-apply-warpfield requires the forward warpfield after -fnirt. The original command used T1w_restore.nii.gz there; instead, it now uses acpc_dc2standard.nii.gz.

Technically, what the `-fnirt` option needs is just a volume with the "before resample" spacing information for that warpfield (which in this case is already the inverse of the warpfield you might expect to use from the naming, because surfaces use a pointcloud instead of a raster grid).  We expect that fnirt warpfields have an inverse available (which requires using that same spacing info in order to be a correct inverse), which made it relatively easy to describe that in the help info instead of going into the details of how a warpfield might have been generated.  Any volume file that uses a standard applyisoxfm MNI grid would work in this specific case.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -1,11 +1,12 @@
-#!/bin/bash 
-
```
coalsont, line 2: Please don't delete this, is has a purpose.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +81,54 @@ if [ -n "${command_line_specified_subject}" ]; then
...
-QUEUE=""
-#QUEUE="hcp_priority.q"
-
-########################################## INPUTS ########################################## 
+LogFolder="$StudyFolder"
+if [[ -n "$command_line_specified_log_folder" ]] ; then
+    LogFolder="$command_line_specified_log_folder"
+fi
+mkdir -p "$LogFolder"
+cd "$LogFolder"
```
coalsont, line 101:
```suggestion
LogFolder=""
if [[ -n "$command_line_specified_log_folder" ]] ; then
    LogFolder="$command_line_specified_log_folder"
fi
if [[ "$LogFolder" != "" ]] ; then
    mkdir -p "$LogFolder"
    cd "$LogFolder"
fi
```

Examples/Scripts/RunPostHippUnfoldHCP.sh
```diff
@@ -51,44 +95,58 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+    RunLocal="FALSE"
+fi
+
+# setting log folder
+LogFolder=""
+if [[ -n "$command_line_specified_log_folder" ]] ; then
+    LogFolder="$command_line_specified_log_folder"
+    mkdir -p "$LogFolder"
+    cd "$LogFolder"
+fi
```
coalsont, line 110: See my suggestion here in the other Run script.  If someone manually sets LogFolder while editing the Run script, the command line argument could still be empty, but we'd still want to do these.

Examples/Scripts/RunHippUnfoldHCP.sh
```diff
@@ -51,44 +81,54 @@ if [ -n "${command_line_specified_subject}" ]; then
...
+        queuing_command+=(-l "$LogFolder")
+    fi
+fi
 
-  echo "set -- --study-folder=$StudyFolder \
-      --subject=$Subject" \
+"${queuing_command[@]}" \
+    "$HCPPIPEDIR"/HippUnfoldHCP/HippUnfoldHCP.sh \
+    --study-folder="$StudyFolder" \
+    --subject="$Subject"
```
coalsont, line 129: This `if` block and command are inside a loop, they should be indented.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -32,15 +33,58 @@ log_Msg "Starting PostHippUnfold pipeline for subject: $Subject"
...
+            ;;
+        *)
+            log_Err_Abort "Unrecognized HippUnfold mesh '$Mesh'"
+            ;;
+    esac
+done
+
+for Mesh in "${BrainMeshes[@]}"; do
+    case "$Mesh" in
+        native|32k|59k|164k)
```
coalsont, line 80: If we add more meshes in the future (or someone else makes a custom mesh for their data), we'd need to edit this sanity check.  If we want to sanity check these in advance, it would probably be better to check that an expected file on that mesh exists, rather than hardcoding acceptable mesh identifiers.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -58,292 +102,246 @@ function PALETTE {
...
+            if [ $Scalar = "surfarea" ] ; then
+                wb_command -surface-vertex-areas ${AtlasHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_midthickness.${Mesh}.surf.gii ${AtlasHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii
+            else
+                cp ${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii ${AtlasHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii
+            fi
+            if $runMMORF && [[ $Scalar = "surfarea" ]] ; then
+                wb_command -surface-vertex-areas ${MMORFHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_midthickness.${Mesh}.surf.gii ${MMORFHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii
+            else
+                cp ${PhysicalHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii ${MMORFHippUnfoldFolder}/${Subject}.${Hemisphere}.${Structure}_${Scalar}.${Mesh}.shape.gii
```
coalsont, line 201: I'm not sure we want to have separate copies of surface-only files for curvature, gyrification, and myelin for different volume spaces (we don't duplicate files this way in the MMORF PR).

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -3,7 +3,7 @@ set -eu
 pipedirguessed=0
 if [[ "${HCPPIPEDIR:-}" == "" ]]
 then
-  # pipedirguessed=1
+   pipedirguessed=1
    #fix this if the script is more than one level below HCPPIPEDIR
    export HCPPIPEDIR="$(dirname -- "$0")/.."
 fi
```
coalsont, line 9: Many text editors don't offer 3-space indentation, please make it 4 like (most of) the rest of the file.

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -32,15 +33,58 @@ log_Msg "Starting PostHippUnfold pipeline for subject: $Subject"
...
+fi
+
+if [ -z ${PhysicalHippUnfoldDIR-} ] ; then
   PhysicalHippUnfoldDIR="${T1wFolder}/HippUnfold"
 fi 
 
-if [ -z ${AtlasHippUnfoldDIR} ] ; then
+if [ -z ${AtlasHippUnfoldDIR-} ] ; then
   AtlasHippUnfoldDIR="${AtlasFolder}/HippUnfold"
 fi
```
coalsont, line 57: Since the code on both sides of this is 4-space indented, I would argue to make these match.  The "files" tab has a "hide whitespace" option anyway (to not mark-up edits that are only whitespace changes).

HippUnfoldHCP/PostHippUnfoldHCP.sh
```diff
@@ -58,292 +102,246 @@ function PALETTE {
...
         elif [ ${Hemisphere} = "R" ] ; then
-          if [ ${Structure} = "dentate" ] ; then
+            if [ ${Structure} = "dentate" ] ; then
             HemiStructure="HIPPOCAMPUS_DENTATE_RIGHT"
-          elif [ ${Structure} = "hipp" ] ; then
+            elif [ ${Structure} = "hipp" ] ; then
             HemiStructure="HIPPOCAMPUS_RIGHT"
-          fi
+            fi
         fi
```
coalsont, line 153: Please make sure the indentation does actually increase inside every loop and conditional section.

PR #412, opened 2026-08-26
Add hippocampal fMRISurface processing

orenpoliva:
fMRI pipeline changes:

GenericHippocampusfMRISurfaceProcessingPipeline.sh:

Re-enabled pipedirguessed=1.

Removed the output-folder and fmri-qc options.

Added the --resample-mesh option. A validation loop was also added to allow only valid mesh values (512, 2k, 8k, 18k). The selected meshes are then passed to the child script HippocampalVolumeToSurfaceMapping.sh.
Moved surface smoothing into a separate child script, HippocampalSurfaceSmoothing.sh.
Moved CIFTI generation into a separate child script, HippocampalCreateCIFTIs.sh.

Shifted to a more compact formatting style, including reducing unnecessary use of line-continuation backslashes.
Removed the variable previously used for wb_command and now call wb_command directly.

The native mesh is added internally to the requested resampling meshes so that native-space processing is always performed before resampling to the user-requested meshes.

HippocampalVolumeToSurfaceMapping.sh:

The script now receives the mesh list as an input parameter.

Similar to its cortical counterpart, RibbonVolumeToSurfaceMapping.sh, intermediate volumetric files are generated inside the HippocampalVolumeToSurfaceMapping working subdirectory. Final surface-related files are generated in the parent fMRI results directory (e.g., rfMRI_REST1_LR).

Previously, generation of the fMRI surface time series and VN surface files was intermixed. These are now handled in separate processing loops.

VN is mapped once from the volume to the native hippocampal surface and then resampled from native to the requested meshes.

Shifted to a more compact formatting style, including reducing unnecessary use of line-continuation backslashes.
Removed the variable previously used for wb_command and now call wb_command directly.

Questions:
1. The HippocampalVolumeToSurfaceMapping working-directory name is currently hardcoded in both the parent and child scripts. Should the working-directory path instead be passed from the parent to the child as an input parameter?

2. If the long-term plan is to combine the cortical and hippocampal fMRISurface pipelines, would it make more sense to begin integrating them now rather than maintaining two parallel pipelines?

3. There is some inconsistency in how wb_command is called across HCP scripts. For example, some scripts, such as RSNRegression.sh, use wb_command, while others, such as SurfaceSmoothing.sh, use ${CARET7DIR}/wb_command. Which format should I use?

### DIFF
--- a/fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+set -eu
 # Requirements for this script
 # installed versions of: FSL, Connectome Workbench (wb_command)
 # environment: HCPPIPEDIR, FSLDIR, CARET7DIR
@@ -24,7 +24,9 @@ then
 fi
 
 #comment the line back in when done
-source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
+#source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/log.shlib" "$@"          # Debugging functions; also sources log.shlib
+
 source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"
 source "${HCPPIPEDIR}/global/scripts/processingmodecheck.shlib" # Check processing mode requirements
 
@@ -40,14 +42,11 @@ opts_AddOptional '--procstring' 'ProcString' 'string' 'processing suffix appende
 
 opts_AddMandatory '--smoothingFWHM' 'SmoothingFWHM' 'number' 'smoothing FWHM (mm)'
 
-opts_AddOptional '--fmri-qc' 'QCMode' 'YES OR NO OR ONLY' "Controls whether to generate a QC scene and snapshots (default=YES). ONLY executes just the QC script." "YES"
-
-opts_AddOptional '--output-directory' 'OutputDirectory' 'path' 'Directory where pipeline outputs will be written' ""
-
 opts_AddOptional '--goodvoxel' 'doGoodVoxels' 'YES OR NO' "Controls whether to do goodVoxel procedure (default = YES)" "YES"
 
-opts_AddOptional '--factor' 'factor' 'number' "Scaling factor for eliminating high COV voxels"
+opts_AddOptional '--factor' 'factor' 'number' "Scaling factor for eliminating high COV voxels (default = 1.5)" "1.5"
 
+opts_AddOptional '--resample_mesh' 'MeshString' 'string' 'Resample native mesh to: 512, 2k, 8k, or 18k. Use quote for multiple meshes, e.g. "512 2k 8k"' "512 2k 8k 18k"
 opts_ParseArguments "$@"
 
 if ((pipedirguessed))
@@ -76,162 +75,59 @@ HCPPIPEDIR_fMRISurf="${HCPPIPEDIR}/fMRISurface/scripts"
 log_Msg "Platform Information Follows:"
 uname -a
 
-QCMode="$(echo "${QCMode}" | tr '[:upper:]' '[:lower:]')"
-
-doProcessing=1
-doQC=1
-
-case "${QCMode}" in
-    yes)
-        ;;
-    no)
-        doQC=0
-        ;;
-    only)
-        doProcessing=0
-        log_Warn "Only generating fMRI QC scene and snapshots from existing data"
-        ;;
-    *)
-        log_Err_Abort "Unrecognized value '${QCMode}' for --fmri-qc; use YES, NO, or ONLY"
-        ;;
-esac
-
-
 # ------------------------------------------------------------------------------
 # Set up paths
 # ------------------------------------------------------------------------------
 
-PipelineScripts="${HCPPIPEDIR_fMRISurf}"
-
 MainSubjectFolder="${Path}/${Subject}"
 AtlasSpaceFolder="${MainSubjectFolder}/MNINonLinear"
 InputResultsFolder="${AtlasSpaceFolder}/Results/${NameOffMRI}"
-ResultsFolderName="Results"
+ResultsFolder="${AtlasSpaceFolder}/Results/${NameOffMRI}"
 
-if [ -n "${OutputDirectory}" ]; then
-    ResultsFolder="${OutputDirectory}/${Subject}/${NameOffMRI}"
-else
-    ResultsFolder="${AtlasSpaceFolder}/${ResultsFolderName}/${NameOffMRI}"
-fi
-
-WorkingDirectory="${ResultsFolder}/tmp"
-
+WorkingDirectory="${ResultsFolder}/HippocampalVolumeToSurfaceMapping" #should 'HippocampalVolumeToSurfaceMapping' be a parameter?'
 mkdir -p "${WorkingDirectory}"
 
-VolumefMRI="${InputResultsFolder}/${NameOffMRI}${ProcString}"
+VolumefMRI="${InputResultsFolder}/${NameOffMRI}${ProcString}.nii.gz"
 
 SBRef="${InputResultsFolder}/${NameOffMRI}_SBRef.nii.gz"
 
 HippUnfoldFolder="${AtlasSpaceFolder}/HippUnfold"
 
-if [ ! -f "${VolumefMRI}.nii.gz" ]; then
-    log_Err_Abort "Cannot find selected fMRI file: ${VolumefMRI}.nii.gz"
+if [ ! -f "${VolumefMRI}" ]; then
+    log_Err_Abort "Cannot find selected fMRI file: ${VolumefMRI}"
 fi
 
 if [ ! -f "${SBRef}" ]; then
     log_Err_Abort "Cannot find selected SBRef file: ${SBRef}"
 fi
 
-log_Msg "Selected fMRI file: ${VolumefMRI}.nii.gz"
+#Parse MeshString into a mesh array
+for Mesh in ${MeshString}; do
+    if [[ " 512 2k 8k 18k " != *" ${Mesh} "* ]]; then
+        log_Err_Abort "Invalid mesh: ${Mesh}"
+    fi
+done
+Meshes="native ${MeshString}"
+
+log_Msg "Selected fMRI file: ${VolumefMRI}"
 log_Msg "Selected SBRef file: ${SBRef}"
 
 # ------------------------------------------------------------------------------
-# Generate fMRI time series within ROIs
+# Combining ROIs into a single CIFTI time series and vn file
 # ------------------------------------------------------------------------------
 
-if ((doProcessing)); then
-
-    # Make fMRI ribbon
-    log_Msg "Make fMRI Ribbon"
-    log_Msg "mkdir -p ${WorkingDirectory}"
-
-    log_Msg "Hippocampal Volume To Surface Mapping"
-    "${PipelineScripts}/HippocampalVolumeToSurfaceMapping.sh" \
-            "${WorkingDirectory}" \
-            "${Subject}" \
-            "${HippUnfoldFolder}" \
-            "${VolumefMRI}" \
-            "${SBRef}" \
-            "${doGoodVoxels}" \
-            "${factor}"
-
-#Surface Smoothing
- # ------------------------------------------------------------------------------
-# Hippocampal Surface Smoothing
-# ------------------------------------------------------------------------------
-
-    log_Msg "Hippocampal Surface Smoothing"
-
-    Sigma=`echo "$SmoothingFWHM / (2 * sqrt(2 * l(2)))" | bc -l`
-
-    Meshes=(native 512 2k 8k 18k)
-    Structures=(hipp dentate)
-
-    for Mesh in "${Meshes[@]}"; do
-        for Structure in "${Structures[@]}"; do
-            for Hemisphere in L R; do
-                "${CARET7DIR}/wb_command" -metric-smoothing \
-                    "${HippUnfoldFolder}/${Mesh}/${Subject}.${Hemisphere}.${Structure}_midthickness.${Mesh}.surf.gii" \
-                    "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_fMRI.${Mesh}.func.gii" \
-                    "${Sigma}" \
-                    "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
-                    -roi "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_ones.${Mesh}.func.gii"
-            done
-        done
-    done
-# ------------------------------------------------------------------------------
-# Combining ROIs into a single CIFTI file
-# ------------------------------------------------------------------------------
+log_Msg "mkdir -p ${WorkingDirectory}"
+log_Msg "Hippocampal Volume To Surface Mapping"
 
 
-    for Mesh in ${Meshes[@]}; do
-        "${CARET7DIR}/wb_command" -cifti-create-dense-timeseries \
-            "${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}.${Mesh}.dtseries.nii" \
-            -metric HIPPOCAMPUS_LEFT "${WorkingDirectory}/${Subject}.L.hipp_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
-                -roi "${WorkingDirectory}/${Subject}.L.hipp_ones.${Mesh}.func.gii" \
-            -metric HIPPOCAMPUS_RIGHT "${WorkingDirectory}/${Subject}.R.hipp_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
-                -roi "${WorkingDirectory}/${Subject}.R.hipp_ones.${Mesh}.func.gii" \
-            -metric HIPPOCAMPUS_DENTATE_LEFT "${WorkingDirectory}/${Subject}.L.dentate_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
-                -roi "${WorkingDirectory}/${Subject}.L.dentate_ones.${Mesh}.func.gii" \
-            -metric HIPPOCAMPUS_DENTATE_RIGHT "${WorkingDirectory}/${Subject}.R.dentate_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
-                -roi "${WorkingDirectory}/${Subject}.R.dentate_ones.${Mesh}.func.gii"
+# Generate fMRI time series for each of the hippocampal structures: L hipp, R hipp, L dentate, R dentate
+"${HCPPIPEDIR_fMRISurf}/HippocampalVolumeToSurfaceMapping.sh" "${ResultsFolder}" "${Subject}" "${HippUnfoldFolder}" "${VolumefMRI}" "${SBRef}" "${doGoodVoxels}" "${factor}" "${Meshes}"
 
-            log_Msg "Generated fMRI time series Cifti file: "${ResultsFolder}/${Subject}.${NameOffMRI}"_AtlasHipp_${ProcString}.${Mesh}.dtseries.nii"
-            
-# --------------------------------------------------------------------------
-# Combine VN maps from all four hippocampal structures
-# --------------------------------------------------------------------------
+#Surface Smoothing of each hippocampal structure
+"${HCPPIPEDIR_fMRISurf}/HippocampalSmoothing.sh" "${HippUnfoldFolder}" "${ResultsFolder}" "${WorkingDirectory}" "${Subject}" "${SmoothingFWHM}" "${Meshes}"
 
-    "${CARET7DIR}/wb_command" -cifti-create-dense-scalar \
-        "${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}_vn.${Mesh}.dscalar.nii" \
-        -metric HIPPOCAMPUS_LEFT \
-            "${WorkingDirectory}/${Subject}.L.hipp_fMRI_vn.${Mesh}.func.gii" \
-            -roi "${WorkingDirectory}/${Subject}.L.hipp_ones.${Mesh}.func.gii" \
-        -metric HIPPOCAMPUS_RIGHT \
-            "${WorkingDirectory}/${Subject}.R.hipp_fMRI_vn.${Mesh}.func.gii" \
-            -roi "${WorkingDirectory}/${Subject}.R.hipp_ones.${Mesh}.func.gii" \
-        -metric HIPPOCAMPUS_DENTATE_LEFT \
-            "${WorkingDirectory}/${Subject}.L.dentate_fMRI_vn.${Mesh}.func.gii" \
-            -roi "${WorkingDirectory}/${Subject}.L.dentate_ones.${Mesh}.func.gii" \
-        -metric HIPPOCAMPUS_DENTATE_RIGHT \
-            "${WorkingDirectory}/${Subject}.R.dentate_fMRI_vn.${Mesh}.func.gii" \
-            -roi "${WorkingDirectory}/${Subject}.R.dentate_ones.${Mesh}.func.gii"
+#Integration of the 4 hippocampal structures into single CIFTI timeseries and vn files
+"${HCPPIPEDIR_fMRISurf}/CreateHippocampalCIFTIs.sh" "${ResultsFolder}" "${Subject}" "${NameOffMRI}" "${ProcString}" "${SmoothingFWHM}" "${Meshes}"
 
-    log_Msg "Generated VN CIFTI file: ${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}_vn.${Mesh}.dscalar.nii"
+log_Msg "GenericHippocampusfMRISurfaceProcessingPipeline Completed!"
 
-    done
-fi
-
-#Uncomment if want to remove the subfoler with intermediate files
-#rm -rf "${WorkingDirectory}"
-
-# if ((doQC)); then
-#     log_Msg "Generating fMRI QC scene and snapshots"
-#     "${PipelineScripts}/GenerateFMRIScenes.sh" \
-#         --study-folder="${Path}" \
-#         --subject="${Subject}" \
-#         --fmriname="${NameOffMRI}${ProcString}" \
-#         --output-folder="${ResultsFolder}/fMRIQC"
-# fi
-
-log_Msg "Completed!"
--- a/fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
@@ -197,6 +197,28 @@ if ((doProcessing)); then
                 -roi "${WorkingDirectory}/${Subject}.R.dentate_ones.${Mesh}.func.gii"
 
             log_Msg "Generated fMRI time series Cifti file: "${ResultsFolder}/${Subject}.${NameOffMRI}"_AtlasHipp_${ProcString}.${Mesh}.dtseries.nii"
+            
+# --------------------------------------------------------------------------
+# Combine VN maps from all four hippocampal structures
+# --------------------------------------------------------------------------
+
+    "${CARET7DIR}/wb_command" -cifti-create-dense-scalar \
+        "${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}_vn.${Mesh}.dscalar.nii" \
+        -metric HIPPOCAMPUS_LEFT \
+            "${WorkingDirectory}/${Subject}.L.hipp_fMRI_vn.${Mesh}.func.gii" \
+            -roi "${WorkingDirectory}/${Subject}.L.hipp_ones.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_RIGHT \
+            "${WorkingDirectory}/${Subject}.R.hipp_fMRI_vn.${Mesh}.func.gii" \
+            -roi "${WorkingDirectory}/${Subject}.R.hipp_ones.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_DENTATE_LEFT \
+            "${WorkingDirectory}/${Subject}.L.dentate_fMRI_vn.${Mesh}.func.gii" \
+            -roi "${WorkingDirectory}/${Subject}.L.dentate_ones.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_DENTATE_RIGHT \
+            "${WorkingDirectory}/${Subject}.R.dentate_fMRI_vn.${Mesh}.func.gii" \
+            -roi "${WorkingDirectory}/${Subject}.R.dentate_ones.${Mesh}.func.gii"
+
+    log_Msg "Generated VN CIFTI file: ${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}_vn.${Mesh}.dscalar.nii"
+
     done
 fi
 
--- a/fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
+++ b/fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
@@ -1,0 +1,215 @@
+#!/bin/bash
+
+# Requirements for this script
+# installed versions of: FSL, Connectome Workbench (wb_command)
+# environment: HCPPIPEDIR, FSLDIR, CARET7DIR
+
+########################################## PIPELINE OVERVIEW ##########################################
+
+# TODO
+
+########################################## OUTPUT DIRECTORIES ##########################################
+
+# TODO
+
+################################################ SUPPORT FUNCTIONS ##################################################
+
+set -eu
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    # Fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+#comment the line back in when done
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"
+source "${HCPPIPEDIR}/global/scripts/processingmodecheck.shlib" # Check processing mode requirements
+
+opts_SetScriptDescription "Run hippocampal fMRISurface processing"
+
+opts_AddMandatory '--studyfolder' 'Path' 'path' "folder containing all subjects" "--path"
+
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+
+opts_AddMandatory '--fmriname' 'NameOffMRI' 'string' 'fMRI run name, for example rfMRI_REST1_LR'
+
+opts_AddOptional '--procstring' 'ProcString' 'string' 'processing suffix appended to the 4D fMRI filename; include the leading underscore' ""
+
+opts_AddMandatory '--smoothingFWHM' 'SmoothingFWHM' 'number' 'smoothing FWHM (mm)'
+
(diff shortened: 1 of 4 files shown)


fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,555 @@
...
+
+        # ------------------------------------------------------------------------------
+        #  Generating an all-ones surface metric
+        # ------------------------------------------------------------------------------
+
+        "${WB}" -metric-math \
+            '1' \
+            "${OnesMetric}" \
+            -var x \
+            "${ThicknessMetric}"
```
coalsont: For math commands, I typically use a more compact style due to the expected compactness of the variable names and the expression.  Also, `wb_command` is on `$PATH`, we don't need a variable in the way.

```suggestion
        wb_command -metric-math '1' "${OnesMetric}" \
            -var x "${ThicknessMetric}"
```

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+
+opts_AddMandatory '--smoothingFWHM' 'SmoothingFWHM' 'number' 'smoothing FWHM (mm)'
+
+opts_AddOptional '--fmri-qc' 'QCMode' 'YES OR NO OR ONLY' "Controls whether to generate a QC scene and snapshots (default=YES). ONLY executes just the QC script." "YES"
+
+opts_AddOptional '--output-directory' 'OutputDirectory' 'path' 'Directory where pipeline outputs will be written' ""
+
+opts_AddOptional '--goodvoxel' 'doGoodVoxels' 'YES OR NO' "Controls whether to do goodVoxel procedure (default = YES)" "YES"
+
+opts_AddOptional '--factor' 'factor' 'number' "Scaling factor for eliminating high COV voxels"
```
coalsont: This should probably have a default value.

Single stage with...apparently half a standard deviation threshold may have been chosen for cortex due to substantial outlier effects (multi-stage and/or dilation to isolate partial voxel effects would probably be a better approach).  The single stage approach may work particularly poorly with a much smaller ROI like hippocampus.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+# ------------------------------------------------------------------------------
+# Combining ROIs into a single CIFTI file
+# ------------------------------------------------------------------------------
+
+
+    for Mesh in ${Meshes[@]}; do
+        "${CARET7DIR}/wb_command" -cifti-create-dense-timeseries \
+            "${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}.${Mesh}.dtseries.nii" \
+            -metric HIPPOCAMPUS_LEFT "${WorkingDirectory}/${Subject}.L.hipp_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+                -roi "${WorkingDirectory}/${Subject}.L.hipp_ones.${Mesh}.func.gii" \
```
coalsont: If these are all-ones files, you don't need to use `-roi` at all.  The reason the `-roi` option exists is because the cortical surfaces are topologically spheres, even though some of the sphere vertices don't actually represent cortex.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,555 @@
...
+
+        Prefix="${Subject}.${Hemisphere}.${Structure}"
+
+        ThicknessMetric="${HippUnfoldFolder}/native/${Prefix}_thickness.native.shape.gii"
+
+        InnerSurface="${HippUnfoldFolder}/native/${Prefix}_inner.native.surf.gii"
+        MidSurface="${HippUnfoldFolder}/native/${Prefix}_midthickness.native.surf.gii"
+        OuterSurface="${HippUnfoldFolder}/native/${Prefix}_outer.native.surf.gii"
+
+        OnesMetric="${WorkingDirectory}/${Prefix}_ones.native.func.gii"
```
coalsont, line 83: I don't think we need to keep any all-ones metrics around, I would use `tempfiles.shlib` to handle it instead (in the `/tmp` folder or wherever the standardized environment variables point), like this:

https://github.com/Washington-University/HCPpipelines/blob/07c74f1f3db196cecd2527560ef2e29fa9f235fd/global/scripts/VolumeToCIFTI.sh#L97

A file created this way gets deleted when the script exits, without any further code.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
glasserm: Using a variable for wb_command would allow someone to specify a particular version of wb_command (e.g., in a CHPC-like environment).  I usually have had a variable for that reason.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
glasserm: I agree.  Oren did an analysis to figure this out, so we should use that as the default.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
glasserm: Agree.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+
+opts_AddMandatory '--fmriname' 'NameOffMRI' 'string' 'fMRI run name, for example rfMRI_REST1_LR'
+
+opts_AddOptional '--procstring' 'ProcString' 'string' 'processing suffix appended to the 4D fMRI filename; include the leading underscore' ""
+
+opts_AddMandatory '--smoothingFWHM' 'SmoothingFWHM' 'number' 'smoothing FWHM (mm)'
+
+opts_AddOptional '--fmri-qc' 'QCMode' 'YES OR NO OR ONLY' "Controls whether to generate a QC scene and snapshots (default=YES). ONLY executes just the QC script." "YES"
```
glasserm: Is this implemented?  If not, it should be taken out.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+
+opts_AddOptional '--output-directory' 'OutputDirectory' 'path' 'Directory where pipeline outputs will be written' ""
```
glasserm:
```suggestion
opts_AddOptional '--output-directory' 'OutputDirectory' 'path' 'Directory where pipeline outputs will be written for testing purposes' ""
```

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+        doQC=0
+        ;;
+    only)
+        doProcessing=0
+        log_Warn "Only generating fMRI QC scene and snapshots from existing data"
+        ;;
+    *)
+        log_Err_Abort "Unrecognized value '${QCMode}' for --fmri-qc; use YES, NO, or ONLY"
+        ;;
+esac
```
glasserm: Remove if we are not doing this.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+
+
+# ------------------------------------------------------------------------------
+# Set up paths
+# ------------------------------------------------------------------------------
+
+PipelineScripts="${HCPPIPEDIR_fMRISurf}"
+
+MainSubjectFolder="${Path}/${Subject}"
```
glasserm, line 82: Is this variable needed?

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+AtlasSpaceFolder="${MainSubjectFolder}/MNINonLinear"
+InputResultsFolder="${AtlasSpaceFolder}/Results/${NameOffMRI}"
+ResultsFolderName="Results"
+
+if [ -n "${OutputDirectory}" ]; then
+    ResultsFolder="${OutputDirectory}/${Subject}/${NameOffMRI}"
+else
+    ResultsFolder="${AtlasSpaceFolder}/${ResultsFolderName}/${NameOffMRI}"
+fi
```
glasserm: This seems specific to your testing.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+                "${CARET7DIR}/wb_command" -metric-smoothing \
+                    "${HippUnfoldFolder}/${Mesh}/${Subject}.${Hemisphere}.${Structure}_midthickness.${Mesh}.surf.gii" \
+                    "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_fMRI.${Mesh}.func.gii" \
+                    "${Sigma}" \
+                    "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+                    -roi "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_ones.${Mesh}.func.gii"
+            done
+        done
+    done
+# ------------------------------------------------------------------------------
```
glasserm: This will need to be moved into a module in the future for parallelism with fMRISurface (and eventual inclusion in fMRISurface).  If you insist on keeping it in the main pipeline, please put it in a function with the appropriate I/O so that it can easily be copied into the appropriate fMRISurface module when the time comes.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+            -metric HIPPOCAMPUS_LEFT "${WorkingDirectory}/${Subject}.L.hipp_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+                -roi "${WorkingDirectory}/${Subject}.L.hipp_ones.${Mesh}.func.gii" \
+            -metric HIPPOCAMPUS_RIGHT "${WorkingDirectory}/${Subject}.R.hipp_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+                -roi "${WorkingDirectory}/${Subject}.R.hipp_ones.${Mesh}.func.gii" \
+            -metric HIPPOCAMPUS_DENTATE_LEFT "${WorkingDirectory}/${Subject}.L.dentate_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+                -roi "${WorkingDirectory}/${Subject}.L.dentate_ones.${Mesh}.func.gii" \
+            -metric HIPPOCAMPUS_DENTATE_RIGHT "${WorkingDirectory}/${Subject}.R.dentate_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+                -roi "${WorkingDirectory}/${Subject}.R.dentate_ones.${Mesh}.func.gii"
+
+            log_Msg "Generated fMRI time series Cifti file: "${ResultsFolder}/${Subject}.${NameOffMRI}"_AtlasHipp_${ProcString}.${Mesh}.dtseries.nii"
```
glasserm: This will also need to be moved into a module for parallelism with fMRISurface and eventual folding into fMRI surface.  If you insist on keeping it in the main pipeline, please put it in a function with the appropriate I/O so that it can easily be copied into the appropriate fMRISurface module when the time comes.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+#rm -rf "${WorkingDirectory}"
+
+# if ((doQC)); then
+#     log_Msg "Generating fMRI QC scene and snapshots"
+#     "${PipelineScripts}/GenerateFMRIScenes.sh" \
+#         --study-folder="${Path}" \
+#         --subject="${Subject}" \
+#         --fmriname="${NameOffMRI}${ProcString}" \
+#         --output-folder="${ResultsFolder}/fMRIQC"
+# fi
```
glasserm: Delete if not doing.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,237 @@
...
+            "${WorkingDirectory}/${Subject}.R.dentate_fMRI_vn.${Mesh}.func.gii" \
+            -roi "${WorkingDirectory}/${Subject}.R.dentate_ones.${Mesh}.func.gii"
+
+    log_Msg "Generated VN CIFTI file: ${ResultsFolder}/${NameOffMRI}_AtlasHipp_${ProcString}_vn.${Mesh}.dscalar.nii"
+
+    done
+fi
+
+#Uncomment if want to remove the subfoler with intermediate files
+#rm -rf "${WorkingDirectory}"
```
glasserm: Seems dangerous.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,555 @@
...
+HippUnfoldFolder="$3"
+VolumefMRI="$4"
+SBRef="$5"
+doGoodVoxels="$6"
+Factor="$7"
+#Factor="1"                 # Factor is a scaling factor for how many std units away from mean to set threshold
+NeighborhoodSmoothing="5"   # Distinguishes large vs small dropout
+dilation_dist="10"
+
+Meshes=(512 2k 8k 18k)
```
glasserm: We don't need to run this on all meshes.  Just the one we choose.  Mesh density should be an input parameter to the overall pipeline.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,555 @@
...
+        #  Establishing input and output files
+        # ------------------------------------------------------------------------------
+
+        Prefix="${Subject}.${Hemisphere}.${Structure}"
+
+        ThicknessMetric="${HippUnfoldFolder}/native/${Prefix}_thickness.native.shape.gii"
+
+        InnerSurface="${HippUnfoldFolder}/native/${Prefix}_inner.native.surf.gii"
+        MidSurface="${HippUnfoldFolder}/native/${Prefix}_midthickness.native.surf.gii"
+        OuterSurface="${HippUnfoldFolder}/native/${Prefix}_outer.native.surf.gii"
```
glasserm, line 81: Native should be capitalized.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,555 @@
...
+                -thr "${Upper}" \
+                -bin \
+                -sub "${WorkingDirectory}/${Prefix}_mask" \
+                -mul -1 \
+                -mas "${BinaryRibbon}" \
+                "${WorkingDirectory}/${Prefix}_goodvoxels"
+        fi
+
+
+        NativeFolder="${HippUnfoldFolder}/native"
```
glasserm, line 133: Native again here.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,555 @@
...
+                -cubic
+
+            "${WB}" -metric-mask \
+                "${TargetfMRIVN}" \
+                "${TargetROI}" \
+                "${TargetfMRIVN}"
+
+            log_Msg "Generated VN file in ${Mesh} mesh at: ${TargetfMRIVN}"
+        done
+    done
```
glasserm: This is disorganized.  The VN stuff should all be in a separate section by itself since it is a hack that will be removed in the future.

glasserm (changes requested):
151 lines became over 500 lines of code.  I find this very hard to follow.  Can we reduce the extent of the excessive carriage returns?

Also I will need to see a test dataset.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
coalsont: Changing `$PATH` to point to it would do that just fine.  There is also a place to do this in `SetUpHCPPipeline.sh`, which results in it being in `$PATH`.  As currently written, it is hardcoded to use `$CARET7DIR`, which is the same as what SetUp... puts on `$PATH` anyway.

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
orenpoliva: Native should be capitalized - are you referring to folder name or substring in filename, or both?

glasserm: Folder.

orenpoliva:
fMRI pipeline changes:

GenericHippocampusfMRISurfaceProcessingPipeline.sh:

Re-enabled pipedirguessed=1.

Removed the output-folder and fmri-qc options.

Added the --resample-mesh option. A validation loop was also added to allow only valid mesh values (512, 2k, 8k, 18k). The selected meshes are then passed to the child script HippocampalVolumeToSurfaceMapping.sh.
Moved surface smoothing into a separate child script, HippocampalSurfaceSmoothing.sh.
Moved CIFTI generation into a separate child script, HippocampalCreateCIFTIs.sh.

Shifted to a more compact formatting style, including reducing unnecessary use of line-continuation backslashes.
Removed the variable previously used for wb_command and now call wb_command directly.

The native mesh is added internally to the requested resampling meshes so that native-space processing is always performed before resampling to the user-requested meshes.

HippocampalVolumeToSurfaceMapping.sh:

The script now receives the mesh list as an input parameter.

Similar to its cortical counterpart, RibbonVolumeToSurfaceMapping.sh, intermediate volumetric files are generated inside the HippocampalVolumeToSurfaceMapping working subdirectory. Final surface-related files are generated in the parent fMRI results directory (e.g., rfMRI_REST1_LR).

Previously, generation of the fMRI surface time series and VN surface files was intermixed. These are now handled in separate processing loops.

VN is mapped once from the volume to the native hippocampal surface and then resampled from native to the requested meshes.

Shifted to a more compact formatting style, including reducing unnecessary use of line-continuation backslashes.
Removed the variable previously used for wb_command and now call wb_command directly.

Questions:
1. The HippocampalVolumeToSurfaceMapping working-directory name is currently hardcoded in both the parent and child scripts. Should the working-directory path instead be passed from the parent to the child as an input parameter?

2. If the long-term plan is to combine the cortical and hippocampal fMRISurface pipelines, would it make more sense to begin integrating them now rather than maintaining two parallel pipelines?

3. There is some inconsistency in how wb_command is called across HCP scripts. For example, some scripts, such as RSNRegression.sh, use wb_command, while others, such as SurfaceSmoothing.sh, use ${CARET7DIR}/wb_command. Which format should I use?

orenpoliva:
> 151 lines became over 500 lines of code. I find this very hard to follow. Can we reduce the extent of the excessive carriage returns?
> 
> Also I will need to see a test dataset.

orenpoliva:
A path to a test dataset is: 
This path link to final files dtseries.nii and _vn.dscalar.nii
/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/103818/MNINonLinear/Results/rfMRI_REST1_LR 
See also subfolder for intermediate files:
/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/103818/MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping

Script was modified to reduce in size

coalsont:
> 1. The HippocampalVolumeToSurfaceMapping working-directory name is currently hardcoded in both the parent and child scripts. Should the working-directory path instead be passed from the parent to the child as an input parameter?

Defining the folder name only once would be better, yes.  Hardcoding it in the top-level script is still good, in case we have a need for the intermediates later.

> 3. There is some inconsistency in how wb_command is called across HCP scripts. For example, some scripts, such as RSNRegression.sh, use wb_command, while others, such as SurfaceSmoothing.sh, use ${CARET7DIR}/wb_command. Which format should I use?

I use (and recommend) bare `wb_command` (and fsl commands, too), because it is more readable and has no substantial disadvantages (as I understand it, fsl uses `$FSLDIR` in the scripts in their installation directory specifically in case the user has called a version of an fsl tool that isn't the first one on `$PATH`, but we control `$PATH` in our SetUpHCP... script, and our scripts aren't part of an FSL installation, so that situation doesn't apply to us).  @glasserm does other things, partly because he made a `wb_command_retry` script long ago, but that can be done by adding a special "retry script" folder to the front of `$PATH` instead of using a different executable name.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,133 @@
...
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    # Fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+#comment the line back in when done
+#source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/log.shlib" "$@"          # Debugging functions; also sources log.shlib
```
coalsont, line 28: If it runs with `set -eu`, then we should be sourcing `debug.shlib`.  I'm not sure what commenting it out was intended to accomplish, as it should merely give more information than `set -eu`, but otherwise similar "exit on error" effects.

```suggestion
source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
```

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,133 @@
...
+
+# TODO
+
+########################################## OUTPUT DIRECTORIES ##########################################
+
+# TODO
+
+################################################ SUPPORT FUNCTIONS ##################################################
+
+set -eu
```
coalsont, line 17: Redundant `set -eu`.  Also, if we aren't going to fill these documentation sections in, I wouldn't bother having them (yet).
```suggestion
```

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
+set -eu
+# Requirements for this script
+# installed versions of: FSL, Connectome Workbench (wb_command)
+# environment: HCPPIPEDIR, FSLDIR, CARET7DIR
```
coalsont, line 5: I'm not sure how useful this list is, as all of these are always set up by `SetUpHCP...`.

fMRISurface/scripts/HippocampalSmoothing.sh
```diff
@@ -0,0 +1,74 @@
...
+Meshes="$6"
+
+Sigma=$(echo "${SmoothingFWHM} / (2 * sqrt(2 * l(2)))" | bc -l)
+
+Structures=(hipp dentate)
+
+for Mesh in ${Meshes}; do
+    for Structure in "${Structures[@]}"; do
+        for Hemisphere in L R; do
+            wb_command -metric-smoothing "${HippUnfoldFolder}/${Mesh}/${Subject}.${Hemisphere}.${Structure}_midthickness.${Mesh}.surf.gii" "${ResultsFolder}/${Subject}.${Hemisphere}.${Structure}_fMRI.${Mesh}.func.gii" "${Sigma}" "${ResultsFolder}/${Subject}.${Hemisphere}.${Structure}_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" -roi "${WorkingDirectory}/${Subject}.${Hemisphere}.${Structure}_ones.${Mesh}.func.gii"
```
coalsont, line 69: This script has 6 arguments, to (effectively) loop over 1 command.  I would consider doing this loop in another script that already contains most/all of the arguments, and deleting this script.  Even doing it in the main pipeline script is fine with me, we don't have to hide absolutely every command in a "submodule" script.

fMRISurface/scripts/CreateHippocampalCIFTIs.sh
```diff
@@ -0,0 +1,82 @@
...
+        -metric HIPPOCAMPUS_DENTATE_LEFT "${ResultsFolder}/${Subject}.L.dentate_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_DENTATE_RIGHT "${ResultsFolder}/${Subject}.R.dentate_fMRI_s${SmoothingFWHM}.${Mesh}.func.gii"
+
+    log_Msg "Generated fMRI time series CIFTI file: ${ResultsFolder}/${NameOffMRI}_AtlasHipp${ProcString}.${Mesh}.dtseries.nii"
+
+    wb_command -cifti-create-dense-scalar "${ResultsFolder}/${NameOffMRI}_AtlasHipp${ProcString}_vn.${Mesh}.dscalar.nii" \
+        -metric HIPPOCAMPUS_LEFT "${ResultsFolder}/${Subject}.L.hipp_fMRI_vn.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_RIGHT "${ResultsFolder}/${Subject}.R.hipp_fMRI_vn.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_DENTATE_LEFT "${ResultsFolder}/${Subject}.L.dentate_fMRI_vn.${Mesh}.func.gii" \
+        -metric HIPPOCAMPUS_DENTATE_RIGHT "${ResultsFolder}/${Subject}.R.dentate_fMRI_vn.${Mesh}.func.gii"
```
coalsont, line 76: Similarly, a 6 argument script to loop over 2 commands.

fMRISurface/scripts/HippocampalSmoothing.sh
glasserm: I asked him to do it this way because of how the code will get merged in with the existing fMRISurface code in the future.

fMRISurface/scripts/CreateHippocampalCIFTIs.sh
glasserm: Same so that the merge with fMRISurface is obvious.

fMRISurface/scripts/HippocampalSmoothing.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+set -eu
+# --------------------------------------------------------------------------------
```
coalsont, line 3:
```suggestion
#based on fMRISurface/scripts/SurfaceSmoothing.sh

# --------------------------------------------------------------------------------
```

fMRISurface/scripts/CreateHippocampalCIFTIs.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+set -eu
+# --------------------------------------------------------------------------------
```
coalsont, line 3:
```suggestion
#similar to fMRISurface/scripts/CreateDenseTimeseries.sh

# --------------------------------------------------------------------------------
```

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
orenpoliva: Sorry about that. I'm writing my code on vs code using bashdb debugger. This specific script crashes my debug. I was planning to uncomment before committing to the PR.

fMRISurface/scripts/HippocampalSmoothing.sh
orenpoliva: Matt requested this command to be placed in its own script to parallel cortical processing.

fMRISurface/scripts/CreateHippocampalCIFTIs.sh
orenpoliva: The newer version have a lot more commands.

coalsont: What do you mean by "newer version", do you have more edits that haven't been pushed?

glasserm:
> > 1. The HippocampalVolumeToSurfaceMapping working-directory name is currently hardcoded in both the parent and child scripts. Should the working-directory path instead be passed from the parent to the child as an input parameter?
> 
> Defining the folder name only once would be better, yes. Hardcoding it in the top-level script is still good, in case we have a need for the intermediates later.
> 
> > 3. There is some inconsistency in how wb_command is called across HCP scripts. For example, some scripts, such as RSNRegression.sh, use wb_command, while others, such as SurfaceSmoothing.sh, use ${CARET7DIR}/wb_command. Which format should I use?
> 
> I use (and recommend) bare `wb_command` (and fsl commands, too), because it is more readable and has no substantial disadvantages (as I understand it, fsl uses `$FSLDIR` in the scripts in their installation directory specifically in case the user has called a version of an fsl tool that isn't the first one on `$PATH`, but we control `$PATH` in our SetUpHCP... script, and our scripts aren't part of an FSL installation, so that situation doesn't apply to us). @glasserm does other things, partly because he made a `wb_command_retry` script long ago, but that can be done by adding a special "retry script" folder to the front of `$PATH` instead of using a different executable name.

Okay if you think that we don't need to specify the wb_command location even for the user who wants to run a non-default wb_command, then we don't have to keep using it.

orenpoliva:
fMRI Pipeline
Matt - per your request, all .func.gii files were merged into CIFTI files and afterward deleted. The final dtseries files are now saved directly in the main fMRI results directory (e.g., MNINonLinear/Results/rfMRI_REST1_LR), and intermediates files into the HippocampalVolumeToSurfaceMapping subdirectory. These changes were incorporated into CreateHippocampalCIFTIs.sh.
Question: Is this a correct division of files into the folders? Do you want all dscalar.nii files in the main fMRI results directory?

Updated the scripts to read data from the ‘Native’ directory instead of the previous ‘native' directory.

Removed remnants of previous analyses from the test subjects' fMRI results directories so that they now contain outputs generated by the current version of the pipeline. 

Two datasets are available for review:
/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/103818/MNINonLinear/Results/rfMRI_REST1_LR

/media/myelin/brainmappers/Connectome_Project/YA_HCP_Final/105923/MNINonLinear/Results/rfMRI_REST1_LR

fMRISurface/scripts/HippocampalVolumeToSurfaceMapping.sh
```diff
@@ -0,0 +1,333 @@
...
+#  Check that HCPPIPEDIR is defined and Load Function Libraries
+# ------------------------------------------------------------------------------
+
+if [[ -z "${HCPPIPEDIR}" ]]; then
+    echo "${script_name}: ABORTING: HCPPIPEDIR environment variable must be set"
+    exit 1
+fi
+
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"         # Debugging functions; also sources log.shlib
+#source "${HCPPIPEDIR}/global/scripts/log.shlib" "$@"            # Debugging functions; also sources log.shlib
```
coalsont, line 34:
```suggestion
```

Or at least remove the incorrect comment on the end.

Alternatively, it shouldn't actually be harmful to source `log.shlib` again (or before `debug.shlib` actually, since `debug.shlib` actually checks whether `log.shlib` has already been sourced), if we want to support the type of debugging that is improved by commenting out the debug line.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,153 @@
...
+    if [[ "${HCPPIPEDIR:-}" == "" ]]
+    then
+        pipedirguessed=1
+        # Fix this if the script is more than one level below HCPPIPEDIR
+        export HCPPIPEDIR="$(dirname -- "$0")/.."
+    fi
+
+    source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
+    source "${HCPPIPEDIR}/global/scripts/log.shlib" "$@"          # Debugging functions; also sources log.shlib
+
```
coalsont, line 49: Why is this entire file indented now?  This has made some of my previous comments no longer show up on the diff ("files") page (github thinks they are outdated because the lines were edited).

orenpoliva: I reformatted the indentation because I understood your earlier feedback to mean that all of my scripts should follow the same style, specifically four-space indentation. This script had previously been using two-space indentation. I did not realize that reformatting the entire file would cause some of your earlier comments to appear outdated in GitHub. I’ll avoid broad formatting changes like this in future revisions so that existing review comments and the substantive diff remain easier to track.

coalsont: To clarify, by "entire file indented", I meant that after this change, there are now 4 spaces before even the first line of the script.  Anything that isn't in any conditional/loop/function/etc context should have zero spaces/tabs before the first character.  Concretely:

```suggestion
################################################ SUPPORT FUNCTIONS ##################################################
set -eu
pipedirguessed=0
if [[ "${HCPPIPEDIR:-}" == "" ]]
then
    pipedirguessed=1
    # Fix this if the script is more than one level below HCPPIPEDIR
    export HCPPIPEDIR="$(dirname -- "$0")/.."
fi

source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"          # Debugging functions; also sources log.shlib
source "${HCPPIPEDIR}/global/scripts/log.shlib" "$@"          # Debugging functions; also sources log.shlib

```

I prefer four spaces per indentation level, so in any new scripts that would be appreciated.  In existing scripts, following whatever convention is already present (as long as it is consistent) means that the per-line commit history remains more useful.

fMRISurface/GenericHippocampusfMRISurfaceProcessingPipeline.sh
```diff
@@ -0,0 +1,153 @@
...
+    opts_AddOptional '--procstring' 'ProcString' 'string' 'processing suffix appended to the 4D fMRI filename; include the leading underscore' ""
+
+    opts_AddMandatory '--smoothingFWHM' 'SmoothingFWHM' 'number' 'smoothing FWHM (mm)'
+
+    opts_AddOptional '--goodvoxel' 'doGoodVoxels' 'YES OR NO' "Controls whether to do goodVoxel procedure (default = YES)" "YES"
+
+    opts_AddOptional '--factor' 'factor' 'number' "Scaling factor for eliminating high COV voxels (default = 1.5)" "1.5"
+
+    opts_AddOptional '--resample_mesh' 'MeshString' 'string' 'Resample native mesh to: 512, 2k, 8k, or 18k. Use quote for multiple meshes, e.g. "512 2k 8k"' "2k"
```
glasserm, line 69: We don't need a carriage return between each option.  This just makes for more scrolling.

glasserm:
With regard to the test dataset.  I don't think we need to save out "ones" files permanently in ${StudyFolder}/${Subject}/MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping.  

What are these files? ${StudyFolder}/${Subject}//MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping/rfMRI_REST1_LR_AtlasHipp_hp2000_clean_rclean_tclean_fMRI.2k.dscalar.nii
${StudyFolder}/${Subject}/MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping/rfMRI_REST1_LR_AtlasHipp_hp2000_clean_rclean_tclean_fMRI.native.dscalar.nii

They seem like something named dubiously that also probably should be deleted.

orenpoliva:
> With regard to the test dataset. I don't think we need to save out "ones" files permanently in ${StudyFolder}/${Subject}/MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping.
> 
> What are these files? ${StudyFolder}/${Subject}//MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping/rfMRI_REST1_LR_AtlasHipp_hp2000_clean_rclean_tclean_fMRI.2k.dscalar.nii ${StudyFolder}/${Subject}/MNINonLinear/Results/rfMRI_REST1_LR/HippocampalVolumeToSurfaceMapping/rfMRI_REST1_LR_AtlasHipp_hp2000_clean_rclean_tclean_fMRI.native.dscalar.nii
> 
> They seem like something named dubiously that also probably should be deleted.

How do you want the output dtseries to be labelled?

PR #413, opened 2026-08-28
ROI to DICOM

coalsont:
New script for users to make a T1w DICOM file with an ROI "burned" into the image values.

### DIFF
--- a/global/scripts/ROI_to_DICOM.sh
+++ b/global/scripts/ROI_to_DICOM.sh
@@ -20,6 +20,7 @@ opts_AddMandatory '--dicom-input' 'DicomIn' 'path' "folder containing the origin
 opts_AddOptional '--dicom-series' 'DicomSeriesIn' 'series number' "if the folder contains multiple series, you must specify the DICOM series number of the appropriate T1w here"
 opts_AddOptional '--grayordinates' 'Grayord' '91282' "the grayordinates cifti space the roi/vertex is based on, default 91282" '91282'
 opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "what surface registration to use"
+opts_AddMandatory '--dicom-output' 'DicomOut' 'path' "location to write modified dicom series"
 opts_AddOptional '--cifti-roi-in' 'roiIn' 'file' "cifti (dscalar) file containing the binary roi"
 opts_AddOptional '--vertex-in' 'vertexIn' 'index' "specify a single vertex index (0-based) to draw an ROI from, requires --vertex-structure and uses the surface mesh implied by --grayordinates"
 opts_AddOptional '--vertex-structure' 'vertexStruct' 'name' "specify what surface the vertex is in, currently supports CORTEX_LEFT or CORTEX_RIGHT"
@@ -29,7 +30,6 @@ opts_AddOptional '--outline' 'outlineDist' 'distance' "use an outline with the g
 opts_AddOptional '--outline-type' 'outlineType' 'name' "outline behavior, valid types:
 INSIDE (default) - do not draw any voxels outside of the ROI, only make some of the ROI interior transparent
 OUTSIDE - draw voxels outside the ROI and make the entire ROI transparent" 'INSIDE'
-opts_AddMandatory '--dicom-output' 'DicomOut' 'path' "location to write modified dicom series"
 
 opts_ParseArguments "$@"
 
--- a/global/scripts/ROI_to_DICOM.sh
+++ b/global/scripts/ROI_to_DICOM.sh
@@ -21,7 +21,9 @@ opts_AddOptional '--dicom-series' 'DicomSeriesIn' 'series number' "if the folder
 opts_AddOptional '--grayordinates' 'Grayord' '91282' "the grayordinates cifti space the roi/vertex is based on, default 91282" '91282'
 opts_AddMandatory '--surf-reg-name' 'RegName' 'MSMAll' "what surface registration to use"
 opts_AddMandatory '--dicom-output' 'DicomOut' 'path' "location to write modified dicom series"
-opts_AddOptional '--cifti-roi-in' 'roiIn' 'file' "cifti (dscalar) file containing the binary roi"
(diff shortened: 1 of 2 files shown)


global/scripts/ROI_to_DICOM.sh
```diff
@@ -0,0 +1,399 @@
...
+opts_AddOptional '--vertex-in' 'vertexIn' 'index' "specify a single vertex index (0-based) to draw an ROI from, requires --vertex-structure and uses the surface mesh implied by --grayordinates"
+opts_AddOptional '--vertex-structure' 'vertexStruct' 'name' "specify what surface the vertex is in, currently supports CORTEX_LEFT or CORTEX_RIGHT"
+opts_AddOptional '--vertex-radius' 'vertexDist' 'number' "a distance in mm around --vertex-in to include in the ROI, default 0 (just the vertex)" '0'
+opts_AddOptional '--burn-value' 'outVal' 'number' "what value to overwrite in-ROI voxels with [default: 20% above voxelwise maximum]"
+opts_AddOptional '--outline' 'outlineDist' 'distance' "use an outline with the given thickness in millimeters instead of occluding the ROI location"
+opts_AddOptional '--outline-type' 'outlineType' 'name' "outline behavior, valid types:
+INSIDE (default) - do not draw any voxels outside of the ROI, only make some of the ROI interior transparent
+OUTSIDE - draw voxels outside the ROI and make the entire ROI transparent" 'INSIDE'
+opts_AddMandatory '--dicom-output' 'DicomOut' 'path' "location to write modified dicom series"
```
glasserm: Should the mandatory options all be first so that this last one does not get missed?

glasserm:
What needs to be done yet to get this merged?  Do you want external users to test it?

coalsont:
Since it is a new script, we don't have to test "did we break something", at least.  I wasn't sure how much review we wanted to do on it.

glasserm:
In the future, we will want to support volume ROIs too (e.g., a subcortical grey matter or white matter structure).  That isn't needed for the initial code though.  Let's merge when we feel we are done with any additional work on this.

global/scripts/ROI_to_DICOM.sh
```diff
@@ -0,0 +1,399 @@
...
+    echo "$vertexIn" > "$vertTxt"
+    #we don't know if the user has requested a smaller radius than the vertex spacing, so use dilation's "one neighbor minimum" to keep the roi centered-ish on the vertex
+    #start with distance of 0 to get just the vertex
+    wb_command -surface-geodesic-rois \
+        "$StudyFolder"/"$Subject"/T1w/fsaverage_LR"$ciftiMesh"k/"$Subject"."$hem".midthickness"$RegString"."$ciftiMesh"k_fs_LR.surf.gii \
+        0 \
+        "$vertTxt" \
+        "$tempVertMetric"
+    
+    if [[ $(echo "$vertexDist == 0" | bc) == "1*" ]]
```
mmilch01: "1*" will be treated as literal string so the expression will likely be always evaluated to False
```suggestion
    if [[ $(echo "$vertexDist == 0" | bc) == 1* ]]
```

global/scripts/ROI_to_DICOM.sh
```diff
@@ -0,0 +1,399 @@
...
+#-x i keeps dicom voxel encoding order
+dcmniicmd=(dcm2niix -z y -x i -b n -f "$rawbase" -o "$rawfolder")
+
+if [[ "$DicomSeriesIn" != "" ]]
+then
+    #first matching volume is selected by default, in the case of several volumes per series
+    echo "Sorting the input DICOM folder"
+    series_crc=($(dcm2niix -o "$rawfolder" -n -1 -b n -f %s "$DicomIn" | \
+        awk -v s="$DicomSeriesIn" '$1~/^[0-9]+([.][0-9]+)?$/{n=$NF;sub(/^.*\//,"",n);sub(/_.*/,"",n);if(n==s)print $1}'))
+    if [[ "$series_crc" == "" ]]
```
mmilch01: This would avoid the 'unbound variable' instead of the intended message, under 'set -u'  when series_src is an empty array.
```suggestion
   if (( ${#series_crc[@]} == 0 ))
```

global/scripts/ROI_to_DICOM.sh
```diff
@@ -0,0 +1,399 @@
...
+    then
+        log_Err_Abort "could not find series $DicomSeriesIn in folder $DicomIn"
+    fi
+    dcmniicmd+=(-n "$series_crc")
+else #determine the input series number to make the output series number
+    echo "Reading the input DICOM folder"
+    DicomSeriesIn=($(dcm2niix -o "$rawfolder" -n -1 -b n -f %s "$DicomIn" | \
+        awk '$1~/^[0-9]+([.][0-9]+)?$/{sub(".*/","",$NF);sub("_.*","",$NF);print $NF;exit}'))
```
mmilch01: As written, this would always return the first series. Remove 'exit' to list all series instead of the first one, to trigger the following error message both when no series and more than one series is found. 
```suggestion
        awk '$1~/^[0-9]+([.][0-9]+)?$/{sub(".*/","",$NF);sub("_.*","",$NF);print $NF}'))
```

mmilch01:
I did a few minor syntax edits. The one above is the only outstanding right now.

coalsont:
> Let's merge when we feel we are done with any additional work on this.

I'm not sure what this means.

Volume support shouldn't be hard to add, it already supports volume components from cifti.  I'm thinking the setting of the ROI volume space should not have a default when the input is a volume file.

glasserm:
Can we try a test with a volume CIFTI component?

coalsont:
I tested with a cerebellum ROI that I drew and manually imported to 91k space - importing deleted a white matter blade from the ROI, which in hindsight is expected.  Technically it should work with any cifti volume space, but .nii.gz is more disk and IO efficient for purely volume ROIs.

glasserm (approved):
Seems like this is ready for merge unless anyone feels that more testing is needed.

mmilch01 (approved):

PR #414, opened 2026-09-11
Tractography pipe line

Azure2003:
This is not done, but I want to get some early comments on this before I write the sample script, especially on options and/or combining options when necessary.

### DIFF
--- a/DiffusionTractography/PostTractography.sh
+++ b/DiffusionTractography/PostTractography.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+set -eu
+
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    #fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+# Load function libraries
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"  # Command line option functions
+
+
+# Perform the steps of the HCP Diffusion Preprocessing Pipeline
+opts_SetScriptDescription "Prepare the data to run Tractography"
+
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'folders delimited by @' 'BEDPOSTX Folders delimited by @'
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'NONE for MSMSulc, else RegName such as MSMAll'
+opts_AddMandatory '--matrix' 'Matrix' '1 or 3' 'Matrix 1 or Matrix 3 seeding strategy'
+
+
+opts_ParseArguments "$@"
+
+if ((pipedirguessed))
+then
+    log_Err_Abort "HCPPIPEDIR is not set, you must first source your edited copy of Examples/Scripts/SetUpHCPPipeline.sh"
+fi
+
+opts_ShowValues
+
+"$HCPPIPEDIR"/show_version
+
+# ------------------------------------------------------------------------------
+#  Verify required environment variables are set and log value
+# ------------------------------------------------------------------------------
+
+log_Check_Env_Var HCPPIPEDIR
+log_Check_Env_Var FSLDIR
+log_Check_Env_Var CARET7DIR
+
+log_Msg "Platform Information Follows: "
+uname -a
+
+
+T1wDiffusionFolder="${StudyFolder}/${Subject}/T1w/Diffusion"
+DiffusionResolution=`${FSLDIR}/bin/fslval ${T1wDiffusionFolder}/data pixdim1`
+DiffusionResolution=`printf "%0.2f" ${DiffusionResolution}`
+ResultsFolder="${T1wFolder}/Results"
+DiffMeshFolder="${T1wFolder}/fsaverage_LR${DiffResMesh}k"
+TractographyResultsFolder="${ResultsFolder}/Matrix${Matrix}WholeBrainTractography"
+BedpostXFolder="${BedpostXFolders}" #TODO: Allow only one bedpostX folder
+
+log_Msg "Converting Probtrackx Matrices"
+
+if [ ${Matrix} -eq 1 ] ; then
+  ${CARET7DIR}/wb_command -probtrackx-dot-convert ${TractographyResultsFolder}/fdt_matrix1.dot WBSPARSE ${TractographyResultsFolder}/fdt_matrix1.dconn.wbsparse -row-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -col-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -transpose
+  ${CARET7DIR}/wb_command -probtrackx-dot-convert ${TractographyResultsFolder}/fdt_matrix1_lengths.dot WBSPARSE ${TractographyResultsFolder}/fdt_matrix1_lengths.dconn.wbsparse -row-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -col-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -transpose
+elif [ ${Matrix} -eq 3 ] ; then
+  ${CARET7DIR}/wb_command -probtrackx-dot-convert ${TractographyResultsFolder}/fdt_matrix3.dot WBSPARSE ${TractographyResultsFolder}/fdt_matrix3.dconn.wbsparse -row-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -col-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -transpose -make-symmetric
+  ${CARET7DIR}/wb_command -probtrackx-dot-convert ${TractographyResultsFolder}/fdt_matrix3_lengths.dot WBSPARSE ${TractographyResultsFolder}/fdt_matrix3_lengths.dconn.wbsparse -row-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -col-cifti ${DiffMeshFolder}/Grey.dscalar.nii COLUMN -transpose -make-symmetric
+else
+  log_Err_Abort "Matrix Type Not Supported"
+fi
+${CARET7DIR}/wb_command -convert-matrix4-to-workbench-sparse ${TractographyResultsFolder}/fdt_matrix4_1.mtx ${TractographyResultsFolder}/fdt_matrix4_2.mtx ${TractographyResultsFolder}/fdt_matrix4_3.mtx ${BedpostXFolder}/Diffusion.bedpostX_Whole_Brain_Trajectory_1.25.fiberTEMP.nii ${TractographyResultsFolder}/tract_space_coords_for_fdt_matrix4 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse -cifti-seeds ${DiffMeshFolder}/Grey.dscalar.nii COLUMN
+${CARET7DIR}/wb_command -convert-matrix4-to-matrix2 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse WBSPARSE ${TractographyResultsFolder}/fdt_matrix2.dconn.wbsparse -distances ${TractographyResultsFolder}/fdt_matrix2_dist.dconn.wbsparse
+
+log_Msg "Completed"
+
--- a/DiffusionTractography/PreTractography.sh
+++ b/DiffusionTractography/PreTractography.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+set -eu
+
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    #fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+# Load function libraries
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"  # Command line option functions
+
+
+# Perform the steps of the HCP Diffusion Preprocessing Pipeline
+opts_SetScriptDescription "Prepare the data to run Tractography"
+
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'name of folder containing fiber estimations' ""
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'RegName such as MSMAll'
+opts_AddMandatory '--results-folder' 'folder' 'The specific folder in which the seed of tractography is located. This should follow HCP standards' ""
+opts_AddMandatory '--group' 'whim' 'true or false' "Indicate if you tractography for averaging or just an individual. true if this is for averaging"
+opts_AddOptional '--warp' 'warp' 'If your intended seed space is not diffusion space' ""
+opts_AddOptional '--groupname' 'GroupName' 'Average Group Name' ""
+
+
+
+opts_ParseArguments "$@"
+
+if ((pipedirguessed))
+then
+    log_Err_Abort "HCPPIPEDIR is not set, you must first source your edited copy of Examples/Scripts/SetUpHCPPipeline.sh"
+fi
+
+opts_ShowValues
+
+"$HCPPIPEDIR"/show_version
+
+# ------------------------------------------------------------------------------
+#  Verify required environment variables are set and log value
+# ------------------------------------------------------------------------------
+
+log_Check_Env_Var HCPPIPEDIR
+log_Check_Env_Var FSLDIR
+log_Check_Env_Var CARET7DIR
+
+log_Msg "Platform Information Follows: "
+uname -a
+
+# Setup PATHS
+#PipelineScripts=${HCPPIPEDIR_dMRITract}
+PipelineScripts=${HCPPIPEDIR}/DiffusionTractography/scripts #TODO: Delete when commited and in setup script
+
+
+WholeBrainTrajectoryLabels=${HCPPIPEDIR_Config}/WholeBrainFreeSurferTrajectoryLabelTableLut.txt
+LeftCerebralTrajectoryLabels=${HCPPIPEDIR_Config}/LeftCerebralFreeSurferTrajectoryLabelTableLut.txt 
+RightCerebralTrajectoryLabels=${HCPPIPEDIR_Config}/RightCerebralFreeSurferTrajectoryLabelTableLut.txt
+WholeBrainSubcorticalLabels=${HCPPIPEDIR_Config}/WholeBrainSubcorticalFreeSurferTrajectoryLabelTableLut.txt
+LeftSubcorticalLabels=${HCPPIPEDIR_Config}/LeftSubcorticalFreeSurferTrajectoryLabelTableLut.txt 
+RightSubcorticalLabels=${HCPPIPEDIR_Config}/RightSubcorticalFreeSurferTrajectoryLabelTableLut.txt
+WholeBrainWhiteLabels=${HCPPIPEDIR_Config}/WholeBrainWhiteFreeSurferTrajectoryLabelTableLut.txt
+FreeSurferLabels=${HCPPIPEDIR_Config}/FreeSurferAllLut.txt
+
+
+T1wDiffusionFolder="${StudyFolder}/${Subject}/T1w/Diffusion"
+DiffusionResolution=`${FSLDIR}/bin/fslval ${T1wDiffusionFolder}/data pixdim1`
+DiffusionResolution=`printf "%0.2f" ${DiffusionResolution}`
+
+log_Msg "MakeTrajectorySpace"
+
+
+${PipelineScripts}/MakeTrajectorySpace.sh \
+    --path="$StudyFolder" --subject="$Subject" \
+    --wholebrainlabels="$WholeBrainTrajectoryLabels" \
+    --folder="${folder}" \
+    --warp="${warp}" \
+    --leftcerebrallabels="$LeftCerebralTrajectoryLabels" \
+    --rightcerebrallabels="$RightCerebralTrajectoryLabels" \
+    --wholebrainsubcorticallabels="$WholeBrainSubcorticalLabels" \
+    --leftsubcorticallabels="$LeftSubcorticalLabels" \
+    --rightsubcorticallabels="$RightSubcorticalLabels" \
+    --wholebrainwhitelabels="$WholeBrainWhiteLabels" \
+    --diffresol="${DiffusionResolution}" \
+    --freesurferlabels="${FreeSurferLabels}" \
+    --whim="$whim"
+
+log_Msg "MakeWorkbenchUODFs"
+#${HCPPIPEDIR_dMRITract}/MakeWorkbenchUODFs.sh --path="${StudyFolder}" --subject="${Subject}" --lowresmesh="${LowResMesh}" --diffresol="${DiffusionResolution}" --bpxdirs="${BedpostXFolders}"
+#The bpxdirs is just the name of the folder, it is not a directory. If you do it, it is wrong.
+#whim is a flag. Whim is single sample so different commands needs to be used.
+${PipelineScripts}/MakeWorkbenchUODFs.sh --path="${StudyFolder}" --subject="${Subject}" --folder="${folder}" --diffresol="${DiffusionResolution}" --bpxdirs="${BedpostXFolders}" --whim="${whim}"
+
+log_Msg "MakeSeeds"
+${PipelineScripts}/MakeSeeds.sh --path="${StudyFolder}" --subject="${Subject}" --folder="${folder}" --diffresmesh="${DiffResMesh}" --diffresol="${DiffusionResolution}" --regname="${RegName}" --whim="${whim}" --groupname="$GroupName"
+
+log_Msg "Completed"
+
--- a/DiffusionTractography/Tractography.sh
+++ b/DiffusionTractography/Tractography.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+set -eu
+
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    #fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+# Load function libraries
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"  # Command line option functions
+
+
+# Perform the steps of the HCP Diffusion Preprocessing Pipeline
+opts_SetScriptDescription "Actual run Tractography"
+
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject"
+opts_AddMandatory '--results-folder' 'Folder' 'folder' "folder"
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'Full path to bedpostx folders' 'BEDPOSTX Folders delimited by @'
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'NONE for MSMSulc, else RegName such as MSMAll'
+opts_AddMandatory '--matrix4' 'Matrix4' 'If we are running matrix 4' 'true'
+opts_AddMandatory '--seeding-strategy' 'SeedingStrategy' 'Matrix3 or Matrix1' 'true'
+opts_AddMandatory '--nsamples' 'nsamples' 'number of samples we run per voxel for speed.' '100'
+opts_AddMandatory '--nsteps' 'nsteps' 'number of steps we take per streamline.' '2000'
+opts_AddOptional '--whim' 'Whim' 'full path to whim folder' ""
+opts_AddOptional '--mask' 'Mask' 'full path to Whim group mask' ""
+opts_ParseArguments "$@"
+
+opts_ShowValues
+
+
+T1wDiffusionFolder="${StudyFolder}/${Subject}/T1w/Diffusion"
+DiffusionResolution=`${FSLDIR}/bin/fslval ${T1wDiffusionFolder}/data pixdim1`
+DiffusionResolution=`printf "%0.2f" ${DiffusionResolution}`
+STEPLENGTH=$(echo "$DiffusionResolution / 4" | bc -l)
+STEPLENGTH=$(printf "%g" "$STEPLENGTH")   # removes trailing zeros
+PipelineScripts=${HCPPIPEDIR}/DiffusionTractography/scripts #TODO: Delete when commited and in setup script
+
+
+if [[ "$SeedingStrategy" == "Matrix3" ]]; then
+    mkdir -p "${StudyFolder}/${Subject}/${Folder}/Results/Matrix3WholeBrainTractography"
+fi
+if [[ "$SeedingStrategy" == "Matrix1" ]]; then
+    mkdir -p "${StudyFolder}/${Subject}/${Folder}/Results/Matrix1WholeBrainTractography"
+fi
+
+if [[ "$SeedingStrategy" == "Matrix3" ]]; then
+    OUTDIR="${StudyFolder}/${Subject}/${Folder}/Results/Matrix3WholeBrainTractography"
+    SEED="${StudyFolder}/${Subject}/${Folder}/Whole_Brain_WhiteMatter_${DiffusionResolution}.nii.gz"
+    mkdir -p "$OUTDIR"
+else
+    OUTDIR="${StudyFolder}/${Subject}/${Folder}/Results/Matrix1WholeBrainTractography"
+    SEED="${StudyFolder}/${Subject}/${Folder}/ROIs/${Subject}.GreyROI_${RegName}.${DiffResMesh}k_fs_LR_${DiffusionResolution}.txt"
+    mkdir -p "$OUTDIR"
+fi
+
+COMMON_ARGS=(
+    --samples="${BedpostXFolders}"
+    --mask="${StudyFolder}/${Subject}/T1w/Whole_Brain_Trajectory_${DiffusionResolution}.nii.gz"
+    --seed="${SEED}"
+    --waypoints="${StudyFolder}/${Subject}/${Folder}/Whole_Brain_WhiteMatter_${DiffusionResolution}.nii.gz"
+    --stop="${StudyFolder}/${Subject}/${Folder}/ROIs/${Subject}.StopROI_${RegName}.${DiffResMesh}k_fs_LR.txt"
+    --wtstop="${StudyFolder}/${Subject}/${Folder}/ROIs/${Subject}.GreyROI_${RegName}.${DiffResMesh}k_fs_LR_${DiffusionResolution}.txt"
+    --seedref="${StudyFolder}/${Subject}/${Folder}/Whole_Brain_Trajectory_${DiffusionResolution}.nii.gz"
+    --dir="${OUTDIR}"
+    --nsamples="${nsamples}"
+    --cthr=0
+    --steplength="${STEPLENGTH}"
+    --distthresh=0
+    --forcedir
+    --meshspace=caret
+    --nsteps="${nsteps}"
+    --fibthresh=0.05
+    --loopcheck
+    --randfib=2
+    --forcefirststep
+    --opd
+    --ompl
+    --sampvox="${DiffusionResolution}"
+    --verbose=1
+)
+
+EXTRA_ARGS=()
+
+if [[ "$SeedingStrategy" == "Matrix3" ]]; then
+    EXTRA_ARGS+=(
+        --target3="${StudyFolder}/${Subject}/T1w/ROIs/${Subject}.GreyROI_${RegName}.${DiffResMesh}k_fs_LR_${DiffusionResolution}.txt"
+        --distthresh3=0
+        --omatrix3
+    )
+else
+    EXTRA_ARGS+=(
+        --distthresh1=0
+        --omatrix1
+    )
+fi
+
+if [[ "$Matrix4" == "true" ]]; then
+    EXTRA_ARGS+=(
+        --target4="${StudyFolder}/${Subject}/T1w/Whole_Brain_Trajectory_${DiffusionResolution}.nii.gz"
+        --omatrix4
+    )
+
+    if [[ "$SeedingStrategy" == "Matrix3" ]]; then
+        EXTRA_ARGS+=(
+            --colmask4="${StudyFolder}/${Subject}/T1w/ROIs/${Subject}.GreyROI_${RegName}.${DiffResMesh}k_fs_LR_${DiffusionResolution}.txt"
+        )
+    fi
+fi
+
+WHIM_ARGS=()
+
+if [[ -n "${Whim:-}" ]]; then
+    WHIM_ARGS+=(
+        --whim="${Whim}"
+        --mask1="${Mask}"
+    )
+fi
+
+
+${FSLDIR}/bin/probtrackx2_gpu \
+    "${COMMON_ARGS[@]}" \
+    "${EXTRA_ARGS[@]}" \
+    "${WHIM_ARGS[@]}"
--- a/DiffusionTractography/groupSubCorticalGray.sh
+++ b/DiffusionTractography/groupSubCorticalGray.sh
@@ -1,0 +1,65 @@
+
+#!/bin/bash
+set -eu
+
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    #fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+# Load function libraries
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"  # Command line option functions
+
+opts_SetScriptDescription "Generate group Subcortical Gray from group Wmparc"
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--results-folder' 'Folder' 'The specific folder in which the seed of tractography is located. This should follow HCP standards' ""
+opts_AddOptional '--groupname' 'GroupName' 'Group Folder_Name' ""
+opts_ParseArguments "$@"
+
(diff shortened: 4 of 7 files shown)


DiffusionTractography/PreTractography.sh
```diff
@@ -0,0 +1,100 @@
...
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'name of folder containing fiber estimations' ""
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'RegName such as MSMAll'
+opts_AddMandatory '--results-folder' 'folder' 'The specific folder in which the seed of tractography is located. This should follow HCP standards' ""
+opts_AddMandatory '--group' 'whim' 'true or false' "Indicate if you tractography for averaging or just an individual. true if this is for averaging"
+opts_AddOptional '--whimmask' 'WhimMask' 'path' "path for group volume labeled whim mask"
+opts_AddOptional '--warp' 'warp' 'If your intended seed space is not diffusion space' ""
+opts_AddOptional '--groupname' 'GroupName' 'Average Group Name' ""
```
coalsont: Having an empty string in a Mandatory line should be rare.  The only reason we do it in the `--subject` line is that we don't think it needs an explanation.  Anywhere else, you are missing an argument.

Additionally, if `--results-folder` is supposed to have a specific naming convention, it probably shouldn't be an argument to the pipeline itself.  If it is intended for testing purposes, then having an optional parameter like `--alternate-output-path` would make more sense.  `folder` is also not a good variable name, it is nonspecific and is a substring of many other variables and comments.

Edit: by convention, the descriptions are in double quotes (`""`), in case someone wants to expand a variable in one.

```suggestion
opts_AddOptional '--bpxdirs' 'BedpostXFolders' 'folder@folder' "names of folders containing fiber estimations, default Diffusion.bedpostX" 'Diffusion.bedpostX'
opts_AddMandatory '--regname' 'RegName' 'Name of Registration' "RegName such as MSMAll"
opts_AddOptional '--whimmask' 'WhimMask' 'path' "path for group volume labeled whim mask"
opts_AddOptional '--warp' 'warp' 'file' "If your intended seed space is not diffusion space"
opts_AddOptional '--groupname' 'GroupName' 'string' "Average Group Name"
```

Azure2003: --results-folder is supposed to designate T1w or MMORF etc so it is not set per se.

Azure2003: Also, the additional question is: should we designate whimmask and whim, or combine the two options?

coalsont: Also, instead of `--warp`, we may instead want a specifier of what volume space to use, like `MNINonLinear` or `HCPMultiModalNonLinear`, so that both the warpfield and output folder can be set properly without the user needing to manually specify the HCP-intended structure.

The setting of `--group` can be inferred by the presence of `--whimmask` instead of being a separate option.

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+set -e
+echo -e "\n START: MakeWorkbenchUODFs"
+
+
+########################################## SUPPORT FUNCTIONS #####################################################
+# function for parsing options
+getopt1() {
```
coalsont: Why are we using parsing conventions from before even `opts.shlib`?

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,90 @@
...
+StudyFolder=`getopt1 "--path" $@`                # "$1" #Path to Generic Study folder
+Subject=`getopt1 "--subject" $@`                 # "$2" #SubjectID
+Folder=`getopt1 "--folder" $@`                   # "$3" #folder so it is not just t1w
+DiffusionResolution=`getopt1 "--diffresol" $@`   # "$4" #Diffusion Resolution in mm
+BedpostXFolders=`getopt1 "--bpxfoldername" $@` #Delimited by @
+Whim=`getopt1 "--whim" $@` #"true" or "false". Though it is just judging if it is true or not.
+WhimMask=`getopt1 "--whimmask" $@` #Path to whim mask
+
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"
```
coalsont, line 15: And then sourcing `newopts.shlib` without using it?

DiffusionTractography/PreTractography.sh
Azure2003: OK

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
Azure2003: From Professor Glasser's code

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
Azure2003: From professor Glasser's Code, we should rewrite it as the getopt is kind of annoying to use

glasserm:
What are the NIFTI files included in the pull request used for?  It looks like they are used in the wb_command -convert-fiber-orientations.  Should psi, ka, and kb made optional @coalsont so that this is not necessary?

DiffusionTractography/scripts/MakeSeeds.sh
```diff
@@ -0,0 +1,103 @@
...
+    ${CARET7DIR}/wb_command -metric-math var-var+1 ${FolderMeshFolder}/${Subject}.${Hemisphere}.all_${RegName}.${DiffResMesh}k_fs_LR.shape.gii -var var ${AtlasFolder}/fsaverage_LR${DiffResMesh}k/${Subject}.${Hemisphere}.atlasroi.${DiffResMesh}k_fs_LR.shape.gii  #Is this really needed?
+    $FSLDIR/bin/surf2surf -i ${FolderMeshFolder}/${Subject}.${Hemisphere}.white_${RegName}.${DiffResMesh}k_fs_LR.surf.gii -o ${ROIsTrajectorySpaceFolder}/${Subject}.${Hemisphere}.white_${RegName}.${DiffResMesh}k_fs_LR.gii --values=${FolderMeshFolder}/${Subject}.${Hemisphere}.roi_${RegName}.${DiffResMesh}k_fs_LR.shape.gii
+    $FSLDIR/bin/surf2surf -i ${FolderMeshFolder}/${Subject}.${Hemisphere}.pial_${RegName}.${DiffResMesh}k_fs_LR.surf.gii -o ${ROIsTrajectorySpaceFolder}/${Subject}.${Hemisphere}.pial_${RegName}.${DiffResMesh}k_fs_LR.gii --values=${FolderMeshFolder}/${Subject}.${Hemisphere}.roi_${RegName}.${DiffResMesh}k_fs_LR.shape.gii
+  fi
+done
+
+#TODO: Currently unique per individual
+##Need group name here
```
glasserm, line 72: Are these done?

DiffusionTractography/scripts/MakeSeeds.sh
```diff
@@ -0,0 +1,103 @@
...
+  ${CARET7DIR}/wb_command -volume-label-import ${ROIsTrajectorySpaceFolder}/Atlas_Whole_Brain_SubCortical_GreyMatter_${DiffusionResolution}.nii.gz ${ROIsTrajectorySpaceFolder}/tmp.txt ${ROIsTrajectorySpaceFolder}/Atlas_Whole_Brain_SubCortical_GreyMatter_${DiffusionResolution}.nii.gz
+else
+  fslmaths ${TrajectorySpaceFolder}/Whole_Brain_SubCortical_GreyMatter_${DiffusionResolution}.nii.gz -bin ${ROIsTrajectorySpaceFolder}/Whole_Brain_SubCortical_GreyMatter_${DiffusionResolution}.nii.gz
+  echo "OTHER" > ${ROIsTrajectorySpaceFolder}/tmp.txt
+  echo "1 255 255 255 255" >> ${ROIsTrajectorySpaceFolder}/tmp.txt
+  ${CARET7DIR}/wb_command -volume-label-import ${ROIsTrajectorySpaceFolder}/Whole_Brain_SubCortical_GreyMatter_${DiffusionResolution}.nii.gz ${ROIsTrajectorySpaceFolder}/tmp.txt ${ROIsTrajectorySpaceFolder}/Whole_Brain_SubCortical_GreyMatter_${DiffusionResolution}.nii.gz
+fi
+rm ${ROIsTrajectorySpaceFolder}/tmp.txt
+
+###Add conditional to detect whether to change this to individual vs group.
```
glasserm, line 86: Is this done?

DiffusionTractography/scripts/MakeTrajectorySpace.sh
```diff
@@ -0,0 +1,196 @@
...
+##Again if not T1w, do apply warp only with xfms
+###We are creating the diffusion resolution reference
+if [ ! -e "$TrajectorySpaceFolder"/"$T1wImage"_"$DiffusionResolution" ] ; then
+  ${FSLDIR}/bin/flirt -interp spline -in "$TrajectorySpaceFolder"/"$T1wImage".nii.gz -ref "$TrajectorySpaceFolder"/"$T1wImage".nii.gz -applyisoxfm "$DiffusionResolution" -out "$TrajectorySpaceFolder"/"$T1wImage"_"$DiffusionResolution"
+fi
+if [[ -n "$WhimMask" ]]; then
+  if [ ! -e "${StudyFolder}"/"${Subject}"/T1w/wmparc_1mm.nii.gz ] ; then
+    FreeSurferFolder="$TrajectorySpaceFolder"/"$Subject"
+    mri_convert -rt nearest -rl "$TrajectorySpaceFolder"/"$T1wImage".nii.gz "$FreeSurferFolder"/mri/wmparc.mgz "$TrajectorySpaceFolder"/wmparc_1mm.nii.gz
+  fi
```
glasserm, line 91: Is this assuming PostFreeSurfer wasn't run?  This file should exist in ${StudyFolder}/${Subject}/T1w.  If it does not exist in MMORFNonLinear, it would not be correct to simply copy it into MMORFNonLinear because it is not in that space.  We need to keep the files in the correct volume spaces within given folders.

DiffusionTractography/scripts/MakeTrajectorySpace.sh
```diff
@@ -0,0 +1,196 @@
...
+  ${Caret7_Command} -volume-label-import "$TrajectorySpaceFolder"/"$wmparc"_"$DiffusionResolution".nii.gz "$FreeSurferLabels" "$TrajectorySpaceFolder"/"$wmparc"_"$DiffusionResolution".nii.gz -drop-unused-labels
+fi
+###That is already done in the pre script. I think for consistensy sake, we need them to provide this.
+
+LeftGreyRibbonValue="3"
+LeftWhiteMaskValue="2"
+RightGreyRibbonValue="42"
+RightWhiteMaskValue="41"
+###OK in theory we don't need this either if we have whim. But it is also good to have??
```
glasserm, line 106: Could keep it or make it conditional on individual space processing.

glasserm:
When does one run [DiffusionTractography/groupSubCorticalGray.sh](https://github.com/Washington-University/HCPpipelines/pull/414/changes#diff-f9db4100217cf843866b8ab8170e4361c93a5858401737db2f0d7eddef31dc5d) and how will we tell users to run it?

DiffusionTractography/PostTractography.sh
```diff
@@ -0,0 +1,81 @@
...
+  ${CARET7DIR}/wb_command -convert-matrix4-to-workbench-sparse ${TractographyResultsFolder}/fdt_matrix4_1.mtx ${TractographyResultsFolder}/fdt_matrix4_2.mtx ${TractographyResultsFolder}/fdt_matrix4_3.mtx ${BedpostXFolderPath}/${BedpostXFolder}_Whole_Brain_Trajectory_1.25.fiberTEMP.nii ${TractographyResultsFolder}/tract_space_coords_for_fdt_matrix4 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse -cifti-seeds ${DiffMeshFolder}/Grey.dscalar.nii COLUMN
+else
+  ${CARET7DIR}/wb_command -convert-matrix4-to-workbench-sparse ${TractographyResultsFolder}/fdt_matrix4_1.mtx ${TractographyResultsFolder}/fdt_matrix4_2.mtx ${TractographyResultsFolder}/fdt_matrix4_3.mtx ${BedpostXFolderPathT1w}/Diffusion.bedpostX_Whole_Brain_Trajectory_1.25.fiberTEMP.nii ${TractographyResultsFolder}/tract_space_coords_for_fdt_matrix4 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse -cifti-seeds ${DiffMeshFolder}/Grey.dscalar.nii COLUMN
+fi
+${CARET7DIR}/wb_command -convert-matrix4-to-matrix2 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse WBSPARSE ${TractographyResultsFolder}/fdt_matrix2.dconn.wbsparse -distances ${TractographyResultsFolder}/fdt_matrix2_dist.dconn.wbsparse
```
glasserm: Are we currently making this?  I think it would be better not to waste space on this copy of the data and instead make sure we can robustly access the original copy in the trajTEMP.wbsparse.

DiffusionTractography/PostTractography.sh
```diff
@@ -0,0 +1,81 @@
...
+else
+  log_Err_Abort "Matrix Type Not Supported"
+fi
+if [ "${whim}" == "true" ]; then
+  ${CARET7DIR}/wb_command -convert-matrix4-to-workbench-sparse ${TractographyResultsFolder}/fdt_matrix4_1.mtx ${TractographyResultsFolder}/fdt_matrix4_2.mtx ${TractographyResultsFolder}/fdt_matrix4_3.mtx ${BedpostXFolderPath}/${BedpostXFolder}_Whole_Brain_Trajectory_1.25.fiberTEMP.nii ${TractographyResultsFolder}/tract_space_coords_for_fdt_matrix4 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse -cifti-seeds ${DiffMeshFolder}/Grey.dscalar.nii COLUMN
+else
+  ${CARET7DIR}/wb_command -convert-matrix4-to-workbench-sparse ${TractographyResultsFolder}/fdt_matrix4_1.mtx ${TractographyResultsFolder}/fdt_matrix4_2.mtx ${TractographyResultsFolder}/fdt_matrix4_3.mtx ${BedpostXFolderPathT1w}/Diffusion.bedpostX_Whole_Brain_Trajectory_1.25.fiberTEMP.nii ${TractographyResultsFolder}/tract_space_coords_for_fdt_matrix4 ${TractographyResultsFolder}/fdt_matrix4.trajTEMP.wbsparse -cifti-seeds ${DiffMeshFolder}/Grey.dscalar.nii COLUMN
+fi
```
glasserm, line 78: Should there be a flag to clean up the .mtx files to save space?

DiffusionTractography/scripts/MakeSeeds.sh
Azure2003: Yes, it is.

DiffusionTractography/scripts/MakeSeeds.sh
Azure2003: This is also done.

DiffusionTractography/scripts/MakeTrajectorySpace.sh
Azure2003: We are not actually copying it. This just creates the T1w PostFreeSurfer. The actual applywarp afterwards corects it into MMORFNonLinear

DiffusionTractography/scripts/MakeTrajectorySpace.sh
Azure2003: We still need some of the files. It was not like every file was used even if it is in the T1w space.

DiffusionTractography/PostTractography.sh
Azure2003: OK

DiffusionTractography/PostTractography.sh
Azure2003: This was just from your original code, I will clean this up.

Azure2003:
> When does one run [DiffusionTractography/groupSubCorticalGray.sh](https://github.com/Washington-University/HCPpipelines/pull/414/changes#diff-f9db4100217cf843866b8ab8170e4361c93a5858401737db2f0d7eddef31dc5d) and how will we tell users to run it?

Post averaging. I will be making a less ambitious version of make average. Also supposedly this is just for whim. Like there are a lot of steps if they are running with whim that is not included in this pipeline. They should just refer to the methods paper.

Azure2003:
> > When does one run [DiffusionTractography/groupSubCorticalGray.sh](https://github.com/Washington-University/HCPpipelines/pull/414/changes#diff-f9db4100217cf843866b8ab8170e4361c93a5858401737db2f0d7eddef31dc5d) and how will we tell users to run it?
> 
> Post averaging. I will be making a less ambitious version of make average. Also supposedly this is just for whim. Like there are a lot of steps if they are running with whim that is not included in this pipeline. They should just refer to the methods paper.

These include running the actual whim, the making the mask for whim, making the mask into labeled volume, averaging.

DiffusionTractography/PostTractography.sh
glasserm: The suggestion was to have an argument that the user could use to control this behavior, not do it always.

coalsont:
> What are the NIFTI files included in the pull request used for? It looks like they are used in the wb_command -convert-fiber-orientations. Should psi, ka, and kb made optional @coalsont so that this is not necessary?

We should make temporary files within the script instead (especially if the voxel resolution might be different), rather than including files of constant value on a specific voxel grid in the repo.

The output format requires values for these, so taking them as mandatory inputs to the command is the obvious approach, and makes it more obvious when a script is faking them.  Even for group orientations, in theory WHIM or a post-processing step could at some point combine the fanning information across subjects and output something useful where it is consistent, so putting in fake values is still a hack even in that context.

For individual WHIM-ordered group-registered orientations, it would be nice if we had a way to generate the real fanning information (it would be more obviously meaningful than in the group orientations).  As a brutal hack (maybe just to help explain the concept), one idea would be to nearest-neighbor resample the bedpost samples (and apply the warpfield's local rotation, and reorder the fibers to match WHIM...) and rerun the bingham fitting (but if combined with the presumably nicer-interpolated dyads orientations, since the main orientation can disagree, the psi angle may result in fanning in somewhat the wrong direction, particularly near "gimbal lock").

DiffusionTractography/scripts/MakeTrajectorySpace.sh
```diff
@@ -0,0 +1,196 @@
...
+  ${FSLDIR}/bin/fslmaths "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white_uthr0.native.nii.gz -mul $WhiteMaskValue "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white_mask.native.nii.gz
+  ${FSLDIR}/bin/fslmaths "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".ribbon.nii.gz -add "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white_mask.native.nii.gz "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".ribbon.nii.gz
+  rm "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white.native.nii.gz "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white_thr0.native.nii.gz "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".pial.native.nii.gz "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".pial_uthr0.native.nii.gz "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white_uthr0.native.nii.gz "$TrajectorySpaceFolder"/"$NativeFolder"/"$Subject"."$Hemisphere".white_mask.native.nii.gz
```
coalsont, line 127: Consider `tempfiles.shlib`, which would also have the side effect of shortening all the arguments that aren't permanent files:

https://github.com/Washington-University/HCPpipelines/blob/122eab1ff935cf08d5dfe78cf7d9519e297a81d2/global/scripts/RSNregression.sh#L117-L118

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+set -e
+echo -e "\n START: MakeWorkbenchUODFs"
```
coalsont, line 3: If we have a line like this, it should go after `opts_ShowValues` (maybe even after the sanity check(s)).

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+set -e
```
coalsont, line 2: It looks like this script should be okay under `set -eu`...

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,85 @@
...
+    ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+    ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+  -fiber \
+    ${BedpostXFolder}/f_3_std.nii.gz \
+    ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+    ${BedpostXFolder}/theta_3_std.nii.gz \
+    ${BedpostXFolder}/phi_3_std.nii.gz \
+    ${HCPPIPEDIR}/global/templates/psi_zero.nii.gz \
+    ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+    ${HCPPIPEDIR}/global/templates/small_values.nii.gz
```
coalsont: When doing lots of line continuations, please indent them, and option tree sections.

```suggestion
    ${Caret7_Command} -convert-fiber-orientations \
      $WhimMask \
      ${BedpostXFolder}/${BedpostXFolderName}_${trajectory}_${DiffusionResolution}.fiberTEMP.nii \
      -fiber \
        ${BedpostXFolder}/f_1_std.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
        ${BedpostXFolder}/theta_1_std.nii.gz \
        ${BedpostXFolder}/phi_1_std.nii.gz \
        ${HCPPIPEDIR}/global/templates/psi_zero.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
      -fiber \
        ${BedpostXFolder}/f_2_std.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
        ${BedpostXFolder}/theta_2_std.nii.gz \
        ${BedpostXFolder}/phi_2_std.nii.gz \
        ${HCPPIPEDIR}/global/templates/psi_zero.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
      -fiber \
        ${BedpostXFolder}/f_3_std.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
        ${BedpostXFolder}/theta_3_std.nii.gz \
        ${BedpostXFolder}/phi_3_std.nii.gz \
        ${HCPPIPEDIR}/global/templates/psi_zero.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
        ${HCPPIPEDIR}/global/templates/small_values.nii.gz
```

DiffusionTractography/PostTractography.sh
```diff
@@ -0,0 +1,95 @@
...
+opts_SetScriptDescription "Prepare the data to run Tractography"
+
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' ""
+opts_AddMandatory '--results-folder' 'folder' 'The specific folder in which the seed of tractography is located. This should follow HCP standards' ""
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdir' 'BedpostXFolder' 'folder that stores bedpostX results or whim results' 'BEDPOSTX Folder or whim folder'
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'NONE for MSMSulc, else RegName such as MSMAll'
+opts_AddMandatory '--matrix' 'Matrix' '1 or 3' 'Matrix 1 or Matrix 3 seeding strategy'
+opts_AddMandatory '--group' 'whim' 'true or false' "Indicate if you tractography for averaging or just an individual. true if this is for averaging"
```
coalsont, line 27: If the user uses WHIM in (Pre)Tractography, but not in Post, what happens (is there an error message, would the user understand it means they needed this flag)?  Should the WHIM and non-WHIM outputs have more-different naming to catch this mistake?

Sanity check all boolean arguments (converts to 0/1 to allow math-type boolean conditionals):

https://github.com/Washington-University/HCPpipelines/blob/122eab1ff935cf08d5dfe78cf7d9519e297a81d2/global/scripts/RSNregression.sh#L72

DiffusionTractography/Tractography.sh
```diff
@@ -0,0 +1,129 @@
...
+opts_AddMandatory '--results-folder' 'Folder' 'folder' "folder"
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'Full path to bedpostx folders' 'BEDPOSTX Folders delimited by @'
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'NONE for MSMSulc, else RegName such as MSMAll'
+opts_AddMandatory '--matrix4' 'Matrix4' 'If we are running matrix 4' 'true'
+opts_AddMandatory '--seeding-strategy' 'SeedingStrategy' 'Matrix3 or Matrix1' 'true'
+opts_AddMandatory '--nsamples' 'nsamples' 'number of samples we run per voxel for speed.' '100'
+opts_AddMandatory '--nsteps' 'nsteps' 'number of steps we take per streamline.' '2000'
+opts_AddOptional '--whim' 'Whim' 'full path to whim folder' ""
+opts_AddOptional '--mask' 'Mask' 'full path to Whim group mask' ""
```
coalsont, line 31: These are missing an argument, what looks like a redundant empty string default is actually the "long description" field.

```suggestion
opts_AddOptional '--whim' 'Whim' 'path' "full path to whim folder"
opts_AddOptional '--mask' 'Mask' 'file' "full path to Whim group mask"
```

DiffusionTractography/Tractography.sh
```diff
@@ -0,0 +1,129 @@
...
+    export HCPPIPEDIR="$(dirname -- "$0")/.."
+fi
+
+# Load function libraries
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"  # Command line option functions
+
+
+# Perform the steps of the HCP Diffusion Preprocessing Pipeline
+opts_SetScriptDescription "Actual run Tractography"
```
coalsont:
```suggestion
opts_SetScriptDescription "Run probabilistic tractography"
```

DiffusionTractography/Tractography.sh
```diff
@@ -0,0 +1,129 @@
...
+    mkdir -p "${StudyFolder}/${Subject}/${Folder}/Results/Matrix1WholeBrainTractography"
+fi
+
+if [[ "$SeedingStrategy" == "Matrix3" ]]; then
+    OUTDIR="${StudyFolder}/${Subject}/${Folder}/Results/Matrix3WholeBrainTractography"
+    SEED="${StudyFolder}/${Subject}/${Folder}/Whole_Brain_WhiteMatter_${DiffusionResolution}.nii.gz"
+    mkdir -p "$OUTDIR"
+else
+    OUTDIR="${StudyFolder}/${Subject}/${Folder}/Results/Matrix1WholeBrainTractography"
+    SEED="${StudyFolder}/${Subject}/${Folder}/ROIs/${Subject}.GreyROI_${RegName}.${DiffResMesh}k_fs_LR_${DiffusionResolution}.txt"
```
coalsont, line 58: We don't want `GreyROI_.32k_fs_LR` or other garbage when using MSMSulc:

https://github.com/Washington-University/HCPpipelines/blob/122eab1ff935cf08d5dfe78cf7d9519e297a81d2/global/scripts/RSNregression.sh#L100-L104

coalsont: The test should probably also include "MSMSulc".  I prefer not to use the old "NONE" convention in new scripts, or at least not mention it in the help info (even if it is accepted to mean MSMSulc).

DiffusionTractography/Tractography.sh
```diff
@@ -0,0 +1,129 @@
...
+STEPLENGTH=$(printf "%g" "$STEPLENGTH")   # removes trailing zeros
+PipelineScripts=${HCPPIPEDIR}/DiffusionTractography/scripts #TODO: Delete when commited and in setup script
+
+
+if [[ "$SeedingStrategy" == "Matrix3" ]]; then
+    mkdir -p "${StudyFolder}/${Subject}/${Folder}/Results/Matrix3WholeBrainTractography"
+fi
+if [[ "$SeedingStrategy" == "Matrix1" ]]; then
+    mkdir -p "${StudyFolder}/${Subject}/${Folder}/Results/Matrix1WholeBrainTractography"
+fi
```
coalsont, line 50: Please check that the user didn't say "matrix3" or "Mtrix3", or other typos.  I would combine these into a `case` and put the error message in `(*)`.

DiffusionTractography/Tractography.sh
```diff
@@ -0,0 +1,129 @@
...
+opts_AddMandatory '--path' 'StudyFolder' 'Path' "path to session's data folder"
+opts_AddMandatory '--subject' 'Subject' 'subject ID' "subject"
+opts_AddMandatory '--results-folder' 'Folder' 'folder' "folder"
+opts_AddMandatory '--diffresmesh' 'DiffResMesh' 'number' 'diffusion res mesh number'
+opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'Full path to bedpostx folders' 'BEDPOSTX Folders delimited by @'
+opts_AddMandatory '--regname' 'RegName' 'Name of Registration' 'NONE for MSMSulc, else RegName such as MSMAll'
+opts_AddMandatory '--matrix4' 'Matrix4' 'If we are running matrix 4' 'true'
+opts_AddMandatory '--seeding-strategy' 'SeedingStrategy' 'Matrix3 or Matrix1' 'true'
+opts_AddMandatory '--nsamples' 'nsamples' 'number of samples we run per voxel for speed.' '100'
+opts_AddMandatory '--nsteps' 'nsteps' 'number of steps we take per streamline.' '2000'
```
coalsont, line 29: On closer inspection, many of these are also missing the "type-ish" argument.  They also attempt to provide a default for mandatory parameters (in the long description field).  `--matrix4` will need to be handled with `opts_StringToBool`.

```suggestion
opts_AddMandatory '--bpxdirs' 'BedpostXFolders' 'path@path' 'BEDPOSTX Folders delimited by @'
opts_AddMandatory '--regname' 'RegName' 'MSMAll' 'RegName such as MSMAll or MSMSulc'
opts_AddOptional '--matrix4' 'Matrix4' 'YES or NO' 'If we are running matrix 4, default YES' 'YES'
opts_AddMandatory '--seeding-strategy' 'SeedingStrategy' 'Matrix3 or Matrix1' 'what seeding strategy to use' #FIXME: tried to have a default of "true"?!?
opts_AddOptional '--nsamples' 'nsamples' 'integer' 'number of samples we run per seed, has linear effect on runtime, default 100' '100'
opts_AddOptional '--nsteps' 'nsteps' 'integer' 'number of steps we take per streamline, default 2000' '2000'
```

It might be nice if we could either hide or explain the "Matrix4 is an output type, Matrix3 is a seeding strategy" confusion.  Maybe use more descriptive identifiers for the seeding strategy, like `surface` and `voxels`?

DiffusionTractography/groupSubCorticalGray.sh
```diff
@@ -0,0 +1,65 @@
...
+    [[ -d "$dir" ]] || continue
+    [[ "$Subject" =~ ^[0-9]+$ ]] || continue
+
+    echo "$Subject"
+    break
+done
+
+T1wDiffusionFolder="${StudyFolder}/${Subject}/T1w/Diffusion"
+DiffusionResolution=`${FSLDIR}/bin/fslval ${T1wDiffusionFolder}/data pixdim1`
+DiffusionResolution=`printf "%0.2f" ${DiffusionResolution}`
```
coalsont, line 37: A list of subjects (or a single example subject) should be a mandatory argument if it is really needed.  In particular, checking that the subject is only a string of digits will break for anything other than HCP-YA.

In this case, it appears to use only the first subject to get only the resolution of the data, which is then string-formatted and used in filenames.  In other pipelines, the resolution string is a mandatory argument to the script, which would remove this fragile code.

DiffusionTractography/scripts/MakeTrajectorySpace.sh
glasserm: That is long-standing code of mine Tim.  Probably not Alex's responsibility to change.

DiffusionTractography/Tractography.sh
glasserm: Agree.

DiffusionTractography/Tractography.sh
glasserm: Matrix1=grey matter seeding
Matrix3=white matter seeding

DiffusionTractography/groupSubCorticalGray.sh
Azure2003: OK, that makes things easy then. Mostly because the pipelines in this folder do not ask the user for this, so  I thought for consistency's sake I will do it this way.

DiffusionTractography/scripts/MakeSeeds.sh
glasserm: Please clean out the comments then.

DiffusionTractography/scripts/MakeSeeds.sh
glasserm: Please clean out the comments then.

DiffusionTractography/scripts/MakeTrajectorySpace.sh
glasserm: It is putting it into the wrong folder if $TrajectorySpaceFolder is MMORFNonLinear.

glasserm:
> > What are the NIFTI files included in the pull request used for? It looks like they are used in the wb_command -convert-fiber-orientations. Should psi, ka, and kb made optional @coalsont so that this is not necessary?
> 
> We should make temporary files within the script instead (especially if the voxel resolution might be different), rather than including files of constant value on a specific voxel grid in the repo.
> 
> The output format requires values for these, so taking them as mandatory inputs to the command is the obvious approach, and makes it more obvious when a script is faking them. Even for group orientations, in theory WHIM or a post-processing step could at some point combine the fanning information across subjects and output something useful where it is consistent, so putting in fake values is still a hack even in that context.
> 
> For individual WHIM-ordered group-registered orientations, it would be nice if we had a way to generate the real fanning information (it would be more obviously meaningful than in the group orientations). As a brutal hack (maybe just to help explain the concept), one idea would be to nearest-neighbor resample the bedpost samples (and apply the warpfield's local rotation, and reorder the fibers to match WHIM...) and rerun the bingham fitting (but if combined with the presumably nicer-interpolated dyads orientations, since the main orientation can disagree, the psi angle may result in fanning in somewhat the wrong direction, particularly near "gimbal lock").

Why can't we fix wb_command to put garbage in these locations in this mode of use instead of needing to create temporary files for this?

DiffusionTractography/groupSubCorticalGray.sh
glasserm: Agree.

Azure2003:
> > What are the NIFTI files included in the pull request used for? It looks like they are used in the wb_command -convert-fiber-orientations. Should psi, ka, and kb made optional @coalsont so that this is not necessary?
> 
> We should make temporary files within the script instead (especially if the voxel resolution might be different), rather than including files of constant value on a specific voxel grid in the repo.
> 
> The output format requires values for these, so taking them as mandatory inputs to the command is the obvious approach, and makes it more obvious when a script is faking them. Even for group orientations, in theory WHIM or a post-processing step could at some point combine the fanning information across subjects and output something useful where it is consistent, so putting in fake values is still a hack even in that context.
> 
> For individual WHIM-ordered group-registered orientations, it would be nice if we had a way to generate the real fanning information (it would be more obviously meaningful than in the group orientations). As a brutal hack (maybe just to help explain the concept), one idea would be to nearest-neighbor resample the bedpost samples (and apply the warpfield's local rotation, and reorder the fibers to match WHIM...) and rerun the bingham fitting (but if combined with the presumably nicer-interpolated dyads orientations, since the main orientation can disagree, the psi angle may result in fanning in somewhat the wrong direction, particularly near "gimbal lock").

Currently, whim and makeaverage etc. is not necessarily covered by this pipeline as there are some things that requires user fine tuning for example thresholding the mask for whim. This is done upon manual inspection and trying different values. This is why we probably won't have a post whim. It will probably be a markdown file or something like that.

We can make it on the fly but it is also really annoying to create one every single time for every single subject especially if there is no subject specific difference in the file. My two cents are that we do it like greyoordinates basically where we have standard resolutions built in.

I highly doubt if whim actually can produce the fanning information in general. It does not use MCMC or some sort of monte carlo based thing (which is what I assume bedpostx uses, though I have not looked at the code, so I don't know), because it is a single minimization problem; thus, in the eyes of the algorithm, there is a singular correct value, basically.

DiffusionTractography/scripts/MakeTrajectorySpace.sh
Azure2003: Oh good point

coalsont:
> We can make it on the fly but it is also really annoying to create one every single time for every single subject especially if there is no subject specific difference in the file. My two cents are that we do it like greyoordinates basically where we have standard resolutions built in.

It can be different across studies or species, so there isn't a compelling reason to have it in the repository.  A study might even use a different resolution for WHIM (to be index compatible with a previous study) and non-WHIM (to use the full resolution) runs of tractography.  The only obvious thing we can count on is that it will be the same space as the `merged_f1samples.nii.gz` or `f_1_std.nii.gz` and related.  There is even a comment about resampling the files anticipating problems of this nature, which would be completely avoided by generating them on the fly in the script.

A file that contains constant values should compress (.nii.gz) rather well (as evidence, note that the ones in the PR are ~150KB rather than ~15MB uncompressed, so a ratio of 1%), and these are single-frame files at less than T1w resolution anyway.  I don't think we need to regenerate the fiberTEMP files very often, and making them should be much faster than any per-subject part of bedpostx or probtrackx.

In short, these constant value dummy files don't belong in the repository (as things stand, they are trivial to generate, and need to be altered for different diffusion resolutions anyway).  I don't think the time savings of modifying the command would have any appreciable effect within the context of the entire diffusion pipeline (and it would potentially also make it less obvious from reading the script that fake values were being used in the outputs).  I will write suggestion comments to generate the desired files on the fly.

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,85 @@
...
+pipedirguessed=0
+if [[ "${HCPPIPEDIR:-}" == "" ]]
+then
+    pipedirguessed=1
+    #fix this if the script is more than one level below HCPPIPEDIR
+    export HCPPIPEDIR="$(dirname -- "$0")/../.."
+fi
+
+source "${HCPPIPEDIR}/global/scripts/debug.shlib" "$@"    # Debugging functions; also sources log.shlib
+source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"
```
coalsont, line 15: (for psi_zero and small_values) To handle temporary files more easily:
```suggestion
source "${HCPPIPEDIR}/global/scripts/newopts.shlib" "$@"  
source "${HCPPIPEDIR}/global/scripts/tempfiles.shlib" "$@"  
```

DiffusionTractography/scripts/MakeWorkbenchUODFs.sh
```diff
@@ -0,0 +1,85 @@
...
+        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+      -fiber \
+        ${BedpostXFolder}/f_3_std.nii.gz \
+        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+        ${BedpostXFolder}/theta_3_std.nii.gz \
+        ${BedpostXFolder}/phi_3_std.nii.gz \
+        ${HCPPIPEDIR}/global/templates/psi_zero.nii.gz \
+        ${HCPPIPEDIR}/global/templates/small_values.nii.gz \
+        ${HCPPIPEDIR}/global/templates/small_values.nii.gz
```
coalsont, line 75: (for psi_zero and small_values) Generate and use temporaries (in `/tmp/` by default, automatically get deleted when script exits):

```suggestion
    BedpostXFolder="${T1wFolder}/${BedpostXFolderName}"
    tempfiles_create diffusion_UODFs_zero_XXXXXX.nii.gz zeroFile
    tempfiles_create diffusion_UODFs_smallval_XXXXXX.nii.gz smallValFile
    wb_command -volume-math '0' "$zeroFile" -var x "$BedpostXFolder"/f_1_std.nii.gz
    wb_command -volume-math '0.001' "$smallValFile" -var x "$BedpostXFolder"/f_1_std.nii.gz
    ${Caret7_Command} -convert-fiber-orientations \
      $WhimMask \
      ${BedpostXFolder}/${BedpostXFolderName}_${trajectory}_${DiffusionResolution}.fiberTEMP.nii \
      -fiber \
        ${BedpostXFolder}/f_1_std.nii.gz \
        "$smallValFile" \
        ${BedpostXFolder}/theta_1_std.nii.gz \
        ${BedpostXFolder}/phi_1_std.nii.gz \
        "$zeroFile" \
        "$smallValFile" \
        "$smallValFile" \
      -fiber \
        ${BedpostXFolder}/f_2_std.nii.gz \
        "$smallValFile" \
        ${BedpostXFolder}/theta_2_std.nii.gz \
        ${BedpostXFolder}/phi_2_std.nii.gz \
        "$zeroFile" \
        "$smallValFile" \
        "$smallValFile" \
      -fiber \
        ${BedpostXFolder}/f_3_std.nii.gz \
        "$smallValFile" \
        ${BedpostXFolder}/theta_3_std.nii.gz \
        ${BedpostXFolder}/phi_3_std.nii.gz \
        "$zeroFile" \
        "$smallValFile" \
        "$smallValFile"
```

glasserm:
> > We can make it on the fly but it is also really annoying to create one every single time for every single subject especially if there is no subject specific difference in the file. My two cents are that we do it like greyoordinates basically where we have standard resolutions built in.
> 
> It can be different across studies or species, so there isn't a compelling reason to have it in the repository. ~A study might even use a different resolution for WHIM (to be index compatible with a previous study) and non-WHIM (to use the full resolution) runs of tractography~ (EDIT: on closer inspection, non-whim doesn't use fake parameters, it uses -estimate-fiber-binghams). The only obvious thing we can count on is that it will be the same space as the ~`merged_f1samples.nii.gz` or~ `f_1_std.nii.gz` and related. There is even a comment about resampling the files anticipating problems of this nature, which would be completely avoided by generating them on the fly in the script.
> 
> A file that contains constant values should compress (.nii.gz) rather well (as evidence, note that the ones in the PR are ~150KB rather than ~15MB uncompressed, so a ratio of 1%), and these are single-frame files at less than T1w resolution anyway. I don't think we need to regenerate the fiberTEMP files very often, and making them should be much faster than any per-subject part of bedpostx or probtrackx.
> 
> In short, these constant value dummy files don't belong in the repository (as things stand, they are trivial to generate, and need to be altered for different diffusion resolutions anyway). I don't think the time savings of modifying the command would have any appreciable effect within the context of the entire diffusion pipeline (and it would potentially also make it less obvious from reading the script that fake values were being used in the outputs). I will write suggestion comments to generate the desired files on the fly.

Tim and I discussed and are in agreement to make these on the fly.

PR #415, opened 2026-09-18
Hippocampal support to RSN regression

orenpoliva:
Adds hippocampal input preparation and hippocampal surface support to RSN regression.


orenpoliva:
RSNregression changes:

1. The script uses HCPCIFTIRWDIR, which is defined in SetUpHCPPipeline.sh. Therefore, the script now sources SetUpHCPPipeline.sh when HCPCIFTIRWDIR is not already set.

2. Added a Structure option with two accepted values: Cortex, which is the default, and Hippocampus. The script validates the supplied value and rejects unsupported options. Cerebellum can be added as another option in the future.
Throughout the script, case statements are used whenever different files or processing steps are required for the cortex and hippocampus.

3. Added a helper script, CreateHippocampalRSNInputs.sh. When Structure==Hippocampus, this script prepares the hippocampal group maps, surfaces, and normalized vertex-area data required for hippocampal RSN regression.

4. For cortical processing, the normalized surface vertex-area data are expanded into the group-map CIFTI space using wb_command -cifti-separate followed by wb_command -cifti-create-dense-from-template. These commands are not used when Structure==Hippocampus because CreateHippocampalRSNInputs.sh instead adapts the group maps and normalized vertex-area data directly to the HippUnfold surface space.

glasserm:
This is not what I had in mind.  We do not need to prepare hippocampal group maps.  We just need to support the hippocampal data as an output space for now.  That means the input is the whole brain CIFTI and the outputs are whole brain CIFITI, volume, hippocampus, and in the future cerebellum.

coalsont:
> 1. The script uses HCPCIFTIRWDIR, which is defined in SetUpHCPPipeline.sh. Therefore, the script now sources SetUpHCPPipeline.sh when HCPCIFTIRWDIR is not already set.

This is wrong.  The user needs to edit `SetUpHCP...` and then ensure it is sourced before running any "pipeline script" (as contrasted with the "launcher examples" in `Examples/Scripts`, which do source `SetUpHCP...`, and the user is expected to substitute in the actual path to their edited version), it is never a pipeline script's job to find the user's environment setup.  The user-edited version of SetUpHCP... is expected to often not have the same name as before (including the user's initials, for instance), and/or to be in a different folder (not necessarily `$HOME/projects/...`).  Users are expected to never edit anything in the repo except files under `Examples/`.  It is true that at the moment, we don't have a launcher example for `RSNregression.sh`, but it still shouldn't break this convention.

The reason that `newopts.shlib` scripts guess at `HCPPIPEDIR` is only so that they can print the help info (because it needs the shlib to define the `opts_Add*` functions which define both the parsing and help info).  They intentionally abort if they had to guess `HCPPIPEDIR` and were provided all of their mandatory arguments, rather than attempting to continue in an improper environment setup.

global/scripts/RSNregression.sh
```diff
@@ -111,18 +133,40 @@ fi
...
+case "$Structure" in
+    (Hippocampus)
+        DownSampleMNIFolder="$MNIFolder/HippUnfold/${LowResMesh}k"
+        DownSampleT1wFolder="$T1wFolder/HippUnfold/${LowResMesh}k"
+        ;;
+    (Cortex)
+        DownSampleMNIFolder="$MNIFolder/fsaverage_LR${LowResMesh}k"
+        DownSampleT1wFolder="$T1wFolder/fsaverage_LR${LowResMesh}k"
+        ;;
+esac
```
coalsont, line 146: This can be combined with the sanity checking `case`.

global/scripts/RSNregression.sh
```diff
@@ -65,6 +79,14 @@ then
...
+case "$Structure" in
+    (Hippocampus | Cortex)
+        ;;
+    (*)
+        log_Err_Abort "unrecognized structure '$Structure', use Hippocampus or Cortex"
+        ;;
+esac
+
 #display the parsed/default values
 opts_ShowValues
```
coalsont, line 91: Please put new code somewhere below the `opts_ShowValues` line, so that argument values get logged before the script can abort.

global/scripts/CreateHippocampalRSNInputs.sh
```diff
@@ -0,0 +1,129 @@
...
+    exit 1
+fi
+
+MNIFolder="$1"
+WorkingDirectory="$2"
+Subject="$3"
+GroupMaps="$4"
+LowResMesh="$5"
+fMRIName="$6"
+ProcString="$7"
```
coalsont, line 22: For new scripts, consider using `newopts.shlib` when there are several arguments:

https://github.com/Washington-University/HCPpipelines/blob/122eab1ff935cf08d5dfe78cf7d9519e297a81d2/global/scripts/newopts.shlib#L4-L46

global/scripts/CreateHippocampalRSNInputs.sh
```diff
@@ -0,0 +1,129 @@
...
+if [[ ! -f "$GroupMaps" ]]; then
+    log_Err_Abort "VN template does not exist: $VNTemplate"
+fi
+# ------------------------------------------------------------------------------
+# Create normalized hippocampal vertex-area weights
+# ------------------------------------------------------------------------------
+
+log_Msg "Creating hippocampal vertex-area CIFTI"
+
+wb_command -cifti-create-dense-from-template "$VNTemplate" "$VARaw" -metric HIPPOCAMPUS_LEFT "${HippUnfoldDirectory}/${Subject}.L.hipp_surfarea.${LowResMesh}k.shape.gii" -metric HIPPOCAMPUS_RIGHT "${HippUnfoldDirectory}/${Subject}.R.hipp_surfarea.${LowResMesh}k.shape.gii" -metric HIPPOCAMPUS_DENTATE_LEFT "${HippUnfoldDirectory}/${Subject}.L.dentate_surfarea.${LowResMesh}k.shape.gii" -metric HIPPOCAMPUS_DENTATE_RIGHT "${HippUnfoldDirectory}/${Subject}.R.dentate_surfarea.${LowResMesh}k.shape.gii"
```
coalsont, line 66: These might be better named based on "vertex area", matching existing conventions (which abbreviate it to "_va").  It appears that they are expected to be generated by a different pipeline, though.

glasserm: We used HippUnfold's convention.

